PR #168, opened 2015-01-05
Cankov/debug brk

ivanbuhov:
Added debug-brk option to build and attach debug tools in Safari for ios platform.

### DIFF
--- a/lib/options.ts
+++ b/lib/options.ts
@@ -17,7 +17,7 @@ var knownOpts:any = {
 		"keyStorePath": String,
 		"keyStorePassword": String,
 		"keyStoreAlias": String,
-		"keyStoreAliasPassword": String
+		"keyStoreAliasPassword": String,
 		"debug-brk": Boolean
 	},
 	shorthands = {
--- a/lib/options.ts
+++ b/lib/options.ts
@@ -18,6 +18,7 @@ var knownOpts:any = {
 		"keyStorePassword": String,
 		"keyStoreAlias": String,
 		"keyStoreAliasPassword": String
+		"debug-brk": Boolean
 	},
 	shorthands = {
 	};
--- a/lib/services/ios-project-service.ts
+++ b/lib/services/ios-project-service.ts
@@ -84,7 +84,6 @@ class IOSProjectService implements  IPlatformProjectService {
 
 	public interpolateData(projectRoot: string): IFuture<void> {
 		return (() => {
-			console.log("interpolate data");
 			this.replaceFileName("-Info.plist", path.join(projectRoot, IOSProjectService.IOS_PROJECT_NAME_PLACEHOLDER)).wait();
 			this.replaceFileName("WithInspector-Info.plist", path.join(projectRoot, IOSProjectService.IOS_PROJECT_NAME_PLACEHOLDER)).wait();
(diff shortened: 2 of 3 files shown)


lib/services/platform-service.ts
```diff
@@ -243,13 +244,19 @@ export class PlatformService implements IPlatformService {
...
 			var packageFile = this.getLatestApplicationPackageForEmulator(platformData).wait().packageName;
 			this.$logger.out("Using ", packageFile);
 
-			var logFilePath = path.join(platformData.projectRoot, this.$projectData.projectName, "emulator.log");
+			var projectName = options["debug-brk"] ? this.$projectData.projectName + "WithInspector" : this.$projectData.projectName;
+			var logFilePath = path.join(platformData.projectRoot, projectName, "emulator.log");
 
 			emulatorServices.startEmulator(packageFile, { stderrFilePath: logFilePath, stdoutFilePath: logFilePath }).wait();
+
+			if (platform === "ios" && options["debug-brk"]) {
```
Fatme: This `if` shouldn't be here. You should add new function in `IPlatformProjectService` interface - for example `runDebugger()` and implement it only in `ios-project-service`. You should be able to call it here - `platformData. platformProjectService.runDebugger()`.

lib/services/ios-project-service.ts
```diff
@@ -20,6 +20,7 @@ class IOSProjectService implements  IPlatformProjectService {
 		private $iOSEmulatorServices: Mobile.IEmulatorPlatformServices) { }
 
 	public get platformData(): IPlatformData {
+		var projectName = options["debug-brk"] ? this.$projectData.projectName + "WithInspector" : this.$projectData.projectName;
```
Fatme: This code is duplicated on 3 places. I suggest you to expose here `projectName` and use it from other places.

lib/options.ts
```diff
@@ -17,7 +17,8 @@ var knownOpts:any = {
 		"keyStorePath": String,
 		"keyStorePassword": String,
 		"keyStoreAlias": String,
-		"keyStoreAliasPassword": String
+		"keyStoreAliasPassword": String,
+		"debug-brk": Boolean
```
rosen-vladimirov: You should add the option as debug-brk and debugBrk here. The reason is that we are using yargs to get command line arguments and it is working this way: 
If you pass value with dash, yargs adds it to yargs.argv in two ways: with dash and without dash, replacing first symbol after it with its toUpper equivalent.
As we are validating all passed arguments, when you try using debug-brk, yargs will pass two values to our validation logic  - debug-brk and debugBrk. Just add it in both ways, so the validation will pass.

lib/services/platform-service.ts
```diff
@@ -243,13 +244,19 @@ export class PlatformService implements IPlatformService {
...
+				var downloadedPackagePath = this.$npm.install("tns-ios", ["g"]).wait();
```
ligaz: This does not respect the semver convention we are trying to follow, because it will install the latest version of the runtime which debugging tools might not be compatible with the current runtime used in the app.

ligaz: Also consider adding the debugging tools in a separate npm package.

Issue #173, opened 2015-01-12
tns does not work when only Android SDK 21 is installed


teobugslayer:
I verified that I have installed Android SDK 21. When I tries adding a platform, got this error:

```Please install Android target 17 the Android newest SDK). Make sure you have the latest Android tools installed as well. Run "android" from your command-line to install/update any missing SDKs or tools.```

Also, note the spelling mistake ;)

**Comment by ErjanGavalji** [comment]
In the original message (I verified that I have installed Android SDK 21.), did you ONLY have Android SDK 21, but not SDK Android SDK 17?

**Comment by fourpixels** [comment]
Unfortunately this keeps happening, despite "Android SDK 19 or later" info in system requirements. Just just fresh install following step by step instructions (from http://docs.nativescript.org/setup/ns-cli-setup/ns-setup-win.html), and ended up here:

```
Please install Android target 17. Make sure you have the latest Android tools in
stalled as well. Run "android" from your command-line to install/update any miss
ing SDKs or tools.
```

Had to play around (in)installing different SDKs from Android SDK Manager, and hit the spot at Android 4.4.2 (API 19).

Win 7, fresh install of everything that's in the instructions (choco, ant, geny, everything).

Alber70g:
@fourpixels I can confirm this.

teobugslayer:
Hi @fourpixels , @Alber70g,

We have a list of verified SDKs which we should be able to use to build {N} projects. The list is [here](https://github.com/NativeScript/nativescript-cli/blob/release/lib/services/android-project-service.ts#L15). Can you verify that you have fully installed one of the supported SDKs?
I have just tried building with MNC SDK and it worked (this code will land in master later).

fourpixels:
Hm, I have tried installing BOTH MNC and standard API22 (Android 5.1.1) - with those two together I wasn't able to make it work. I'm not sure if this is kind of my fault, but I finally did it with API 19 (4.4.2).

The bad thing is that the error says 'Use Android 17', which is from 2013, and it seems like different from the docs - 19 (requirements)..

toddanglin:
Just experienced this, too, with NS 1.1.0 running on Mac. Is this a regression?

rosen-vladimirov:
Hi @toddanglin, 
[...]

PR #176, opened 2015-01-12
Fix the message for missing Android target 17

teobugslayer:
Add warning on install that it is needed.

See https://github.com/NativeScript/nativescript-cli/issues/173


Fatme:
:+1:

teobugslayer:
:-1: After talking to @ErjanGavalji and @slavchev, we decided to update the target while preparing

PR #179, opened 2015-01-13
Use vladimirov/fibers_1_0_5_fixes branch

rosen-vladimirov:
Change package.json to use vladimirov/fibers_1_0_5_fixes branch where node-fibers are built with all latest fixes. This is a temporary change that has to be removed after merging vladimirov/fibers_1_0_5_fixes into master of node-fibers repo. The temp change is required for testing purposes.


teobugslayer:
:+1:

PR #180, opened 2015-01-13
Allow installing with sudo node

teobugslayer:
See https://github.com/NativeScript/nativescript-cli/issues/171


ligaz:
The code looks similar to the one in AppBuilder CLI. Can it be moved to the common lib and used from there?

rosen-vladimirov:
:+1:

Issue #167, opened 2014-12-23
Genymotion support


tjvantoll:
I think it's important to support Genymotion from the CLI, but I haven't yet been able to get it to work. Specifically, as soon as I have a Genymotion emulator up and running all relevant `tns` commands start returning this error:

```
Command failed: ADB server didn't ACK
* failed to start daemon *
```

What's weird is `adb devices` shows my Genymotion emulator as a device, but `tns list-devices` just returns that error. I'm not sure what's up because I'm able to use the Cordova CLI with Genymotion without issues. I've been using `cordova run android` to deploy apps to the same emulator.

I found [this help document from Titanium](https://wiki.appcelerator.org/display/guides2/Installing+Genymotion#InstallingGenymotion-Failedtostartdaemon) that seemed relevant, but it's recommended workaround steps didn't work for me either.

Any help on this would be appreciated. Thanks.

PR #182, opened 2015-01-14
Fix android emulator starting

rosen-vladimirov:
Add appId option when starting android emulator in order to fix starting of the application (it is installed correctly on the device, but last call, that has to start the app, uses appid). Update common lib where support for genymotion emulator has been added.

https://github.com/NativeScript/nativescript-cli/issues/167


Fatme:
:+1:

ErjanGavalji:
:+1:

PR #183, opened 2015-01-15
Allow updating target android sdk during "platform add"

teobugslayer:
The tns-android template uses by default sdk 17. If the user does not have it installed, we try to update it.
To do so safely, we have a white list of well-tested sdks. We match the white list with the installed sdks and take the newest one.

This commit depends on https://github.com/telerik/tns-android/commit/990f725d845364e33ca77d4e54212e0c2342bc55

Implements https://github.com/NativeScript/nativescript-cli/issues/173

### DIFF
--- a/lib/services/android-project-service.ts
+++ b/lib/services/android-project-service.ts
@@ -3,12 +3,14 @@
 import path = require("path");
 import shell = require("shelljs");
 import util = require("util");
+import Future = require("fibers/future");
 import options = require("./../options");
 import constants = require("./../constants");
 import hostInfo = require("../common/host-info");
 import helpers = require("./../common/helpers");
 
 class AndroidProjectService implements IPlatformProjectService {
+	private SUPPORTED_TARGETS = ["android-17", "android-18", "android-19", "android-21"];
 	private targetApi: string;
 
 	constructor(private $androidEmulatorServices: Mobile.IEmulatorPlatformServices,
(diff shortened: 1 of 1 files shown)


lib/services/android-project-service.ts
```diff
@@ -196,13 +201,58 @@ class AndroidProjectService implements IPlatformProjectService {
...
+				this.$errors.fail("Please install Android target %s. Make sure you have the latest Android tools installed as well." +
+									" Run \"android\" from your command-line to install/update any missing SDKs or tools.",
+									validTarget.split('-')[1]);
+			}
 		}
+
+		return newTarget;
+	}
+
+	private getCompatibleTarget(): string {
```
rosen-vladimirov: The return type must be IFuture<string> as you have .wait() in its body.

dtopuzov:
:+1:

lib/services/android-project-service.ts
```diff
@@ -196,13 +201,58 @@ class AndroidProjectService implements IPlatformProjectService {
...
+		var installedTargets = this.getInstalledTargets().wait();
+		var compatibleTarget = _(this.SUPPORTED_TARGETS).sort().reverse().find(supportedTarget => _.contains(installedTargets, supportedTarget));
```
rosen-vladimirov: Maybe you can use findLast instead of reverse and find
 https://lodash.com/docs#findLast

lib/services/android-project-service.ts
```diff
@@ -196,13 +201,58 @@ class AndroidProjectService implements IPlatformProjectService {
...
+				}
+			});
+			future.wait();
+		}).future<void>()();
+	}
+
+	private installedTargetsCache: string[] = null;
+	private getInstalledTargets(): IFuture<string[]> {
+		return (() => {
+			if (this.installedTargetsCache === null) {
```
rosen-vladimirov: Maybe:
```Typescript
if (!this.installedTargetsCache)
```

lib/services/android-project-service.ts
teobugslayer: We shall wait and see (pun intended).

rosen-vladimirov:
:+1:

PR #184, opened 2015-01-16
Update common lib

rosen-vladimirov:
In common lib the following changes are applied:
Add function to print message after timeout: add new function to helpers.ts which prints a message on the same line (using printInfoMessageOnSameLine method from the same helpers.ts) after a specific timeout had passed. Using future to guarantee that the function will not finish until timeout is passed and message is printed.

Fix missing ";" in options.ts.


teobugslayer:
:+1:

ErjanGavalji:
:+1:

PR #185, opened 2015-01-19
Use ios-sim from master

rosen-vladimirov:
Update package.json to use ios-sim from its master branch. This is temporary in order to be able to test latest changes of node fibers (which are merged in ios-sim master).


teobugslayer:
:+1:

PR #186, opened 2015-01-19
Fix usage of shorthand options

rosen-vladimirov:
Fix usage of shorthands, so when you specify -v to use verbose mode. Update common lib where shorthands for help command had been added.


teobugslayer:
:+1:

PR #191, opened 2015-01-20
Fix exception in getCommandName

rosen-vladimirov:
Add toString() call before calling toLowerCase() to make sure the value is a string. This prevents exception when passed value is a number.

Fixes https://huboard.com/NativeScript/nativescript-cli#/issues/54850524


ErjanGavalji:
:+1:

teobugslayer:
:+1:

PR #194, opened 2015-01-26
Share 7zip between CLIs

rosen-vladimirov:
Switch to external 7zip binary to uncompress files. This supports both zip and lzma compression methods.

Update common lib to fix dev-post-install issue.


teobugslayer:
:+1:

PR #197, opened 2015-01-27
Fix ts compilation

Fatme:


rosen-vladimirov:
:+1:  after green build :)

PR #201, opened 2015-01-30
Add check if 7za can be executed

rosen-vladimirov:
If user cannot use 7zip, he is unable to create projects, unable to build etc. So on post-install we check if we are able to execute the 7za file and if ENOENT is return we show the user where he can find system requirements.

Fix host-info paths in some files and remove host-info reference from some other files.


Fatme:
:+1:  after green build

PR #204, opened 2015-02-02
Use adb from PATH

rosen-vladimirov:
Update common lib where the following changes have been applied:
Add helpers function that checks if "adb" is part of the PATH. In this case all android related commands will use the installed adb instead of the one from our resources. This way we prevent errors due to new released version of the adb (for example in our resources the version is 1.0.31, but there's already 1.0.32) when executing our adb. Errors like:
Command failed: ADB server didn't ACK
* failed to start daemon *
error:

Update adb for windows to version 1.0.32

https://huboard.com/NativeScript/nativescript-cli#/issues/56225098


Fatme:
:+1:

PR #206, opened 2015-02-03
Fix genymotion check

rosen-vladimirov:
When you specify --geny option of $ appbuilder emulate android command, we check for genyshell in the PATH environment variable. As on MacOS genyshell and the real program that we need for genymotion (player) are in different directories, we fail with incorrect message.


Fatme:
:+1:

PR #209, opened 2015-02-06
Update fibers and ios-sim usage

rosen-vladimirov:
Update package.json to use fibers version 1.0.5.1 (it is the same as the currently used from branch vladimirov/fibers_1_0_5_fixes). The only addition are prebuilt binaries that allow usage of fibers with nodejs 0.12
Use ios-sim-portable version 1.0.1 - it is the same as the master branch.

IMPORTANT: This pull request can be merged ONLY AFTER we publish version 1.0.1 of ios-sim-portable.  This requires merge of https://github.com/telerik/ios-sim-portable/pull/16

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,10 +27,10 @@
     "byline": "4.1.1",
     "colors": "0.6.2",
     "ffi": "https://github.com/icenium/node-ffi/tarball/master",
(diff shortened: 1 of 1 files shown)


teobugslayer:
:+1:

rosen-vladimirov:
It is expected to have failed tests until we publish version 1.0.1 of ios-sim-portable

PR #210, opened 2015-02-12
Add cli-table as dependency

rosen-vladimirov:
Add cli-table as dependency as it is used in common lib. Merge all latest changes from common-lib.


ErjanGavalji:
:+1:

Fatme:
:+1:

PR #212, opened 2015-02-12
Don't trigger build if emulate iOS --availableDevices command is executed

Fatme:


teobugslayer:
Is there any change aside the blank spaces?

teobugslayer:
:+1:

PR #213, opened 2015-02-12
Fix usage on Linux

rosen-vladimirov:
Update common lib where On Linux we use IOSDeviceDiscoveryStub as we cannot execute iosDevice related commands.


lib/services/platform-service.ts
```diff
@@ -10,15 +10,19 @@ import options = require("./../options");
...
 		private $logger: ILogger,
 		private $npm: INodePackageManager,
 		private $platformsData: IPlatformsData,
 		private $projectData: IProjectData,
 		private $projectDataService: IProjectDataService,
-		private $prompter: IPrompter) { }
+		private $prompter: IPrompter,
+		private $injector: IInjector) { }
+
+	private get devicesServices(): Mobile.IDevicesServices {
```
Fatme: It should be better to add comment here

teobugslayer:
:+1:

PR #218, opened 2015-02-13
* Remove device option

Fatme:
* Build for emulator by default

Currently device option is added both in common options and in options. I removed the device option from ns options and changed the default behavior of build command.

### DIFF
--- a/lib/services/ios-project-service.ts
+++ b/lib/services/ios-project-service.ts
@@ -130,7 +130,14 @@ class IOSProjectService implements  IPlatformProjectService {
 			];
 			var args: string[] = [];
 
-			if(options.device) {
+			if(options.emulator) {
+				args = basicArgs.concat([
+					"-sdk", "iphonesimulator",
+					"-arch", "i386",
+					"VALID_ARCHS=\"i386\"",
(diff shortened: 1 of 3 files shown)


rosen-vladimirov:
The description says - build for emulator by default, but the code says - build for device? Is the description incorrect?

lib/services/ios-project-service.ts
```diff
@@ -138,19 +145,12 @@ class IOSProjectService implements  IPlatformProjectService {
...
-					"CONFIGURATION_BUILD_DIR=" + path.join(projectRoot, "build", "emulator")
-				]);
 			}
 
 			this.$childProcess.spawnFromEvent("xcodebuild", args, "exit", {cwd: options, stdio: 'inherit'}).wait();
 
-			if(options.device) {
-				var buildOutputPath = path.join(projectRoot, "build", options.device ? "device" : "emulator");
+			if(!options.emulator) {
+				var buildOutputPath = path.join(projectRoot, "build", options.emulator ? "emulator" : "device");
```
rosen-vladimirov: we get here only when options.emulator is false (undefined), why do we check again, shouldn't we use "device"directly?

rosen-vladimirov:
:+1:  just one minor comment for improvement

Issue #219, opened 2015-02-13
Build in release mode for android is throwing an error.


valentinstoychev:
BUILD FAILED
C:\Users\valentin.stoychev\AppData\Local\Android\android-sdk\tools\ant\build.xml:1139: The following error occurred while executing this line:
C:\Users\valentin.stoychev\AppData\Local\Android\android-sdk\tools\ant\build.xml:1151: C:\Users\valentin.stoychev\myAppDemo\platforms\android\u
ndefined (The system cannot find the file specified)

Total time: 9 seconds
Command cmd failed with exit code 1

PR #220, opened 2015-02-13
Fix android release build

Fatme:
Fixes https://github.com/NativeScript/nativescript-cli/issues/219
Send keyStore* options to ant only if they are specified.


rosen-vladimirov:
:+1:

PR #230, opened 2015-02-19
Library add

slavchev:


slavchev:
ping @Fatme @teobugslayer @mbektchiev

lib/commands/add-library.ts
```diff
@@ -0,0 +1,28 @@
+///<reference path="../.d.ts"/>
+"use strict";
+
+import path = require("path");
+import Future = require("fibers/future");
```
ErjanGavalji: Consider using lower-case for the "fibers/future module"...

slavchev: I prefer lower-case syntax as well, but it seems capital-letter syntax is already widely used in the project so I sticked up with the existing convention.

lib/definitions/osenv.d.ts
```diff
@@ -1,3 +1,4 @@
+/*
 declare module "osenv" {
 	function home(): string;
-}
\ No newline at end of file
+}*/
```
rosen-vladimirov: why is this commented?

lib/definitions/platform.d.ts
```diff
@@ -13,6 +13,7 @@ interface IPlatformService {
 	deployOnEmulator(platform: string): IFuture<void>;
 	validatePlatformInstalled(platform: string): void;
 	validatePlatform(platform: string): void;
+    addLibrary(platform: string, libraryPath: string): IFuture<void>;
```
rosen-vladimirov: mixed tabs/spaces maybe

lib/services/android-project-service.ts
```diff
@@ -118,19 +119,135 @@ class AndroidProjectService implements IPlatformProjectService {
...
+    }
 
 	public buildProject(projectRoot: string): IFuture<void> {
-		return (() => {
+        return (() => {
 			var buildConfiguration = options.release ? "release" : "debug";
-			var args = this.getAntArgs(buildConfiguration, projectRoot);
-			this.spawn('ant', args).wait();
+            var args = this.getAntArgs(buildConfiguration, projectRoot);
+            var args2 = this.getAntArgs(buildConfiguration, projectRoot);
```
rosen-vladimirov: why do we need same args saved in another variable?

lib/services/android-project-service.ts
```diff
@@ -118,19 +119,135 @@ class AndroidProjectService implements IPlatformProjectService {
...
+
+        var projProp = path.join(projDir, "project.properties");
+
+        if (!this.$fs.exists(projProp).wait()) {
+            this.$logger.warn("File %s does not exist", projProp);
+            return;
+        }
+
+        var lines = fs.readFileSync(projProp, { encoding: "utf-8" }).split("\n");
+        var thiz = this;
```
rosen-vladimirov: is this needed?

lib/services/android-project-service.ts
```diff
@@ -118,19 +119,135 @@ class AndroidProjectService implements IPlatformProjectService {
...
+        // TODO: parametrize targetSdk
+        var targetSdk = "android-17";
+        this.generateBuildFile(targetDir, targetSdk);
+    }
+
+    private getProjectReferences(projDir: string): ILibRef[]{
+        var projProp = path.join(projDir, "project.properties");
+
+        var lines = fs.readFileSync(projProp, { encoding: "utf-8" }).split("\n");
+        var thiz = this;
```
rosen-vladimirov: is this needed?

lib/services/android-project-service.ts
```diff
@@ -118,19 +119,135 @@ class AndroidProjectService implements IPlatformProjectService {
...
+    private parseProjectProrperies(projDir: string, destDir: string): void {
+
+        var projProp = path.join(projDir, "project.properties");
+
+        if (!this.$fs.exists(projProp).wait()) {
+            this.$logger.warn("File %s does not exist", projProp);
+            return;
+        }
+
+        var lines = fs.readFileSync(projProp, { encoding: "utf-8" }).split("\n");
```
rosen-vladimirov: maybe we should split project properties by os.EOL as if the file is created on Windows, we'll end up with each line finished with \r

rosen-vladimirov: but this is valid only in case where projProp file is created on customer's machine. In case we create this file on Linux/Mac machine, disregard my comment

lib/services/android-project-service.ts
slavchev: Good question. It turned `args` is mutable and it is updated in the program flow.

lib/services/android-project-service.ts
```diff
@@ -118,19 +119,135 @@ class AndroidProjectService implements IPlatformProjectService {
...
-	}
+    }
+
+    private generateBuildFile(projDir: string, targetSdk: string): void {
+        this.$logger.info("Generate build.xml for %s", projDir);
+        var cmd = util.format("android update project -p %s --target %s --subprojects", projDir, targetSdk);
+        this.$childProcess.exec(cmd).wait();
+    }
+
+    private parseProjectProrperies(projDir: string, destDir: string): void {
```
rosen-vladimirov: parseProjectProperties maybe?

lib/services/android-project-service.ts
slavchev: I am not sure what you are asking.

rosen-vladimirov: why do we need "thiz" variable? I couldn't found a place where it is used

lib/services/android-project-service.ts
```diff
@@ -118,19 +119,135 @@ class AndroidProjectService implements IPlatformProjectService {
...
+        return refs;
+    }
+
+    private updateProjectReferences(projDir: string, libraryPath: string): void {
+        var refs = this.getProjectReferences(projDir);
+        var maxIdx = refs.length > 0 ? _.max(refs, r => r.idx).idx : 0
+
+        var relLibDir = path.relative(projDir, libraryPath).split("\\").join("/");
+
+        var libRefExists = _.filter(refs, r => path.normalize(r.path) == path.normalize(relLibDir)).length > 0;
```
rosen-vladimirov:
```Typescript
var libRefExists = _.any(refs, r => path.normalize(r.path) === path.normalize(relLibDir));
```
maybe?

lib/services/android-project-service.ts
slavchev: splitting by EOL fixed with https://github.com/NativeScript/nativescript-cli/commit/326f5ecf4e02404b1a397e46394995a30be18268. You should adhere to your own standards ;)

lib/services/android-project-service.ts
slavchev: 10x, fixed with https://github.com/fealebenpae/nativescript-cli/commit/2e2f1b33cf2e1c98c2461dd3d5870314d1b7a10d

lib/services/android-project-service.ts
slavchev: a leftover, fixed with https://github.com/fealebenpae/nativescript-cli/commit/a1da6ce7871326b779c5cc7d74204ce0400fe70b

lib/services/android-project-service.ts
slavchev: fixed with https://github.com/fealebenpae/nativescript-cli/commit/c978d6148bc0b808d1cfb6418964ffa4274131c1

lib/definitions/platform.d.ts
slavchev: fixed with https://github.com/fealebenpae/nativescript-cli/commit/a76438269b6f87173f4fdfd275eca29296ee4e64

lib/services/android-project-service.ts
```diff
@@ -118,19 +119,135 @@ class AndroidProjectService implements IPlatformProjectService {
...
+
+        shell.cp("-f", path.join(libraryPath, "*.jar"), targetPath);
+        var projectLibsDir = path.join(platformData.projectRoot, "libs");
+        this.$fs.ensureDirectoryExists(projectLibsDir).wait();
+        shell.cp("-f", path.join(libraryPath, "*.jar"), projectLibsDir);
+
+        var targetLibPath = path.join(targetPath, path.basename(libraryPath));
+
+        this.updateProjectReferences(platformData.projectRoot, targetLibPath);
+        return Future.fromResult();
```
rosen-vladimirov: instead of returning ```Future.fromResult()``` it's better to wrap whole body of the method inside 
```Typescript
return (() => { /*body here*/ }).future<void>()();
```

lib/services/ios-project-service.ts
```diff
@@ -169,6 +171,89 @@ class IOSProjectService implements  IPlatformProjectService {
...
+
+    private validateDynamicFramework(libraryPath: string): IFuture<void> {
+        return (() => {
+            var infoPlistPath = path.join(libraryPath, "Info.plist");
+            if (!this.$fs.exists(infoPlistPath).wait()) {
+                this.$errors.failWithoutHelp("The bundle at %s does not contain an Info.plist file.", libraryPath);
+            }
+
+            var packageType = this.$childProcess.exec(util.format("/usr/libexec/PlistBuddy -c \"Print :CFBundlePackageType\" %s", infoPlistPath)).wait().trim();
+            if (packageType != "FMWK") {
```
rosen-vladimirov: !==

lib/services/android-project-service.ts
slavchev: fixed with https://github.com/fealebenpae/nativescript-cli/commit/fef99b33c26e4024804fdaa2d393dfd69571d0b2

lib/services/ios-project-service.ts
slavchev: fixed with https://github.com/NativeScript/nativescript-cli/commit/0533482e52e796fe44c3c991e0e2f9cccd6bc413

lib/commands/add-library.ts
```diff
@@ -0,0 +1,28 @@
...
+    allowedParameters: ICommandParameter[] = [];
+
+    execute(args: string[]): IFuture<void> {
+        var platform = args[0];
+        var libraryPath = path.resolve(args[1]);
+        return this.$platformService.addLibrary(platform, libraryPath);
+    }
+
+    canExecute(args: string[]): IFuture<boolean> {
+        if (args.length != 2) {
```
Fatme: `!==`

lib/definitions/project.d.ts
```diff
@@ -29,4 +29,5 @@ interface IPlatformProjectService {
 	prepareProject(platformData: IPlatformData): IFuture<string>;
 	buildProject(projectRoot: string): IFuture<void>;
 	isPlatformPrepared(projectRoot: string): IFuture<boolean>;
+    addLibrary(platformData: IPlatformData, libraryPath: string): IFuture<void>;
```
Fatme: mixed spaces and tabs

lib/commands/add-library.ts
slavchev: fixed with https://github.com/fealebenpae/nativescript-cli/commit/3a61db44116be17d37548b7438ba82575d5fcf0d

lib/services/android-project-service.ts
```diff
@@ -118,19 +120,135 @@ class AndroidProjectService implements IPlatformProjectService {
...
+
+        var lines = fs.readFileSync(projProp, { encoding: "utf-8" }).split(os.EOL);
+        var thiz = this;
+
+        lines.forEach((elem, idx, arr) => {
+            var match = elem.match(/android\.library\.reference\.(\d+)=(.*)/);
+            if (match) {
+                var libRef: ILibRef = { idx: parseInt(match[1]), path: match[2].trim() };
+                libRef.adjustedPath = thiz.$fs.isRelativePath(libRef.path) ? path.join(projDir, libRef.path) : libRef.path;
+                thiz.parseProjectProperties(libRef.adjustedPath, destDir);
```
Fatme: `this.parseProjectProperties`

lib/definitions/project.d.ts
slavchev: fixed with https://github.com/fealebenpae/nativescript-cli/commit/09b76ea9240859cd8abcfb2e4c4d0b12b4ea0e8c

lib/services/android-project-service.ts
```diff
@@ -118,19 +120,135 @@ class AndroidProjectService implements IPlatformProjectService {
...
+        var targetDir = path.join(destDir, path.basename(projDir));
+        // TODO: parametrize targetSdk
+        var targetSdk = "android-17";
+        this.generateBuildFile(targetDir, targetSdk);
+    }
+
+    private getProjectReferences(projDir: string): ILibRef[]{
+        var projProp = path.join(projDir, "project.properties");
+
+        var lines = fs.readFileSync(projProp, { encoding: "utf-8" }).split("\n");
```
rosen-vladimirov: ```split(os.EOL)``` maybe

lib/services/android-project-service.ts
slavchev: fixed with https://github.com/fealebenpae/nativescript-cli/commit/09b76ea9240859cd8abcfb2e4c4d0b12b4ea0e8c

lib/services/android-project-service.ts
```diff
@@ -118,19 +120,135 @@ class AndroidProjectService implements IPlatformProjectService {
...
+                thiz.parseProjectProperties(libRef.adjustedPath, destDir);
+            }
+        });
+
+        this.$logger.info("Copying %s", projDir);
+        shell.cp("-Rf", projDir, destDir);
+
+        var targetDir = path.join(destDir, path.basename(projDir));
+        // TODO: parametrize targetSdk
+        var targetSdk = "android-17";
```
Fatme: We always generate build file using `android-17`. I was wondering if we shouldn't get the latest installed android sdk and generate build file using it.

lib/services/android-project-service.ts
slavchev: fixed with https://github.com/fealebenpae/nativescript-cli/commit/9d10b30b6960b14af1133f58fb41822afec6fd39

lib/services/android-project-service.ts
slavchev: We are considering this change as well, though we are not sure if it is a good idea since we mainly test against api level 17.

lib/services/android-project-service.ts
```diff
@@ -118,19 +120,134 @@ class AndroidProjectService implements IPlatformProjectService {
...
 	public buildProject(projectRoot: string): IFuture<void> {
-		return (() => {
+        return (() => {
 			var buildConfiguration = options.release ? "release" : "debug";
-			var args = this.getAntArgs(buildConfiguration, projectRoot);
-			this.spawn('ant', args).wait();
+            var args = this.getAntArgs(buildConfiguration, projectRoot);
+            var args2 = this.getAntArgs(buildConfiguration, projectRoot);
+            this.spawn('ant', args).wait();
+            this.generateMetadata(projectRoot);
```
Fatme: Do we need to generate metadata on every build? Shouldn't do it only when `library add` command is executed?

lib/services/android-project-service.ts
```diff
@@ -118,19 +120,134 @@ class AndroidProjectService implements IPlatformProjectService {
...
+    private parseProjectProperties(projDir: string, destDir: string): void {
+
+        var projProp = path.join(projDir, "project.properties");
+
+        if (!this.$fs.exists(projProp).wait()) {
+            this.$logger.warn("File %s does not exist", projProp);
+            return;
+        }
+
+        var lines = fs.readFileSync(projProp, { encoding: "utf-8" }).split(os.EOL);
```
Fatme: Shouldn't we use here `this.$fs` instead `fs`?

lib/commands/add-library.ts
teobugslayer: It depends how do you plan on using the API. It is available both as a constructor (i.e. new Future()) and as a function (i.e. future()). Import it the way you plan to use it. See https://github.com/laverdet/node-fibers#api-documentation

lib/services/android-project-service.ts
slavchev: Yes, `tns library add` does not compile the library being added. Since the metadata generator takes *.jar file as input we need to do on `build` step. It takes less than 2 seconds on average. We consider other options as well.

lib/definitions/xcode.d.ts
```diff
@@ -0,0 +1,23 @@
+declare module "xcode" {
+    interface FrameworkOptions {
+        [key: string]: any;
```
teobugslayer: This is IDictionary\<any\>, you can extend it if you find it more appealing.

lib/services/android-project-service.ts
```diff
@@ -18,7 +20,7 @@ class AndroidProjectService implements IPlatformProjectService {
 		private $errors: IErrors,
 		private $fs: IFileSystem,
 		private $logger: ILogger,
-		private $projectData: IProjectData,
+        private $projectData: IProjectData,
```
teobugslayer: Please revert this change

lib/services/android-project-service.ts
slavchev: No, because `IFileSystem` does not expose `readFileSync` method.

rosen-vladimirov: You can use:
```Typesript
this.$fs.readText(projProp, "utf-8").wait().split(os.EOL);
```

lib/services/android-project-service.ts
```diff
@@ -118,19 +120,134 @@ class AndroidProjectService implements IPlatformProjectService {
 			return assetsDirectory;
 
 		}).future<string>()();
-	}
+    }
+
+    private updateMetadata(projectRoot: string): void {
+        var projMetadataDir = path.join(projectRoot, "assets", "metadata");
+        var libsmetadataDir = path.join(projectRoot, "../../lib", this.platformData.normalizedPlatformName, "__metadata");
```
teobugslayer: you can extract __metadata as a constant (due to typescript limitations, a private static member) because it is shared among methods in the class.

lib/services/android-project-service.ts
rosen-vladimirov: But in case you do so, you have to modify parseProjectProperties to return IFuture<void> and wrap the body inside ```return (() => { /*body*/ }).future<void>()();

and add ```.wait()``` where you are calling parseProjectProperties

lib/services/android-project-service.ts
slavchev: fixed with https://github.com/fealebenpae/nativescript-cli/commit/3e3e69c94a9d7954d27668bb8ca7444ec37c3c4e

lib/services/android-project-service.ts
```diff
@@ -118,19 +120,134 @@ class AndroidProjectService implements IPlatformProjectService {
...
+        var metadataDirName = "__metadata";
+        var inputDir = libsFolder;
+        var outDir = path.join(libsFolder, metadataDirName);
+        this.$fs.ensureDirectoryExists(outDir).wait();
+
+        shell.cp("-f", path.join(__dirname, "../../resources/tools/android.jar"), inputDir);
+        shell.cp("-f", path.join(__dirname, "../../resources/tools/android-support-v4.jar"), inputDir);
+        shell.cp("-f", path.join(projectRoot, "libs/*.jar"), inputDir);
+
+        this.spawn('java', ['-jar', metadataGeneratorPath, inputDir, outDir]).wait();
```
teobugslayer: Is it possible that 'java' is not in the path? generally, when we call external binaries, we have a check if they are runnable. Ex, adb, geny player, etc. We can do this for java as well. Of course, if android SDK depends on java in the path, this won't be necessary.

lib/services/android-project-service.ts
```diff
@@ -118,19 +120,134 @@ class AndroidProjectService implements IPlatformProjectService {
...
+    }
 
 	public buildProject(projectRoot: string): IFuture<void> {
-		return (() => {
+        return (() => {
```
teobugslayer: Please revert this line

lib/services/android-project-service.ts
teobugslayer: Then perhaps a more descriptive name? 'argsSaved'?

lib/services/android-project-service.ts
```diff
@@ -118,19 +120,134 @@ class AndroidProjectService implements IPlatformProjectService {
...
 			var buildConfiguration = options.release ? "release" : "debug";
-			var args = this.getAntArgs(buildConfiguration, projectRoot);
-			this.spawn('ant', args).wait();
+            var args = this.getAntArgs(buildConfiguration, projectRoot);
+            var args2 = this.getAntArgs(buildConfiguration, projectRoot);
+            this.spawn('ant', args).wait();
```
teobugslayer: Is it possible that 'ant' is not in the path? generally, when we call external binaries, we have a check if they are runnable. Ex, adb, geny player, etc. We can do this for java as well. Of course, if android SDK depends on ant in the path, this won't be necessary.

lib/services/android-project-service.ts
```diff
@@ -118,19 +120,134 @@ class AndroidProjectService implements IPlatformProjectService {
...
+        lines.forEach((elem, idx, arr) => {
+            var match = elem.match(/android\.library\.reference\.(\d+)=(.*)/);
+            if (match) {
+                var libRef: ILibRef = { idx: parseInt(match[1]), path: match[2] };
+                libRef.adjustedPath = path.join(projDir, libRef.path);
+                refs.push(libRef);
+            }
+        });
+
+        return refs;
```
Fatme: Just a suggestion: `return _.filter(lines, line => line.match(/android\.library\.reference\.(\d+)=(.*)/)).map(line => {
idx: parseInt(match[1]), 
path: match[2],
adjustedPath: path.join(projDir, libRef.path)
}).value()` 
The didn't test the code!

lib/services/android-project-service.ts
rosen-vladimirov: we have these checks when validating the platform, so this should be safe
this.checkAnt().wait() && this.checkAndroid().wait() && this.checkJava().wait();

lib/services/android-project-service.ts
rosen-vladimirov: we have these checks when validating the platform, so this should be safe
this.checkAnt().wait() && this.checkAndroid().wait() && this.checkJava().wait();

lib/services/android-project-service.ts
slavchev: fixed as suggested https://github.com/fealebenpae/nativescript-cli/commit/5d84f49015bdf806f18909eb48870a65397b6c76

lib/services/android-project-service.ts
```diff
@@ -118,19 +120,134 @@ class AndroidProjectService implements IPlatformProjectService {
...
+    }
+
+    private generateBuildFile(projDir: string, targetSdk: string): void {
+        this.$logger.info("Generate build.xml for %s", projDir);
+        var cmd = util.format("android update project -p %s --target %s --subprojects", projDir, targetSdk);
+        this.$childProcess.exec(cmd).wait();
+    }
+
+    private parseProjectProperties(projDir: string, destDir: string): void {
+
```
teobugslayer: unneeded empty line

lib/services/android-project-service.ts
slavchev: fixed with https://github.com/fealebenpae/nativescript-cli/commit/5d84f49015bdf806f18909eb48870a65397b6c76

lib/services/android-project-service.ts
Fatme: Build function is called when for example `deploy` command is executed(or `emulate`, or `build`). Do we need to generate metadata when deploy is executed?

lib/services/android-project-service.ts
```diff
@@ -118,19 +121,134 @@ class AndroidProjectService implements IPlatformProjectService {
...
+        var projProp = path.join(projDir, "project.properties");
+
+        if (!this.$fs.exists(projProp).wait()) {
+            this.$logger.warn("File %s does not exist", projProp);
+            return;
+        }
+
+		var lines = this.$fs.readText(projProp, "utf-8").wait().split(os.EOL);
+
+        lines.forEach((elem, idx, arr) => {
```
teobugslayer: idx and arr are not used

lib/services/android-project-service.ts
```diff
@@ -118,19 +121,134 @@ class AndroidProjectService implements IPlatformProjectService {
...
+        var targetDir = path.join(destDir, path.basename(projDir));
+        // TODO: parametrize targetSdk
+        var targetSdk = "android-17";
+        this.generateBuildFile(targetDir, targetSdk);
+    }
+
+    private getProjectReferences(projDir: string): ILibRef[]{
+        var projProp = path.join(projDir, "project.properties");
+
+        var lines = fs.readFileSync(projProp, { encoding: "utf-8" }).split(os.EOL);
```
rosen-vladimirov: var lines = this.$fs.readText(projProp, "utf-8").wait().split(os.EOL);

lib/services/android-project-service.ts
```diff
@@ -8,9 +8,12 @@ import options = require("../common/options");
 import constants = require("../constants");
 import hostInfo = require("../common/host-info");
 import helpers = require("../common/helpers");
+import fs = require("fs");
```
rosen-vladimirov: remove this require, after using readText everywhere :)

lib/services/android-project-service.ts
```diff
@@ -118,19 +121,134 @@ class AndroidProjectService implements IPlatformProjectService {
...
+
+        if (!this.$fs.exists(projProp).wait()) {
+            this.$logger.warn("File %s does not exist", projProp);
+            return;
+        }
+
+		var lines = this.$fs.readText(projProp, "utf-8").wait().split(os.EOL);
+
+        lines.forEach((elem, idx, arr) => {
+            var match = elem.match(/android\.library\.reference\.(\d+)=(.*)/);
```
teobugslayer: Can we move the regexp out of the loop? afaik, it is recreated on each iteration.

lib/services/android-project-service.ts
slavchev: fixed with https://github.com/fealebenpae/nativescript-cli/commit/5630cdece66840f2f67bef5622a3d409ddf88814

lib/services/android-project-service.ts
```diff
@@ -118,19 +121,134 @@ class AndroidProjectService implements IPlatformProjectService {
...
+        var projMetadataDir = path.join(projectRoot, "assets", "metadata");
+        var libsmetadataDir = path.join(projectRoot, "../../lib", this.platformData.normalizedPlatformName, AndroidProjectService.METADATA_DIRNAME);
+        shell.cp("-f", path.join(libsmetadataDir, "*.dat"), projMetadataDir);
+    }
+
+    private generateMetadata(projectRoot: string): void {
+        var metadataGeneratorPath = path.join(__dirname, "../../resources/tools/metadata-generator.jar");
+        var libsFolder = path.join(projectRoot, "../../lib", this.platformData.normalizedPlatformName);
+		var metadataDirName = AndroidProjectService.METADATA_DIRNAME;
+        var inputDir = libsFolder;
```
Fatme: Do we need `inputDir` variable? Can we use just `libsFolder`?

lib/services/android-project-service.ts
teobugslayer: utf-8 is the default encoding, you can nuke it.

lib/services/android-project-service.ts
```diff
@@ -118,19 +121,134 @@ class AndroidProjectService implements IPlatformProjectService {
...
+    }
+
+    private getProjectReferences(projDir: string): ILibRef[]{
+        var projProp = path.join(projDir, "project.properties");
+
+        var lines = fs.readFileSync(projProp, { encoding: "utf-8" }).split(os.EOL);
+
+        var refs: ILibRef[] = [];
+
+        lines.forEach((elem, idx, arr) => {
```
teobugslayer: idx and arr are not used, please remove them

lib/services/android-project-service.ts
```diff
@@ -118,19 +121,134 @@ class AndroidProjectService implements IPlatformProjectService {
...
+            var match = elem.match(/android\.library\.reference\.(\d+)=(.*)/);
```
teobugslayer: Gain, consider moving the regexp out of scope

lib/services/android-project-service.ts
slavchev: rename variable `args2` to `argsSaved` as suggested https://github.com/fealebenpae/nativescript-cli/commit/ff1273404ad1225163fde0d4982f075b363cfe0d

lib/services/android-project-service.ts
slavchev: fixed with https://github.com/fealebenpae/nativescript-cli/commit/bd9c46e660ac615fd97c0eec201bd1d62ba79131

lib/services/android-project-service.ts
Fatme: Can we use `propertiesParser` here? https://github.com/NativeScript/nativescript-cli/blob/65b49ae86c0a4d0be76d11706d1c835a6362acee/lib/services/android-project-service.ts#L267

lib/services/android-project-service.ts
slavchev: Yes, for the same reasons. It guarantees the the deployed app contains the latest metadata.

lib/services/android-project-service.ts
slavchev: any suggestions are welcome :)

lib/services/ios-project-service.ts
```diff
@@ -169,6 +171,89 @@ class IOSProjectService implements  IPlatformProjectService {
...
+                this.$errors.failWithoutHelp("The bundle at %s does not appear to be a dynamic framework.", libraryPath);
+            }
+        }).future<void>()();
+    }
+
+    private getUmbrellaHeaderFromDynamicFramework(libraryPath: string): IFuture<string> {
+        return (() => {
+            var modulemapPath = path.join(libraryPath, "Modules", "module.modulemap");
+            if (!this.$fs.exists(modulemapPath).wait()) {
+                this.$errors.failWithoutHelp("The framework at %s does not contain a module.modulemap file.");
```
Fatme: We have `%s` but don't have a value for it.

lib/services/android-project-service.ts
slavchev: fixed as suggested https://github.com/fealebenpae/nativescript-cli/commit/906aa01a69c58151c570704fdd644235b5cfb67c

lib/services/android-project-service.ts
slavchev: `fs` is still used for `fs.appendFileSync`. I checked `IFileSystem` and I does not offer method for appending. Any suggestions?

lib/services/ios-project-service.ts
```diff
@@ -169,6 +171,89 @@ class IOSProjectService implements  IPlatformProjectService {
...
+            this.$fs.createDirectory(yamlOut).wait();
+
+            var tempHeader = path.join(yamlOut, "Metadata.h");
+            this.$fs.writeFile(tempHeader, util.format("#import <%s/%s>", frameworkName, umbrellaHeader)).wait();
+
+            this.$logger.info("Generating metadata for %s.framework. This can take a minute.", frameworkName);
+            var sdkPath = this.$childProcess.exec("xcrun -sdk iphoneos --show-sdk-path").wait().trim();
+            var generatorExecOptions = {
+                env: {
+                    DYLD_FALLBACK_LIBRARY_PATH: this.$childProcess.exec("xcode-select -p").wait().trim() + "/Toolchains/XcodeDefault.xctoolchain/usr/lib"
```
Fatme: This is a little bit tricky. Can we add a comment why we need this options.

lib/services/android-project-service.ts
slavchev: fixed with https://github.com/fealebenpae/nativescript-cli/commit/98e5598bf16746f23d4c063d8ac5f9ed2cb1f6cd

lib/services/android-project-service.ts
teobugslayer: Or, (again untested, because there should be some fun left for you too):

```
return _(lines)
		.map(elem => 
    		_.each(elem.match(/android\.library\.reference\.(\d+)=(.*)/), match => { 
	        	idx: parseInt(match[1]), 
	        	path: match[2],
	        	adjustedPath: path.join(projDir, libRef.path)
	        })
)
.filter(val => !!val)
.value();
```

lib/services/android-project-service.ts
slavchev: Do you write program for the compiler/computer or for the humans? As suggested https://github.com/fealebenpae/nativescript-cli/commit/e8aac716ef5a8e48a0b69a6cd18235113b4967b5

lib/services/android-project-service.ts
slavchev: Explicit encoding won't harm.

lib/services/android-project-service.ts
slavchev: as suggested https://github.com/fealebenpae/nativescript-cli/commit/1e70c55ec1253b836330eceb4826cb3763ad5745

lib/services/android-project-service.ts
slavchev: extract regex variable as suggested https://github.com/fealebenpae/nativescript-cli/commit/2c8b6950b652d778a2c0005862b269055bdf98f9

slavchev: Good question. Apparently not because we need the last/max ref index.

lib/services/ios-project-service.ts
slavchev: Good catch. Fixed with https://github.com/fealebenpae/nativescript-cli/commit/eb1d28e5264c111cf5e84899d7e7e826c3e6a565

lib/services/android-project-service.ts
teobugslayer: @fatme, your code can call result of calling ```null.map``` because match can return null. Imho, this is one of the cases, where the procedural code is shorter than the functional one. I would only get rid of the libref.adjustedPath line and move it in the returned object itself:

```
var refs: ILibRef[] = [];
lines.forEach(elem => {
	var match = elem.match(/android\.library\.reference\.(\d+)=(.*)/);
	if (match) {
            var libRef: ILibRef = { idx: parseInt(match[1]), path: match[2], adjustedPath: path.join(projDir, libRef.path) };
            refs.push(libRef);
        }
    });
return refs;
```

lib/services/android-project-service.ts
```diff
@@ -118,19 +121,134 @@ class AndroidProjectService implements IPlatformProjectService {
...
+                refs.push(libRef);
+            }
+        });
+
+        return refs;
+    }
+
+    private updateProjectReferences(projDir: string, libraryPath: string): void {
+        var refs = this.getProjectReferences(projDir);
+        var maxIdx = refs.length > 0 ? _.max(refs, r => r.idx).idx : 0
```
teobugslayer: please add ;

lib/services/android-project-service.ts
```diff
@@ -118,19 +121,134 @@ class AndroidProjectService implements IPlatformProjectService {
...
+
+        var relLibDir = path.relative(projDir, libraryPath).split("\\").join("/");
+
+        var libRefExists = _.any(refs, r => path.normalize(r.path) === path.normalize(relLibDir));
+
+        if (!libRefExists) {
+            var projRef = util.format("\nandroid.library.reference.%d=%s", maxIdx + 1, relLibDir);
```
rosen-vladimirov:
```Typescript
var projRef = util.format("%sandroid.library.reference.%d=%s", os.EOL, maxIdx + 1, relLibDir);
```
We should append files with current os End-of-file symbols I think

lib/services/android-project-service.ts
```diff
@@ -118,19 +121,134 @@ class AndroidProjectService implements IPlatformProjectService {
...
+        });
+
+        return refs;
+    }
+
+    private updateProjectReferences(projDir: string, libraryPath: string): void {
+        var refs = this.getProjectReferences(projDir);
+        var maxIdx = refs.length > 0 ? _.max(refs, r => r.idx).idx : 0
+
+        var relLibDir = path.relative(projDir, libraryPath).split("\\").join("/");
```
teobugslayer: Are you sure \\ is valid on all OSes? why not using os.pathSeparator?

lib/services/android-project-service.ts
slavchev: fixed as suggested https://github.com/fealebenpae/nativescript-cli/commit/a179dd7e01fe7b871f88e06de941f7bd8cf8fbfa

lib/services/android-project-service.ts
slavchev: fixed with https://github.com/fealebenpae/nativescript-cli/commit/b02c6d7f8772f9c27826778ab1b423be02edb421

lib/services/android-project-service.ts
slavchev: Actually the code replaces `\` with `/` which is the point.

slavchev: Actually the code replace `\` with `/` which is the point.

lib/services/android-project-service.ts
slavchev: Please bare with mere mortals.

lib/services/android-project-service.ts
rosen-vladimirov: Well, we can implement appendFile in our FileSystem

lib/services/ios-project-service.ts
```diff
@@ -169,6 +171,89 @@ class IOSProjectService implements  IPlatformProjectService {
...
+
+            var frameworkName = path.basename(libraryPath, path.extname(libraryPath));
+            var targetPath = path.join(this.$projectData.projectDir, "lib", platformData.normalizedPlatformName, frameworkName);
+            this.$fs.ensureDirectoryExists(targetPath).wait();
+            shell.cp("-R", libraryPath, targetPath);
+
+            this.generateFrameworkMetadata(platformData.projectRoot, targetPath, frameworkName, umbrellaHeader).wait();
+
+            var pbxProjPath = path.join(platformData.projectRoot, this.$projectData.projectName + ".xcodeproj", "project.pbxproj");
+            var project = new xcode.project(pbxProjPath);
```
teobugslayer: newable functions should be named with uppercase. Can we do it here?

lib/services/ios-project-service.ts
```diff
@@ -169,6 +171,89 @@ class IOSProjectService implements  IPlatformProjectService {
...
+    }
+
+    private validateDynamicFramework(libraryPath: string): IFuture<void> {
+        return (() => {
+            var infoPlistPath = path.join(libraryPath, "Info.plist");
+            if (!this.$fs.exists(infoPlistPath).wait()) {
+                this.$errors.failWithoutHelp("The bundle at %s does not contain an Info.plist file.", libraryPath);
+            }
+
+            var packageType = this.$childProcess.exec(util.format("/usr/libexec/PlistBuddy -c \"Print :CFBundlePackageType\" %s", infoPlistPath)).wait().trim();
```
teobugslayer: You may omit util.format in this case.

lib/services/android-project-service.ts
rosen-vladimirov: It should be something like:
```Typescript
	public appendFile(filename: string, data: any, encoding?: string): IFuture<void> {
		return (() => {
			this.createDirectory(path.dirname(filename)).wait();
			var future = new Future<void>();
			fs.appendFile(filename, data, { encoding: encoding },(err: Error) => {
				if(err) {
					future.throw(err);
				} else {
					future.return();
				}
			});
			future.wait();
		}).future<void>()();
	}
```

lib/services/ios-project-service.ts
```diff
@@ -169,6 +171,89 @@ class IOSProjectService implements  IPlatformProjectService {
...
+            this.$fs.writeFile(tempHeader, util.format("#import <%s/%s>", frameworkName, umbrellaHeader)).wait();
+
+            this.$logger.info("Generating metadata for %s.framework. This can take a minute.", frameworkName);
+            var sdkPath = this.$childProcess.exec("xcrun -sdk iphoneos --show-sdk-path").wait().trim();
+            var generatorExecOptions = {
+                env: {
+                    DYLD_FALLBACK_LIBRARY_PATH: this.$childProcess.exec("xcode-select -p").wait().trim() + "/Toolchains/XcodeDefault.xctoolchain/usr/lib"
+                }
+            };
+            this.$childProcess.exec(util.format("%s/Metadata/MetadataGenerator -s %s -u %s -o %s -cflags=\"-F%s\"", projectRoot, sdkPath, tempHeader, yamlOut, frameworkDir), generatorExecOptions).wait();
```
teobugslayer: You can use single quotes to avoid escaping

lib/services/ios-project-service.ts
```diff
@@ -169,6 +171,89 @@ class IOSProjectService implements  IPlatformProjectService {
...
+
+            this.$fs.copyFile(path.join(yamlOut, "Metadata-armv7", frameworkName + ".yaml"), path.join(projectRoot, "Metadata", "Metadata-armv7", frameworkName + ".yaml")).wait();
+            this.$fs.copyFile(path.join(yamlOut, "Metadata-arm64", frameworkName + ".yaml"), path.join(projectRoot, "Metadata", "Metadata-arm64", frameworkName + ".yaml")).wait();
+
```
teobugslayer: consider extracting common paths like frameworkName + ".yaml" for easier reading.

lib/services/android-project-service.ts
slavchev: Thank you for the code. I will consider it for future improvements.

lib/services/ios-project-service.ts
fealebenpae: Blame the xcode module, then. I can't change their API.

lib/services/ios-project-service.ts
slavchev: fixed with https://github.com/fealebenpae/nativescript-cli/commit/c502ab0c94de0626680c25bfb3d67a1406fe987a

lib/services/ios-project-service.ts
slavchev: as suggested https://github.com/fealebenpae/nativescript-cli/commit/93e35d5e7b77cd04934d6099a1d33724ce783570

lib/services/ios-project-service.ts
slavchev: see https://github.com/NativeScript/nativescript-cli/commit/cea494b456ee5146acf7df41a4215c1cfbb4daca

lib/definitions/osenv.d.ts
slavchev: mistake, fixed with https://github.com/fealebenpae/nativescript-cli/commit/bffb555a508b1b7d10d1f51b9aec77ceb7303cc2

lib/services/ios-project-service.ts
slavchev: fixed with https://github.com/fealebenpae/nativescript-cli/commit/a8ac0fe1720027554c08ecc16a72bc7ec72d0255

package.json
```diff
@@ -58,7 +58,8 @@
     "unzip": "0.1.9",
     "xmlhttprequest": "https://github.com/telerik/node-XMLHttpRequest/tarball/master",
     "yargs": "1.2.2",
-    "node-inspector":  "0.7.4"
+    "node-inspector":  "0.7.4",
+    "xcode": "https://github.com/fealebenpae/node-xcode/tarball/master"
```
teobugslayer: We talked in person with @fealebenpae and decided to move this repo in the NS organization and depend against a github release created in the new repo.

lib/services/ios-project-service.ts
teobugslayer: It appears that we can use it without new: https://github.com/fealebenpae/node-xcode/blob/master/lib/pbxProject.js#L14

package.json
fealebenpae: And this is now done.

lib/services/android-project-service.ts
teobugslayer: I am stupid, sorry!

teobugslayer:
You need to rebase, otherwise, :+1:

Fatme:
:+1:

rosen-vladimirov:
:+1:

PR #239, opened 2015-02-24
* Update common lib so we can use debug command for iOS devices

Fatme:
* Move tns related code from common lib


teobugslayer:
```
hudson.plugins.git.GitException: Command "git submodule update --init --recursive" returned status code 1:
stdout: 
stderr: fatal: reference is not a tree: b04fabf16677fdd0b5e6dca8641d6dab9f642661
Unable to checkout 'b04fabf16677fdd0b5e6dca8641d6dab9f642661' in submodule path 'lib/common'
```

Issue #257, opened 2015-02-27
Arguments with whitespaces don't get passed right


Mitko-Kerezov:
For example
$tns create "My App" results in 
The parameter App is not valid for this command.

PR #258, opened 2015-02-27
Fix Bash passing unescaped arguments to node

Mitko-Kerezov:
This fixes https://github.com/NativeScript/nativescript-cli/issues/257


rosen-vladimirov:
:+1:

PR #264, opened 2015-02-28
Updated system requirements

ikoevska:


README.md
```diff
@@ -136,17 +137,16 @@ If you have installed Chocolatey, you can complete these steps to set up JDK, Ap
 > On OS X systems, you can develop, build, and deploy NativeScript projects that target iOS and Android.
 
 * OS X Mavericks
-* [Node.js 0.10.26] or a later stable official release
+* [Node.js 0.10.26][Node.js 0.10.26] or a later stable official release
 * For iOS development
-	* [iOS 7.0 SDK][12] or a later stable official release
-	* [Xcode 5][12] or a later stable official release
+	* [Xcode 6][12] or a later stable official release
 	* [Xcode command-line tools][12]
```
teobugslayer: If I remember correctly, only 64 bit Mono works. Someone from the NS team should confirm.

atanasovg: The 64-bit Mono is required in case users want to use the new library add command. We may include this as well.

ikoevska: This last comment is unclear. What may we include? the Mono req or that Mono is req-ed for library add or something about library add?

If the comments is about the Mono req being listed as a whole, please, take a look at line 144 - it's already there.

And finally, should it or shoildn't it be 64-bit?

atanasovg: Finally, it should be 64-bit; and it is required only when library add is used.

dtopuzov:
:100:

ikoevska:
Reopening as another PR https://github.com/NativeScript/nativescript-cli/pull/272

PR #267, opened 2015-03-04
Fix starting of genymotion emulator

rosen-vladimirov:
Use lodash's startsWith method as the extension method startsWith (it was part of extensions.ts) had been removed.


teobugslayer:
:+1:

Issue #268, opened 2015-03-04
OSX issue with node 0.12.0


emazv72:
I'm getting the following error when creating a new app:

> tns create TestApp

Error: "name" and "value" are required for setHeader().
    at ClientRequest.OutgoingMessage.setHeader (_http_outgoing.js:333:11)
    at new ClientRequest (_http_client.js:101:14)
    at Object.exports.request (http.js:49:10)
    at Object.exports.request (https.js:136:15)
    at Request.start (/usr/local/lib/node_modules/nativescript/node_modules/npm/node_modules/request/request.js:583:30)
    at Request.end (/usr/local/lib/node_modules/nativescript/node_modules/npm/node_modules/request/request.js:1237:28)
    at /usr/local/lib/node_modules/nativescript/node_modules/npm/node_modules/request/request.js:413:12
    at process._tickCallback (node.js:355:11)

rosen-vladimirov:
Hi @emazv72 , 
Thank you for reporting this problem. It is related to old npm version that we are using in our code. I've prepared a fix, but until it's published, you have to use node 0.10.x (0.10.36 for example). Please excuse us for the caused inconvenience.

dtopuzov:
Available in 0.9.1

PR #269, opened 2015-03-04
Update npm in order to support node 0.12

rosen-vladimirov:
Update npm to 2.6.1 and change two calls according to latest npm.cache changes. If we do not update npm, we cannot work with node 0.12
Update ios-sim-portable to 1.0.4

Fixes https://github.com/NativeScript/nativescript-cli/issues/268


Fatme:
cool! :+1:

dtopuzov:
:100:

PR #274, opened 2015-03-04
Updated sys req for Xcode to run/debug

ikoevska:


rosen-vladimirov:
:+1:  after green build

PR #284, opened 2015-03-06
updated metadata generator with proxygenerator support

blagoev:


rosen-vladimirov:
I'm not sure - what should we see here - what is the change, why it is required. I see only a changed .jar, but even if I take this branch locally, I'm still not sure what does it affect. Could you share some details?
Also it looks like a rebase is required (automatic merge is not possible at the moment).

rosen-vladimirov:
I'm closing this PR as the change is already part of master branch.

PR #295, opened 2015-03-11
Fix hierarchical commands

rosen-vladimirov:
Update common lib where a fix for hierarchical commands had been applied. Add help for new commands - '$ tns device android' and '$ tns device ios'


dtopuzov:
:100:

PR #296, opened 2015-03-11
Advance the common to adopt the ios-debug-brk notification

PanayotCankov:


rosen-vladimirov:
:+1:  after green builds

PR #298, opened 2015-03-11
Update d.ts definitions to node 0.12

teobugslayer:


rosen-vladimirov:
:+1:  after green build

Issue #293, opened 2015-03-10
Deploy and Run commands has the same behavior and different code implementations


dtopuzov:
It looks Deploy and Run both do the same:
1. Prepare
2. Build
3. Deploy
4. Run 

Both commands have different implementations.
We should make them aliases and go with one single code implementation 
Personally,  I think deploy implementation is better (at least it handles invalid input better)

dtopuzov:
Found a difference in iOS:
Run can run the app in simulator, Deploy can't deploy in simulator.

emiloberg:
To make things easier, wouldn't it be a good idea to remove one of the choices; run or deploy.

My vote is on removing the `run` as that would give us `deploy` for running on actual devices and 'emulate' for virtual devices. Clear separation.

rosen-vladimirov:
Hi @emiloberg ,
I'm wondering if we should duplicate the behavior or remove one of the commands. What do you think about making the deploy command just deploying to connected devices, without starting the application, while run will make deploy + starting of application. Emulate will make deploy on emulator + starting on emulator?

emiloberg:
You're proposal works - if you can see a use case for _just_ deploying an app without starting it. Personally I can't see why I want to do that, but there might be one.

Doing it this way still separates the commands for emulating and deploying (and running) which is I see as the most important part.

rosen-vladimirov:
As you've said - trying to find a use case for just deploying is really difficult. The only case I could think about is when you want to deploy several applications on different devices and check them after that.
In fact we have widely discussed command's behavior in [this issue](https://github.com/NativeScript/nativescript-cli/issues/475), so I'll close this one and we can continue our communication there. Thank you for the valuable feedback!

PR #301, opened 2015-03-12
Improve commands

Fatme:


rosen-vladimirov:
It looks like you have to rebase :P
Anyway, with these changes ```run``` command will build your project and start it on the device (deployOnDevice method now starts the app).
```deploy``` command uses directly deploy on device, so it will do exactly the same thing as ```run``` command for device.

Maybe this is the right time to fix this https://github.com/NativeScript/nativescript-cli/issues/293 ? What do you think - I suggest to remove deploy command and just make an alias of run commands for android and ios?

Fatme:
I totally agree with you and will implement it in this PR.

Issue #302, opened 2015-03-12
Does not work behind proxy


briangabriel:
I can install NativeScript via npm with no issues, but there is no way to configure the http-client with the proper proxy settings.

E:\LocalDev\nativescript>tns create hello-world --log trace
Starting watch on killswitch C:\Users\gabrielb\AppData\Local\Temp\gabrielb\KillSwitches\cli
Creating a new NativeScript project with name hello-world and id org.nativescript.helloworld at location E:\LocalDev\nativescript\hello-world
Using NativeScript hello world application
User-Agent: AppBuilderCLI/0.9.1 (Node.js 0.10.28; win32; x64)
httpRequest: { method: 'GET',
  host: 'registry.npmjs.org',
  port: null,
  path: '/tns-template-hello-world',
  headers:
   { Accept: 'application/json; charset=UTF-8, */*;q=0.8',
     'User-Agent': 'AppBuilderCLI/0.9.1 (Node.js 0.10.28; win32; x64)',
     'Accept-Encoding': 'gzip,deflate' } }
httpRequest: Sending:
[nothing]
{ [Error: connect ETIMEDOUT] stack: [Getter] }
Error: connect ETIMEDOUT

rosen-vladimirov:
Hi @briangabriel 
It looks like there's a problem with the http-proxy. Could you try the suggestion from [this blog post](http://wil.boayue.com/blog/2013/06/14/using-npm-behind-a-proxy/): 
```npm config set proxy http://proxy.company.com:8080``` for example
Hopefully this will solve the issue.

briangabriel:
I have no problems with npm, the issue is that the `NodePackageManager.getAvailableVersions` method calls the npm registry via a straight up HTTP request:

    var url = NodePackageManager.NPM_REGISTRY_URL + packageName;
    var response = this.$httpClient.httpRequest(url).wait().body;

rosen-vladimirov:
Hi @briangabriel ,

You are absolutely correct. We are working on adding support for proxy, but meanwhile, can you try one more thing:
Can you try installing [this custom package](https://www.dropbox.com/s/wrn77fgmmdvxg6c/nativescript-0.9.1.tgz?dl=0) of nativescript-cli. Just download the file and execute:
```$ npm i nativescript-0.9.1.tgz -g```
[...]

Issue #297, opened 2015-03-11
npm install infinite build


Kyozatch:
Hi,

I'm trying to install nativescript but each time (whatever the nodejs version) it fails because i never pass this line (even after 20min..). I currently use nodejs 0.10.36.

"
npm WARN deprecated byline@4.1.1: critical bug fixed in v4.2.0
npm WARN excluding symbolic link docs\stylesheets\hightlight.css -> ../../node_m
odules/highlight.js/src/styles/solarized_light.css
npm WARN excluding symbolic link docs\stylesheets\hightlight.css -> ../../node_m
odules/highlight.js/src/styles/solarized_light.css
npm WARN excluding symbolic link docs\assets\ir_black.css -> ../../node_modules/
highlight.js/src/styles/ir_black.css
npm WARN excluding symbolic link docs\stylesheets\hightlight.css -> ../../node_m
odules/highlight.js/src/styles/solarized_light.css
|

fibers@1.0.5 install C:\Users\Test\AppData\Roaming\npm\node_modules\native
script\node_modules\fibers
node ./build.js "   <---- ...

rosen-vladimirov:
Hi @Kyozatch ,

This is really strange. I assume you are working on Windows, am I right? Could you try executing 
$npm install nativescript --verbose and send the output? Are you using 32-bit or 64-bit node?

Kyozatch:
Yes you are right, windows. Node 64-bit.

This is a part of last ouputs :

npm verb tar unpack C:\Users\KNSD1821\AppData\Roaming\npm-cache\debug\2.1.2\pack
age.tgz
npm verb lock tar://C:\Users\KNSD1821\myApp\node_modules\nativescript\node_modul
es\ref-struct\node_modules\debug C:\Users\KNSD1821\AppData\Roaming\npm-cache\2a7
b36c2-es-ref-struct-node-modules-debug.lock
npm verb lock tar://C:\Users\KNSD1821\AppData\Roaming\npm-cache\debug\2.1.2\pack
age.tgz C:\Users\KNSD1821\AppData\Roaming\npm-cache\fe2edf10-pm-cache-debug-2-1-
2-package-tgz.lock
npm info preinstall winston@0.6.2
npm info preinstall semver@1.1.4
npm info preinstall async@0.1.15
npm info preinstall buffers@0.1.1
npm info preinstall fibers@1.0.5
npm verb readDependencies using package.json deps
npm verb readDependencies using package.json deps
[...]

PR #310, opened 2015-03-13
Enable proxy using for http requests

rosen-vladimirov:
Add config.json file inside new config directory. This file is required in order to allow user specific configurations. Modify nativescript-cli.ts in order to wrap resolving of config and errors inside fiber as the new implementation of config has .wait().

Update common lib, where the following changes are applied:
Rename the following config options:
 - FIDDLER_HOSTNAME to PROXY_HOSTNAME
 - PROXY_TO_FIDDLER to USE_PROXY

Add PROXY_PORT option to config with default value 8888. Add ConfigBase class which should be used as a base for CLI specific configs.

Required for https://github.com/NativeScript/nativescript-cli/issues/297 and https://github.com/NativeScript/nativescript-cli/issues/302

### DIFF
--- a/config/config.json
+++ b/config/config.json
@@ -1,0 +1,8 @@
+{
+	"DEBUG": false,
+	"USE_PROXY": false,
+	"PROXY_PORT": 8888,
+	"PROXY_HOSTNAME": "127.0.0.1",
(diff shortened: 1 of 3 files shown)


rosen-vladimirov:
This is not for master, closing it and opening it against release branch

PR #311, opened 2015-03-13
Enable proxy using for http requests

rosen-vladimirov:
Add config.json file inside new config directory. This file is required in order to allow user specific configurations. Modify nativescript-cli.ts in order to wrap resolving of config and errors inside fiber as the new implementation of config has .wait().

Update common lib, where the following changes are applied:
Rename the following config options:
 - FIDDLER_HOSTNAME to PROXY_HOSTNAME
 - PROXY_TO_FIDDLER to USE_PROXY

Add PROXY_PORT option to config with default value 8888. Add ConfigBase class which should be used as a base for CLI specific configs.

Required for https://github.com/NativeScript/nativescript-cli/issues/297 and https://github.com/NativeScript/nativescript-cli/issues/302

### DIFF
--- a/lib/config.ts
+++ b/lib/config.ts
@@ -4,12 +4,21 @@
 import path = require("path");
 import util = require("util");
 import staticConfigBaseLibPath = require("./common/static-config-base");
+import configBaseLib = require("./common/config-base");
 
-$injector.register("config", {
-	CI_LOGGER: false,
-	DEBUG: process.env.NATIVESCRIPT_DEBUG,
-	TYPESCRIPT_COMPILER_OPTIONS: { }
-});
+export class Configuration extends configBaseLib.ConfigBase { // User specific config
+	CI_LOGGER = false;
+	DEBUG = false;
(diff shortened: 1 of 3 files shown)


lib/config.ts
```diff
@@ -4,12 +4,21 @@
...
+import configBaseLib = require("./common/config-base");
 
-$injector.register("config", {
-	CI_LOGGER: false,
-	DEBUG: process.env.NATIVESCRIPT_DEBUG,
-	TYPESCRIPT_COMPILER_OPTIONS: { }
-});
+export class Configuration extends configBaseLib.ConfigBase { // User specific config
+	CI_LOGGER = false;
+	DEBUG = process.env.NATIVESCRIPT_DEBUG;
```
Fatme: `process.env.NATIVESCRIPT_DEBUG` is unneeded, we can set `DEBUG = false`

Fatme:
:+1:

teobugslayer:
I think this must be documented and communicated to users of tns. @ikoevska @ErjanGavalji your opinion?

PR #314, opened 2015-03-13
Fix analytics OS tracking

rosen-vladimirov:
Update common lib, where the following change is applied: Set userAgentString to (osType). For example on Linux the agent will be (Linux).


Fatme:
:+1:

PR #315, opened 2015-03-13
Fix Config implementation

rosen-vladimirov:
ConfigBase implements Config.IConfig. Fix Configuration extends ConfigBase and implements IConfiguration.


Fatme:
:+1:

Issue #251, opened 2015-02-26
[OSX] No validation for command options


dtopuzov:
If you have a typo in --frameworkPath CLI defaults to npmjs.org without any error.
If some of the CLI params is wrong I expect error instead of default behavior without error.

rosen-vladimirov:
This should work fine, could you give some more details how to reproduce the problem.

dtopuzov:
Valication for --frameworkPath is ok on Windows, but when run this on OSX:
tns platform add android --framerowkpaht tns-android.tgz --path TNSApp

Result is:
No error message, platform added (the one from npmjs.org)

Expected:
Error message, because "--framerowkpaht" is not a valid option

dtopuzov:
Same applies also for symlink:
tns platform add android --frameworkPath tns-android.tgz --simlink --path TNS_App

Windows: Error message (expected, because there is a typo in --simlink)
OSX: No validation, command executed and --simlink is just ignored

dtopuzov:
Same for:
node_modules/.bin/tns build android --debug --path TNS_App

In Windows environment result is:
The option 'debug' is not supported.

In OSX result is:
```Shell
shell.js: internal error
Error: ENOENT, no such file or directory '/Users/topuzov/Git/nativescript-cli-tests/TNS_App/platforms/android/assets/App_Resources'
    at Object.fs.mkdirSync (fs.js:653:18)
    at /Users/topuzov/Git/nativescript-cli-tests/node_modules/nativescript/node_modules/shelljs/src/cp.js:174:14
    at Array.forEach (native)
    at Object._cp (/Users/topuzov/Git/nativescript-cli-tests/node_modules/nativescript/node_modules/shelljs/src/cp.js:157:11)
    at Object.cp (/Users/topuzov/Git/nativescript-cli-tests/node_modules/nativescript/node_modules/shelljs/src/common.js:186:23)
    at /Users/topuzov/Git/nativescript-cli-tests/node_modules/nativescript/lib/services/android-project-service.js:101:19
    at Function.settle (/Users/topuzov/Git/nativescript-cli-tests/node_modules/nativescript/node_modules/fibers/future.js:220:26)
    at FiberFuture.Future.wait (/Users/topuzov/Git/nativescript-cli-tests/node_modules/nativescript/node_modules/fibers/future.js:487:10)
    at /Users/topuzov/Git/nativescript-cli-tests/node_modules/nativescript/lib/services/platform-service.js:121:88
[...]

PR #321, opened 2015-03-16
Remove duplicate -- options

rosen-vladimirov:
Yargs adds -- options to yargs.argv in two ways (--profile-dir is added as profile-dir and profileDir). In order to workaround this behavior, we had to add  all dashed options in both ways to our known options. Fix our validation logic, so we'll have to add only profile-dir.
Remove check for client name when validating -- options. This way options will be validated no matter if CLIENT_NAME( tns) or CLIENT_NAME_ALIAS (nativescript) is used on command line.

Fixes https://github.com/NativeScript/nativescript-cli/issues/251

### DIFF
--- a/lib/options.ts
+++ b/lib/options.ts
@@ -9,13 +9,10 @@ import hostInfo = require("./common/host-info");
 var knownOpts:any = {
 		"frameworkPath": String,
 		"copy-from": String,
-		"copyFrom": String,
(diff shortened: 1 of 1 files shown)


teobugslayer:
:+1:

dtopuzov:
+1

Issue #299, opened 2015-03-11
Build failure for helloworld project on nativescript.org


hy9be:
I setup the environment as instructed by the page: http://docs.nativescript.org/setup/ns-cli-setup/ns-setup-os-x.html
And then I try to build the hello world project on nativescript.org: http://docs.nativescript.org/hello-world/hello-world-ns-cli.html. But I was given the error below:

BUILD FAILED
/usr/local/Cellar/android-sdk/24.0.2/tools/ant/build.xml:653: The following error occurred while executing this line:
/usr/local/Cellar/android-sdk/24.0.2/tools/ant/build.xml:698: null returned: 1

Total time: 0 seconds
Command ant failed with exit code 1

Fatme:
Hi @hy9be,

Could you try executing $tns build android --log trace and send the output?

Fatme:
Could you also send the hardware info of your machine? I experienced such issue on machines with less than 512MB RAM.

hy9be:
Thanks! I will try to collect the log. The machine has 16GB RAM so should be fine in terms of available memory.

hy9be:
Here you go, the build trace:

TypeError: Cannot read property 'isLessThanOrEqualTo' of undefined
    at FiberFuture.Future.wait (/usr/local/lib/node_modules/nativescript/node_modules/fibers/future.js:488:15)
    at /usr/local/lib/node_modules/nativescript/lib/nativescript-cli.js:15:45
    - - - - -
    at FiberFuture.Future.wait (/usr/local/lib/node_modules/nativescript/node_modules/fibers/future.js:488:15)
    at /usr/local/lib/node_modules/nativescript/lib/common/dispatchers.js:30:46
    at Function.settle (/usr/local/lib/node_modules/nativescript/node_modules/fibers/future.js:220:26)
    at FiberFuture.Future.wait (/usr/local/lib/node_modules/nativescript/node_modules/fibers/future.js:487:10)
    at /usr/local/lib/node_modules/nativescript/lib/nativescript-cli.js:15:45
    - - - - -
    at Logger.isLevelEnabled (/usr/local/lib/node_modules/nativescript/node_modules/log4js/lib/logger.js:58:20)
    at Logger.(anonymous function) [as trace] (/usr/local/lib/node_modules/nativescript/node_modules/log4js/lib/logger.js:69:16)
    at Logger.trace (/usr/local/lib/node_modules/nativescript/lib/common/logger.js:73:33)
[...]

PR #326, opened 2015-03-16
Fix build for android below 21

rosen-vladimirov:
In res directory there's values directory that is required for all android versions and directory called values-v21, which is required only for Android version 21. In case you do not have installed latest Android version, this directory should not be added to your project.
Add copyResValues method, which copies res/values directory and res/values-v<version_number> directory to your project. This way when we add directories for other android versions, they will not break builds for older ones.

Fixes https://github.com/NativeScript/nativescript-cli/issues/299


dtopuzov:
:+1:

Fatme:
:+1:

PR #331, opened 2015-03-16
Get back availableDevices option

rosen-vladimirov:
Update common lib. where the following change is applied:
Get back the availableDevices option that had been removed by mistake.


Fatme:
:+1:

PR #338, opened 2015-03-17
Update to latest common lib - Show warning if the device and the provided image signature are not compatible

Fatme:


rosen-vladimirov:
:+1:

PR #339, opened 2015-03-17
Move the NativeScript iOS debug from common to the NativeScript CLI, add debug on emulator

PanayotCankov:


lib/commands/debug.ts
```diff
@@ -1,18 +1,26 @@
 ﻿///<reference path="../.d.ts"/>
 "use strict";
 
-import helpers = require("./../common/helpers");
-import util = require("util")
+export class DebugPlatformCommand implements ICommand {
```
rosen-vladimirov: shouldn't you remove this command?

rosen-vladimirov: Ahh, sorry, my mistake

lib/definitions/debug.d.ts
```diff
@@ -0,0 +1,8 @@
+
+interface IPlatformDebugService {
```
rosen-vladimirov: ok, this time I'm almost sure you don't need this

lib/services/android-debug-service.ts
```diff
@@ -0,0 +1,53 @@
...
+	public debugOnEmulator(): IFuture<void> {
+		return (() => {
+			this.$platformService.deployOnEmulator(this.platform).wait();
+			this.debugOnDevice().wait();
+		}).future<void>()();
+	}
+
+	public debugOnDevice(): IFuture<void> {
+		return (() => {
+            var platformData = this.$platformsData.getPlatformData(this.platform);
```
rosen-vladimirov: space/tab issues

lib/services/ios-debug-service.ts
```diff
@@ -0,0 +1,290 @@
...
+	}
+
+	get platform(): string {
+		return "ios";
+	}
+
+	public debug(): IFuture<void> {
+		return (() => {
+			if ((!options.debugBrk && !options.start) || (options.debugBrk && options.start)) {
+				this.$logger.error("Expected exactly one of the --debug-brk or --start options.");
```
rosen-vladimirov: this.$errors.failWithoutHelp seems more appropraite here. So you can omit the return.

lib/services/ios-debug-service.ts
```diff
@@ -0,0 +1,290 @@
...
+		}).future<void>()();
+	}
+
+	private emulatorDebugBrk(): void {
+		var emulatorDebugging = new iOSXcrunSimctl(this.$childProcess);
+		var device = emulatorDebugging.getRunningEmulator().wait();
+		if (device) {
+			this.$logger.info("Install and run on: " + device.uuid);
+		} else {
+			this.$logger.error("This command expects a running emulator.");
```
rosen-vladimirov: this.$errors.failWithoutHelp

lib/services/ios-debug-service.ts
```diff
@@ -0,0 +1,290 @@
...
+	}
+
+	private emulatorStart() {
+		this.executeOpenDebuggerClient().wait();
+
+		var emulatorDebugging = new iOSXcrunSimctl(this.$childProcess);
+
+		var device = emulatorDebugging.getRunningEmulator().wait();
+		if (!device) {
+			this.$logger.error("This command expects a running emulator.");
```
rosen-vladimirov: This will not break the process? Do you want to break it here?

lib/services/ios-debug-service.ts
```diff
@@ -0,0 +1,290 @@
...
+		}
+
+		var projectId = this.$projectData.projectId;
+		var attachRequestMessage = notification.attachRequest(projectId);
+		device.postNotification(attachRequestMessage).wait();
+	}
+
+	private deviceDebugBrk() {
```
rosen-vladimirov: missing return type of the method

lib/services/ios-debug-service.ts
```diff
@@ -0,0 +1,290 @@
...
+		this.$devicesServices.execute(device => (() => {
+			this.$platformService.deployOnDevice(this.platform).wait();
+			var deviceDebugging = new IOSDeviceDebugging(this.$projectData.projectId, <iOSDevice.IOSDevice>device, this.$logger, this.$injector);
+			deviceDebugging.debugApplicationOnStart();
+			this.executeOpenDebuggerClient().wait();
+			device.runApplication(this.$projectData.projectId).wait();
+		}).future<void>()()).wait();
+	}
+
+	private deviceStart() {
```
rosen-vladimirov: missing return type of the method

lib/services/ios-debug-service.ts
```diff
@@ -0,0 +1,290 @@
...
+		var emulatorPackage = this.$platformService.getLatestApplicationPackageForEmulator(platformData).wait();
+
+		device.installApp(emulatorPackage.packageName).wait();
+
+		this.executeOpenDebuggerClient().wait();
+
+		device.launchApp(this.$projectData.projectId, "--nativescript-debug-brk").wait();
+	}
+
+	private emulatorStart() {
```
rosen-vladimirov: missing return type of the method

rosen-vladimirov: also this method has wait(), so you should wrap its body in ```return (() => { .... }).future<void>()()```

lib/services/ios-debug-service.ts
```diff
@@ -0,0 +1,290 @@
...
+
+		this.$devicesServices.execute(device => (() => {
+			this.executeOpenDebuggerClient().wait();
+			var deviceDebugging = new IOSDeviceDebugging(this.$projectData.projectId, <iOSDevice.IOSDevice>device, this.$logger, this.$injector);
+			deviceDebugging.debugRunningApplication();
+		}).future<void>()()).wait();
+	}
+
+	public executeOpenDebuggerClient(): IFuture<void> {
+		if (options.client === false) {
```
rosen-vladimirov: can you write a comment that this means --no-client is specified

lib/services/ios-debug-service.ts
```diff
@@ -0,0 +1,290 @@
...
+	public installApp(appPackage: string): IFuture<void> {
+		return this.$childProcess.exec("xcrun simctl install " + this.uuid + " " + appPackage);
+	}
+
+	public launchApp(appId: string, args?: string): IFuture<void> {
+		var cmd = "xcrun simctl launch " + this.uuid + " " + appId;
+		if (args) {
+			cmd = cmd + " " + args;
+		}
+		console.log("cmd: " + cmd);
```
rosen-vladimirov: Use $logger instead of console.log

resources/help.txt
```diff
@@ -411,45 +411,78 @@ Options:
         You cannot use --device and --emulator simultaneously.
     --release - If set, produces a release build. Otherwise, produces a debug build.
 --[/]--
-    
+
 --[debug]--
+Usage:
```
rosen-vladimirov: Do we have debug command now? From your PR it seems we've removed it?

lib/definitions/debug.d.ts
PanayotCankov: Will fix.

lib/services/android-debug-service.ts
PanayotCankov: Will fix.

resources/help.txt
PanayotCankov: Well, `tns debug` will print the debug information specified here. And we do not have a `debug` command. We have `debug|ios` and `debug|android`. Is this a problem?

lib/services/ios-debug-service.ts
teobugslayer: Also, maybe you can use ```if (!options.client)```

lib/services/ios-debug-service.ts
```diff
@@ -0,0 +1,295 @@
...
+			this.$logger.info("Got ReadyForAttach");
+
+			// NOTE: We will try to provide command line options to select ports, at least on the localhost.
+			var devicePort = 8080;
+			var localPort = 8080;
+
+			this.printHowToTerminate();
+
+			var localServer = net.createServer((localSocket) => {
+				// localServer.close();
```
rosen-vladimirov: Is this comment required for some debugging purposes or its just left behind?

resources/help.txt
rosen-vladimirov: Well... By default we have help only for commands that exist. In case you write:
$ tns debug
we'll show error message that such command does not exist:

But in case you write $ tns debug -h we'll not validate the command, we'll directly check help.txt for such 'article'. And as you have left it here, we'll show help and when the customer tries ```$ tns debug ``` the command will fail.
I'm not sure which one is better solution. We are working hard on improving our help at the moment and I'll consider this case, but for the moment I prefer to remove it. Ping @teobugslayer , @Fatme  - what do you think about this?

rosen-vladimirov:
:+1:  after removing/uncommenting the comment in ios-debug-service.ts that I've commented
I hope you'll understand my comment as its really commenting some comments

lib/services/ios-debug-service.ts
PanayotCankov: fixed

lib/services/ios-debug-service.ts
```diff
@@ -0,0 +1,294 @@
...
+
+			device.launchApp(this.$projectData.projectId, "--nativescript-debug-brk").wait();
+		}).future<void>()();
+	}
+
+	private emulatorStart(): IFuture<void> {
+		return (() => {
+			this.executeOpenDebuggerClient().wait();
+
+			var emulatorDebugging = new iOSXcrunSimctl(this.$childProcess);
```
Fatme: You can use `this.$injector.resolve(iOSXcrunSimctl)`

lib/services/ios-debug-service.ts
```diff
@@ -0,0 +1,294 @@
...
+			device.postNotification(attachRequestMessage).wait();
+		}).future<void>()();
+	}
+
+	private deviceDebugBrk(): IFuture<void> {
+		return (() => {
+			this.$devicesServices.initialize({ platform: this.platform, deviceId: options.device }).wait();
+			this.$devicesServices.execute(device => (() => {
+				this.$platformService.deployOnDevice(this.platform).wait();
+				var deviceDebugging = new IOSDeviceDebugging(this.$projectData.projectId, <iOSDevice.IOSDevice>device, this.$logger, this.$injector);
```
Fatme: You can use `this.$injector.resolve(IOSDeviceDebuggins, { projectId: this.$projectData.projectId})`

lib/services/ios-debug-service.ts
```diff
@@ -0,0 +1,294 @@
...
+			}).future<void>()();
+		} else {
+			return this.openDebuggingClient();
+		}
+	}
+
+	private openDebuggingClient(): IFuture<void> {
+		return (() => {
+			var cmd = "open -a Safari " + this.getSafariPath();
+			this.$childProcess.exec(cmd).wait();
```
Fatme: You can remove `return (() => { .. }).future<void>()()` and just ` var cmd = "open -a Safari " + this.getSafariPath();
return this.$childProcess.exec(cmd)`

lib/services/ios-debug-service.ts
```diff
@@ -0,0 +1,294 @@
...
+	private getSafariPath(): string {
+		var tnsIosPackage = "";
+		if (options.frameworkPath) {
+			if (this.$fs.getFsStats(options.frameworkPath).wait().isFile()) {
+				this.$errors.failWithoutHelp("frameworkPath option must be path to directory which contains tns-ios framework");
+			}
+
+			tnsIosPackage = path.resolve(options.frameworkPath);
+		} else {
+			tnsIosPackage = this.$npm.install("tns-ios").wait();
```
Fatme: Instead of hardcoded name you can user `platformData.frameworkPackageName`

lib/services/ios-debug-service.ts
```diff
@@ -0,0 +1,294 @@
...
+class iOSXcrunSimctl {
+	constructor(private $childProcess: IChildProcess) { }
+
+	public getRunningEmulator(): IFuture<iOSXcrunSimctlDevice> {
+		return (() => {
+			var runningEmulators: string = this.$childProcess.exec("xcrun simctl list | grep Booted | cut -d \"(\" -f2 | cut -d \")\" -f1").wait();
+			var runningEmulator = runningEmulators.split("\n")[0];
+
+			if (runningEmulator) {
+				return new iOSXcrunSimctlDevice(runningEmulator, this.$childProcess);
```
Fatme: `this.$injector.resolve(iOSXcrunSimctlDevice , {deviceUUID: runningEmulator})`

lib/services/ios-debug-service.ts
```diff
@@ -0,0 +1,294 @@
...
+			} else {
+				return null;
+			}
```
Fatme: You don't need `else`. Just `return null;`

lib/services/ios-debug-service.ts
PanayotCankov: Ops, the getSafariPath() must be a future, since it may install the "tns-ios". Good point.

lib/services/ios-debug-service.ts
PanayotCankov: Poluting the $inspector in 3, 2, 1...

PanayotCankov:
@Fatme all the injector suggestions were implemented.

Fatme:
If I run `tns debug ios --debug-brk --emulator` or `tns debug ios --start --emulator` commands, I get `This command expects a running emulator.` message. I was wondering if it is possible to start simulator automatically?

lib/services/ios-debug-service.ts
```diff
@@ -0,0 +1,299 @@
...
+		});
+	}
+
+	public debugRunningApplication() {
+		this.proxyDebuggingTrafic();
+		var attachRequestMessage = notification.attachRequest(this.bundleId);
+		this.$notificationProxyClient.postNotification(attachRequestMessage);
+	}
+
+	private proxyDebuggingTrafic(): void {
```
Fatme: A typo? Shouldn't be `proxyDebuggingTraffic`?

lib/services/ios-debug-service.ts
```diff
@@ -0,0 +1,299 @@
...
+		}).future<void>()();
+	}
+
+	private emulatorStart(): IFuture<void> {
+		return (() => {
+			this.executeOpenDebuggerClient().wait();
+			var device = this.$iosXcrunSimctl.getRunningEmulator().wait();
+			if (!device) {
+				this.$errors.failWithoutHelp("This command expects a running emulator.");
+			}
```
Fatme: Shouldn't  this `this.executeOpenDebuggerClient().wait();` after `if (!device) {`

lib/services/ios-debug-service.ts
```diff
@@ -0,0 +1,299 @@
...
+		}).future<void>()();
+	}
+
+	private deviceStart(): IFuture<void> {
+		return (() => {
+			this.$devicesServices.initialize({ platform: this.platform, deviceId: options.device }).wait();
+
+			this.$devicesServices.execute(device => (() => {
+				this.executeOpenDebuggerClient().wait();
+				var deviceDebugging = new IOSDeviceDebugging(this.$projectData.projectId, <iOSDevice.IOSDevice>device, this.$logger, this.$injector);
```
Fatme: `this.$injecor.resolve(IOSDeviceDebugging, {bundleId: this.$projectData.projectId})`

lib/commands/debug.ts
ligaz: Is there a need for this class to be `exported`?

PanayotCankov: Of course, it is exposed by the ios and android debug commands.
It would throw TS errors if it is not exported. 

I can copy the implementation of the base class into the two derived classes if you insist.

ligaz: I was just curious if the TS compiler will error out.

PR #341, opened 2015-03-17
Merge release 0.9.2 into master

Fatme:
Conflicts:
	lib/definitions/project.d.ts
	lib/services/android-project-service.ts
	lib/services/ios-project-service.ts
	test/stubs.ts


dtopuzov:
:100:

PR #345, opened 2015-03-18
Set version to 0.9.3

rosen-vladimirov:
Set version to 0.9.3


Fatme:
:+1:  after green build

PR #356, opened 2015-03-20
Fix help of debug command

rosen-vladimirov:
Add help for debug command. It is useful to have it, even if there's no "debug" command (it is 'debug android' or 'debug ios' now).
Update sha of common lib to latest (pointing the merge commit of iOSDebugging PR).


Fatme:
:+1:

PR #367, opened 2015-03-23
Emit system information of every run of the CLI

teobugslayer:
Gather system information about the runtime environment.
At install time, warn about potential problems due to mis-configured environment.


package.json
```diff
@@ -55,11 +56,11 @@
     "shelljs": "0.3.0",
     "tabtab": "https://github.com/Icenium/node-tabtab/tarball/master",
     "temp": "0.8.1",
-    "unzip": "0.1.9",
+    "unzip": "0.1.11",
```
rosen-vladimirov: Is this change required - do we use unzip anywhere? If not, let's just remove it.

teobugslayer: I do not know, I simply It is require'd in the FS but not actually used

rosen-vladimirov:
:+1:

ErjanGavalji:
:+1:

package.json
```diff
@@ -58,11 +59,10 @@
     "shelljs": "0.3.0",
     "tabtab": "https://github.com/Icenium/node-tabtab/tarball/master",
     "temp": "0.8.1",
-    "unzip": "0.1.9",
+    "winreg": "0.0.12",
+    "xcode": "https://github.com/NativeScript/node-xcode/archive/NativeScript-0.9.tar.gz",
```
ErjanGavalji: why `xcode` but not `node-xcode`?

teobugslayer: This is not my code, I have no idea :)

rosen-vladimirov: because the name of the module in package.json is xcode

fealebenpae: Basically what Rosen said.

PR #370, opened 2015-03-24
Show HTML Help

rosen-vladimirov:
## Use case:
CLIs should be able to show HTML help when using ```<cli name> help <command name>```
When using ```<cli name> <command name> --help``` console help should be shown.
When an error is raised in command, only console help is shown.

## Implementation
* ```HelpCommand``` is modified to show console help when ```options.help``` is specified and html help when it is not.
* Add ```HtmlHelpService``` which handles both console and html help by parsing markdown help files.
* Each commands help is placed in a separate file. The files are markdown and they are used for both html and console help.
* Html pages are generated from markdown on post-install. As post-install command could fail before generating html pages, when html help should be shown, we check if command name is valid and if the command exists, but its html help is missing, we generate all html pages.
* Modify ```CommandsService``` to show console help by specifying options.help before calling ```HelpCommand```
* Modify ```DynamicHelpService``` and ```DynamicHelpProvider``` to accept isHtml option. When ```isHtml``` is specified, all localVariables will be set to true. This way html help will show full help, no matter of the conditions.
* Add ```basic_page.html``` page and used three placeholders in it. The page is used to generate all html files, by replacing the placeholders with commands name, help content and path to css file.
* Add ```styles.css``` file which contains styles used in html pages.
* Add definitions for marked module.
* Add ```appendFile``` method to ```FileSystem```
* Fix issue in yok, which was not requiring the first hierarchical command if it was not default one (if the hierarchical command doesn't have default command to execute) - this was leading to incorrect behavior when checking if command is valid.
* Remove help.txt file

## Still TODO:
* Update marked-terminal in package.json in order to support tableOptions
* styles.css must be improved
* basic_page.html could be improved with logo, favicon, etc.
* Improve markdown of all commands
* Currently console help is not rendered very well when table is shown. There are two issues here: problem with < and > symbols and problem with cli-table for long rows


rosen-vladimirov:
run ci

teobugslayer:
:+1:

PR #377, opened 2015-03-26
Change app structure - move tns_modules inside app folder

Fatme:
Should be merged after this https://github.com/NativeScript/template-hello-world/pull/16

### DIFF
--- a/lib/services/android-project-service.ts
+++ b/lib/services/android-project-service.ts
@@ -30,12 +30,16 @@ class AndroidProjectService implements IPlatformProjectService {
 	}
 
 	public get platformData(): IPlatformData {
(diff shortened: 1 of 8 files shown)


lib/services/android-project-service.ts
```diff
@@ -30,12 +30,16 @@ class AndroidProjectService implements IPlatformProjectService {
 	}
 
 	public get platformData(): IPlatformData {
+		var projectRoot = path.join(this.$projectData.platformsDir, "android");
```
rosen-vladimirov: what will happen on case sensitive file systems?

Fatme: it should work, because we've created this folder and named it "android"

rosen-vladimirov:
:+1:

PR #392, opened 2015-03-30
Fix marked-terminal parsing and tables in console

rosen-vladimirov:
Remove the hack when parsing markdown text that has to be show on the console. The problem was with code in table, but it had been fixed in version 1.3.0 of marked-terminal.
Use telerik fork of cli-table in order to show tables correctly.


Fatme:
:+1:

Issue #394, opened 2015-03-30
Unable to build on Linux/Mac when app contains symlink


rosen-vladimirov:
In case you have symlink in your app, you can build it only the first time. Each consecutive try leads to 
```
shell.js: internal error
Error: EEXIST, file already exists 'main-page.xml'
    at Object.fs.symlinkSync (fs.js:742:18)
    at cpdirSyncRecursive (/home/rvladimirov/Work/nativescript-cli/node_modules/shelljs/src/cp.js:76:10)
    at cpdirSyncRecursive (/home/rvladimirov/Work/nativescript-cli/node_modules/shelljs/src/cp.js:73:7)
    at /home/rvladimirov/Work/nativescript-cli/node_modules/shelljs/src/cp.js:180:9
    at Array.forEach (native)
    at Object._cp (/home/rvladimirov/Work/nativescript-cli/node_modules/shelljs/src/cp.js:157:11)
    at Object.cp (/home/rvladimirov/Work/nativescript-cli/node_modules/shelljs/src/common.js:186:23)
    at /home/rvladimirov/Work/nativescript-cli/lib/services/platform-service.js:128:19
    at Function.settle (/home/rvladimirov/Work/nativescript-cli/node_modules/fibers/future.js:220:26)
    at FiberFuture.Future.wait (/home/rvladimirov/Work/nativescript-cli/node_modules/fibers/future.js:487:10)
```

### Steps to reproduce:
* Create project: `tns create app1`
* `cd app1`
* `tns platform add android`
* `cd app/app/`
* `ln -s main-page.xml symlinkToMainPage.xml`
* `tns build android` // this one will pass
* `tns build android` // this one will fail

This is very important when you are using node_modules in your app, as node_modules dir contains .bin directory with symlinks to actual files.

The issue is __not__ reproducible on Windows.

vbresults:
@dtopuzov @rosen-vladimirov Still seeing this; rborn/nativescript-i18n#28 rborn/nativescript-i18n#24

PR #395, opened 2015-03-30
Delete destination app dir when preparing project

rosen-vladimirov:
Delete dir where app will be copied on prepare. If we do not delete it, on Linux and Mac we receive EEXIST errors for symlinks inside app directory. Symlinks are common scenario when node_modules are used.

Fixes https://github.com/NativeScript/nativescript-cli/issues/394


Fatme:
:+1:

PR #398, opened 2015-03-30
Logging

jasssonpet:
Introduce `--logging` flag for `emulate`, `run`, `debug` and `device run` commands that outputs the standard output/logcat to the running terminal.

### DIFF
--- a/docs/man_pages/project/testing/debug-ios.md
+++ b/docs/man_pages/project/testing/debug-ios.md
@@ -2,7 +2,7 @@ debug ios
 ==========
 
 Usage:
-    `$ tns debug ios [--debug-brk | --start] [--device <Device ID> | --emulator <Emulator Options>] [--no-client]`
+    `$ tns debug ios [--debug-brk | --start] [--device <Device ID> | --emulator <Emulator Options>] [--print-app-output] [--no-client]`
 
 Example usage:
     `$ tns debug ios --debug-brk`
@@ -22,6 +22,7 @@ Options:
 * `--debug-brk` - Shorthand for prepare, build and deploy. Prepares, builds and deploys the application package on a device or in an emulator, launches the developer tools of your Safari browser.
 * `--start` - Attaches the debug tools to a deployed and running app. Your app must be running on device or emulator, launches the developer tools of your Safari browser.
 * `--emulator` - Debug on already running emulator. Requires `xcrun` from Xcode 6 or later.
+* `--print-app-output` - If set, prints the standard output of the running application. (Device only)
 * `--no-client` - Suppresses the launch of the developer tools in Safari.
 <% if(isHtml) { %> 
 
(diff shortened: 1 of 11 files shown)


jasssonpet:
Also `tns device run` doesn't block on iOS.

package.json
```diff
@@ -34,7 +34,7 @@
     "filesize": "2.0.3",
     "gaze": "0.5.1",
     "iconv-lite": "0.4.4",
-    "ios-sim-portable": "1.0.4",
+    "ios-sim-portable": "https://github.com/jasssonpet/ios-sim-portable/tarball/master",
```
teobugslayer: Should we use 1.0.6 here? @fatme ?

.gitmodules
```diff
@@ -1,3 +1,3 @@
 [submodule "lib/common"]
 	path = lib/common
-	url = git@github.com:telerik/mobile-cli-lib.git
+	url = git@github.com:jasssonpet/mobile-cli-lib.git
```
rosen-vladimirov: can you revert this change :)

jasssonpet: I will revert it and I will also update the change in `package.json`, once the dependencies are published/merged.

docs/man_pages/project/testing/emulate-ios.md
```diff
@@ -13,6 +13,7 @@ Before running the iOS Simulator, verify that you have met the following require
 * You have installed Xcode. The version of Xcode must be compatible with the ios-sim-portable npm package on which the NativeScript CLI depends. For more information, visit https://www.npmjs.org/package/ios-sim-portable
 
 Options:
+* `--logging` - If set, prints the standard output of the running application.
```
ikoevska: Why is this "standard" output remark? How does the "standard output" differ from the "output".

docs/man_pages/project/testing/debug-ios.md
```diff
@@ -22,6 +22,7 @@ Options:
 * `--debug-brk` - Shorthand for prepare, build and deploy. Prepares, builds and deploys the application package on a device or in an emulator, launches the developer tools of your Safari browser.
 * `--start` - Attaches the debug tools to a deployed and running app. Your app must be running on device or emulator, launches the developer tools of your Safari browser.
 * `--emulator` - Debug on already running emulator. Requires `xcrun` from Xcode 6 or later.
+* `--logging` - If set, prints the standard output of the running application. (Device only)
```
ikoevska: What is the (Device only) remark for and why is it featured only here?

docs/man_pages/project/testing/emulate-ios.md
jasssonpet: In android `logcat` is used to show the output, which I think differs from the standard output of the process of iOS. I don't really know about android, so I changed it to be more generic: output. Feel free to reword this as you see fit :)

docs/man_pages/project/testing/debug-ios.md
jasssonpet: At the moment you can't use debugging and logging at the same time on iOS simulator.

docs/man_pages/project/testing/emulate-ios.md
jasssonpet: With Misho, we made it more accurate for android.

ligaz:
What do you think if we made this the other way around and change the default behavior. The `run` command will block the shell and display the app output and if you do not want this you can specify an option `--nolog` or something similar.

PR #400, opened 2015-03-31
Use correct function from postNotificationClient

Fatme:
Should be merged after this https://github.com/telerik/mobile-cli-lib/pull/251


rosen-vladimirov:
:+1:

PR #401, opened 2015-03-31
Update to latest common lib

Fatme:
Introduced `commandsServiceProvider`


rosen-vladimirov:
:+1:  after green build

PR #403, opened 2015-04-01
Add preuninstall script

rosen-vladimirov:
Add preuninstall script that is executed before installation of the new version. It calls dev-preuninstall command and tries to kill adb server (as it may prevent us from successfully installing nativescript-cli when it is used from our resources).


Fatme:
:+1:

PR #405, opened 2015-04-01
Stop calling metadata generator and build twice for android platform.

teobugslayer:
Instead, the project template will supply a custom ants hook to do the work

Also, cache the platformData
Also, starting ant through the user command interpreter falling back to cmd

See https://github.com/NativeScript/android-runtime/issues/47

### DIFF
--- a/lib/services/android-project-service.ts
+++ b/lib/services/android-project-service.ts
@@ -29,24 +29,29 @@ class AndroidProjectService implements IPlatformProjectService {
 		private $propertiesParser: IPropertiesParser) {
 	}
 
+	private _platformData: IPlatformData = null;
 	public get platformData(): IPlatformData {
-		var projectRoot = path.join(this.$projectData.platformsDir, "android");
+		if (!this._platformData) {
+			var projectRoot = path.join(this.$projectData.platformsDir, "android");
 
-		return {
-			frameworkPackageName: "tns-android",
-			normalizedPlatformName: "Android",
-			appDestinationDirectoryPath: path.join(projectRoot, "assets"),
-			appResourcesDestinationDirectoryPath: path.join(projectRoot, "res"),
-			platformProjectService: this,
-			emulatorServices: this.$androidEmulatorServices,
-			projectRoot: projectRoot,
-			deviceBuildOutputPath: path.join(this.$projectData.platformsDir, "android", "bin"),
-			validPackageNamesForDevice: [
-				util.format("%s-%s.%s", this.$projectData.projectName, "debug", "apk"),
-				util.format("%s-%s.%s", this.$projectData.projectName, "release", "apk")
-			],
-			frameworkFilesExtensions: [".jar", ".dat", ".so"]
-		};
+			this._platformData = {
+				frameworkPackageName: "tns-android",
+				normalizedPlatformName: "Android",
+				appDestinationDirectoryPath: path.join(projectRoot, "assets"),
+				appResourcesDestinationDirectoryPath: path.join(projectRoot, "res"),
+				platformProjectService: this,
+				emulatorServices: this.$androidEmulatorServices,
+				projectRoot: projectRoot,
+				deviceBuildOutputPath: path.join(this.$projectData.platformsDir, "android", "bin"),
+				validPackageNamesForDevice: [
+					util.format("%s-%s.%s", this.$projectData.projectName, "debug", "apk"),
+					util.format("%s-%s.%s", this.$projectData.projectName, "release", "apk")
+				],
+				frameworkFilesExtensions: [".jar", ".dat", ".so"]
+			};
+		}
+
+		return this._platformData;
 	}
 
 	public validate(): IFuture<void> {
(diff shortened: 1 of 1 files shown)


teobugslayer:
ping @plamen5kov, @atanasovg

lib/common
```diff
@@ -1 +1 @@
-Subproject commit 202ed586eb28867903230fca9b53f11e349be80c
+Subproject commit 5d37e4203d0b155640210addb66fd047c1b9ad4d
```
rosen-vladimirov: why is the submodule changed?

rosen-vladimirov:
:+1:

lib/services/android-project-service.ts
```diff
@@ -311,6 +290,10 @@ class AndroidProjectService implements IPlatformProjectService {
 			}
 		}
 
+		// metadata generation support
+		args = args.concat(["-Dns.resources", path.join(__dirname, "../../resources/tools")]);
+		args = args.concat((["-Dns.project.libs", this.platformData.libDir]));
```
Fatme: Can we do this inside custom-rules.xml?

lib/services/android-project-service.ts
```diff
@@ -29,24 +29,30 @@ class AndroidProjectService implements IPlatformProjectService {
...
-			emulatorServices: this.$androidEmulatorServices,
-			projectRoot: projectRoot,
-			deviceBuildOutputPath: path.join(this.$projectData.platformsDir, "android", "bin"),
-			validPackageNamesForDevice: [
-				util.format("%s-%s.%s", this.$projectData.projectName, "debug", "apk"),
-				util.format("%s-%s.%s", this.$projectData.projectName, "release", "apk")
-			],
-			frameworkFilesExtensions: [".jar", ".dat", ".so"]
-		};
+		if (!this._platformData) {
```
Fatme: Why do we need to cache this?

lib/common
teobugslayer: No idea, I just pulled the latest common lib

lib/services/android-project-service.ts
teobugslayer: For now, custom-rules.xml depend on us to cook both params.

lib/services/android-project-service.ts
teobugslayer: Because this.platformData is being called quite many times. While string operations are not the slowest ones, creating new objects is a bit taxing in V8.

lib/services/android-project-service.ts
Plamen5kov: From custom_rules.xml we don't know where resources are installed, and i need to use the metadata-generator.jar  in the custom ant step

teobugslayer: Actually, it turned out that we do need it and I have removed it

Plamen5kov:
:+1:

teobugslayer:
**Do not merge before the new project templates!**

teobugslayer:
This issue was moved to NativeScript/cross-platform-modules#241

ligaz:
Should we delete the contents of https://github.com/NativeScript/nativescript-cli/tree/master/resources/tools as well?

teobugslayer:
@ligaz, no, we are using them

ligaz:
@teobugslayer Yep we are using them but shouldn't the be part of the Android runtime package and not the CLI?

teobugslayer:
@ligaz, sounds very good idea to me. @atanasovg , @Plamen5kov , @slavchev  - your opinion?

atanasovg:
We need to evaluate this thoroughly. As of now our thinking is that the metadata generator is a stand-alone tool that generates an artifact for the runtime to consume. In theory, the runtime will be able to consume any such metadata as long as it is in the appropriate format. With that said, I am not sure whether the runtime itself should be responsible for producing the metadata artifacts. Anyways, we are open for a discussion in some greater details.

ErjanGavalji:
Considering the [Plugin model](/NativeScript/NativeScript/issues/25), the metadata generator would be used to generate the metadata for the plugins. When working with a framework, one would expect that its tooling aids in the production of add-ons. That said, either NativeScript should have some king of a toolset, or the Metadata generation would exist in the CLI.

If we take the CLI as an entry to the tooling system (the way we have `android` an entry point to the `adb` and to other android tools) it should be the exact thing to be used for metadata generation.

I'm not sure if the metadata generator should be packed with the CLI though. It could be an addition which the CLI would search upon need. That would make the CLI package smaller. Additionally, NativeScript CLI could inform users for missing tools and suggest an automatic installation...

ligaz:
The main idea behind moving the metadata generation as part of the runtime is that the metadata format is tightly coupled with the version of the runtime that is being used. If we decide to change this format we will be in forced to push new version of the CLI but what if users want to use the older format (due to some breaking changes). It will be nightmare to support this if it is not bundled in each runtime package.

atanasovg:
@ligaz Makes sense. We had a quick discussion about this within the team and what we agreed is on par with your concern - versioning and maintenance. Originally our idea was to ship the metadata generator as a separate NPM package and the Android-specific CLI logic to download it upon request. We just didn't have the time to prepare the package for the previous release and we put it in the CLI directly. But we will make the package for the next version.

teobugslayer:
This PR should be re-targeted for the release branch

teobugslayer:
This PR is obsolete, closing for now

PR #408, opened 2015-04-03
Move html styles to common

rosen-vladimirov:
Move all styles of HTML to common lib. Modify replacing in HtmlHelpService to use global match in order to replace all strings in the html. Replace HTML_HELPERS_DIR from StaticConfig with HTML_COMMON_HELPERS_DIR and HTML_CLI_HELPERS_DIR. The idea is to have specific helpers for each CLI and common helpers.
Add favicon and logo to docs/helpers dir.


teobugslayer:
:+1:

PR #409, opened 2015-04-06
Fix errors print in debug

rosen-vladimirov:
Sync with latest common lib, where the following change is applied:
In case you are using Node 0.12.x and in config.json "debug" is set to true, the real error message is not printed to the output. When we are in debug mode, we print only the callstack and in Node 0.10.x , when fibers throw exception, the top row of the callstack is the error message. With Node 0.12.x, when error is thrown from fibers, the message is not part of the callstack for some reason. Workaround this behavior by checking the callstack and add the message in case it is missing.


dtopuzov:
:+1:

Issue #175, opened 2015-01-12
Support Google images for emulate


teobugslayer:
Right now, only pure android images are supported

robert52:
is there any workaround for this? or when will the Google images be supported? 
Thanks.

teobugslayer:
Hi @robert52,
The bug title is wrong - I meant "Support google images for running in the emulator". I fixed the title.
There's no workaround for now - there's some code to be written. We can try fixing this for the 1.0 release but it may be postponed. Of course, pull requests are welcome :)

teobugslayer:
Implemented in https://github.com/telerik/mobile-cli-lib/pull/268

robert52:
Hi @teobugslayer nice job :+1: that was fast!. I was digging around in the source code, but it would have taken me a while. But anyways I find this project very useful so keep it up. I'll try and contribute with something.

teobugslayer:
Hi @robert52, please share with us if you have any problems getting up with the project - we would like to make it every easy to start hacking on it!

PR #411, opened 2015-04-06
Support for running emulator with Google images

teobugslayer:
see https://github.com/NativeScript/nativescript-cli/issues/175


rosen-vladimirov:
:+1:

PR #412, opened 2015-04-06
Stop calling metadata generator and build twice for android platform

teobugslayer:
Instead, the project template will supply a custom ants hook to do the work

Also, cache the platformData
Also, starting ant through the user command interpreter falling back to cmd

See NativeScript/android-runtime#47

This pR is against the release branch. The one against master will be closed (https://github.com/NativeScript/nativescript-cli/pull/405)

### DIFF
--- a/lib/services/android-project-service.ts
+++ b/lib/services/android-project-service.ts
@@ -29,24 +29,29 @@ class AndroidProjectService implements IPlatformProjectService {
 		private $propertiesParser: IPropertiesParser) {
 	}
 
+	private _platformData: IPlatformData = null;
 	public get platformData(): IPlatformData {
-		var projectRoot = path.join(this.$projectData.platformsDir, "android");
+		if (!this._platformData) {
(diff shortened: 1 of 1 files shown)


rosen-vladimirov:
:+1:

lib/services/android-project-service.ts
```diff
@@ -422,16 +403,6 @@ class AndroidProjectService implements IPlatformProjectService {
 		}).future<void>()();
 	}
 
-	private checkJava(): IFuture<void> {
```
Fatme: Do you need to remove this function?

ligaz:
:+1:

lib/services/android-project-service.ts
teobugslayer: Yes, it is unused. We used to check for java because we needed to run it manually. Now, we do not this.

Plamen5kov:
:+1:

PR #420, opened 2015-04-09
Improve autocompletion

rosen-vladimirov:
Move autocompletion code to a new shell file. In .bash_profile, .bashrc and .zshrc add code that executes this code when we want to enable autocompletion.
Autocomplete is now hierarchical command. The subcommands and their behavior is:
* `autocomplete` - checks if autocompletion is enabled. In case it is not, ask the user for confirmation. In case autocompletion is enabled, but it is the old one (where we've written some data to .bash_profile, etc.), silently remove the old code and use the new approach. In case the user chooses to not use autocompletion, call disableAutocCompletion in order to remove all autocompletion code from all shell profiles.
* `autocomplete enable` - enables autocompletion in case it is not already enabled.
* `autocomplete disable` - disables autocompletion in case it is not already disabled.
* `autocomplete status` - prints the current status of autocompletion.

Add help for autocomplete commands.

### DIFF
--- a/docs/man_pages/general/autocomplete-disable.md
+++ b/docs/man_pages/general/autocomplete-disable.md
@@ -1,17 +1,15 @@
-autocomplete
+autocomplete disable
 ==========
 
 Usage | Syntax
 ------|-------
-General | `$ appbuilder autocomplete`
-Get status | `$ appbuilder autocomplete status`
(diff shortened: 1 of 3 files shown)


dtopuzov:
:+1:

Issue #421, opened 2015-04-09
Typescript


rhg1968:
This is a question about using typescript with Nativescript. I created the hello world example and it worked correctly, but I wanted to see how it would work with typescript. I decided to change the main-page.js to main-page.ts. This broke the application and the taps that were being counted on the screen now show with just the binding expression {{message}}. I even tried changing the the new ts file to use import and export typescript functions without any luck.

Does the CLI automatically compile the ts files into javascript ?

george-jones:
I had the exact same experience.  Did you find a resolution to this?

rhg1968:
I didn't find a solution unfortunately. I closed this because I posted it instead on the Native script forums.
https://groups.google.com/forum/#!topic/nativescript/r2jwG4fIXt4

teobugslayer:
Hello @rhg1968,

You are right that at the moment we do not auto-compile ts files. I am preparing a fix for it. 

We have already started preparing for releasing version 0.10 of the CLI and my fix probably won't be in it. You'll need to wait a little longer or work against master ;)

ctartamella:
His this available in 1.0.1?  I am not seeing any of my .ts files being compiled.

teobugslayer:
Hello @ctartamella,

No, this is not available in {N} 1.0.1

We decided that the proposed implementation is not good enough and I closed my PRs. We will implement it in a better way, with ability to customize TypeScript compilation per-project in one of the future releases of {N} CLI.

In the mean time, you can use one of the node.js automation tools, like [grunt](http://gruntjs.com/) or [gulp](http://gulpjs.com/) to automate ts compilation.

ligaz:
We have written down our current thinking in this issue: https://github.com/NativeScript/nativescript-cli/issues/493. I'm going to close this down and let's continue the discussion there.

PR #422, opened 2015-04-11
TypeScript compilation in NativeScript

teobugslayer:
see https://github.com/NativeScript/nativescript-cli/issues/421

### DIFF
--- a/lib/fiber-bootstrap.ts
+++ b/lib/fiber-bootstrap.ts
@@ -1,0 +1,15 @@
+import Fiber = require("fibers");
+import Future = require("fibers/future");
+import errors = require("./common/errors");
+
+export function run(action: Function) {
+	var fiber = Fiber(() => {
+		errors.installUncaughtExceptionListener();
+		action();
+		$injector.dispose();
+		Future.assertNoFutureLeftBehind();
+	});
+
+	fiber.run();
+}
+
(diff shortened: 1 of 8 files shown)


ligaz:
We have been discussing that we might want to add the TypeScript as project level hook and not global one. The problem with the global one is that it cannot easily be overwritten so that users can create their own workflows using task runners like grunt, gulp or broccoli. 

Another benefit of having it per project is that users might use different version of the TypeScript compiler in each project along with different `tsconfig.json`

lib/fiber-bootstrap.ts
```diff
@@ -0,0 +1,15 @@
+import Fiber = require("fibers");
```
Fatme: This file is duplicated and should be moved to common lib

lib/hooks/before-livesync.ts
```diff
@@ -0,0 +1,3 @@
+require("./typescript-compilation.js");
```
Fatme: This is unneeded here

lib/hooks/typescript-compilation.ts
```diff
@@ -0,0 +1,22 @@
+///<reference path="../.d.ts"/>
```
Fatme: This service should be in common lib

teobugslayer:
We will support TypeScript per-project, closing this PR for now.

PR #423, opened 2015-04-14
Switch the debugging on emulator for iOS to use the iOSEmulatorService

PanayotCankov:
Switch the debugging on emulator for iOS to use the iOSEmulatorService so that it can benefit from logging in the terminal.

> **NOTE:** This requires the ios-sim-portable 1.0.6


jasssonpet:
The "device only note" should be removed from: https://github.com/NativeScript/nativescript-cli/blob/master/docs/man_pages/project/testing/debug-ios.md

teobugslayer:
ping @ikoevska

PanayotCankov:
@jasssonpet done

PR #431, opened 2015-04-20
Move KillSwitches dir to tmp

rosen-vladimirov:
Move KillSwitches dir to os.temp directory as when CLI is installed with sudo, we need correct privileges.


dtopuzov:
:+1:

PR #432, opened 2015-04-21
Updated the help for the help command as it now opens the browser

ikoevska:
ping @rosen-vladimirov


rosen-vladimirov:
:+1:  after green build

PR #433, opened 2015-04-21
Removing mention of nodes-legacy

ikoevska:
and pointing to the official installation instructions instead


rosen-vladimirov:
:+1:

PR #437, opened 2015-04-27
* Update common lib

Fatme:
* Add dynamicCommandsPrefix function in commandsService
* Remove options-service.ts
* Update version to 1.0.0


rosen-vladimirov:
:+1:  after green build

PR #449, opened 2015-04-27
Set version to 1.0.0-RC

teobugslayer:
Set the build version in package.json to 1.0.0-RC for the next public release.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "nativescript",
   "preferGlobal": true,
-  "version": "1.0.0",
+  "version": "1.0.0-rc0",
   "author": "Telerik <support@telerik.com>",
   "description": "Command-line interface for building NativeScript projects",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "nativescript",
   "preferGlobal": true,
-  "version": "1.0.0",
+  "version": "1.0.0-RC",
```
rosen-vladimirov: RC0, RC1 maybe? What will be the next version in case we decide to publish additional one before we refactor our project infrastructure? In case we keep 1.0.0-RC here and after that we publish 1.0.0-RC2, will it work for npm?

rosen-vladimirov:
:+1:

package.json
rosen-vladimirov: check point 9 -> http://semver.org/ current code is fine

PR #451, opened 2015-04-28
Breaking changes in 0.10.0, bumping version to 1.0.0-RC

ikoevska:


rosen-vladimirov:
:+1:  after green build

PR #457, opened 2015-05-04
Fix library add ios with spaces in path

jasssonpet:
Fixes https://github.com/NativeScript/nativescript-cli/issues/459 in the CLI

### DIFF
--- a/lib/services/ios-project-service.ts
+++ b/lib/services/ios-project-service.ts
@@ -230,7 +230,7 @@ class IOSProjectService implements  IPlatformProjectService {
                 this.$errors.failWithoutHelp("The bundle at %s does not contain an Info.plist file.", libraryPath);
             }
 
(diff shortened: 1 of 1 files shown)


lib/services/ios-project-service.ts
```diff
@@ -230,7 +230,7 @@ class IOSProjectService implements  IPlatformProjectService {
                 this.$errors.failWithoutHelp("The bundle at %s does not contain an Info.plist file.", libraryPath);
             }
 
-            var packageType = this.$childProcess.exec('/usr/libexec/PlistBuddy -c "Print :CFBundlePackageType" ' + infoPlistPath).wait().trim();
+            var packageType = this.$childProcess.exec(util.format('/usr/libexec/PlistBuddy -c "Print :CFBundlePackageType" "%s"', infoPlistPath)).wait().trim();
```
ligaz: You can use string interpolation here. It makes the code more readable.

jasssonpet: Nice :)

teobugslayer:
For this to work, you need to use back quotes (`)

Fatme:
:+1:

PR #469, opened 2015-05-07
Update lib/common to latest

jasssonpet:


rosen-vladimirov:
:+1:  after green build

PR #471, opened 2015-05-08
Fix iOS device debug-brk.

jasssonpet:
The deploy command now runs the application and the debug break command is starting it twice which locks the app.


teobugslayer:
It is a bit too late for 1.0.1 - please re-target for master.

PR #482, opened 2015-05-15
Use cancellation only on Windows

teobugslayer:
Due to different security models on Windows and *ix, this code cannot be made to work reliably on all OSes.
Because it exists only because Windows keeps executable code locked, the right thing to do is to execute it only on Windows.


rosen-vladimirov:
:+1:  after green build

PR #483, opened 2015-05-15
Fixes

teobugslayer:
Console not released when only 1 CLI instance is running.
Incorrect validation for iTunes on Linux - we consider it existing.


rosen-vladimirov:
:+1:

PR #484, opened 2015-05-18
Update semver module and move it to common

Mitko-Kerezov:
Merge along with https://github.com/telerik/mobile-cli-lib/pull/335


package.json
```diff
@@ -58,7 +58,7 @@
     "ref": "https://github.com/icenium/ref/tarball/master",
     "ref-struct": "https://github.com/telerik/ref-struct/tarball/master",
     "rimraf": "2.2.6",
-    "semver": "3.0.1",
+    "semver": "^4.3.4",
```
Fatme: We prefer to keep a exact version so this should be `4.3.4`

Fatme:
:+1:

PR #486, opened 2015-05-18
Enable command autocompletion when local installation is used

rosen-vladimirov:
When you do not specify -g flag, we could still enable autocompletion, but currently this process fails as we are executing globally installed CLI (for example tns completion >> ...). Use the javascript from bin directory instead.
Fix dev-preuninstall command, which was failing due to incorrect service called in it (cancellationService is not called at all, so I removed it. In fact we have $cancellation, not $cancellationService).


teobugslayer:
:+1: after reasonably approximate to a green build

PR #487, opened 2015-05-19
Enable npm as package manager

Fatme:
Should be merged after this one https://github.com/telerik/mobile-cli-lib/pull/351


lib/definitions/semver.d.ts
```diff
@@ -0,0 +1,16 @@
+
+declare module "semver" {
```
rosen-vladimirov: I think we already have the semver definitions in common

lib/services/platform-service.ts
```diff
@@ -137,7 +140,7 @@ export class PlatformService implements IPlatformService {
 
 	public preparePlatform(platform: string): IFuture<void> {
 		return (() => {
-			platform = platform.toLowerCase();
+		 	platform = platform.toLowerCase();
```
teobugslayer: Is this added space needed?

lib/services/platform-service.ts
```diff
@@ -168,14 +171,16 @@ export class PlatformService implements IPlatformService {
...
 
+			var tnsModulesDestinationPath = path.join(platformData.appDestinationDirectoryPath, constants.APP_FOLDER_NAME, PlatformService.TNS_MODULES_FOLDER_NAME);
+			this.$broccoliBuilder.prepareNodeModules(tnsModulesDestinationPath, this.$projectData.projectDir).wait();
+
 			this.$logger.out("Project successfully prepared");
 
 		}).future<void>()();
 	}
-
+
```
teobugslayer: Is this change needed?

lib/tools/broccoli/broccoli-plugin-wrapper.ts
```diff
@@ -0,0 +1,59 @@
...
+  constructor(private pluginClass: any, wrappedPluginArguments: any) { 
+      this.inputTree = wrappedPluginArguments[0];
+      this.description = this.pluginClass.name;
+      this.absoluteOutputPath = wrappedPluginArguments[1];
+      this.treeRootDirName = wrappedPluginArguments[2];
+      this.projectDir = wrappedPluginArguments[3];
+      this.$injector = $injector.resolve("injector");
+  }
+
+  rebuild(): any {
```
teobugslayer: The code does not return but function is declared to return any. Consider declaring it as void.

lib/tools/broccoli/builder.ts
```diff
@@ -0,0 +1,62 @@
...
+
+  private rebuild(builder: any): IFuture<any> {
+    let future = new Future<any>();
+    builder.build()
+        .then((result: any) => {
+              future.return(result);
+        })
+        .catch((err: any) => {
+              if(err.file) {
+                console.error("File: " + err.file);
```
teobugslayer: Looks like debug code to me. Is it needed? If yes, consider using $loggger instead.

lib/tools/broccoli/node-modules-dest-copy.ts
```diff
@@ -0,0 +1,84 @@
...
+		  let dirs = fs.readdirSync(this.outputRoot);
+		  _.each(dirs, dir => {
+			  let fileContent = require(path.join(dir, constants.PACKAGE_JSON_FILE_NAME));
+			  result[fileContent.name] = fileContent;
+			 });
+	  } catch(err) {
+		  result = Object.create(null);
+	  }
+	  
+	  return result;
```
teobugslayer: As the function is written, it silently skips all dependencies in case any of them is corrupted. This means that code after that will randomly fail. Maybe it is better to let the exception cancel the current operation so that the user can have contextual information what's wrong and can fix it.

lib/tools/broccoli/trees/node-modules-tree.ts
```diff
@@ -0,0 +1,16 @@
+///<reference path="../../../.d.ts"/>
+"use strict";
+
+let Funnel = require('broccoli-funnel');
+let path = require("path");
```
teobugslayer: this can be import'ed

teobugslayer:
You need to rebase on master

lib/definitions/semver.d.ts
Fatme: Yes, they were moved to common.

ErjanGavalji:
:+1:

lib/tools/broccoli/builder.ts
```diff
@@ -0,0 +1,63 @@
...
+    let future = new Future<any>();
+    builder.build()
+        .then((result: any) => {
+              future.return(result);
+        })
+        .catch((err: any) => {
+              if(err.file) {
+                this.$logger.error("File: " + err.file); 
+              }
+              if(err.statck) {
```
teobugslayer: maybe err.stack? (a spelling mistake)

lib/tools/broccoli/builder.ts
```diff
@@ -0,0 +1,63 @@
...
+                this.$logger.error(err.stack);
+              }
+              future.throw(err.toString());
```
teobugslayer: Why are you throwing a string instead of an Error?

lib/tools/broccoli/node-modules-dest-copy.ts
```diff
@@ -0,0 +1,108 @@
...
+				};
+				
+				let addedDependency = dependencies[currentDependency.name];
+				if (addedDependency) {
+					if (semver.gt(currentDependency.version, addedDependency.version)) {
+						let currentDependencyMajorVersion = semver.major(currentDependency.version);
+						let addedDependencyMajorVersion = semver.major(addedDependency.version);
+		
+						let message = util.format("Old version %s, new version %s. Old location %s, new location %s.", addedDependency.version, currentDependency.version, addedDependency.directory, currentDependency.directory);
+						currentDependencyMajorVersion === addedDependencyMajorVersion ? console.log(message) : console.log(message);
```
teobugslayer: This code does not look write - the if and else code is the same.

lib/services/plugins-service.ts
```diff
@@ -0,0 +1,196 @@
...
+				}
+			};
+			
+			this.executeForAllInstalledPlatforms(action);
+		}).future<void>()();
+	}
+	
+	public getAllInstalledPlugins(): IFuture<IPluginData[]> {
+		return (() => {
+			let nodeModules = this.$fs.readDirectory(this.nodeModulesPath).wait();
```
rosen-vladimirov: make sure `this.nodeModulesPath` directory exists

rosen-vladimirov:
In case you remove node_modules dir manually, the plugin is not added to the project after build. I suggest to call `npm install` in project dir on prepare step. This way all dependencies will be installed in node_modules dir before building the project. 
Add/Remove plugin commands can check the package.json's dependencies instead of listing node_modules

lib/services/plugins-service.ts
```diff
@@ -0,0 +1,196 @@
...
+	}
+	
+	private parseNpmCommandResult(npmCommandResult: string): string { // The npmCommandResult is in the following format: [<name>@<version node_modules/<name>]
+		return npmCommandResult.split("@")[0]; // returns plugin name
+	}
+	
+	private executeForAllInstalledPlatforms(action: (pluginDestinationPath: string, platform: string, platformData: IPlatformData) => void): void {
+		let availablePlatforms = _.keys(this.$platformsData.availablePlatforms);
+		_.each(availablePlatforms, platform => {
+			let isPlatformInstalled = this.$fs.exists(path.join(this.$projectData.platformsDir, platform.toLowerCase())).wait();
```
teobugslayer: There is such function in platform service but it is private. Maybe we can reuse it here?

lib/services/platform-service.ts
```diff
@@ -305,6 +315,10 @@ export class PlatformService implements IPlatformService {
 			}
 		}).future<void>()();
 	}
+	
+	private isPlatformInstalled(platform: string): IFuture<boolean> {
```
teobugslayer: It seems that platform name is not always lower-cased. Consider adding toLowerCase().

lib/services/plugins-service.ts
Fatme: Actually I wanted to reuse it. The problem is that `pluginsService` is dependency of `platformsService` so here I am not able to inject `pluginsService` :smile:

lib/services/platform-service.ts
Fatme: Done :)

lib/declarations.ts
```diff
@@ -39,7 +39,8 @@ interface IOptions extends ICommonOptions {
 	emulator: boolean;
 	symlink: boolean;
 	forDevice: boolean;
-	client: boolean;
+	client: boolean
```
rosen-vladimirov: missing `;` after client

lib/services/platform-service.ts
```diff
@@ -169,14 +177,17 @@ export class PlatformService implements IPlatformService {
...
+			// Process node_modules folder
+			this.$pluginsService.ensureAllDependenciesAreInstalled().wait();
+			var tnsModulesDestinationPath = path.join(platformData.appDestinationDirectoryPath, constants.APP_FOLDER_NAME, PlatformService.TNS_MODULES_FOLDER_NAME);
+			this.$broccoliBuilder.prepareNodeModules(tnsModulesDestinationPath, this.$projectData.projectDir).wait();
 
+			this.$logger.out("Project successfully prepared");
 		}).future<void>()();
 	}
-
+
```
rosen-vladimirov: some unneeded tabs :)

lib/services/plugins-service.ts
teobugslayer: We discussed in person and agreed to leave it as-is.

PR #488, opened 2015-05-19
Support for Visual studio code

Fatme:


.gitignore
```diff
@@ -17,6 +17,7 @@ lib-cov
 *.tgz
 *.tmp
 *.sublime-workspace
+*.tsconfig.json
```
teobugslayer: We actually can use tsconfig

teobugslayer:
:+1: after green build

PR #494, opened 2015-05-22
Fix profileDir path

Fatme:


lib/options.ts
```diff
@@ -25,7 +25,7 @@ export class Options extends commonOptionsLibPath.OptionsBase {
 			keyStoreAlias: { type: OptionType.String },
 			keyStoreAliasPassword: { type: OptionType.String }
 		},
-		path.join($hostInfo.isWindows ? process.env.LocalAppData : path.join(osenv.home(), ".local/share", "Telerik", "BlackDragon", ".nativescript-cli")),
+		path.join($hostInfo.isWindows ? process.env.LocalAppData : path.join(osenv.home(), ".local/share", ".nativescript-cli")),
```
rosen-vladimirov: this is not correct, the profile dir should be 
```TypeScript
process.env.LocalAppData/.nativescript-cli
```
on Windows
With this code it will be just process.env.LocalAppData

lib/options.ts
```diff
@@ -25,7 +25,7 @@ export class Options extends commonOptionsLibPath.OptionsBase {
 			keyStoreAlias: { type: OptionType.String },
 			keyStoreAliasPassword: { type: OptionType.String }
 		},
-		path.join($hostInfo.isWindows ? process.env.LocalAppData : path.join(osenv.home(), ".local/share", "Telerik", "BlackDragon", ".nativescript-cli")),
+		path.join($hostInfo.isWindows ? path.join(process.env.LocalAppData, ".nativescript-cli") : path.join(osenv.home(), ".local/share", ".nativescript-cli")),
```
rosen-vladimirov: or maybe:
```TypeScript
path.join($hostInfo.isWindows ? process.env.LocalAppData : path.join(osenv.home(), ".local/share"), ".nativescript-cli"),
```

rosen-vladimirov:
:+1:

PR #506, opened 2015-05-29
Initial version of plugin commands

Fatme:


lib/commands/plugin/add-plugin.ts
```diff
@@ -0,0 +1,24 @@
+///<reference path="../../.d.ts"/>
+"use strict";
+
+export class AddPluginCommand implements ICommand {
+	constructor(private $pluginsService: IPluginsService,
+		private $errors: IErrors) { }
+	
+	execute(args: string[]): IFuture<void> {
+		if(!args[0]) {
```
teobugslayer: I think this if belongs to canExecute

lib/commands/plugin/remove-plugin.ts
```diff
@@ -0,0 +1,24 @@
+///<reference path="../../.d.ts"/>
+"use strict";
+
+export class RemovePluginCommand implements ICommand {
+	constructor(private $pluginsService: IPluginsService,
+		private $errors: IErrors) { }
+	
+	execute(args: string[]): IFuture<void> {
+		if(!args[0]) {
```
teobugslayer: Consider moving this check into canExecute

lib/definitions/plugins.d.ts
```diff
@@ -0,0 +1,5 @@
+interface IPluginsService {
```
teobugslayer: Consider adding JSDocs for the interface and its members

lib/services/plugins-service.ts
```diff
@@ -0,0 +1,48 @@
...
+		private $childProcess: IChildProcess) { }
+		
+	public get modulesDestinationPath() {
+		let platformData = this.$platformsData.getPlatformData("ios");
+		return path.join(platformData.appDestinationDirectoryPath, constants.APP_FOLDER_NAME, "tns_modules");
+	} 
+		
+	public add(pluginName: string): IFuture<void> {
+		return (() => {
+			let command = `npm install ${pluginName} --save`;
```
teobugslayer: Maybe it is a good idea to support --production as well. I mean, our command tns plugin add should have a --production option, which we pass to npm install. Your opinion?

lib/services/plugins-service.ts
```diff
@@ -0,0 +1,48 @@
...
+			this.$childProcess.exec(command, { cwd: this.$projectData.projectDir }).wait();
+			this.prepare(pluginName).wait();	
+		}).future<void>()();
+	}
+	
+	public prepare(pluginName: string): IFuture<void> {
+		return (() => {
+			// Process .js files 
+			// TODO: get correctly pluginName if plugin is installed with url or folder path
```
teobugslayer: Until this is implemented, you should reject such plugins.

Fatme:
I'm closing and will reopen against correct branch.

PR #509, opened 2015-05-29
Initial version of plugin commands

Fatme:
`tns plugin add <plugin>` where `plugin` can be `local folder path`, `github url`, `node module that is published in npm` - This command should install plugin locally in node_modules folder, save it to `package.json` and prepare it if some platform is added. If no platforms are added the installed plugins are prepared on `platform add step`.
`tns plugin remove pluginName` removes the specified plugin

### DIFF
--- a/lib/definitions/plugins.d.ts
+++ b/lib/definitions/plugins.d.ts
@@ -1,0 +1,22 @@
+interface IPluginsService {
+	add(plugin: string): IFuture<void>; // adds plugin by name, github url, local path and et.
+	remove(pluginName: string): IFuture<void>; // removes plugin only by name
+	prepare(pluginData: IPluginData): IFuture<void>;
+	getAllInstalledPlugins(): IFuture<IPluginData[]>;
+}
+
+interface IPluginData extends INodeModuleData {
+	platformsData: IPluginPlatformsData; 
+}
+
+interface INodeModuleData {
+	name: string;
+	version: string;
+	fullPath: string;
+	isPlugin: boolean;
+}
+
+interface IPluginPlatformsData {
+	ios: string;
+	android: string;
+}
--- a/lib/definitions/plugins.d.ts
+++ b/lib/definitions/plugins.d.ts
@@ -14,6 +14,7 @@ interface INodeModuleData {
 	version: string;
 	fullPath: string;
 	isPlugin: boolean;
+	moduleInfo: any;
 }
 
 interface IPluginPlatformsData {
--- a/lib/services/platform-service.ts
+++ b/lib/services/platform-service.ts
@@ -317,9 +317,7 @@ export class PlatformService implements IPlatformService {
 	}
 	
 	private isPlatformInstalled(platform: string): IFuture<boolean> {
-		return (() => {
-			return this.$fs.exists(path.join(this.$projectData.platformsDir, platform)).wait();
-		}).future<boolean>()();
+		return this.$fs.exists(path.join(this.$projectData.platformsDir, platform));
 	}
 
 	private isValidPlatform(platform: string) {
--- a/lib/services/platform-service.ts
+++ b/lib/services/platform-service.ts
@@ -22,7 +22,8 @@ export class PlatformService implements IPlatformService {
 		private $prompter: IPrompter,
 		private $commandsService: ICommandsService,
 		private $options: IOptions,
-		private $broccoliBuilder: IBroccoliBuilder) { }
+		private $broccoliBuilder: IBroccoliBuilder,
+		private $pluginsService: IPluginsService) { }
 
 	public addPlatforms(platforms: string[]): IFuture<void> {
 		return (() => {
@@ -86,6 +87,10 @@ export class PlatformService implements IPlatformService {
 				this.$fs.deleteDirectory(platformPath).wait();
 				throw err;
 			}
+			
+			// Prepare installed plugins
+			let installedPlugins = this.$pluginsService.getAllInstalledPlugins().wait();
+			_.each(installedPlugins, pluginData => this.$pluginsService.prepare(pluginData).wait());
 
 			this.$logger.out("Project successfully created.");
 
@@ -310,6 +315,12 @@ export class PlatformService implements IPlatformService {
 			}
 		}).future<void>()();
 	}
+	
+	private isPlatformInstalled(platform: string): IFuture<boolean> {
+		return (() => {
+			return this.$fs.exists(path.join(this.$projectData.platformsDir, platform)).wait();
+		}).future<boolean>()();
+	}
 
 	private isValidPlatform(platform: string) {
 		return this.$platformsData.getPlatformData(platform);
@@ -323,12 +334,6 @@ export class PlatformService implements IPlatformService {
 		}
 
 		return false;
-	}
-
-	private isPlatformInstalled(platform: string): IFuture<boolean> {
-		return (() => {
-			return this.$fs.exists(path.join(this.$projectData.platformsDir, platform)).wait();
-		}).future<boolean>()();
 	}
 
 	private isPlatformPrepared(platform: string): IFuture<boolean> {
--- a/lib/services/plugins-service.ts
+++ b/lib/services/plugins-service.ts
@@ -28,11 +28,11 @@ export class PluginsService implements IPluginsService {
 			if(!nodeModuleData.isPlugin) {
 				// We should remove already downloaded plugin and show an error message
 				this.executeNpmCommand(PluginsService.UNINSTALL_COMMAND_NAME, pluginName).wait();
-				this.$errors.failWithoutHelp(`The specified plugin ${plugin} is not a valid NativeScript plugin. Ensure that the plugin contains nativescript key in package.json file and try again.`);
+				this.$errors.failWithoutHelp(`${plugin} is not a valid NativeScript plugin. Verify that the plugin package.json file contains a nativescript key and try again.`);
 			}
 			
 			this.prepare(this.convertToPluginData(nodeModuleData)).wait();
-			this.$logger.out(`Succsessfully installed plugin with name ${nodeModuleData.name}.`);	
+			this.$logger.out(`Successfully installed plugin ${nodeModuleData.name}.`);	
 		}).future<void>()();
 	}
 	
@@ -42,7 +42,7 @@ export class PluginsService implements IPluginsService {
 			let showMessage = true;
 			let action = (modulesDestinationPath: string, platform: string, platformData: IPlatformData) => {
 				shelljs.rm("-rf", path.join(modulesDestinationPath, pluginName));
-				this.$logger.out(`Successfully removed plugin ${pluginName} for ${platform} platform`);
+				this.$logger.out(`Successfully removed plugin ${pluginName} for ${platform}.`);
 				showMessage = false;
 			};
 			this.executeForAllInstalledPlatforms(action);
@@ -61,7 +61,7 @@ export class PluginsService implements IPluginsService {
 				let installedFrameworkVersion = this.getInstalledFrameworkVersion(platform).wait();
 				let pluginVersion = (<any>pluginData.platformsData)[platform];
 				if(semver.gt(pluginVersion, installedFrameworkVersion)) {
-					this.$logger.warn(`Plugin ${pluginData.name} with specified version ${pluginVersion} for ${platform} platform is not compatible for currently installed framework with version ${installedFrameworkVersion}.`);
+					this.$logger.warn(`${pluginData.name} ${pluginVersion} for ${platform} is not compatible with the currently installed framework version ${installedFrameworkVersion}.`);
 					skipExecution = true;  
 				}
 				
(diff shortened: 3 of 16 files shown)


lib/definitions/plugins.d.ts
```diff
@@ -0,0 +1,5 @@
+interface IPluginsService {
```
teobugslayer: Please add JSDocs for the interface and its members

lib/services/plugins-service.ts
```diff
@@ -0,0 +1,48 @@
...
+		private $childProcess: IChildProcess) { }
+		
+	public get modulesDestinationPath() {
+		let platformData = this.$platformsData.getPlatformData("ios");
+		return path.join(platformData.appDestinationDirectoryPath, constants.APP_FOLDER_NAME, "tns_modules");
+	} 
+		
+	public add(pluginName: string): IFuture<void> {
+		return (() => {
+			let command = `npm install ${pluginName} --save`;
```
teobugslayer: Do you think it is a good idea to support --production flag for tns plugin add as well? It will simply be passed to npm.
At any rate this functionality is for another PR.

lib/services/plugins-service.ts
```diff
@@ -0,0 +1,48 @@
...
+
+export class PluginsService implements IPluginsService {
+	constructor(private $npm: INodePackageManager,
+		private $fs: IFileSystem,
+		private $projectData: IProjectData,
+		private $platformsData: IPlatformsData,
+		private $childProcess: IChildProcess) { }
+		
+	public get modulesDestinationPath() {
+		let platformData = this.$platformsData.getPlatformData("ios");
```
teobugslayer: This crashes with only added android platform

lib/services/plugins-service.ts
```diff
@@ -0,0 +1,190 @@
...
+		private $childProcess: IChildProcess,
+		private $options: IOptions,
+		private $logger: ILogger,
+		private $errors: IErrors) { }
+		
+	public add(plugin: string): IFuture<void> {
+		return (() => {
+			let pluginName = this.executeNpmCommand(PluginsService.INSTALL_COMMAND_NAME, plugin).wait();
+			let nodeModuleData = this.getNodeModuleData(pluginName);
+			if(!nodeModuleData.isPlugin) {
```
Fatme: I wasn't able to figure out a better way for validating `plugin` :disappointed:

ErjanGavalji: This is indeed an interesting topic. Let's brainstorm of what is a plugin.

I think this is an npm package that has a `nativescript` complex property in its package.json. @ligaz, @teobugslayer, your input?

ErjanGavalji:
A brief example of the usage in the pull-request description?

Fatme:
I've just updated the description and I'm planning to add help :)

ErjanGavalji:
:+1:

lib/services/plugins-service.ts
```diff
@@ -0,0 +1,190 @@
...
+				// We should remove already downloaded plugin and show an error message
+				this.executeNpmCommand(PluginsService.UNINSTALL_COMMAND_NAME, pluginName).wait();
+				this.$errors.failWithoutHelp(`The specified plugin ${plugin} is not a valid NativeScript plugin. Ensure that the plugin contains nativescript key in package.json file and try again.`);
```
ikoevska: ${plugin} is not a valid NativeScript plugin. Verify that the plugin package.json file contains a nativescript key and try again.

lib/services/platform-service.ts
```diff
@@ -310,6 +315,12 @@ export class PlatformService implements IPlatformService {
 			}
 		}).future<void>()();
 	}
+	
+	private isPlatformInstalled(platform: string): IFuture<boolean> {
+		return (() => {
+			return this.$fs.exists(path.join(this.$projectData.platformsDir, platform)).wait();
```
rosen-vladimirov: you can return the future from fs.exists directly without wrapping it in new future

lib/services/plugins-service.ts
```diff
@@ -0,0 +1,190 @@
...
+			}
+			
+			this.prepare(this.convertToPluginData(nodeModuleData)).wait();
+			this.$logger.out(`Succsessfully installed plugin with name ${nodeModuleData.name}.`);
```
ikoevska: Successfully installed plugin ${nodeModuleData.name}.

lib/services/plugins-service.ts
```diff
@@ -0,0 +1,190 @@
...
+		}).future<void>()();
+	}
+	
+	public remove(pluginName: string): IFuture<void> {
+		return (() => {
+			this.executeNpmCommand(PluginsService.UNINSTALL_COMMAND_NAME, pluginName).wait();
+			let showMessage = true;
+			let action = (modulesDestinationPath: string, platform: string, platformData: IPlatformData) => {
+				shelljs.rm("-rf", path.join(modulesDestinationPath, pluginName));
+				this.$logger.out(`Successfully removed plugin ${pluginName} for ${platform} platform`);
```
ikoevska: Successfully removed plugin ${pluginName} for ${platform}.

lib/services/plugins-service.ts
```diff
@@ -0,0 +1,190 @@
...
+	
+	public prepare(pluginData: IPluginData): IFuture<void> {
+		return (() => {
+			let action = (pluginDestinationPath: string, platform: string, platformData: IPlatformData) => {
+				let skipExecution = false;
+				// Process .js files 				
+				let installedFrameworkVersion = this.getInstalledFrameworkVersion(platform).wait();
+				let pluginVersion = (<any>pluginData.platformsData)[platform];
+				if(semver.gt(pluginVersion, installedFrameworkVersion)) {
+					this.$logger.warn(`Plugin ${pluginData.name} with specified version ${pluginVersion} for ${platform} platform is not compatible for currently installed framework with version ${installedFrameworkVersion}.`);
```
ikoevska: ${pluginData.name} ${pluginVersion} for ${platform} is not compatible with the currently installed framework version ${installedFrameworkVersion}.

lib/services/plugins-service.ts
```diff
@@ -0,0 +1,190 @@
...
+					}
+					
+					if(this.$fs.exists(pluginPlatformsFolderPath).wait()) {
+						shelljs.rm("-rf", pluginPlatformsFolderPath);			
+					}						
+					
+					// TODO: Add libraries
+					
+					// Show message
+					this.$logger.out(`Successfully prepared plugin ${pluginData.name} for ${platform} platform`);
```
ikoevska: Successfully prepared plugin ${pluginData.name} for ${platform}.

test/plugins-service.ts
```diff
@@ -0,0 +1,415 @@
...
+		testInjector = createTestInjector();
+		testInjector.registerCommand("plugin|add", AddPluginCommandLib.AddPluginCommand);
+	});
+	
+	describe("plugin add", () => {
+		it("fails when no param is specified to plugin add command", () => {
+			addPluginWhenExpectingToFail(testInjector, null, "You must specify plugin name.");
+		}); 
+		it("fails when invalid nativescript plugin name is specified", () => {
+			addPluginWhenExpectingToFail(testInjector, "lodash", "The specified plugin lodash is not a valid NativeScript plugin. Ensure that the plugin contains nativescript key in package.json file and try again.");
```
ikoevska: Apply the identical message from above.

test/plugins-service.ts
```diff
@@ -0,0 +1,415 @@
...
+		});
+		it("fails when the plugin is already installed", () => {
```
ikoevska: Is this a customer-facing message?

test/plugins-service.ts
```diff
@@ -0,0 +1,415 @@
...
+	
+			isErrorThrown = false;			
+			let commandsService = testInjector.resolve(CommandsServiceLib.CommandsService);
+			commandsService.tryExecuteCommand("plugin|add", [pluginName]).wait();
+			
+			assert.isTrue(isErrorThrown);
+		});
+		it("fails when the plugin does not support the installed framework", () => {
+			let isWarningMessageShown = false;
+			let expectedWarningMessage = "Plugin mySamplePlugin with specified version 1.0.1 for android platform is not compatible for currently installed framework with version 1.0.0.";
```
ikoevska: Apply the identical message from above.

test/plugins-service.ts
Fatme: No, this is the name of test :smile:

lib/services/plugins-service.ts
```diff
@@ -0,0 +1,190 @@
...
+				}
+				
+				if(!skipExecution) {
+					this.$fs.ensureDirectoryExists(pluginDestinationPath).wait();
+					shelljs.cp("-R", pluginData.fullPath, pluginDestinationPath);
+				
+					let pluginPlatformsFolderPath = path.join(pluginDestinationPath, pluginData.name, "platforms", platform);
+					let pluginConfigurationFilePath = path.join(pluginPlatformsFolderPath, platformData.configurationFileName);					
+					if(this.$fs.exists(pluginConfigurationFilePath).wait()) {
+						let pluginConfigurationFileContent = this.$fs.readFile(pluginConfigurationFilePath).wait().toString();
```
rosen-vladimirov: you can use readText instead of toString-ing the result of readFile

lib/services/plugins-service.ts
```diff
@@ -0,0 +1,190 @@
...
+						let configurationFileContent = this.$fs.readFile(platformData.configurationFilePath).wait().toString();
```
rosen-vladimirov: readText maybe

lib/services/plugins-service.ts
```diff
@@ -0,0 +1,190 @@
...
+			this.executeForAllInstalledPlatforms(action);
+		}).future<void>()();
+	}
+	
+	public getAllInstalledPlugins(): IFuture<IPluginData[]> {
+		return (() => {
+			let nodeModules = this.$fs.readDirectory(this.nodeModulesPath).wait();
+			let plugins: IPluginData[] = [];
+			_.each(nodeModules, nodeModuleName => {
+				var nodeModuleData = this.getNodeModuleData(nodeModuleName);
```
rosen-vladimirov: var -> let

rosen-vladimirov: also consider using filter + map:
```TypeScript
return _(nodeModules)
          .map(nm => this.getNodeModuleData(nm))
          .filter(nm => nm.isPlugin)
          .map(nm => this.convertToPluginData(nodeModuleData))
          .value();
```
After writing this I'm not sure my suggestion is correct. Maybe your code is even faster than this chain...

lib/services/plugins-service.ts
```diff
@@ -0,0 +1,190 @@
...
+	private executeNpmCommand(npmCommandName: string, npmCommandArguments: string): IFuture<string> {
+		return (() => {
+			let command = this.composeNpmCommand(npmCommandName, npmCommandArguments);
+			let result = this.$childProcess.exec(command, { cwd: this.$projectData.projectDir }).wait();
+			return this.parseNpmCommandResult(result);
+		}).future<string>()();
+	}
+	
+	private composeNpmCommand(npmCommandName: string, npmCommandArguments: string): string {
+		let command = `npm ${npmCommandName} ${npmCommandArguments} --save `;
```
rosen-vladimirov: what will happen in case one of the arguments has space in it ?

lib/services/plugins-service.ts
```diff
@@ -0,0 +1,190 @@
...
+	}
+	
+	private getNodeModuleData(moduleName: string): INodeModuleData {
+		let fullNodeModulePath = path.join(this.nodeModulesPath, moduleName);
+		let packageJsonFilePath = path.join(fullNodeModulePath, "package.json");
+		let data = require(packageJsonFilePath);
+		let result = {
+			name: data.name,
+			version: data.version,
+			isPlugin: data.nativescript,
```
teobugslayer: isPlugin suggests a boolean value, but here is the entire object. I suggest you either give this property a boolean value, or rename it to something like "pluginInfo".

lib/services/plugins-service.ts
```diff
@@ -0,0 +1,190 @@
...
+			isPlugin: data.nativescript,		
+			fullPath: fullNodeModulePath,
+		};
+		
+		return result;
+	}
+	
+	private convertToPluginData(nodeModuleData: INodeModuleData): IPluginData {
+		let pluginData: any = _.extend({}, nodeModuleData);
+		let data = <any>(nodeModuleData.isPlugin);
```
teobugslayer: Why the typecast?

lib/services/plugins-service.ts
```diff
@@ -0,0 +1,190 @@
...
+			let platformData = this.$platformsData.getPlatformData(platform);
+			this.$projectDataService.initialize(this.$projectData.projectDir);
+			let frameworkData = this.$projectDataService.getValue(platformData.frameworkPackageName).wait();
+			return frameworkData.version;
+		}).future<string>()();
+	}
+	
+	private mergeXml(xml1: string, xml2: string): IFuture<string> {
+		let future = new Future<string>();
+		xmlmerge.merge(xml1, xml2, "", (mergedXml: string) => { // TODO: process errors
```
teobugslayer: The API does not provide errors (https://github.com/heysdk/xmlmerge-js/blob/master/xmlmerge.js#L243) so we cannot handle them.

lib/services/plugins-service.ts
teobugslayer: Seems reasonable to me. If a need arise, we can make the definition stricter.

lib/services/plugins-service.ts
```diff
@@ -0,0 +1,196 @@
...
+	
+	private getNodeModuleData(moduleName: string): INodeModuleData {
+		let fullNodeModulePath = path.join(this.nodeModulesPath, moduleName);
+		let packageJsonFilePath = path.join(fullNodeModulePath, "package.json");
+		let data = require(packageJsonFilePath);
+		let result = {
+			name: data.name,
+			version: data.version,
+			fullPath: fullNodeModulePath,
+			isPlugin: data.nativescript !== undefined,
```
rosen-vladimirov: maybe `isPlugin: !!data.nativescript`, but not a merge stopper

rosen-vladimirov:
:+1:  just one minor comment, but anyway, great work
:100: :100: :100:

Issue #507, opened 2015-05-29
Android MNC support


teobugslayer:

teobugslayer:
The official Android 6.0 SDK is out and this story is obsolete.

juju470:
Duplicate of #3825

PR #513, opened 2015-06-01
Support for android MNC

teobugslayer:
Whitelist MNC SDK
change all vars to lets
When building with MNC SDK, change minSdkVersion to be MNC, required for succesful build
Use latest common lib to support MNC emulator

See https://github.com/NativeScript/nativescript-cli/issues/507

### DIFF
--- a/lib/services/android-project-service.ts
+++ b/lib/services/android-project-service.ts
@@ -10,8 +10,7 @@ import fs = require("fs");
 import os = require("os");
 
 class AndroidProjectService implements IPlatformProjectService {
-	private SUPPORTED_TARGETS = ["android-17", "android-18", "android-19", "android-21"];
-	private static METADATA_DIRNAME = "__metadata";
+	private SUPPORTED_TARGETS = ["android-17", "android-18", "android-19", "android-21", "android-MNC"];
 	private static RES_DIRNAME = "res";
 	private static VALUES_DIRNAME = "values";
 	private static VALUES_VERSION_DIRNAME_PREFIX = AndroidProjectService.VALUES_DIRNAME + "-v";
@@ -32,7 +31,7 @@ class AndroidProjectService implements IPlatformProjectService {
 	private _platformData: IPlatformData = null;
 	public get platformData(): IPlatformData {
 		if (!this._platformData) {
-			var projectRoot = path.join(this.$projectData.platformsDir, "android");
+			let projectRoot = path.join(this.$projectData.platformsDir, "android");
 
 			this._platformData = {
 				frameworkPackageName: "tns-android",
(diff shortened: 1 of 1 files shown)


lib/services/android-project-service.ts
```diff
@@ -111,27 +110,35 @@ class AndroidProjectService implements IPlatformProjectService {
...
 			this.runAndroidUpdate(projectRoot, targetApi).wait();
+			this.adjustMinSdk(projectRoot);
 		}).future<void>()();
 	}
 
+	private adjustMinSdk(projectRoot: string): void {
+		let manifestPath = path.join(projectRoot, "AndroidManifest.xml");
+		let apiLevel = this.getTarget(projectRoot).wait().split('-')[1];
+		if (apiLevel === "MNC") { // MNC SDK requires that minSdkVersion is set to "MNC"
+			shell.sed('-i', /android:minSdkVersion=".*"/, `android:minSdkVersion="${apiLevel}"`, manifestPath);
```
rosen-vladimirov: should you include ? in the regex. If you have line android:minSdkVersion="L" android:smthElse="Value", the match will be everything up to Value".

lib/services/android-project-service.ts
```diff
@@ -179,25 +184,25 @@ class AndroidProjectService implements IPlatformProjectService {
 		this.$logger.info("Copying %s", projDir);
 		shell.cp("-Rf", projDir, destDir);
 
-		var targetDir = path.join(destDir, path.basename(projDir));
+		let targetDir = path.join(destDir, path.basename(projDir));
 		// TODO: parametrize targetSdk
-		var targetSdk = "android-17";
+		let targetSdk = this.targetApi || "android-17";
```
rosen-vladimirov: is it possible to have `this.targetApi` still unpopulated? Shouldn't we call this.getTarget (, just to make sure targetApi has correct value?
As far as I see this.targetApi is populated only in this.getTarget method:
https://github.com/NativeScript/nativescript-cli/blob/totev/mnc/lib/services/android-project-service.ts#L379

rosen-vladimirov:
:+1:  after green build

lib/services/android-project-service.ts
```diff
@@ -179,25 +184,25 @@ class AndroidProjectService implements IPlatformProjectService {
 		this.$logger.info("Copying %s", projDir);
 		shell.cp("-Rf", projDir, destDir);
 
-		var targetDir = path.join(destDir, path.basename(projDir));
+		let targetDir = path.join(destDir, path.basename(projDir));
 		// TODO: parametrize targetSdk
-		var targetSdk = "android-17";
+		let targetSdk = this.getTarget(projDir).wait() || "android-17";
```
Fatme: Why do we need this change here? This will update only `project.properties` file

teobugslayer: After personal discussion we decided to fix this in a separate PR.

Fatme:
:+1:

Issue #514, opened 2015-06-02
Prepare command renames non-project files


Fatme:
1. Create project with name `my-ios-app`.
2. Add `iOS` platform
3. Run `tns prepare ios` 

**Actually**: 
The `-Info.plist` and `-Prefix.pch` are processes as platform specific files and are renamed to `myapp-Info.plist` and `myapp-Prefix.pch`.

**Expected**:
 Only the files in `app` folder should be treated as platform specific.

PR #515, opened 2015-06-02
Process as platform specific files only these located in app folder

Fatme:
https://github.com/NativeScript/nativescript-cli/issues/514

### DIFF
--- a/lib/services/platform-service.ts
+++ b/lib/services/platform-service.ts
@@ -157,15 +157,16 @@ export class PlatformService implements IPlatformService {
 			}
 
 			// Process platform specific files
-			var contents = this.$fs.readDirectory(platformData.appDestinationDirectoryPath).wait();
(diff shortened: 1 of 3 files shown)


lib/services/platform-service.ts
```diff
@@ -157,15 +157,16 @@ export class PlatformService implements IPlatformService {
 			}
 
 			// Process platform specific files
-			var contents = this.$fs.readDirectory(platformData.appDestinationDirectoryPath).wait();
+			var contents = this.$fs.readDirectory(path.join(platformData.appDestinationDirectoryPath, constants.APP_FOLDER_NAME)).wait();
```
teobugslayer: let :)

teobugslayer:
:+1: after addressing my small remark.

lib/services/platform-service.ts
Fatme: I'm planning to replace all `var` usages with `let` in another PR.

ErjanGavalji:
:+1:

PR #519, opened 2015-06-03
Forbid MNC SDK

teobugslayer:
Programs built with it can run only on MNC devices.
We disable it until we create a good UI for the user to select target/min SDK


rosen-vladimirov:
:crying_cat_face: 
:+1: 
:cry:

PR #533, opened 2015-06-08
Process only dependencies in projectDir/node_modules

Fatme:


rosen-vladimirov:
:+1:  after green build

PR #540, opened 2015-06-08
Fixed grunt task name to "tests"

nemephx:
There is no registered task named "test". It should be "tests"


teobugslayer:
Hello @nemephx,

There is "test" task - see https://github.com/NativeScript/nativescript-cli/blob/master/Gruntfile.js#L149

nemephx:
You're right, teo. My mistake!

teobugslayer:
No problem, it was a bit hidden. Thank you for your help!

Issue #538, opened 2015-06-08
File names wrongly cut on deploy


karaahmed:
1.Create a hello world {N} app with iOS platform 
2.Create a module file pair in form  <somename>ios.ts / <somename>ios.xml  (or just rename the main-page.js/xml to main-pageios.js/xml , note the missing . before iOS)
e.g.: abadabaios.js/abadabaios.xml 
3. Execute: tns prepare ios 
4. check the deployed files in prepared solution
Result: the example file are deployed with name: 
abadab.js/abadab.xml
This leads to error during execution: "JS ERROR Error: Failed to load Page from entry.moduleName: views/abadabaios" , since the file with name abadabaios.xml doesn't exist

teobugslayer:
Is this related to https://github.com/NativeScript/nativescript-cli/pull/515 and https://github.com/NativeScript/nativescript-cli/pull/514 ?

PR #541, opened 2015-06-09
Fix rule for filtering platform specific files.

teobugslayer:
Treat only files with ".platform" right before file extension are treated as platform specific.

See https://github.com/NativeScript/nativescript-cli/issues/538


Fatme:
:+1:

PR #542, opened 2015-06-09
Fix android debug --debug-brk to use proper internal files dir instead of sdcard location which may not be present

Fatme:


Fatme:
I'm closing this for now because it is not working in android runtime

PR #546, opened 2015-06-10
Ikoevska/post 110 fixes

ikoevska:
* Fixed --justlaunh
* Added note about commands showing the application output by default
* Fixed wrong platform in the debug android synopsis table


rosen-vladimirov:
:+1:

PR #553, opened 2015-06-11
Polish xml merge

Fatme:
https://github.com/NativeScript/nativescript-cli/issues/555

### DIFF
--- a/lib/services/plugins-service.ts
+++ b/lib/services/plugins-service.ts
@@ -6,6 +6,7 @@ import semver = require("semver");
 import Future = require("fibers/future");
 import constants = require("./../constants");
 let xmlmerge = require("xmlmerge-js");
+let DOMParser = require('xmldom').DOMParser;
 
 export class PluginsService implements IPluginsService {
 	private static INSTALL_COMMAND_NAME = "install";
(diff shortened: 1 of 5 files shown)


test/plugins-service.ts
```diff
@@ -419,4 +445,142 @@ describe("Plugins service", () => {
...
+			createAndroidManifestFile(projectFolder, fs);		
+			
+			// Mock plugins service
+			let pluginsService = testInjector.resolve("pluginsService");
+			pluginsService.getAllInstalledPlugins = () => { 
+				return (() => {
+					return [{
+						name: ""
+					}];
+				}).future<IPluginData[]>()();
```
teobugslayer: Not a merge stopper, but you may use future.fromResult to cut some code

test/plugins-service.ts
```diff
@@ -419,4 +445,142 @@ describe("Plugins service", () => {
...
+				return {
+					appDestinationDirectoryPath: appDestinationDirectoryPath,
+					frameworkPackageName: "tns-android",
+					configurationFileName: "AndroidManifest.xml"
+				}
+			}
+			
+			// Ensure the pluginDestinationPath folder exists
+			let pluginPlatformsDirPath = path.join(appDestinationDirectoryPath, "app", "tns_modules", pluginName, "platforms", "android");
+			fs.ensureDirectoryExists(pluginPlatformsDirPath).wait();
```
teobugslayer: Is there a need to do actual FS I/O? Can we mock the FS instead?

test/plugins-service.ts
```diff
@@ -419,4 +445,142 @@ describe("Plugins service", () => {
...
+				
+			// Expected error message. The assertion happens in mockBeginCommand
+			let expectedErrorMessage = `Exception: Invalid xml file ${pluginConfigurationFilePath}. Additional technical information: element parse error: Exception: Invalid xml file ` + 
+				`${pluginConfigurationFilePath}. Additional technical information: unclosed xml attribute` + 
+				`\n@#[line:1,col:39].` + 
+				`\n@#[line:1,col:39].`;
+			mockBeginCommand(testInjector, expectedErrorMessage);				
+			
+			let commandsService = testInjector.resolve(CommandsServiceLib.CommandsService);		
+			commandsService.tryExecuteCommand("plugin|add", [pluginFolderPath]).wait();
```
teobugslayer: Why are we executing the command itself? Can't we test a service instead?

test/plugins-service.ts
Fatme: I need to use the actual, real `fs` because I really create platforms folder

test/plugins-service.ts
Fatme: Actually because I have already created mocks and want to reuse them :smile:

lib/services/plugins-service.ts
```diff
@@ -77,12 +78,24 @@ export class PluginsService implements IPluginsService {
...
+							locator: {},
+							errorHandler: (level: any, msg: string) => {
+								this.$errors.fail(`Invalid xml file ${pluginConfigurationFilePath}. Additional technical information: ${msg}.` )
+							}
+						});
+						doc.parseFromString(pluginConfigurationFileContent, 'text/xml');
+						
+						// Merge xml
+						let configurationFileContent = this.$fs.readText(configurationFilePath).wait();
+						let resultXml = this.mergeXml(configurationFileContent, pluginConfigurationFileContent, platformData.mergeXmlConfig || []).wait();
```
rosen-vladimirov: shouldn't you verify the resultXml as well? This way we will not save incorrect xml, which the user will have to modify manually after that.

rosen-vladimirov:
:+1:

test/plugins-service.ts
teobugslayer: Please reconsider leaving the test as-is. Currently you are testing the entire command infrastructure and the plugins service at the same time, which is not the correct way of writing unit tests.

Issue #552, opened 2015-06-11
Add option to specify Android Target SDK


rosen-vladimirov:
When using `$ tns platform add android` I would like to be able to specify target SDK. For example
`$ tns platform add android --sdk MNC` should create a project, targeting MNC API Level,
`$ tns platform add android --sdk 19` should create a project, targeting API Level 19.

Issue #551, opened 2015-06-11
Add support for Android 5.1.1 (API Level 22)


rosen-vladimirov:
CLI has list of supported and verified android API Levels. In the moment Android 5.1.1 is not part of this list, which is causing failure when you have only this API Level installed from Android SDK. The error message is:
`Please install Android target 17. Make sure you have the latest Android tools installed as well. Run "android" from your command-line to install/update any missing SDKs or tools.`

> NOTE: In order to test this, you must remove all other SDKs and add only API Level 22 (but not MNC) and make sure you can successfully build and deploy the app on device/emulator.

PR #554, opened 2015-06-11
Add support for different Android SDKs

rosen-vladimirov:
* Add Android API Level 22 to supported versions. This way in case you have only API Level 22 installed, you will be able to use it and create Android project. Implements https://github.com/NativeScript/nativescript-cli/issues/551
* Add --sdk option to `$ tns platform add android` command. This option gives you the ability to specify the SDK that you would like to use. In case the value is below 17, an error will be raised. In case the value is not in the supported versions, a warning will be shown, that some functionality may not work as expected. Implements https://github.com/NativeScript/nativescript-cli/issues/552

### DIFF
--- a/lib/services/android-project-service.ts
+++ b/lib/services/android-project-service.ts
@@ -10,7 +10,9 @@ import fs = require("fs");
 import os = require("os");
 
 class AndroidProjectService implements IPlatformProjectService {
-	private SUPPORTED_TARGETS = ["android-17", "android-18", "android-19", "android-21"]; // forbidden for now: "android-MNC"
+	private static MIN_SUPPORTED_VERSION = 17;
+	private SUPPORTED_TARGETS = ["android-17", "android-18", "android-19", "android-21", "android-22"]; // forbidden for now: "android-MNC"
+	private static ANDROID_TARGET_PREFIX = "android";
 	private static RES_DIRNAME = "res";
 	private static VALUES_DIRNAME = "values";
 	private static VALUES_VERSION_DIRNAME_PREFIX = AndroidProjectService.VALUES_DIRNAME + "-v";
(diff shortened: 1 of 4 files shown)


lib/services/android-project-service.ts
```diff
@@ -331,6 +334,25 @@ class AndroidProjectService implements IPlatformProjectService {
...
 
+	private getAndroidTarget(frameworkDir: string): IFuture<string> {
+		return ((): string => { 
+			let newTarget = this.$options.sdk ? `${AndroidProjectService.ANDROID_TARGET_PREFIX}-${this.$options.sdk}` : this.getLatestValidAndroidTarget(frameworkDir).wait();
+			if(!_.contains(this.SUPPORTED_TARGETS, newTarget)) {
+				let versionNumber = parseInt(_.last(newTarget.split("-")));
+				if(versionNumber) {
+					let minSupportedVersion = _.first(this.SUPPORTED_TARGETS.sort());
+					this.$logger.trace(`Minimum supported Android target is ${minSupportedVersion}.`);
+					if(versionNumber < parseInt(_.last(minSupportedVersion.split("-")))) {
```
Fatme: This `_.last(minSupportedVersion.split("-"))` is always `17`. Isn't it better to hardcode it ?

rosen-vladimirov: I prefer to keep it this way, so in the future, when we decide to support at least 19, just to remove android-17 from SUPPORTED_TARGETS and everything will work as expected (on theory of course :smile_cat: )

Fatme: If we decide to support at least 19, we should change the constant and everything will work as expected (on theory of course :smile_cat: ) but the code will be more readable and will avoid unneeded operations :smile:

Fatme:
:+1:

Issue #281, opened 2015-03-06
"tns platform remove" does not report anything on success 


dtopuzov:
Steps to reproduce:
tns create TNSApp
tns platform add android --path TNSApp
tns platform remove android --path TNSApp

Result:
Androdi platform is removed successfully, but there is no output.
May be it will be better if where is success message.

PR #563, opened 2015-06-14
Report message when platform gets successfully removed

Fatme:
https://github.com/NativeScript/nativescript-cli/issues/281


rosen-vladimirov:
:+1:

PR #564, opened 2015-06-15
Use prebuilt binaries for bufferutil and utf-8-validate

rosen-vladimirov:
"ws" module requires bufferutil and utf-8-validate modules, which have to be build during installation. On Windows this requires Python, Visual Studio 2010, etc. Prebuilt the two modules and use them from Telerik forks.


teobugslayer:
:+1:

Fatme:
:+1:

PR #566, opened 2015-06-15
Copy styles for supported Android versions

rosen-vladimirov:
Inside tns-android resources, there are values-v<API_LEVEL> directories. Currently we were copying only values-v21 in case the current project targets API_LEVEL 21. Instead we should copy it when the project targets API_LEVEL 21 or upper (in case there's no newer directory, for example values-v22).
For example:
* `$ tns platform add android --sdk 22` - should copy values and values-v21 in your platforms/android/res dir
* `$ tns platform add android --sdk 21` - should copy values and values-v21 in your platforms/android/res dir
* `$ tns platform add android --sdk MNC` - should copy values and values-v21 in your platforms/android/res dir
* `$ tns platform add android --sdk 17` - should copy values in your platforms/android/res dir


dtopuzov:
:+1:

teobugslayer:
:+1: The lodash chain is awesome!

PR #571, opened 2015-06-16
Show user friendly message when the plugin is not supported for installed platform

Fatme:


lib/services/plugins-service.ts
```diff
@@ -67,7 +67,12 @@ export class PluginsService implements IPluginsService {
 				let pluginPlatformsData = pluginData.platformsData;
 				if(pluginPlatformsData) {
 					let pluginVersion = (<any>pluginPlatformsData)[platform];
-					if(semver.gt(pluginVersion, installedFrameworkVersion)) {
+					if(!pluginVersion) {
+						this.$logger.warn(`${pluginData.name} is not supported for ${platform}.`);
+						skipExecution = true;
```
teobugslayer: Can't you simply ```return ``` from the function?

Fatme: Yes, we can. I've just fixed it.

teobugslayer:
:+1: after green build

dtopuzov:
:+1:

Issue #539, opened 2015-06-08
Running --log with no args with results in undefined exception instead of good help description


blagoev:
tns debug android --debug-brk --device x.y.z --log

TypeError: Cannot call method 'toString' of undefined
    at FiberFuture.Future.wait (c:\Users\X\Documents\Temp\testCLI1.1\node_
modules\nativescript\node_modules\fibers\future.js:488:15)
    at c:\Users\X\Documents\Temp\testCLI1.1\node_modules\nativescript\lib\
nativescript-cli.js:17:45
    - - - - -
    at Logger.getLevel (c:\Users\X\Documents\Temp\testCLI1.1\node_modules\
nativescript\lib\common\logger.js:37:40)
    at c:\Users\X\Documents\Temp\testCLI1.1\node_modules\nativescript\lib\
common\dispatchers.js:20:31
    at Function.settle (c:\Users\X\Documents\Temp\testCLI1.1\node_modules\
nativescript\node_modules\fibers\future.js:220:26)
    at FiberFuture.Future.wait (c:\Users\X\Documents\Temp\testCLI1.1\node_
modules\nativescript\node_modules\fibers\future.js:487:10)
    at c:\Users\X\Documents\Temp\testCLI1.1\node_modules\nativescript\lib\
nativescript-cli.js:17:45

Notice: there nothing after "--log"

PR #582, opened 2015-06-22
Add dashedOptions for each command

rosen-vladimirov:
Each command can use specific dashed options (`-- avd <name>` for example is valid for one command, but not for other). Add dashedOptions parameter to ICommand interface and make sure to use it for validation when it exists. In case there's no dashedOptions paramter, use the common options. In order to validate the options, delete yargs from require and require it again as yargs.options(this.options).argv is caching the results and calling it more than once with different options is giving strange results.
Also fixes https://github.com/NativeScript/nativescript-cli/issues/539


Fatme:
:+1:

Issue #520, opened 2015-06-03
Enable image resources from App_Resources in iOS


vakrilov:
Image resources added in the `App_Resources` should be available in the application.

Currently, all files from App_Resources are copied in the `Resources/icons` folder inside the xCode project, but they are not included in the xCode project itself, so they are not accessible in the application.

**Workaround**
1. Run `tns prepare ios` after the images are added in the `App_Resources` folder. The images will be copied in the `platforms\ios\<project-name>\Resources\Icons`
2. Open the xCode project from the `platforms\ios\` folder and add the images to the project manually:
![ios-resources](https://cloud.githubusercontent.com/assets/4092076/8226292/0f0bfd36-15a6-11e5-9c6a-ecb693955c48.png)

PR #590, opened 2015-06-24
Enable image resources from App_Resources in iOS

Fatme:
Fixes https://github.com/NativeScript/nativescript-cli/issues/520
https://github.com/NativeScript/ios-runtime/pull/202

### DIFF
--- a/lib/services/android-project-service.ts
+++ b/lib/services/android-project-service.ts
@@ -51,7 +51,7 @@ class AndroidProjectService implements IPlatformProjectService {
 				frameworkFilesExtensions: [".jar", ".dat", ".so"],
 				configurationFileName: "AndroidManifest.xml",
 				configurationFilePath: path.join(this.$projectData.platformsDir, "android", "AndroidManifest.xml"),
-				mergeXmlConfig: [{ "nodename": "manifest", "attrname": "*" }]
+				mergeXmlConfig: [{ "nodename": "manifest", "attrname": "*" }, { "application": "manifest", "attrname": "*" }]
 			};
 		}
 
@@ -272,6 +272,10 @@ class AndroidProjectService implements IPlatformProjectService {
 
 	public getFrameworkFilesExtensions(): string[] {
 		return [".jar", ".dat"];
+	}
+	
+	public prepareProject(): IFuture<void> {
+		return (() => { }).future<void>()();
 	}
 
 	private copy(projectRoot: string, frameworkDir: string, files: string, cpArg: string): IFuture<void> {
--- a/lib/services/ios-project-service.ts
+++ b/lib/services/ios-project-service.ts
@@ -182,13 +182,11 @@ class IOSProjectService implements  IPlatformProjectService {
             this.$fs.ensureDirectoryExists(targetPath).wait();
             shell.cp("-R", libraryPath, targetPath);
 
-            var pbxProjPath = path.join(platformData.projectRoot, this.$projectData.projectName + ".xcodeproj", "project.pbxproj");
-            var project = new xcode.project(pbxProjPath);
-            project.parseSync();
+            let project = this.createPbxProj();
 
             project.addFramework(path.join(targetPath, frameworkName + ".framework"), { customFramework: true, embed: true });
             project.updateBuildProperty("IPHONEOS_DEPLOYMENT_TARGET", "8.0");
-            this.$fs.writeFile(pbxProjPath, project.writeSync()).wait();
+            this.savePbxProj(project).wait();
             this.$logger.info("The iOS Deployment Target is now 8.0 in order to support Cocoa Touch Frameworks.");
         }).future<void>()();
 	}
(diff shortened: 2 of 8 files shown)


ErjanGavalji:
:+1:

lib/services/ios-project-service.ts
```diff
@@ -221,11 +219,58 @@ class IOSProjectService implements  IPlatformProjectService {
...
+			this.$logger.trace("Current images from App_Resources");
+			this.$logger.trace(currentImageNames);
+
+			if(imageNames.length !== currentImageNames.length) {
+				let imagesToAdd = _.difference(currentImageNames, imageNames);
+				this.$logger.trace(`New images to add into xcode project: ${ imagesToAdd.length !== 0 ? imagesToAdd.join(", ") : "" }`);			
+				_.each(imagesToAdd, image => project.addResourceFile(path.join(this.platformData.appResourcesDestinationDirectoryPath, image)));
+				
+				let imagesToRemove = _.difference(imageNames, currentImageNames);
+				this.$logger.trace(`Images to remove from xcode project: ${ imagesToRemove.length !== 0 ? imagesToRemove.join(", ") : "" }`);
```
teobugslayer: You can omit the length check.

```
builder@ab-linuxbpc-1:~/.ssh$ node
> [].join(", ")
''
>
```

lib/services/ios-project-service.ts
```diff
@@ -221,11 +219,58 @@ class IOSProjectService implements  IPlatformProjectService {
...
+		return (() => {
+			let project = this.xcodeproj;
+			let resources = project.pbxGroupByName("Resources");
+			let references = project.pbxFileReferenceSection();
+			
+			let images = _(<any[]>resources.children)
+				.map(resource => references[resource.value])
+				.value();
+			
+			let imageNames = _.map(images, image => image.name.replace(/\"/g, ""));
```
teobugslayer: You use images only here, so you can join the two maps together

teobugslayer: I think this variable is poorly named. the log.trace line bellow suggests a better name: xcodeProjectImages

lib/services/ios-project-service.ts
```diff
@@ -221,11 +219,58 @@ class IOSProjectService implements  IPlatformProjectService {
...
+			let currentImageNames = this.$fs.readDirectory(this.platformData.appResourcesDestinationDirectoryPath).wait();
```
teobugslayer: Again, the name should be something like "appResourcesImages"

lib/services/ios-project-service.ts
```diff
@@ -221,11 +219,58 @@ class IOSProjectService implements  IPlatformProjectService {
...
+			
+			this.$logger.trace("Images from xcode project");
+			this.$logger.trace(imageNames);
+			this.$logger.trace("Current images from App_Resources");
+			this.$logger.trace(currentImageNames);
+
+			if(imageNames.length !== currentImageNames.length) {
```
teobugslayer: What happens if I delete a file in App_Resources and add another one? Or rather, if I rename an existing file?

lib/services/ios-project-service.ts
rosen-vladimirov: Or maybe:
```TypeScript
let imageNames = _.map(<any[]>resources.children, resource => references[resource.value].name.replace(/\"/g, ""));
```

lib/services/ios-project-service.ts
rosen-vladimirov: is it correct to have this check? What if the count is the same, but some of the names are different?

lib/services/android-project-service.ts
```diff
@@ -273,6 +273,10 @@ class AndroidProjectService implements IPlatformProjectService {
 	public getFrameworkFilesExtensions(): string[] {
 		return [".jar", ".dat"];
 	}
+	
+	public prepareProject(): IFuture<void> {
+		return (() => { }).future<void>()();
```
ligaz: Maybe `_.noop()` here.

lib/services/ios-project-service.ts
```diff
@@ -182,13 +182,11 @@ class IOSProjectService implements  IPlatformProjectService {
             this.$fs.ensureDirectoryExists(targetPath).wait();
             shell.cp("-R", libraryPath, targetPath);
 
-            var pbxProjPath = path.join(platformData.projectRoot, this.$projectData.projectName + ".xcodeproj", "project.pbxproj");
-            var project = new xcode.project(pbxProjPath);
-            project.parseSync();
+           let project = this.xcodeproj;
```
ligaz: Indent me.

lib/services/ios-project-service.ts
```diff
@@ -221,11 +219,58 @@ class IOSProjectService implements  IPlatformProjectService {
...
 		}).future<void>()();
 	}
+	
+	public prepareProject(): IFuture<void> {
+		return (() => {
+			let project = this.xcodeproj;
+			let resources = project.pbxGroupByName("Resources");
+			let references = project.pbxFileReferenceSection();
+			
+			let images = _(<any[]>resources.children)
```
ligaz: Why complicating this with lodash when map is built-in?

lib/services/ios-project-service.ts
```diff
@@ -221,11 +219,58 @@ class IOSProjectService implements  IPlatformProjectService {
...
+				.map(resource => references[resource.value])
+				.value();
+			
+			let imageNames = _.map(images, image => image.name.replace(/\"/g, ""));
+			let currentImageNames = this.$fs.readDirectory(this.platformData.appResourcesDestinationDirectoryPath).wait();
+			
+			this.$logger.trace("Images from xcode project");
```
ligaz: Xcode is the correct spelling.

lib/services/android-project-service.ts
rosen-vladimirov: we have to return future, maybe ```Future.fromResult()``` :)

lib/services/ios-project-service.ts
ligaz: Maybe `_.difference`

lib/services/ios-project-service.ts
```diff
@@ -221,11 +219,58 @@ class IOSProjectService implements  IPlatformProjectService {
...
+	}
+	
+	private get xcodeproj(): any {
+		let project = new xcode.project(this.pbxProjPath);
+		project.parseSync();
+		
+		return project;
+	}
+	
+	private saveXcodeproj(project: any): IFuture<void> {
```
ligaz: This is better named `savePbxProj`. The same goes for the getter.

lib/services/ios-project-service.ts
```diff
@@ -221,11 +219,55 @@ class IOSProjectService implements  IPlatformProjectService {
...
+			
+		}).future<void>()();
+	}
+	
+	private get pbxProjPath(): string {
+		return path.join(this.platformData.projectRoot, this.$projectData.projectName + ".xcodeproj", "project.pbxproj");
+	}
+	
+	private get pbxProj(): any {
+		let project = new xcode.project(this.pbxProjPath);
```
rosen-vladimirov: is it expected to return new project every time you try using the property pbxProj?

Fatme: This function is called only one time - https://github.com/NativeScript/nativescript-cli/pull/590/files#diff-b673f137deb4afccccc0bd5f1572c2a0R229

ligaz: So we can make it a function instead of property then.

teobugslayer:
:+1: after addressing @rosen-vladimirov and @ligaz comments.

rosen-vladimirov:
:+1:

PR #597, opened 2015-06-26
Fix broken android prepare

Fatme:


rosen-vladimirov:
:+1:  after green build

PR #603, opened 2015-06-29
Do not init new project when install command is executed

Fatme:


dtopuzov:
:100:

rosen-vladimirov:
https://github.com/NativeScript/nativescript-cli/blob/fatme/install/lib/commands/install.ts#L4 - this line should be removed, it is not used anymore

lib/commands/install.ts
```diff
@@ -9,8 +9,10 @@ export class InstallCommand implements ICommand {
 	constructor(private $fs: IFileSystem,
 		private $errors: IErrors,
 		private $logger: ILogger,
```
rosen-vladimirov: `fs`, `errors` and `logger` are not used anywhere

lib/commands/install.ts
```diff
@@ -9,8 +9,10 @@ export class InstallCommand implements ICommand {
...
-		private $options: IOptions,
-		private $injector: IInjector,
+		private $platformsData: IPlatformsData,
+		private $platformService: IPlatformService,
+		private $projectData: IProjectData,
+		private $projectDataService: IProjectDataService,
 		private $staticConfig: IStaticConfig) { }
```
rosen-vladimirov: staticConfig is not used anywhere

rosen-vladimirov:
Maybe this is out of the scope for current PR, but I've tried using `*` for tns-android:
```JSON
{
        "nativescript": {
                "id": "org.nativescript.app1004",
                "tns-android": {
                        "version": "*"
                }
        },
        "dependencies": {
                "lodash": "^3.9.3"
        }
}
```
And the result is:
```
Copying template files...
Failed to retrieve data from npm. Please try again a little bit later.. Error: Error: ENOENT, stat 'C:\Users\vladimirov\AppData\Roaming\npm-cache\tns-android\*\package'
```
Same happened when I tried `^1.1.0`

teobugslayer:
@rosen-vladimirov -definitely outside the scope of this PR. I think we should support such versions but we may wait until the need raises.

Fatme:
I totally agree with @teobugslayer .

rosen-vladimirov:
:+1:  after green build

PR #604, opened 2015-06-30
Open web inspector frontend in a standalone application instead of Safari

KristinaKoeva:
The update of JavaScriptCore that we use in the iOS bridge goes with an updated web inspector frontend. The updated frontend is not rendered properly in latest Safari, so as a workaround open an application that would be distributed with the iOS bridge package

Related to https://github.com/NativeScript/ios-runtime/pull/201

### DIFF
--- a/lib/services/ios-debug-service.ts
+++ b/lib/services/ios-debug-service.ts
@@ -204,12 +204,15 @@ class IOSDebugService implements IDebugService {
 
     private openDebuggingClient(): IFuture<void> {
         return (() => {
-            var cmd = "open -a Safari " + this.getSafariPath().wait();
+            let inspectorPath = this.getInspectorPath().wait();
+            let inspectorApplicationPath = path.join(inspectorPath, "NativeScript Inspector.app");
+            let inspectorSourceLocation = path.join(inspectorPath, "Safari/Main.html");
+            let cmd = `open -a '${inspectorApplicationPath}' --args '${inspectorSourceLocation}' '${this.$projectData.projectName}'`;
             this.$childProcess.exec(cmd).wait();
         }).future<void>()();
     }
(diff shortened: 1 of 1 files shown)


Fatme:
:+1:

lib/services/ios-debug-service.ts
```diff
@@ -204,12 +205,15 @@ class IOSDebugService implements IDebugService {
 
     private openDebuggingClient(): IFuture<void> {
         return (() => {
-            var cmd = "open -a Safari " + this.getSafariPath().wait();
+            var inspectorPath = this.getInspectorPath().wait();
```
ligaz: It is better to use `let` instead of `var` here.

lib/services/ios-debug-service.ts
```diff
@@ -204,12 +205,15 @@ class IOSDebugService implements IDebugService {
...
+            var inspectorApplicationPath = path.join(inspectorPath, "NativeScript Inspector.app");
+            var inspectorSourceLocation = path.join(inspectorPath, "Safari/Main.html");
+            var cmd = util.format("open -a '%s' --args '%s' '%s'", inspectorApplicationPath, inspectorSourceLocation, this.$projectData.projectName);
```
ligaz: You can use [template strings](https://developers.google.com/web/updates/2015/01/ES6-Template-Strings) to simplify the code here.

teobugslayer: If you use them, then you won't need ```require "util"``` any more.

KristinaKoeva:
ping @ligaz @teobugslayer

ligaz:
:+1:

teobugslayer:
:+1:

Issue #600, opened 2015-06-29
tns init command


Fatme:
Creates a new NativeScript project in current directory or from specified package.json file

PR #605, opened 2015-06-30
Init command

Fatme:
Should be merged after this https://github.com/telerik/mobile-cli-lib/pull/371
https://github.com/NativeScript/nativescript-cli/issues/600

### DIFF
--- a/docs/man_pages/project/creation/create.md
+++ b/docs/man_pages/project/creation/create.md
@@ -17,3 +17,12 @@ Creates a new project for native development with NativeScript from the default 
 For projects that target iOS, you can use uppercase or lowercase letters, numbers, and hyphens.<% } %>    
 * `<App ID>` is the application identifier for your project. It must be a domain name in reverse and must meet the requirements of all platforms that you want to target. If not specified, the application identifier is set to `org.nativescript.<App name>` <% if(isConsole) { %>For more information about the `<App ID>` requirements, run `$ tns help create`<% } %><% if(isHtml) { %>For projects that target Android, you can use uppercase or lowercase letters, numbers, and underscores in the strings of the reversed domain name, separated by a dot. Strings must be separated by a dot and must start with a letter. For example: `com.nativescript.My_Andro1d_App`  
 For projects that target iOS, you can use uppercase or lowercase letters, numbers, and hyphens in the strings of the reversed domain name. Strings must be separated by a dot. For example: `com.nativescript.My-i0s-App`.<% } %>
+
+<% if(isHtml) { %> 
+### Related Commands
+
+Command | Description
+----------|----------
+[init](init.html) | Initializes a project for development. The command prompts you to provide your project configuration interactively and uses the information to create a new package.json file or update the existing one.
+[install](install.html) | Installs all platforms and dependencies described in the `package.json` file in the current directory.
+<% } %> 
--- a/docs/man_pages/project/creation/init.md
+++ b/docs/man_pages/project/creation/init.md
@@ -1,0 +1,21 @@
+init
+==========
+
+Usage | Synopsis
+---|---
+General | `$ tns init [--path <Directory>] [--force]`
+
+Initializes a project for development. The command prompts you to provide your project configuration interactively and uses the information to create a new `package.json` file or update the existing one.
+
+### Options
+* `--path` - Specifies the directory where you want to initialize the project, if different from the current directory. The directory must be empty.
+* `--force` - If set, applies the default project configuration and does not show the interactive prompt. The default project configuration targets the latest official runtimes and sets `org.nativescript.<folder_name>` for application identifier.
+
+<% if(isHtml) { %> 
+### Related Commands
+
+Command | Description
+----------|----------
+[create](create.html) | Creates a new project for native development with NativeScript from the default template or from an existing NativeScript project.
+[install](install.html) | Installs all platforms and dependencies described in the `package.json` file in the current directory.
+<% } %> 
--- a/lib/commands/init.ts
+++ b/lib/commands/init.ts
@@ -1,0 +1,16 @@
+///<reference path="../.d.ts"/>
+"use strict";
+
+import Future = require("fibers/future");
+
+export class InitCommand implements ICommand {
+	constructor(private $initService: IInitService) { }
+		
+	public allowedParameters: ICommandParameter[] = [];
+	public enableHooks = false;
+	
+	public execute(args: string[]): IFuture<void> {
+		return this.$initService.initialize();
+	}
+}
+$injector.registerCommand("init", InitCommand);
--- a/lib/services/init-service.ts
+++ b/lib/services/init-service.ts
@@ -1,0 +1,120 @@
+///<reference path="../.d.ts"/>
+"use strict";
+
+import constants = require("./../constants");
+import helpers = require("./../common/helpers");
+import path = require("path");
+import semver = require("semver");
+
+export class InitService implements IInitService {
+	private static MIN_SUPPORTED_FRAMEWORK_VERSIONS: IStringDictionary = {
+		"tns-ios": "1.1.0",
+		"tns-android": "1.1.0"	
+	};
+	
+	private _projectFilePath: string;
+	
+	constructor(private $fs: IFileSystem,
+		private $errors: IErrors,
+		private $logger: ILogger,
+		private $options: IOptions,
+		private $injector: IInjector,
+		private $staticConfig: IStaticConfig,
+		private $projectHelper: IProjectHelper,
+		private $prompter: IPrompter,
+		private $npm: INodePackageManager,
+		private $npmInstallationManager: INpmInstallationManager) { }
+	
+	public initialize(): IFuture<void> {
+		return (() => {
+			let projectData: any = { };
+			
+			if(this.$fs.exists(this.projectFilePath).wait()) {
+				projectData = this.$fs.readJson(this.projectFilePath).wait();
+			}
+			
+			let projectDataBackup = _.extend({}, projectData);			
+			
+			if(!projectData[this.$staticConfig.CLIENT_NAME_KEY_IN_PROJECT_FILE]) {
+				projectData[this.$staticConfig.CLIENT_NAME_KEY_IN_PROJECT_FILE] = { };
+				this.$fs.writeJson(this.projectFilePath, projectData).wait(); // We need to create package.json file here in order to prevent "No project found at or above and neither was a --path specified." when resolving platformsData				
+			}
+			
+			try {
+				
+				projectData[this.$staticConfig.CLIENT_NAME_KEY_IN_PROJECT_FILE]["id"] = this.getProjectId().wait();
+				
+				if(this.$options.frameworkName && this.$options.frameworkVersion) {
+					projectData[this.$staticConfig.CLIENT_NAME_KEY_IN_PROJECT_FILE][this.$options.frameworkName] = this.buildVersionData(this.$options.frameworkVersion);
+				} else {
+					let $platformsData = this.$injector.resolve("platformsData");
+					_.each($platformsData.platformsNames, platform => {
+						let platformData: IPlatformData = $platformsData.getPlatformData(platform);
+						if(!platformData.targetedOS || (platformData.targetedOS && _.contains(platformData.targetedOS, process.platform))) {
+							projectData[this.$staticConfig.CLIENT_NAME_KEY_IN_PROJECT_FILE][platformData.frameworkPackageName] = this.getVersionData(platformData.frameworkPackageName).wait();
+						}
+					});
+				}
+				
+				this.$fs.writeJson(this.projectFilePath, projectData).wait();
+			} catch(err) {
+				this.$fs.writeJson(this.projectFilePath, projectDataBackup).wait();
+				throw err;
+			}
+			
+			this.$logger.out("Project successfully initialized."); 
+		}).future<void>()();
+	}
+	
+	private get projectFilePath(): string {
+		if(!this._projectFilePath) {
+			let projectDir = path.resolve(this.$options.path || ".");
+			this._projectFilePath = path.join(projectDir, constants.PACKAGE_JSON_FILE_NAME);
+		}
+		
+		return this._projectFilePath;
+	}
+	
+	private getProjectId(): IFuture<string> {
+		return (() => {
+			if(this.$options.appid) {
+				return this.$options.appid;
+			}
+			
+			let defaultAppId = this.$projectHelper.generateDefaultAppId(path.basename(path.dirname(this.projectFilePath)), constants.DEFAULT_APP_IDENTIFIER_PREFIX);
+			if(this.useDefaultValue) {
+				return defaultAppId;
+			}
+			
+			return this.$prompter.getString("Id:", () => defaultAppId).wait();
+		}).future<string>()();
+	}
+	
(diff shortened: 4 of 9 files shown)


docs/man_pages/project/creation/init.md
```diff
@@ -0,0 +1,20 @@
+init
+==========
+
+Usage | Synopsis
+---|---
+General | `$ tns init [--path <Directory>] [--force]`
+
+Initializes a project for development. This will ask you a bunch of questions, and then write a package.json for you.
```
teobugslayer: "ask you a bunch of questions" does not look very professional to me. Perhaps "it will ask you interactively to configure the project". Ping @ikoevska

lib/commands/init.ts
```diff
@@ -0,0 +1,108 @@
...
+			});
+			
+			this.$fs.writeJson(this.projectFilePath, projectData).wait();
+			
+			this.$logger.out("Project successfully initialized."); 
+		}).future<void>()();
+	}
+	
+	public canExecute(args: string[]): IFuture<boolean> {
+		return (() => {
```
teobugslayer: you can simplify it: return feature.fromResult(true);

lib/commands/init.ts
```diff
@@ -0,0 +1,108 @@
...
+			
+			let $platformsData = this.$injector.resolve("platformsData");
+			_.each($platformsData.platformsNames, platform => {
+				let platformData: IPlatformData = $platformsData.getPlatformData(platform);
+				if(!platformData.targetedOS || (platformData.targetedOS && _.contains(platformData.targetedOS, process.platform))) {
+					projectData[this.$staticConfig.CLIENT_NAME_KEY_IN_PROJECT_FILE][platformData.frameworkPackageName] = this.getVersionData(platformData.frameworkPackageName).wait();
+				}
+			});
+			
+			this.$fs.writeJson(this.projectFilePath, projectData).wait();
```
teobugslayer: I think, when this write fails, we should undo the 1st writeJson as well.

lib/commands/init.ts
```diff
@@ -0,0 +1,108 @@
...
+		return (() => {
+			let latestVersion = this.$npmInstallationManager.getLatestVersion(packageName).wait();
+			if(this.canUseDefaultValue) {
+				return this.buildVersionData(latestVersion);
+			}
+			
+			let data = this.$npm.view(packageName, "versions").wait();
+			let versions = data[latestVersion].versions;
+			let sortedVersions = versions.sort(helpers.versionCompare).reverse();
+			let version = this.$prompter.promptForChoice(`${packageName} version:`, sortedVersions).wait();
```
teobugslayer: How can we avoid this prompt? A command line parameter, perhaps?

lib/commands/init.ts
```diff
@@ -0,0 +1,108 @@
...
+			let defaultAppId = this.$projectHelper.generateDefaultAppId(path.basename(path.dirname(this.projectFilePath)), constants.DEFAULT_APP_IDENTIFIER_PREFIX);
+			if(this.canUseDefaultValue) {
+				return defaultAppId;
+			}
+			
+			return this.$prompter.getString("Id:", () => defaultAppId).wait();
+		}).future<string>()();
+	}
+	
+	private getVersionData(packageName: string): IFuture<any> {
```
teobugslayer: You can return IStringDictionary instead of any

lib/commands/init.ts
```diff
@@ -0,0 +1,114 @@
...
+			}
+			
+			if(!projectData[this.$staticConfig.CLIENT_NAME_KEY_IN_PROJECT_FILE]) {
+				projectData[this.$staticConfig.CLIENT_NAME_KEY_IN_PROJECT_FILE] = { };
+			}
+			
+			let projectDataBackup = _.extend({}, projectData);
+			
+			try {
+				this.$fs.writeJson(this.projectFilePath, projectData).wait(); // We need to create package.json file here in order to prevent "No project found at or above and neither was a --path specified." when resolving platformsData
```
rosen-vladimirov: maybe this should be inside 
```TypeScript
if(!projectData[this.$staticConfig.CLIENT_NAME_KEY_IN_PROJECT_FILE])
```
as this is the only case when we do not have package.json or it is missing required information. In all cases when we do not "enter" the mentioned if, we should not write the package.json again as it is already existing and correct.

lib/commands/init.ts
```diff
@@ -0,0 +1,114 @@
...
+			
+			if(this.$fs.exists(this.projectFilePath).wait()) {
+				projectData = this.$fs.readJson(this.projectFilePath).wait();
+			}
+			
+			if(!projectData[this.$staticConfig.CLIENT_NAME_KEY_IN_PROJECT_FILE]) {
+				projectData[this.$staticConfig.CLIENT_NAME_KEY_IN_PROJECT_FILE] = { };
+			}
+			
+			let projectDataBackup = _.extend({}, projectData);
```
rosen-vladimirov: This should be before
```TypeScript
if(!projectData[this.$staticConfig.CLIENT_NAME_KEY_IN_PROJECT_FILE]) {
```
At this point we might have changed the projectData, so this is not a real backup of the original data.

lib/commands/init.ts
```diff
@@ -0,0 +1,114 @@
...
+	private getVersionData(packageName: string): IFuture<IStringDictionary> {
+		return (() => {
+			let latestVersion = this.$npmInstallationManager.getLatestVersion(packageName).wait();
+			if(this.canUseDefaultValue) {
+				return this.buildVersionData(latestVersion);
+			}
+			
+			let data = this.$npm.view(packageName, "versions").wait();
+			let versions = data[latestVersion].versions;
+			let sortedVersions = versions.sort(helpers.versionCompare).reverse();
```
rosen-vladimirov: I suggest to have minSupportedVersion, because currently we show all versions and the user may set something that will not be buildable after that, for example for tns-android I set `0.1.0`

Fatme: I'm not sure how can I select "minSupportedVersion"...

lib/commands/init.ts
```diff
@@ -0,0 +1,114 @@
...
+			let version = this.$prompter.promptForChoice(`${packageName} version:`, sortedVersions).wait();
+			return this.buildVersionData(version);
+		}).future<IStringDictionary>()();
+	}
+	
+	private buildVersionData(version: string) {
+		 return { "version": version };
+	}
+	
+	private get canUseDefaultValue(): boolean {
```
rosen-vladimirov: I think this name is confusing - from the code it looks like this property defines when we **must** use the default value. From the current name it looks like we just check if we can use it, but we'll decide later if we'll use default value or not.

Fatme: I'm opening to any suggestions.

lib/commands/init.ts
rosen-vladimirov: My suggestion is to define it in the platformData and pass it as parameter to this method. This way we'll set it to the minimum verified version that we are able to work with.

lib/commands/init.ts
rosen-vladimirov: Maybe just `useDefaultValue` is enough. Also it could be `shouldUseDefaultValue` or `mustUseDefaultValue`.

lib/commands/init.ts
Fatme: " This way we'll set it to the minimum verified version that we are able to work with." - every next version has some improvements and fixes. I'm not able to choice if for example 1.0.0 is suitable to be "minSupportedVersion".

rosen-vladimirov:
`npm init` checks node_modules and adds each of them directly in the dependencies section of the created package.json
It will be nice to support this in case you do not have package.json (in case it already exists, I assume you have already set the dependencies you want). 
@ligaz  what do you think (we can consider this for another PR)

Fatme:
It is strange to have node_modules and doesn't have `package.json` file.

lib/commands/init.ts
rosen-vladimirov: Well, if I set 0.1.0 and execute `tns install`, the result is:
```
~/Work/nativescript-cli/scratch/app16 $ cat package.json 
{
	"nativescript": {
		"id": "org.nativescript.app16",
		"tns-android": {
			"version": "0.1.0"
		}
	}
}~/Work/nativescript-cli/scratch/app16 $ tns install
Project app16 was successfully created
Adding platforms...
Copying template files...
ENOENT, scandir '/home/rvladimirov/.npm/tns-android/0.1.0/package/framework/res'
```
Of course all versions have improvements, but when you use init, I think we must show you only recent versions, why would you init project with version 0.4.2 for example?

Fatme: This means that the prompt is unneeded?

lib/commands/init.ts
teobugslayer: I like ```useDefaultValue``` the most.

lib/commands/init.ts
rosen-vladimirov: My idea is to guide the people to use latest versions.

docs/man_pages/project/creation/create.md
```diff
@@ -16,4 +16,12 @@ Creates a new project for native development with NativeScript from the default
...
-For projects that target iOS, you can use uppercase or lowercase letters, numbers, and hyphens in the strings of the reversed domain name. Strings must be separated by a dot. For example: `com.nativescript.My-i0s-App`.<% } %>
\ No newline at end of file
+For projects that target iOS, you can use uppercase or lowercase letters, numbers, and hyphens in the strings of the reversed domain name. Strings must be separated by a dot. For example: `com.nativescript.My-i0s-App`.<% } %>
+
+<% if(isHtml) { %> 
+### Related Commands
+
+Command | Description
+----------|----------
+[init](init.html) | Initializes a project for development. It will ask you interactively to configure the project and will write a package.json for you.
```
ikoevska: Remove the second sentence from the description.

docs/man_pages/project/creation/init.md
```diff
@@ -0,0 +1,20 @@
+init
+==========
+
+Usage | Synopsis
+---|---
+General | `$ tns init [--path <Directory>] [--force]`
+
+Initializes a project for development. It will ask you interactively to configure the project and will write a package.json for you.
```
ikoevska: Initializes a project for development. The command prompts you to provide your project configuration interactively and uses the information to create a new `package.json` file or update the existing one.

docs/man_pages/project/creation/init.md
```diff
@@ -0,0 +1,20 @@
...
+
+### Options
+* `--path` - Specifies the directory where you want to initialize the project, if different from the current directory. The directory must be empty.
+* `--force` - Will use only defaults and not prompt you for any options.
```
ikoevska: If set, applies the default project configuration and does not show the interactive prompt. The default project configuration targets the latest official runtimes and sets `org.nativescript.<folder_name>` for application identifier.

docs/man_pages/project/creation/init.md
```diff
@@ -0,0 +1,20 @@
...
+
+<% if(isHtml) { %> 
+### Related Commands
+
+Command | Description
+----------|----------
+[create](create.html) | Creates a new project for native development with NativeScript from the default template or from an existing NativeScript project.
+<% } %>
```
ikoevska: Add the install command here, as it is actually related to the init workflow.

docs/man_pages/project/creation/create.md
Fatme: install command is in `project/configuration` section and the init command is in `project/creation`. Do I need to relate them?

lib/services/init-service.ts
```diff
@@ -0,0 +1,125 @@
...
+		if(!this._projectFilePath) {
+			let projectPath = path.resolve(this.$options.path || ".");
+			this._projectFilePath = path.join(projectPath, constants.PACKAGE_JSON_FILE_NAME);
+		}
+		
+		return this._projectFilePath;
+	}
+	
+	private set projectFilePath(newProjectFilePath: string) {
+		this._projectFilePath = newProjectFilePath;
```
teobugslayer: Should we verify that newProjectFilePath is correct? I.e. is non-empty string and possibly that exists on the fs?

lib/services/init-service.ts
```diff
@@ -0,0 +1,125 @@
...
+				throw err;
+			}
+			
+			this.$logger.out("Project successfully initialized."); 
+		}).future<void>()();
+	}
+	
+	private get projectFilePath(): string {
+		if(!this._projectFilePath) {
+			let projectPath = path.resolve(this.$options.path || ".");
```
teobugslayer: This is more a "projectDir", consider renaming it.

lib/commands/init.ts
```diff
@@ -0,0 +1,20 @@
...
+	constructor(private $initService: IInitService) { }
+		
+	public allowedParameters: ICommandParameter[] = [];
+	public enableHooks = false;
+	
+	public execute(args: string[]): IFuture<void> {
+		return this.$initService.initialize();
+	}
+	
+	public canExecute(args: string[]): IFuture<boolean> {
```
rosen-vladimirov: As you are always returning true, I believe it is better to use `
```TypeScript
public allowedParameters: ICommandParameter[] = [];
```
and remove canExecute. 
This way you'll get the benefit of commandsService canExecute logic, which will validate the number of parameters.
In case you keep this canExecute, the user will be able to write:
`tns init goshko peshko android test` and it will work just fine.

lib/services/init-service.ts
```diff
@@ -0,0 +1,125 @@
+///<reference path="../.d.ts"/>
+"use strict";
+
+import constants = require("./../constants");
+import helpers = require("./../common/helpers");
+import Future = require("fibers/future");
```
rosen-vladimirov: Future is not used anywhere, you can remove it

docs/man_pages/project/creation/init.md
```diff
@@ -0,0 +1,20 @@
+init
+==========
+
+Usage | Synopsis
+---|---
+General | `$ tns init [--path <Directory>] [--force]`
+
+Initializes a project for development. The command prompts you to provide your project configuration interactively and uses the information to create a new package.json file or update the existing one.
```
ikoevska: code formatting for package.json is missing.

docs/man_pages/project/creation/init.md
```diff
@@ -0,0 +1,20 @@
...
+
+### Options
+* `--path` - Specifies the directory where you want to initialize the project, if different from the current directory. The directory must be empty.
+* `--force` - If set, applies the default project configuration and does not show the interactive prompt. The default project configuration targets the latest official runtimes and sets org.nativescript.<folder_name> for application identifier.
```
ikoevska: code formatting for org.nativescript... otherwise, the stuff in the triangle brackets is considered an HTML tag and breaks the output

lib/services/init-service.ts
```diff
@@ -0,0 +1,120 @@
...
+				this.$logger.info(`Only ${versions[0]} version is available for ${packageName} framework.`);
+				return this.buildVersionData(versions[0]);
+			}
+			let sortedVersions = versions.sort(helpers.versionCompare).reverse();
+			let version = this.$prompter.promptForChoice(`${packageName} version:`, sortedVersions).wait();
+			return this.buildVersionData(version);
+		}).future<IStringDictionary>()();
+	}
+	
+	private buildVersionData(version: string) {
```
teobugslayer: Consider declaring return type for the function

rosen-vladimirov:
:+1:

ikoevska:
:+1:

PR #610, opened 2015-07-01
Ensure all npm deps are installed when no platform is specified in package.json and aggregate errors

Fatme:


rosen-vladimirov:
:+1:

PR #611, opened 2015-07-02
Fix help for install command

rosen-vladimirov:
Fix help for install command, --path option description changed.


ikoevska:
:+1:

PR #616, opened 2015-07-02
Add ignoreScripts option

rosen-vladimirov:
Add ignoreScripts option, which should be used when during installation of modules you want to skip all scripts declared in package.json's of these modules. When you specify --ignoreScripts, we just add `--ignore-scripts` to npm install command.


rosen-vladimirov:
Closing as I need this for release branch

Issue #510, opened 2015-05-31
NativeScript Plugins support


ligaz:
This issue outlines the work breakdown for supporting [NativeScript Plugins in the CLI](https://github.com/NativeScript/NativeScript/issues/25)

### Plugin commands
- [x] `add` - will support everything `npm install` supports. Will save the installed plugin in `package.json`. The plugin itself will be installed in `node_modules`.
- [x] `remove` - will remove the plugin. Basically `npm uninstall ... --save` with some other magic.
- [x] `prepare` - hidden command that will do all the required mumbo-jumbo to install a plugin. Actually adding a plugin will be decomposed to `npm install` followed by `tns plugin prepare`

### Supported functionality
- [x] NPM based workflow.
- [x] `require` of JavaScript files from the plugin. `prepare` will ensure the script files are in the native platform `tns_modules` folder, so that they can be required. The `package.json` of the plugin should also be there in case there is a `main` declared.
- [x] Merging of native configuration files (`AndroidManifest.xml` and `Info.plist`)

## v1.2
- Ability to package native libraries and frameworks in the plugin.
  - Android
    - [x] Respect Android libraries added to `platforms/android`
    - [x] Respect jars added to `platforms/android/libs`
    - [x] Remove the library when removing the plugin
  - iOS
    - [x] Support for dynamic libraries (*.framework)

## v1.3
- [x] Support for Cocoapods (#729)
- [x] Support for new Android Library format (AAR)

## v1.4
- [x] Support for static iOS libraries

## v1.5

- [ ] Ability to package resources (images, xml files declaring native UI, etc.) in the plugin.
- [x] Ability to define plugin variables (ex: API keys that should be defined per plugin installation)

valentinstoychev:
related:
https://github.com/NativeScript/nativescript-cli/issues/592
https://github.com/NativeScript/nativescript-cli/issues/591

Plamen5kov:
I don't believe it's a good idea to:
`Ability to package resources (images, xml files declaring native UI, etc.) in the plugin.`
At least not directly. The preferred way is to have only an `.aar` file that includes all of the elements described above.

PR #638, opened 2015-07-13
Support for dynamic frameworks in iOS from plugins

Fatme:
https://github.com/NativeScript/nativescript-cli/issues/510

### DIFF
--- a/lib/definitions/project.d.ts
+++ b/lib/definitions/project.d.ts
@@ -20,6 +20,10 @@ interface IProjectTemplatesService {
 	defaultTemplatePath: IFuture<string>;
 }
 
+interface IPlatformProjectServiceBase {
+	getPluginPlatformsFolderPath(pluginData: IPluginData, platform: string): string;
+}
+
 interface IPlatformProjectService {
 	platformData: IPlatformData;
 	validate(): IFuture<void>;
@@ -30,7 +34,7 @@ interface IPlatformProjectService {
 	prepareProject(): IFuture<void>;
 	prepareAppResources(appResourcesDirectoryPath: string): IFuture<void>;
 	isPlatformPrepared(projectRoot: string): IFuture<boolean>;
-	addLibrary(platformData: IPlatformData, libraryPath: string): IFuture<void>;
+	addLibrary(libraryPath: string): IFuture<void>;
 	canUpdatePlatform(currentVersion: string, newVersion: string): IFuture<boolean>;
 	updatePlatform(currentVersion: string, newVersion: string): IFuture<void>;
 	preparePluginNativeCode(pluginData: IPluginData): IFuture<void>;
--- a/lib/services/android-project-service.ts
+++ b/lib/services/android-project-service.ts
@@ -10,8 +10,9 @@ import fs = require("fs");
 import os = require("os");
 
 import androidProjectPropertiesManagerLib = require("./android-project-properties-manager");
+import projectServiceBaseLib = require("./platform-project-service-base");
 
-class AndroidProjectService implements IPlatformProjectService {
+class AndroidProjectService extends projectServiceBaseLib.PlatformProjectServiceBase implements IPlatformProjectService {
 	private static MIN_SUPPORTED_VERSION = 17;
 	private SUPPORTED_TARGETS = ["android-17", "android-18", "android-19", "android-21", "android-22"]; // forbidden for now: "android-MNC"
 	private static ANDROID_TARGET_PREFIX = "android";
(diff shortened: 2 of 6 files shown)


lib/services/android-project-service.ts
```diff
@@ -273,13 +275,13 @@ class AndroidProjectService implements IPlatformProjectService {
 				let libsFolderContents = this.$fs.readDirectory(libsFolderPath).wait();
 				_(libsFolderContents)
 					.filter(libsFolderItem => path.extname(libsFolderItem) === ".jar")
-					.map(jar => this.addLibrary(this.platformData, path.join(libsFolderPath, jar)).wait())
+					.map(jar => this.addLibrary(path.join(libsFolderPath, jar)).wait())
```
teobugslayer: .map converts (maps) its input values to its output values. In this case, .each is a better choice.

lib/services/android-project-service.ts
```diff
@@ -28,13 +29,14 @@ class AndroidProjectService implements IPlatformProjectService {
 	constructor(private $androidEmulatorServices: Mobile.IEmulatorPlatformServices,
 		private $childProcess: IChildProcess,
 		private $errors: IErrors,
-		private $fs: IFileSystem,
+		$fs: IFileSystem,
```
rosen-vladimirov: as `fs` can be used only in constructor, can you move it below private dependencies

lib/services/android-project-service.ts
```diff
@@ -273,13 +275,13 @@ class AndroidProjectService implements IPlatformProjectService {
 				let libsFolderContents = this.$fs.readDirectory(libsFolderPath).wait();
```
rosen-vladimirov: is it possible to have jars not on the root level of the lib? As far as I understand from this code, we are taking only the files at the root level

lib/services/ios-project-service.ts
```diff
@@ -282,11 +286,26 @@ class IOSProjectService implements  IPlatformProjectService {
...
 	
 	public removePluginNativeCode(pluginData: IPluginData): IFuture<void> {
-		return Future.fromResult();
+		return (() => {
+			let pluginPlatformsFolderPath = pluginData.pluginPlatformsFolderPath(IOSProjectService.IOS_PLATFORM_NAME);
+			let project = this.createPbxProj();
+						
+			this.getAllDynamicFrameworksForPlugin(pluginData).wait().map(fileName => project.removeFramework(path.join(pluginPlatformsFolderPath, fileName + ".framework"), { customFramework: true, embed: true }));
+			
+	        this.savePbxProj(project).wait();
```
rosen-vladimirov: indent me, please :crying_cat_face:

lib/services/ios-project-service.ts
```diff
@@ -282,11 +286,26 @@ class IOSProjectService implements  IPlatformProjectService {
 	}
 	
 	public preparePluginNativeCode(pluginData: IPluginData): IFuture<void> {
-		return Future.fromResult();
+		return (() => {
+			let pluginPlatformsFolderPath = pluginData.pluginPlatformsFolderPath(IOSProjectService.IOS_PLATFORM_NAME);
+			this.getAllDynamicFrameworksForPlugin(pluginData).wait().map(fileName => this.addLibrary(path.join(pluginPlatformsFolderPath, fileName)).wait());
```
Mitko-Kerezov: You can also use `_.each` instead of `_.map` here

lib/services/ios-project-service.ts
```diff
@@ -282,11 +286,26 @@ class IOSProjectService implements  IPlatformProjectService {
...
+		}).future<void>()();
 	}
 	
 	public removePluginNativeCode(pluginData: IPluginData): IFuture<void> {
-		return Future.fromResult();
+		return (() => {
+			let pluginPlatformsFolderPath = pluginData.pluginPlatformsFolderPath(IOSProjectService.IOS_PLATFORM_NAME);
+			let project = this.createPbxProj();
+						
+			this.getAllDynamicFrameworksForPlugin(pluginData).wait().map(fileName => project.removeFramework(path.join(pluginPlatformsFolderPath, fileName + ".framework"), { customFramework: true, embed: true }));
```
Mitko-Kerezov: `_.each`

lib/services/platform-project-service-base.ts
```diff
@@ -0,0 +1,25 @@
...
+	public getPluginPlatformsFolderPath(pluginData: IPluginData, platform: string) {
+		return pluginData.pluginPlatformsFolderPath(platform);
+	}
+	
+	public getAllNativeLibrariesForPlugin(pluginData: IPluginData, platform: string, filter: (fileName: string, pluginPlatformsFolderPath: string) => boolean): IFuture<string[]> {
+		return (() => {
+			let pluginPlatformsFolderPath = this.getPluginPlatformsFolderPath(pluginData, platform);
+			if(pluginPlatformsFolderPath && this.$fs.exists(pluginPlatformsFolderPath).wait()) {
+				let platformsContents = this.$fs.readDirectory(pluginPlatformsFolderPath).wait();
+				return _(platformsContents)
```
Mitko-Kerezov: this logic can be rewritten with a local `string[]` variable  which is assigned to and then returned as to avoid a return statement inside the if block and to have only one return statement at the end of the function

lib/definitions/project.d.ts
```diff
@@ -20,6 +20,10 @@ interface IProjectTemplatesService {
 	defaultTemplatePath: IFuture<string>;
 }
 
+interface IPlatformProjectServiceBase {
+	getPluginPlatformsFolderPath(pluginData: IPluginData, platform: string): string;
```
Mitko-Kerezov: Some JSDocumentation would be nice for this interface and the `addLibrary` function below

Mitko-Kerezov:
:+1: After addressing comments

lib/services/android-project-service.ts
Mitko-Kerezov: This is the behavior we are aiming for - all libraries have to be at root level

PR #641, opened 2015-07-13
Do not delete values dirs

rosen-vladimirov:
During project preparation, we should not delete values directories as they contain specific information, which is important for android builds (strings.xml file for example).

NOTE: This fix is mandatory for integration of {N} inside AppBuilder

### DIFF
--- a/lib/services/android-project-service.ts
+++ b/lib/services/android-project-service.ts
@@ -185,9 +185,10 @@ class AndroidProjectService implements IPlatformProjectService {
 	public prepareAppResources(appResourcesDirectoryPath: string): IFuture<void> {
 		return (() => {
 			let resourcesDirPath = path.join(appResourcesDirectoryPath, this.platformData.normalizedPlatformName);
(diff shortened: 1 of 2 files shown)


lib/services/android-project-service.ts
```diff
@@ -185,9 +185,9 @@ class AndroidProjectService implements IPlatformProjectService {
 	public prepareAppResources(appResourcesDirectoryPath: string): IFuture<void> {
 		return (() => {
 			let resourcesDirPath = path.join(appResourcesDirectoryPath, this.platformData.normalizedPlatformName);
-			let resourcesDirs = this.$fs.readDirectory(resourcesDirPath).wait();
+			let resourcesDirs = this.$fs.readDirectory(resourcesDirPath).wait().filter(resDir => !resDir.match(/^values/));
```
teobugslayer: You can move the regexp outside the lambda, but not a merge stopper

teobugslayer:
:+1:

Issue #632, opened 2015-07-09
Android debugger does not work on some OSX systems


blagoev:
Moved https://github.com/NativeScript/android-runtime/issues/143 to CLI project

nmladenov:
tns debug android --debug-brk not working with attached device. No debugger started, error is thrown. Commented with NS QA team repproducible on their side too. "Command ant failed with exit code 1"

dtopuzov:
If name of Chrome Executable is not "Google Chrome" you should change ANDROID_DEBUG_UI_MAC entry in config/config.json in order to get debugger working.

vchimev:
Android debugger does not work when "ANDROID_DEBUG_UI_MAC": "Google Chrome" specified in config/config.json and Chrome executable is "Google Chrome".

teobugslayer:
The issue was that the fix was only in master branch. I ported it to release as well and the bug is ready for testing.

PR #642, opened 2015-07-13
Allow the user to manually set Chrome app name

teobugslayer:
fixes https://github.com/NativeScript/nativescript-cli/issues/632

### DIFF
--- a/docs/man_pages/project/testing/debug-android.md
+++ b/docs/man_pages/project/testing/debug-android.md
@@ -30,10 +30,10 @@ Debugs your project on a connected device, in a native emulator or in Genymotion
 * `<Emulator Options>` is any valid combination of options as listed by `$ tns help emulate android`
 * `<GenyName>` is the name of the Genymotion virtual device that you want to use as listed by `$ genyshell -c "devices list"`
  
-<% if(isHtml) { %> 
-### Prerequisites
+<% if(isHtml) { %>
+###Prerequisites
 
-* You must have Chrome installed on your system.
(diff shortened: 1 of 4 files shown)


docs/man_pages/project/testing/debug-android.md
```diff
@@ -35,6 +35,11 @@ Debugs your project on a connected device, in a native emulator or in Genymotion
 
 * You must have Chrome installed on your system.
 
+<% if(isMacOS) { %>
```
ikoevska: This tag won't work in the current isHtml block as the OS tags work in the console. My advice is to go with a `<br/>` right next to the period above. In this case, however, you would need to make this "If you are using a non-standard named Chrome on an OS X system (for example, a nightly Canary update), you need to set this name in the...

Also, please apply code formatting to the setting name.

```
`this is how you do code formatting`
```

docs/man_pages/project/testing/debug-android.md
```diff
@@ -30,10 +30,11 @@ Debugs your project on a connected device, in a native emulator or in Genymotion
...
 * `<GenyName>` is the name of the Genymotion virtual device that you want to use as listed by `$ genyshell -c "devices list"`
  
-<% if(isHtml) { %> 
-### Prerequisites
+<% if(isHtml) { %>
+###Prerequisites
 
-* You must have Chrome installed on your system.
+* You must have Chrome installed on your system.<br/>If you are using a non-standard named Chrome app on an OS X system (for example, a nightly Canary update),
+you need to set this name in the `ANDROID_DEBUG_UI_MAC` setting in the NativeScript [config.json](file:///<%= #{config.getConfigPath(config)} %>).
```
ikoevska: There's an unneeded new line here.

rosen-vladimirov:
:+1:

PR #651, opened 2015-07-16
Fix livesync help

rosen-vladimirov:
Fix livesync help and html generation which was failing due to missing <% if(isHtml) { <%> sections.
Update common lib.


Mitko-Kerezov:
:+1:

PR #668, opened 2015-07-21
Fix ios debugger for projects created with framework version lower than 1.2.0…

Fatme:
If the project is created with framework version lower than 1.2, `tns debug ios` command fails with the following error: "Command failed: /bin/sh -c open -a '~/.npm/tns-ios/1.1.2/package/WebInspectorUI/NativeScript Inspector.app' --args '~/.npm/tns-ios/1.1.2/package/WebInspectorUI/Safari/Main.html' 'newAppRelease'
FSPathMakeRef(~/.npm/tns-ios/1.1.2/package/WebInspectorUI/NativeScript Inspector.app) failed with error -43."

### DIFF
--- a/lib/services/ios-debug-service.ts
+++ b/lib/services/ios-debug-service.ts
@@ -6,6 +6,7 @@ import stream = require("stream");
 import path = require("path");
 import http = require("http");
 import Future = require("fibers/future");
+import semver = require("semver");
 
 module notification {
     function formatNotification(bundleId: string, notification: string) {
@@ -71,7 +72,8 @@ class IOSDebugService implements IDebugService {
         private $errors: IErrors,
         private $injector: IInjector,
         private $npmInstallationManager: INpmInstallationManager,
-        private $options: IOptions) { }
+        private $options: IOptions,
+        private $projectDataService: IProjectDataService) { }
 
     get platform(): string {
         return "ios";
@@ -142,6 +144,7 @@ class IOSDebugService implements IDebugService {
                     });
                     awaitNotification(npc, notification.readyForAttach(projectId), 5000).wait();
                 } catch(e) {
+                    this.$logger.trace(`Timeout error: ${e}`);
                     this.$errors.failWithoutHelp("Timeout waiting for NativeScript debugger.");
                 }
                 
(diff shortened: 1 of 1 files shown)


lib/services/ios-debug-service.ts
```diff
@@ -205,9 +208,20 @@ class IOSDebugService implements IDebugService {
...
-            let inspectorSourceLocation = path.join(inspectorPath, "Safari/Main.html");
-            let cmd = `open -a '${inspectorApplicationPath}' --args '${inspectorSourceLocation}' '${this.$projectData.projectName}'`;
+            let cmd: string = null;
+            
+            this.$projectDataService.initialize(this.$projectData.projectDir);
+            let platformData = this.$platformsData.getPlatformData(this.platform);
+            let frameworkVersion = this.$projectDataService.getValue(platformData.frameworkPackageName).wait().version;
+            if(semver.lt(frameworkVersion, "1.2.0")) {
+                let safariPath = path.join(inspectorPath, "Safari/Main.html");
+                cmd = "open -a Safari " + safariPath;
```
teobugslayer: is it possible safariPath to have spaces in it? Should we escape it?

lib/services/ios-debug-service.ts
```diff
@@ -205,9 +208,20 @@ class IOSDebugService implements IDebugService {
...
+            
+            this.$projectDataService.initialize(this.$projectData.projectDir);
+            let platformData = this.$platformsData.getPlatformData(this.platform);
+            let frameworkVersion = this.$projectDataService.getValue(platformData.frameworkPackageName).wait().version;
+            if(semver.lt(frameworkVersion, "1.2.0")) {
+                let safariPath = path.join(inspectorPath, "Safari/Main.html");
+                cmd = "open -a Safari " + safariPath;                
+            } else {
+                let inspectorApplicationPath = path.join(inspectorPath, "NativeScript Inspector.app");
+                let inspectorSourceLocation = path.join(inspectorPath, "Safari/Main.html");
```
teobugslayer: this path is calculated both in if and else, consider moving it out and calculating it once.

lib/services/ios-debug-service.ts
Fatme: It is possible only when the user changes the default npm cache path :smile_cat:  I'll escape it.

teobugslayer:
:+1:

dtopuzov:
run ci

PR #672, opened 2015-07-22
Fix missing warning when adding library

Fatme:


rosen-vladimirov:
:+1:  after green build

dtopuzov:
:100: after green build

PR #681, opened 2015-07-23
Respect correctly .jar files from plugin

Fatme:
1. Create new project
2. Add plugin that contains .jar file

Expected: The .jar file should  be added in libs folder of native project and in lib folder at project root
Actual:  the .jar file is not respected from cli.


rosen-vladimirov:
:+1:

Issue #679, opened 2015-07-23
Library add command adds the framework with its full path into xcode project


Fatme:
1. Create new project
2. Add dynamic framework
3. Move the project to another machine and try to build.

The build will fail because the framework is added with its full path.

PR #694, opened 2015-07-27
Add library with relative path to .xcodeproj file

Fatme:
Fixes https://github.com/NativeScript/nativescript-cli/issues/679

### DIFF
--- a/lib/services/ios-project-service.ts
+++ b/lib/services/ios-project-service.ts
@@ -181,14 +181,15 @@ class IOSProjectService extends projectServiceBaseLib.PlatformProjectServiceBase
 			this.validateDynamicFramework(libraryPath).wait();
 			var umbrellaHeader = this.getUmbrellaHeaderFromDynamicFramework(libraryPath).wait();
 
-			var frameworkName = path.basename(libraryPath, path.extname(libraryPath));
-			var targetPath = path.join(this.$projectData.projectDir, "lib", this.platformData.normalizedPlatformName, frameworkName);
-			this.$fs.ensureDirectoryExists(targetPath).wait();
-			shell.cp("-R", libraryPath, targetPath);
+			let frameworkName = path.basename(libraryPath, path.extname(libraryPath));
+			let targetPath = path.join("lib", this.platformData.normalizedPlatformName, frameworkName);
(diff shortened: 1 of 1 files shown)


lib/services/ios-project-service.ts
```diff
@@ -181,14 +181,15 @@ class IOSProjectService extends projectServiceBaseLib.PlatformProjectServiceBase
...
+			let frameworkName = path.basename(libraryPath, path.extname(libraryPath));
+			let targetPath = path.join("lib", this.platformData.normalizedPlatformName, frameworkName);
+			let fullTargetPath = path.join(this.$projectData.projectDir, targetPath);
+			this.$fs.ensureDirectoryExists(fullTargetPath).wait();
+			shell.cp("-R", libraryPath, fullTargetPath);
 
 			let project = this.createPbxProj();
-
-			project.addFramework(path.join(targetPath, frameworkName + ".framework"), { customFramework: true, embed: true });
+			let frameworkPath = path.relative("platforms/ios", path.join(targetPath, frameworkName + ".framework"));
```
rosen-vladimirov: `platforms/ios` - will this work on Windows?
Maybe: 
```TypeScript
let frameworkPath = path.relative(path.join("platforms", "ios"), path.join(targetPath, frameworkName + ".framework"));
```

rosen-vladimirov: Also, don't you need full paths on both places:
```TypeScript
let frameworkPath = path.relative(path.join(this.$projectData.projectDir, "platforms", "ios"), path.join(fullTargetPath, frameworkName + ".framework"));
```

Fatme: platforms/ios - will this work on Windows? - This service is executed only on OSX.
Also, don't you need full paths on both places: - No, I don't need the fullPath, a just need the relative path to native project.

rosen-vladimirov: :+1:

rosen-vladimirov:
:+1:

lib/services/ios-project-service.ts
teobugslayer: By the way, it works - but path.join will return a string with backslashes.

PR #698, opened 2015-07-27
Respect new files when livesyncing to device

Fatme:


teobugslayer:
:+1: but you need to rebase

Issue #470, opened 2015-05-07
Command failed due to space in file path


Lukosenka:
The error occurs with most of the commands.
```
Command failed: C:\Windows\system32\cmd.exe /s /c "c:\Users\Lukas Rakauskas\AppD
ata\Roaming\npm\node_modules\nativescript\lib\common\resources\platform-tools\an
droid\win32\adb start-server"
'c:\Users\Lukas' is not recognized as an internal or external command,
operable program or batch file.
```

RangerMauve:
It's happening because you have a space in your name username. Not sure how it'd be fixed though :open_mouth:

teobugslayer:
Hi @Lukosenka,

we have reproduced the issue on our side and will fix it in one of our next public releases.

Lukosenka:
Thank you very much, can't wait to start app developing. :+1:

rosen-vladimirov:
Hi @Lukosenka,
As a workaround you may install Android SDK and add it to your PATH environment variable. This way we'll not call adb from nativescript-cli resources and everything should work fine.
I'm reopening this issue and I'll take care to close it once we have permanent fix for it in our code.
Hope this helps.

saadeez:
This issue is still exist. As I'm learning the Nativescript's Hello World but in 2nd step I stucked in "tns platform add android" it gives me the same error even I have set environmental variable for Android SDk but still getting this error dont know why?

teobugslayer:
Hello @saadeez, please run ```tns platform add --log trace > log.txt``` and send us the log.txt file (you can create a Gist, for example)

rosen-vladimirov:
The steps to reproduce this issue are:
1) Remove path to android-sdk from your PATH environment variable.
2) Install CLI locally in directory which has spaces in its name.
3) Try executing `tns device android` command (make sure that when you write `adb` in your terminal, you get output like: "command not found")

PR #700, opened 2015-07-27
Fix adb related commands when path has spaces

rosen-vladimirov:
When CLI is installed in directory, which has space (or the path above this dir has spaces) and android-sdk is not installed, we are trying to execute adb from our resources. This is failing due to missing `"` in our commands.

Fixes https://github.com/NativeScript/nativescript-cli/issues/470


teobugslayer:
:+1: after green build

Issue #652, opened 2015-07-17
Debug command for iOS does not load inspector from the added ios-runtime.


vchimev:
With global installation it loads latest official one or explicitly specified by --frameworkPath option.

Update - add steps to reproduce:
1. npm install -g \<Path To nativescript.tgz\>
2. tns create \<App Name\>
3. cd \<App Name\>
4. tns platform add ios --frameworkPath \<Path To tns-ios.tgz>
5. tns debug ios --debug-brk **--frameworkPath** \<Path To Unpacked tns-ios.tgz\>

So the question is why do you need to specify the path to unpacked tns-ios.tgz in step 5 in order to use the debugger from the iOS runtime added in step 4?
IMO, `tns debug ios` should use the debugger from the already added platform, not the debugger on live.

Issue #671, opened 2015-07-22
ios debugger doesn't work for projects created with framework version lower than 1.2.0


Fatme:
If the project is created with framework version lower than 1.2, tns debug ios command fails with the following error: "Command failed: /bin/sh -c open -a '~/.npm/tns-ios/1.1.2/package/WebInspectorUI/NativeScript Inspector.app' --args '~/.npm/tns-ios/1.1.2/package/WebInspectorUI/Safari/Main.html' 'newAppRelease'
FSPathMakeRef(~/.npm/tns-ios/1.1.2/package/WebInspectorUI/NativeScript Inspector.app) failed with error -43."

Steps to reproduce:
1. Create new project
2. Add ios platform and ensure that the framework version is lower than 1.2.0
3. Try to execute `tns debug ios`.

PR #701, opened 2015-07-27
Open correct debugger url when framework version is lower than 1.2.0

Fatme:
Fixes 
https://github.com/NativeScript/nativescript-cli/issues/652 and https://github.com/NativeScript/nativescript-cli/issues/671

### DIFF
--- a/lib/services/ios-debug-service.ts
+++ b/lib/services/ios-debug-service.ts
@@ -207,13 +207,14 @@ class IOSDebugService implements IDebugService {
 
     private openDebuggingClient(): IFuture<void> {
         return (() => {
-            let inspectorPath = this.getInspectorPath().wait();
-            let inspectorSourceLocation = path.join(inspectorPath, "Safari/Main.html");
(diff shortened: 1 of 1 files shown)


lib/services/ios-debug-service.ts
```diff
@@ -238,7 +239,7 @@ class IOSDebugService implements IDebugService {
                 tnsIosPackage = path.resolve(this.$options.frameworkPath);
             } else {
                 var platformData = this.$platformsData.getPlatformData(this.platform);
-                tnsIosPackage = this.$npmInstallationManager.install(platformData.frameworkPackageName).wait();
+                tnsIosPackage = this.$npmInstallationManager.getCachedPackagePath(platformData.frameworkPackageName, frameworkVersion);
```
rosen-vladimirov: if the cache had been cleaned, will this add the dependency with its correct version to the cache?

rosen-vladimirov:
:+1:

PR #707, opened 2015-07-27
Print connected devices as a table

teobugslayer:
Implements https://github.com/telerik/mobile-cli-lib/pull/402


rosen-vladimirov:
:+1:

Issue #699, opened 2015-07-27
"tns platform update android" does not work


dtopuzov:
Precondition:
npm install -g nativescript@1.1.2
tns create MyApp
tns platform add android@1.1.0 --path MyApp

Steps to reproduce:
npm update -g nativescript
tns platform update android

Result:
ENOENT, scandir 'C:\Users\topuzov\AppData\Roaming\npm-cache\tns-android\1.2.0\package\framework'

dtopuzov:
Workaround:
tns platform remove android
tns platform add android

rosen-vladimirov:
Unable to reproduce

rosen-vladimirov:
It looks like the issue is caused by npm cache. It can be reproduced when you manually delete the framework directory under `<npm_cache_folder>\tns-android\1.2.0\package`

PR #709, opened 2015-07-28
Make sure framework dir is part of the cached framework

rosen-vladimirov:
In some cases framework dir is missing in the npm cached package of the framework. Check if it exists and if not, remove the package from the cache and add it again.
Fixes https://github.com/NativeScript/nativescript-cli/issues/699

### DIFF
--- a/lib/npm-installation-manager.ts
+++ b/lib/npm-installation-manager.ts
@@ -6,7 +6,7 @@ import semver = require("semver");
 import npm = require("npm");
 import constants = require("./constants");
 
-export class NpmInstallationManager {
+export class NpmInstallationManager implements INpmInstallationManager {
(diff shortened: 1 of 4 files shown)


lib/npm-installation-manager.ts
```diff
@@ -27,13 +27,22 @@ export class NpmInstallationManager {
 	public getCachedPackagePath(packageName: string, version: string): string {
 		return path.join(this.getCacheRootPath(), packageName, version, "package");
 	}
-	
-	public addToCache(packageName: string, version: string): IFuture<void> {
+
+	public addToCache( packageName: string, version: string): IFuture<void> {
```
teobugslayer: The horror - added an empty space after the ( :scream_cat:

Fatme:
:+1:

PR #714, opened 2015-07-29
Remove plugin platforms folder

Fatme:
Should be merged after this PR https://github.com/NativeScript/node-xcode/pull/4

If the platform is added before the plugin, the plugin platforms directory inside `tns_modules` is not deleted. This way the native libraries are copied twice in project - in `lib` folder and `tns_modules` folder, which increases the package size of an application a lot.


rosen-vladimirov:
:+1:

PR #733, opened 2015-08-03
Allow using of renamed Chrome on Mac

teobugslayer:
This ports the fix from master to release branches

See https://github.com/NativeScript/nativescript-cli/issues/632


rosen-vladimirov:
:+1:

Issue #682, opened 2015-07-23
Enable "tns debug" default option


valentinstoychev:
Lets have the ```--debug-brk``` to be the default option of the ```debug``` command.

tjvantoll:
+1

vchimev:
```tns debug``` without ```--debug-brk``` option does not start the debugger

Possibly causes: https://github.com/NativeScript/nativescript-cli/issues/722

Tested with: ```tns debug android --emulator```

Console Log: https://gist.github.com/vchimev/2c97c3ec6da5c7feec77

PR #740, opened 2015-08-04
Add default option for android debugger

Fatme:
https://github.com/NativeScript/nativescript-cli/issues/682

### DIFF
--- a/lib/services/android-debug-service.ts
+++ b/lib/services/android-debug-service.ts
@@ -55,9 +55,12 @@ class AndroidDebugService implements IDebugService {
 		return (() => {
 			let packageFile = "";
 
-			if (this.$options.debugBrk) {
-				this.$platformService.preparePlatform(this.platform).wait();
+			if(!this.$options.debugBrk && !this.$options.start) {
+				this.$logger.warn("Neither --debug-brk nor --start option was specified. Defaulting to --debug-brk.");                
+				this.$options.debugBrk = true;
+			}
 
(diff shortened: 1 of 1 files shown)


lib/services/android-debug-service.ts
```diff
@@ -54,10 +54,13 @@ class AndroidDebugService implements IDebugService {
 	public debugOnDevice(): IFuture<void> {
 		return (() => {
 			let packageFile = "";
+            
+            if(!this.$options.debugBrk && !this.$options.start) {
```
rosen-vladimirov: space / tabs issue

lib/services/android-debug-service.ts
```diff
@@ -54,10 +54,13 @@ class AndroidDebugService implements IDebugService {
...
+                this.$logger.warn("Neither --debug-brk nor --start option was specified. Defaulting to --debug-brk.");                
+                this.$options.debugBrk = true;
+            }
 
 			if (this.$options.debugBrk) {
-				this.$platformService.preparePlatform(this.platform).wait();
```
rosen-vladimirov: why have you removed this one ?

Fatme: because prepare function is called from build - https://github.com/NativeScript/nativescript-cli/pull/740/files#diff-c8a25497a509cafa1678ac3583ceb7e5R66. This way the project was prepared 2 times :smile_cat:

rosen-vladimirov:
:+1:

Issue #722, opened 2015-07-30
Debug command builds and installs the app twice


vchimev:
```tns debug <platform> --debug-brk``` build and installs the app twice

Steps to reproduce:
tns create testApp
cd testApp
tns platform add android --frameworkPath <Path To Latest Runtime>
tns debug android **--debug-brk** --emulator

Console Log: https://gist.github.com/vchimev/9587d957355e40d1bc7a

vchimev:
Related to https://github.com/NativeScript/nativescript-cli/issues/475.

vchimev:
The issue with building and installing the app twice is fixed, but following the steps above the debugger does not launch.

Console Log: https://gist.github.com/vchimev/68086fd0b535e5e4b361

vchimev:
It looks that the default timeout for the debugger to establish connection with the device/emulator when launching an emulator for the first time could not be enough. Therefore for version 1.2.2 the default timeout will increase from 60 to 90 seconds. If this would not be enough, --timeout option could be used.

PR #741, opened 2015-08-04
Do not build and install the app twice when debug command is executed

Fatme:
https://github.com/NativeScript/nativescript-cli/issues/722


rosen-vladimirov:
:+1:

PR #744, opened 2015-08-04
Modify test infrastructure and grunt tasks

rosen-vladimirov:
As common lib is now a separate npm package, it has its own tests, that must be run when executing `grunt test` or `npm test` in nativescript-cli. Modify test command in package.json in order to run the tests.
Change Gruntfile.js in order to ignore/build the correct typescript files in common lib.
Add Bluebird as dependency as it is required for common lib.
Update mocha version.
Remove `chai` and `mocha` definitions as they are part of common lib now.


teobugslayer:
:+1:

PR #766, opened 2015-08-10
Fix livesync docs

rosen-vladimirov:
Fix documentation for iOS livesync (iOS Simulator).


teobugslayer:
:+1: and :+1: from Fatme

Issue #730, opened 2015-08-01
LiveSync for Android throws stdout maxBuffer exceeded


cime:
When running *tns livesync android* I get

    Error: stdout maxBuffer exceeded.

After the *Applying changes...* message.

Tablet: Asus MeMO Pad 7 K013

My solution - I changed the maxBuffer size inside lib/child-process.js

    ChildProcess.prototype.exec = function (command, options) {
		options = options || {};
		options.maxBuffer = 1024 * 2048;

iamajaymakwana:
Where is child-process.js located? 
I am not able to find it.

teobugslayer:
@iamajaymakwana  - look at https://github.com/telerik/mobile-cli-lib/blob/master/child-process.ts

iamajaymakwana:
Sorry for asking for again but I an not able to find it. But where is child-process located in my system.
I am using the windows 7 and run the application on Samsung Note3 Neo.

Fatme:
C:\\Users\\{username}\\AppData\\Roaming\\npm\lib\node_modules\nativescript\lib\common\child-process.js

maganius:
I have same problem I can't fix it, I put the maxBuffer on exec command but it still fails.
`
exec(command, options, execOptions) {
        return __awaiter(this, void 0, void 0, function* () {
            return new Promise((resolve, reject) => {
				options = options || {};
				options.maxBuffer = 1024 * 2048;
                const callback = (error, stdout, stderr) => {

                    this.$logger.trace("Exec %s \n stdout: %s \n stderr: %s", command, stdout.toString(), stderr.toString());
                    if (error) {
                        reject(error);
                    }
                    else {
                        const output = execOptions && execOptions.showStderr ? { stdout, stderr } : stdout;
                        resolve(output);
                    }
                };
                if (options) {
                    child_process.exec(command, options, callback);
                }
                else {
                    child_process.exec(command, callback);
                }
            });
        });
    }
`

PR #767, opened 2015-08-10
Get all adb output, even for large operations

teobugslayer:
Speed up adb start

Fixes NativeScript/nativescript-cli#730


rosen-vladimirov:
:+1:

PR #769, opened 2015-08-11
Cocoapods support

Fatme:


lib/services/ios-project-service.ts
```diff
@@ -262,6 +262,13 @@ class IOSProjectService extends projectServiceBaseLib.PlatformProjectServiceBase
 				this.savePbxProj(project).wait();
 			}
 			
+			if(this.$fs.exists(this.projectPodFilePath).wait() && this.$fs.getFsStats(this.projectPodFilePath).wait().mtime < this.$fs.getFsStats(this.$projectData.projectFilePath).wait().mtime) {
+				try {
+					this.$childProcess.exec("pod install", { cwd: this.platformData.projectRoot }).wait();
+				} catch(e) {
+					this.$errors.failWithoutHelp("CocoaPods are not installed. Run `sudo gem install cocoapods` and try again.");
```
ligaz: Shouldn't we display the error as well? For example CocoaPods might be installed but the command `pod install` to fail for some reason?

ligaz:
Consider adding checks for CocoaPods in the `doctor` command.

lib/services/ios-project-service.ts
```diff
@@ -303,6 +314,12 @@ class IOSProjectService extends projectServiceBaseLib.PlatformProjectServiceBase
 		return (() => {
 			let pluginPlatformsFolderPath = pluginData.pluginPlatformsFolderPath(IOSProjectService.IOS_PLATFORM_NAME);
 			_.each(this.getAllDynamicFrameworksForPlugin(pluginData).wait(), fileName => this.addLibrary(path.join(pluginPlatformsFolderPath, fileName)).wait());
+			
+			let pluginPodFilePath = path.join(pluginPlatformsFolderPath, "Podfile");
+			if(this.$fs.exists(pluginPodFilePath).wait()) {
+				let pluginPodFileContent = this.$fs.readText(pluginPodFilePath).wait();
+				this.$fs.appendFile(this.projectPodFilePath, pluginPodFileContent).wait();
```
ligaz: Consider adding a comment in the general pod file mentioning that the code is from this given file.

ligaz:
Adding some unit tests for this functionality will be great. :cop:

lib/services/ios-project-service.ts
teobugslayer: This check must also be introduced in `doctor` command.

package.json
```diff
@@ -10,7 +10,7 @@
   },
   "main": "./lib/nativescript-cli.js",
   "scripts": {
-    "test": "node_modules\\.bin\\_mocha --ui mocha-fibers --recursive --reporter spec --require test/test-bootstrap.js --timeout 60000 test/ lib/common/test/unit-tests",
+    "test": "node_modules\\.bin\\_mocha --ui mocha-fibers --recursive --reporter spec --require test/test-bootstrap.js --timeout 60000 test/ios-project-service",
```
teobugslayer: I guess this change is leftover from debugging - but it currently breaks our test suite.

test/ios-project-service.ts
```diff
@@ -0,0 +1,148 @@
+/// <reference path=".d.ts" />
+"use strict";
+
+import Future = require("fibers/future");
+import path = require("path");
```
teobugslayer: Consider ES6 style imports

lib/services/ios-project-service.ts
```diff
@@ -5,12 +5,13 @@ import Future = require("fibers/future");
 import path = require("path");
 import shell = require("shelljs");
 import util = require("util");
+import os = require("os");
```
teobugslayer: Consider ES6 style imports

lib/services/ios-project-service.ts
```diff
@@ -303,22 +307,32 @@ class IOSProjectService extends projectServiceBaseLib.PlatformProjectServiceBase
...
+	}
+	
+	public afterPrepareAllPlugins(): IFuture<void> {
+		return (() => {
+			if(this.$fs.exists(this.projectPodFilePath).wait()) {
+				// Check availability
+				try {
+					this.$childProcess.exec("gem which cocoapods").wait();
+				} catch(e) {
+					this.$errors.failWithoutHelp("CocoaPods are not installed. Run `sudo gem install cocoapods` and try again.");
```
teobugslayer: The message on my machine is `ERROR:  Can't find ruby library file or shared library cocoapods`. We must find another way to check for existence.

lib/services/ios-project-service.ts
```diff
@@ -379,5 +393,57 @@ class IOSProjectService extends projectServiceBaseLib.PlatformProjectServiceBase
...
+		}).future<void>()();
+	}
+	
+	private prepareCocoapods(pluginPlatformsFolderPath: string): IFuture<void> {
+		return (() => {
+			let pluginPodFilePath = path.join(pluginPlatformsFolderPath, "Podfile");
+			if(this.$fs.exists(pluginPodFilePath).wait()) {
+				let pluginPodFileContent = this.$fs.readText(pluginPodFilePath).wait();
+				let contentToWrite = this.buildPodfileContent(pluginPodFilePath, pluginPodFileContent);
+				this.$fs.appendFile(this.projectPodFilePath, contentToWrite).wait();
```
teobugslayer: This potentially generate invalid podfile.
 - two dependencies may have incompatible versions, see [spec](https://guides.cocoapods.org/syntax/podfile.html#pod)
 - two hooks like post_install

Are we going to address these? perhaps a warning to check the resulting podfile?

lib/services/ios-project-service.ts
```diff
@@ -2,15 +2,16 @@
 "use strict";
 
 import Future = require("fibers/future");
-import path = require("path");
+import * as path from "path";
```
teobugslayer: Almost all other imports can also be converted (Future can't).

lib/services/ios-project-service.ts
```diff
@@ -303,22 +307,32 @@ class IOSProjectService extends projectServiceBaseLib.PlatformProjectServiceBase
...
+			if(this.$fs.exists(this.projectPodFilePath).wait()) {
+				// Check availability
+				try {
+					this.$childProcess.exec("gem which cocoapods").wait();
+				} catch(e) {
+					this.$errors.failWithoutHelp("CocoaPods are not installed. Run `sudo gem install cocoapods` and try again.");					
+				}
+				
+				this.$logger.info("Installing pods...");
+				this.$childProcess.exec("pod install", { cwd: this.platformData.projectRoot }).wait();
```
teobugslayer: What happens when this program fails? We should check its exit code or error output.

lib/services/ios-project-service.ts
```diff
@@ -379,5 +393,57 @@ class IOSProjectService extends projectServiceBaseLib.PlatformProjectServiceBase
...
+	
+	private removeCocoapods(pluginPlatformsFolderPath: string): IFuture<void> {
+		return (() => {
+			let pluginPodFilePath = path.join(pluginPlatformsFolderPath, "Podfile");
+			if(this.$fs.exists(pluginPodFilePath).wait()) {
+				let pluginPodFileContent = this.$fs.readText(pluginPodFilePath).wait();
+				let projectPodFileContent = this.$fs.readText(this.projectPodFilePath).wait();
+				let contentToRemove= this.buildPodfileContent(pluginPodFilePath, pluginPodFileContent);
+				projectPodFileContent = helpers.stringReplaceAll(projectPodFileContent, contentToRemove, "");
+				if(_.isEmpty(projectPodFileContent)) {
```
teobugslayer: You may trim the content - there is no need to keep files consisting of empty space. This may happen when the user manually adds empty space.

lib/services/plugins-service.ts
```diff
@@ -145,6 +144,14 @@ export class PluginsService implements IPluginsService {
 		}).future<IPluginData[]>()();
 	}
 	
+	public afterPrepareAllPlugins(): IFuture<void> {
+		let action = (pluginDestinationPath: string, platform: string, platformData: IPlatformData) => {
```
teobugslayer: Is this lambda really needed? Can we just invoke the wrapped function?

lib/tools/broccoli/node-modules-dest-copy.ts
```diff
@@ -76,6 +76,10 @@ export class DestCopy implements IBroccoliPlugin {
 			this.$pluginsService.prepare(dependency).wait();
 		}
 	});
+	
+	if(_.keys(this.dependencies).length > 0) {
```
teobugslayer: _.isEmpty, perhaps?

test/ios-project-service.ts
```diff
@@ -0,0 +1,148 @@
+/// <reference path=".d.ts" />
+"use strict";
+
+import Future = require("fibers/future");
+import * as path from "path";
+import temp = require("temp");
```
teobugslayer: This leaks the created directories. Consider executing `temp.track()`

test/ios-project-service.ts
```diff
@@ -0,0 +1,148 @@
...
+import FileSystemLib = require("../lib/common/file-system");
+import HostInfoLib = require("../lib/common/host-info");
+import iOSProjectServiceLib = require("../lib/services/ios-project-service");
+import LoggerLib = require("../lib/common/logger");
+import OptionsLib = require("../lib/options");
+import ProjectDataLib = require("../lib/project-data");
+
+import yok = require("../lib/common/yok");
+
+let assert = require("chai").assert;
```
teobugslayer: `import { assert } from "chai";`

lib/services/ios-project-service.ts
teobugslayer: We discussed the issue in person and we decided that the code is correct.

lib/services/ios-project-service.ts
Fatme: If we have two dependencies with incompatible versions `pod install` command fails with descriptive message.

If Podfile has more than one `post_install` sections only the last one is executed.

lib/services/ios-project-service.ts
Fatme: If the command fails the error will be printed on the console.

lib/services/ios-project-service.ts
Fatme: I prefer to keep the content as is from original Podfile.

test/ios-project-service.ts
Fatme: Actually `temp.track` is used from another tests and the created directories are removed.

lib/services/ios-project-service.ts
teobugslayer: We should document this behavior and give a warning to the user. Ping @ikoevska

lib/services/ios-project-service.ts
teobugslayer: The issue is when the user has manually edit the podfile.
In this case, we can leave the code as is and let the users manually remove the empty file.

test/ios-project-service.ts
teobugslayer: Other tests can be rewritten in the future. Do not depend on side effects of unrelated code.

Fatme: We've decided to implement this in another PR.

lib/services/ios-project-service.ts
Fatme: We discussed the issue in person and decided to keep the current code.

lib/services/ios-project-service.ts
Fatme: I've added a warning message.

teobugslayer:
Lets merge this, :+1: !

PR #771, opened 2015-08-11
Add support for iojs 2.x

rosen-vladimirov:
Add experimental support for iojs 2.x. Currently node-inspector cannot be build with iojs and during installation some error messages are shown, but the installation is successful.


Mitko-Kerezov:
:+1:

PR #775, opened 2015-08-12
Fixes for android related commands

teobugslayer:
See http://teampulse.telerik.com/view#item/299038

### DIFF
--- a/lib/services/android-debug-service.ts
+++ b/lib/services/android-debug-service.ts
@@ -96,21 +96,21 @@ class AndroidDebugService implements IDebugService {
 	
 	private printDebugPort(packageName: string): IFuture<void> {
         return (() => {
-            let res = this.$childProcess.spawnFromEvent(this.$staticConfig.getAdbFilePath().wait(), ["shell", "am", "broadcast", "-a", packageName + "-GetDgbPort"], "exit").wait();
-            this.$logger.info(res.stdout);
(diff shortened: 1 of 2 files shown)


lib/services/android-debug-service.ts
```diff
@@ -142,11 +142,11 @@ class AndroidDebugService implements IDebugService {
             let packageDir = util.format(AndroidDebugService.PACKAGE_EXTERNAL_DIR_TEMPLATE, packageName);
             let envDebugOutFullpath = this.$mobileHelper.buildDevicePath(packageDir, AndroidDebugService.ENV_DEBUG_OUT_FILENAME);
             
-            this.device.adb.executeShellCommand(`rm "${envDebugOutFullpath}"`).wait();
-            this.device.adb.executeShellCommand(`mkdir -p "${packageDir}"`).wait();
+            this.device.adb.executeShellCommand(["rm", `${envDebugOutFullpath}`]).wait();
+            this.device.adb.executeShellCommand(["mkdir", "-p", `${packageDir}`]).wait();
```
Fatme: `packageDir` should be in `""`

Fatme:
:+1:  after addressing my comments

lib/services/android-debug-service.ts
teobugslayer: We agreed in person that the code is correct.

Issue #776, opened 2015-08-12
tns livesync ios --emulator --watch doesn't sync changes


tjvantoll:
Hey all,

After 1.2.2 `tns livesync ios --emulator` works great, but if I add the `--watch` flag the iOS simulator no longer reflects my changes. The watcher works, and the simulator reloads, but my changes are not present.

`tns livesync android --emulator --watch` works as expected.

Thanks.

Fatme:
Hi @tjvantoll ,

Can you give us more information about your iOS simulator - device type, sdk version? I've just tested `tns livesync ios --emulator --watch` command with `iPhone-4s/ iOS 7.1` and `iPhone 6 Plus/ iOS 8.4` and it works as expected.

tjvantoll:
I'm using the default on my machine, which is an iPhone-4s running iOS 8.4. It's odd because `tns livesync ios --emulator` works, but with the `--watch` flag those changes don't make it to the app. I'm going to try playing with some other device combinations.

tjvantoll:
I made a short video to show this in action: https://www.dropbox.com/s/odfkf9rvrj26dqp/livesync-issue.mov?dl=0

Fatme:
I've tested `iPhone-4s/ iOS 8.4` and it works as expected. Can you execute `tns livesync ios --emulator --watch --log trace` and send us the output?

tjvantoll:
Looking at the logs myself it seems I'm actually running the iOS 8.1 simulator, which I'm not sure how that happened. Let me get an iOS 8.4 simulator running and get back with you.

Thanks for the quick help.

tjvantoll:
Hmmmm... I made sure to run an iOS 8.4 simulator and I had the same issue. Here's a log of me making a single change: https://gist.github.com/tjvantoll/6a29fb6d77eb22050b97.

tjvantoll:
Just tried on a new project and this command works as expected, which is odd. I had updated the previous project to use the 1.2.1 iOS runtime. I'm still playing with this to try to see what's up.

Fatme:
We've reproduced the problem using `Groceries sample`. The problem is with nested folders. We've moved `login.js` file from `views/login/login.js` file to `app/login.js` which is wrong location. I'm preparing a fix that will be released ASAP.

xuezier:
tns --version:1.6.1
tns livesync ios --emulator --watch, js file has changed,but not sync

vchimev:
[...]

PR #778, opened 2015-08-13
Fix livesync ios --emulator --watch if the file is not at project root level

Fatme:
Fixes https://github.com/NativeScript/nativescript-cli/issues/776


teobugslayer:
:+1:

PR #780, opened 2015-08-13
Fix add plugin unit test

teobugslayer:
Npm has a global configuration to save dependencies with exact versions or as ^ ones. This unit test must work in both configurations. Therefore, we check for one of the two possible outcomes.

Also, small code cleanups.

### DIFF
--- a/test/plugins-service.ts
+++ b/test/plugins-service.ts
@@ -18,16 +18,14 @@ import PlatformsDataLib = require("../lib/platforms-data");
 import ProjectDataServiceLib = require("../lib/services/project-data-service");
 import helpers = require("../lib/common/helpers");
 import ProjectFilesManagerLib = require("../lib/services/project-files-manager");
-import os = require("os");
-
+import { EOL } from "os";
 import PluginsServiceLib = require("../lib/services/plugins-service");
 import AddPluginCommandLib = require("../lib/commands/plugin/add-plugin");
-
-import path = require("path");
-import temp = require("temp");
+import { assert }  from "chai";
+import * as path from "path";
+import * as temp from "temp";
 temp.track();
 
-let assert = require("chai").assert;
 let isErrorThrown = false;	
 
 function createTestInjector() {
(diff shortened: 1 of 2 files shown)


test/plugins-service.ts
```diff
@@ -283,10 +279,9 @@ describe("Plugins service", () => {
 			// Asserts that the plugin is added in package.json file
 			let packageJsonContent = fs.readJson(path.join(projectFolder, "package.json")).wait();
 			let actualDependencies = packageJsonContent.dependencies;
-			let expectedDependencies = {
-				"plugin1": "^1.0.3"
-			};
-			assert.deepEqual(actualDependencies, expectedDependencies); 
+			let expectedDependencies = { "plugin1": "^1.0.3" };
+			let expectedDependenciesExact = { "plugin1": "1.0.3" };
+			assert.isTrue(_.isEqual(actualDependencies, expectedDependencies) || _.isEqual(actualDependencies, expectedDependenciesExact));
```
Fatme: Can we use `deepEqual` here?

Fatme:
:+1:

test/plugins-service.ts
teobugslayer: _.isEqual compares deeply

rosen-vladimirov: so we consider both syntax are correct (`^1.0.3` and `1.0.3`) ?

test/plugins-service.ts
```diff
@@ -299,10 +294,9 @@ describe("Plugins service", () => {
 						name: ""
 					}];
 				}).future<IPluginData[]>()();
-			}
+			};
 			
-			let commandsService = testInjector.resolve(CommandsServiceLib.CommandsService);
-			commandsService.tryExecuteCommand("plugin|add", [pluginName+"@1.0.0"]).wait();
+			pluginsService.add(pluginName+"@1.0.0").wait();
```
rosen-vladimirov: or maybe `${pluginName}@1.0.0`

rosen-vladimirov:
:+1:

test/plugins-service.ts
teobugslayer: Yes, I have updated PR description.

test/plugins-service.ts
teobugslayer: We agreed to leave the code as is.

Issue #774, opened 2015-08-12
```tns run android``` throws an error if android platform is not added before that


valentinstoychev:
It give you an error message - ```c:\work\SideDrawerPlayGround\platforms\android\build.xml does not exist!```.

The thing is that after that you can't add the android platform because it seems like the ```tns run android ``` command is creating an ```android``` folder.

vchimev:
This issue is resolved as we "add platform on each platform related command" - #785.

Issue #475, opened 2015-05-12
Unify and simplify workflows of build-related commands


teobugslayer:
Problem: Currently, the commands which include build - related steps differ slightly in their semantics, parameters and work flows.

Solution: Unify as much as possible but only where make sense.

Remark: All commands should work without specifying a platform if it is already added to the project. If more than one platform is added we should execute the command on all of them. If a user wants to run a particular platform she will be able to specify it as it currently works.

Prepare - stays as is

Build - stays as is

Deploy - does not work for emulators, does not run the app
  -- device

Run - based on deploy
  -- device
  -- emulator android (device)
  -- emulator ios

by default - build for simulator. Can be overridden by command line parameter. run [ios] --device. For android, running --geny instance must be reused

Debug = run + start debugger. works for android & ios, devices & emulators

Emulate = run --emulator. emulate --geny = run --geny

rosen-vladimirov:
Based on the description above, I've prepared the exepected behavior of the commands. 
I've found some cases, that are confusing and I'm not sure what is the expected behavior (check the table at the end).

## platform add
Usage:
* `platform add <platform1>` - add specified platform
* `platform add <platform1> <platform2>` - add all specified platforms

## prepare
Usage: 
* `tns prepare` - prepares the project for all installed platforms, for example if the project is targeting android and ios, this command will prepare it for both of them
* `tns prepare <platform>` - prepares the project for all specified platforms. In case the specified platform is not added, add it and prepare it.

## build
Usage:
* `tns build` - prepare and build the project for all installed platforms
* `tns build <platform>` - prepare and build the project for specified platform

### build for device or for emulator
check the table at the end for information which build should be executed.

## deploy
Usage:
* `tns deploy` - prepare and build the project for all installed platforms. After that deploy the project on all currently running devices and emulators.
* `tns build <platform>` - prepare and build the project for specified platform. After that deploy the project on all currently running devices and emulators of specified platform.
* `tns build <platform> --device <deviceId>` - prepare and build the project for specified device (for Android make sure to check is this device or emulator). Deploy the project on the specified device.

## run
Usage:
* `tns run` - prepare and build the project for all installed platforms. After that deploy the project on all currently running devices and emulators. Finally run the project on all of them.
* `tns run <platform>` - prepare and build the project for specified platform. After that deploy the project on all currently running devices and emulators of specified platform. Finally run the project on all of them.
[...]

PR #782, opened 2015-08-13
Add platform on all commands

rosen-vladimirov:
When you execute `tns run <platform>`, but you have forgotten to add the platform before that, you go in inconsistent state. Make sure to add the platform for the user when one of the following is executed:
* `tns prepare <platform>`
* `tns build <platform>`
* `tns run <platform>`
* `tns update <platform>@<version>` - this will add the specified version in case you have not added the platform before that.
* `tns emulate <platform>`

This way we simplify the workflow, now you can just do:
1) `tns create app1`
2) `tns run android --path app1`
And the application will run on your attached device.

Resolves https://github.com/NativeScript/nativescript-cli/issues/774
Part of the implementation of: https://github.com/NativeScript/nativescript-cli/issues/475

### DIFF
--- a/lib/services/platform-service.ts
+++ b/lib/services/platform-service.ts
@@ -146,6 +146,7 @@ export class PlatformService implements IPlatformService {
 			this.validatePlatform(platform);
 			
 			platform = platform.toLowerCase();
+			this.ensurePlatformInstalled(platform).wait();
 
 			var platformData = this.$platformsData.getPlatformData(platform);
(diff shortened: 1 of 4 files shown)


lib/services/platform-service.ts
```diff
@@ -204,7 +205,7 @@ export class PlatformService implements IPlatformService {
 	public runPlatform(platform: string): IFuture<void> {
 		return (() => {
 			platform = platform.toLowerCase();
-
+
```
rosen-vladimirov: :crying_cat_face: 
I'll remove this change

teobugslayer:
Excellent! :+1:

PR #786, opened 2015-08-14
Bump up the version to 1.2.3

Fatme:


rosen-vladimirov:
:+1: 
May the `alphabet release` be with you!
memeyoda

PR #790, opened 2015-08-14
Write errors on stderr

rosen-vladimirov:
Update common lib where the following change is applied - errors are written on stderr instead of stdout.


rosen-vladimirov:
this should be for master :(

PR #792, opened 2015-08-14
Write errors on stderr

rosen-vladimirov:
Write errors on stderr instead of stdout and this way will be proper cli tool.


Fatme:
:+1:

teobugslayer:
Another PR already updated the common lib.

Issue #797, opened 2015-08-17
Plugin add does not merge plugin's info.plist


teobugslayer:
Consider following scenario: A developer wants to consume a {N} plugin which has custom info.plist using plugin add command.

In this scenario, {N} CLI looks for incorrectly named info.plist file in the plugin. Furthermore, the code does not correctly merge the contents of plugin info.plist file.

vchimev:
The issue is reproducible with 1.3.0 release candidate.

PR #798, opened 2015-08-17
Fix merging info.plist during plugin add

teobugslayer:
Fix: look for correctly named info.plist when adding a plugin
Fix: correctly merge plugin content when merging its info.plist

Fixes https://github.com/NativeScript/nativescript-cli/issues/797


Fatme:
:+1:

PR #802, opened 2015-08-19
Warn users when creating applications named 'app'.

teobugslayer:
See https://github.com/NativeScript/nativescript-cli/issues/760

### DIFF
--- a/lib/commands/create-project.ts
+++ b/lib/commands/create-project.ts
@@ -3,6 +3,7 @@
 
 export class ProjectCommandParameter implements ICommandParameter {
 	constructor(private $errors: IErrors,
+		private $logger: ILogger,
 		private $projectNameValidator: IProjectNameValidator) { }
 
 	mandatory = true;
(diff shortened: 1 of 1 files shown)


lib/commands/create-project.ts
```diff
@@ -12,6 +13,10 @@ export class ProjectCommandParameter implements ICommandParameter {
 				this.$errors.fail("You must specify <App name> when creating a new project.");
 			}
 
+			if (value === "app") {
```
ligaz: If the value is `App` we will not show the warning but we have to.

teobugslayer: Great catch, thanks!

teobugslayer:
ping @ligaz

lib/commands/create-project.ts
```diff
@@ -12,6 +13,10 @@ export class ProjectCommandParameter implements ICommandParameter {
 				this.$errors.fail("You must specify <App name> when creating a new project.");
 			}
 
+			if (value.toUpperCase() === "APP") {
```
rosen-vladimirov: Consider checking the OS as well, the message is useful only on MacOS

rosen-vladimirov:
:+1: But I think the message must be shown only on MacOS

lib/commands/create-project.ts
teobugslayer: The project may be shared through version control with other people who has Mac OS. I think it is best to be shown always.

Fatme:
:+1:

Issue #390, opened 2015-03-27
Support tns_modules from npm


ErjanGavalji:
A NativeScript application package must not contain the `tns_modules` it uses. These should be available through [npmjs.org](http://www.npmjs.org). Utilizing the dependencies in the package.json file (see #362), the CLI would retrieve the modules the application depends on.

ErjanGavalji:
Package name is [tns-core-modules](https://github.com/NativeScript/NativeScript/blob/master/package.json)

vchimev:
https://github.com/NativeScript/nativescript-cli/issues/854

PR #805, opened 2015-08-20
Use tns_modules from npm

e2l3n:
Use npm to fetch tns-core-modules package. 
Create tns_modules folder from tns-core-modules package.
See issue https://github.com/NativeScript/nativescript-cli/issues/390

### DIFF
--- a/lib/definitions/project.d.ts
+++ b/lib/definitions/project.d.ts
@@ -1,3 +1,4 @@
+
 interface IProjectService {
 	createProject(projectName: string): IFuture<void>;
 }
--- a/lib/node-package-manager.ts
+++ b/lib/node-package-manager.ts
@@ -5,7 +5,8 @@ import Future = require("fibers/future");
 import npm = require("npm");
 
 export class NodePackageManager implements INodePackageManager {
-	constructor(private $logger: ILogger,
+	constructor(private $childProcess: IChildProcess,
+		private $logger: ILogger,
 		private $errors: IErrors,
 		private $fs: IFileSystem,
 		private $lockfile: ILockFile,
@@ -52,6 +53,10 @@ export class NodePackageManager implements INodePackageManager {
 	
 	public view(packageName: string, propertyName: string): IFuture<any> {
 		return this.loadAndExecute("view", [[packageName, propertyName], [false]]);
+	}
+
+	public executeNpmCommand(npmCommandName: string, currentWorkingDirectory: string): IFuture<any> {
+		return this.$childProcess.exec(npmCommandName, { cwd: currentWorkingDirectory });
 	}
 	
 	private loadAndExecute(commandName: string, args: any[], opts?: { config?: any, subCommandName?: string }): IFuture<any> {
--- a/lib/services/platform-service.ts
+++ b/lib/services/platform-service.ts
@@ -175,9 +175,15 @@ export class PlatformService implements IPlatformService {
 			platformData.platformProjectService.prepareProject().wait();
 
 			// Process node_modules folder
-			this.$pluginsService.ensureAllDependenciesAreInstalled().wait();
-			let tnsModulesDestinationPath = path.join(platformData.appDestinationDirectoryPath, constants.APP_FOLDER_NAME, PlatformService.TNS_MODULES_FOLDER_NAME);
-			this.$broccoliBuilder.prepareNodeModules(tnsModulesDestinationPath, this.$projectData.projectDir, platform, lastModifiedTime).wait();
+			try {
+				this.$pluginsService.ensureAllDependenciesAreInstalled().wait();
+				let tnsModulesDestinationPath = path.join(platformData.appDestinationDirectoryPath, constants.APP_FOLDER_NAME, PlatformService.TNS_MODULES_FOLDER_NAME);
+				this.$broccoliBuilder.prepareNodeModules(tnsModulesDestinationPath, this.$projectData.projectDir, platform, lastModifiedTime).wait();
+			} catch(error) {
+				this.$logger.debug(error);
+				this.$errors.fail(`Processing node_modules failed. Error:${error}`);
+				shell.rm("-rf", appResourcesDirectoryPath);
+			}
 			
 			// Process platform specific files
 			let directoryPath = path.join(platformData.appDestinationDirectoryPath, constants.APP_FOLDER_NAME);
--- a/lib/services/project-service.ts
+++ b/lib/services/project-service.ts
@@ -10,7 +10,8 @@ import util = require("util");
 
 export class ProjectService implements IProjectService {
 
-	constructor(private $errors: IErrors,
+	constructor(private $npm: INodePackageManager,
+		private $errors: IErrors,
 		private $fs: IFileSystem,
 		private $logger: ILogger,
 		private $projectDataService: IProjectDataService,
@@ -83,13 +84,12 @@ export class ProjectService implements IProjectService {
 				throw err;
 			}
 
-
 			this.$logger.out("Project %s was successfully created", projectName);
 
 		}).future<void>()();
 	}
 
-	private createProjectCore(projectDir: string, appSourcePath: string,  projectId: string): IFuture<void> {
+	private createProjectCore(projectDir: string, appSourcePath: string, projectId: string): IFuture<void> {
 		return (() => {
 			this.$fs.ensureDirectoryExists(projectDir).wait();
 
@@ -101,6 +101,7 @@ export class ProjectService implements IProjectService {
 			} else {
 				shell.cp('-R', path.join(appSourcePath, "*"), appDestinationPath);
 			}
+
 			this.createBasicProjectStructure(projectDir,  projectId).wait();
 		}).future<void>()();
 	}
@@ -111,6 +112,14 @@ export class ProjectService implements IProjectService {
 
 			this.$projectDataService.initialize(projectDir);
 			this.$projectDataService.setValue("id", projectId).wait();
+			
+			let tnsModulesVersion = this.$options.tnsModulesVersion;
+			let packageName = constants.TNS_CORE_MODULES_NAME;
+			if (tnsModulesVersion) {
+				packageName = `${packageName}@${tnsModulesVersion}`;
+			}
+
+			this.$npm.executeNpmCommand(`npm install ${packageName} --save --save-exact`, projectDir).wait();
 		}).future<void>()();
 	}
 
--- a/lib/tools/broccoli/node-modules-dest-copy.ts
+++ b/lib/tools/broccoli/node-modules-dest-copy.ts
@@ -75,6 +75,12 @@ export class DestCopy implements IBroccoliPlugin {
 		if(isPlugin) {
 			this.$pluginsService.prepare(dependency).wait();
 		}
+
+		if (dependency.name === constants.TNS_CORE_MODULES_NAME) {
+			let tnsCoreModulesResourcePath = path.join(this.outputRoot, constants.TNS_CORE_MODULES_NAME);
+			shelljs.cp("-Rf", path.join(tnsCoreModulesResourcePath, "*"), this.outputRoot);
+			this.$fs.deleteDirectory(tnsCoreModulesResourcePath).wait();
+		} 
 	});
   }
 
--- a/test/project-service.ts
+++ b/test/project-service.ts
@@ -3,6 +3,8 @@
 
 import yok = require('../lib/common/yok');
 import stubs = require('./stubs');
+import * as constants from "./../lib/constants";
+import * as ChildProcessLib from "../lib/common/child-process";
 
 import ProjectServiceLib = require("../lib/services/project-service");
 import ProjectDataServiceLib = require("../lib/services/project-data-service");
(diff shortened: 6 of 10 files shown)


lib/definitions/project.d.ts
```diff
@@ -1,3 +1,8 @@
+interface ITNSModulesService {
+	tnsModulesInstallationPath(projectRoot: string): IFuture<string>;
+	NPM_TNS_CORE_MODULES_NAME: string;
```
ligaz: The convention for property names is to use `camelCase`.

teobugslayer:
The build fails executing unit tests with `Error: unable to resolve tnsModulesService`. Can you please verify that unit tests work?

lib/services/tns-modules-service.ts
```diff
@@ -0,0 +1,24 @@
+///<reference path="../.d.ts"/>
+"use strict";
+
+import util = require("util");
```
teobugslayer: Consider using ES2015 style imports where possible

ligaz:
I'm not sure that the currently implemented workflow is the right one. The process should be like:
* create new project from template
* check if the template has `tns_modules` as dependencies and if not add them in the `package.json`
* `npm install`

lib/services/tns-modules-service.ts
```diff
@@ -0,0 +1,24 @@
...
+import path = require("path");
+import shell = require("shelljs");
+import npm = require("npm");
+var helpers = require("../common/helpers");
```
teobugslayer: You can `import` here.

lib/services/project-service.ts
```diff
@@ -101,16 +104,25 @@ export class ProjectService implements IProjectService {
 			} else {
 				shell.cp('-R', path.join(appSourcePath, "*"), appDestinationPath);
 			}
-			this.createBasicProjectStructure(projectDir,  projectId).wait();
+			
+			// Copy tns-core-modules package to node_modules destination directory
+			var nodeModulesDestination = path.join(appDestinationPath, constants.NODE_MODULES_FOLDER_NAME);
```
teobugslayer: Please use `let` everywhere instead of `var`.

teobugslayer:
Consider writing unit tests for TNSModulesService.

lib/services/project-service.ts
ligaz: Let's use a tslinter or at least document our coding styles.

e2l3n: See latest commit. It has a check for `tns-core-modules` version. 
There are two options to pass the version number. 
The first one is to use an optional parameter `--tns-modules-version` when executing `create` command. The other possibility is to have defined the version in `package.json` file which comes with the app contents. 
In both cases the intallation of tns_modules happens on `create` phase.

lib/services/project-service.ts
```diff
@@ -76,20 +77,32 @@ export class ProjectService implements IProjectService {
...
 			}
 
+            var tnsModulesVersion: string = null;
+            // Optional parameter has a higher priority than defined version in package.json
+            if (this.$options.tnsModulesVersion) {
+                tnsModulesVersion = this.$options.tnsModulesVersion;
+            }
+            else {
+                let coreModulesName = this.$tnsModulesService.npmTnsCoreModulesName;
+                tnsModulesVersion = this.$fs.readJson(path.join(appPath, constants.PACKAGE_JSON_FILE_NAME)).wait()[coreModulesName]["version"];
```
Fatme: Consider using `$projectDataService` instead `readJson`

lib/services/project-service.ts
```diff
@@ -101,16 +114,25 @@ export class ProjectService implements IProjectService {
 			} else {
 				shell.cp('-R', path.join(appSourcePath, "*"), appDestinationPath);
 			}
-			this.createBasicProjectStructure(projectDir,  projectId).wait();
+			
+			// Copy tns-core-modules package to node_modules destination directory
+			let nodeModulesDestination = path.join(appDestinationPath, constants.NODE_MODULES_FOLDER_NAME);
+            this.$fs.ensureDirectoryExists(nodeModulesDestination).wait();
+            shell.cp('-Rf', path.join(tnsModulesPath, '*'), nodeModulesDestination);
```
Fatme: You can substitute the logic above with `npm install`

lib/services/project-service.ts
```diff
@@ -111,6 +113,9 @@ export class ProjectService implements IProjectService {
 
 			this.$projectDataService.initialize(projectDir);
 			this.$projectDataService.setValue("id", projectId).wait();
+			//Start child process because --save --save-exact option is necessary when installing tns-core-modules
+			//in order to create "dependencies" key in package.json. 
+			this.$childProcess.exec("npm install " + constants.TNS_CORE_MODULES_NAME + " --save --save-exact", { cwd: projectDir }).wait();
```
ligaz: Should it be better to use `INodePackageManager` and expose this functionality there?

lib/services/platform-service.ts
```diff
@@ -178,7 +178,9 @@ export class PlatformService implements IPlatformService {
 			this.$pluginsService.ensureAllDependenciesAreInstalled().wait();
 			let tnsModulesDestinationPath = path.join(platformData.appDestinationDirectoryPath, constants.APP_FOLDER_NAME, PlatformService.TNS_MODULES_FOLDER_NAME);
 			this.$broccoliBuilder.prepareNodeModules(tnsModulesDestinationPath, this.$projectData.projectDir, platform, lastModifiedTime).wait();
-			
+			shell.cp("-Rf", path.join(tnsModulesDestinationPath, constants.TNS_CORE_MODULES_NAME, "*"), tnsModulesDestinationPath);
+			this.$fs.deleteDirectory(constants.TNS_CORE_MODULES_NAME).wait();
```
ligaz: It will be better to expose `moveDirectory` method in `IFileSystem` abstraction.

ligaz: Would it be better if this is handled in the Broccoli pipeline, in order to avoid excessive file system access by moving files around?

e2l3n:
I will squash all commits right before merging this pull request.

lib/tools/broccoli/node-modules-dest-copy.ts
```diff
@@ -75,6 +75,11 @@ export class DestCopy implements IBroccoliPlugin {
 		if(isPlugin) {
 			this.$pluginsService.prepare(dependency).wait();
 		}
+
+		if (dependency.name == constants.TNS_CORE_MODULES_NAME) {
```
Fatme: Use `===` instead `==`

lib/tools/broccoli/node-modules-dest-copy.ts
```diff
@@ -75,6 +75,11 @@ export class DestCopy implements IBroccoliPlugin {
...
+			shelljs.cp("-Rf", path.join(this.outputRoot, constants.TNS_CORE_MODULES_NAME, "*"), this.outputRoot);
+			this.$fs.deleteDirectory(path.join(this.outputRoot, constants.TNS_CORE_MODULES_NAME)).wait();
```
Fatme: Use can extract `path.join(this.outputRoot, constants.TNS_CORE_MODULES_NAME)` to separate variable

lib/node-package-manager.ts
```diff
@@ -53,6 +54,12 @@ export class NodePackageManager implements INodePackageManager {
 	public view(packageName: string, propertyName: string): IFuture<any> {
 		return this.loadAndExecute("view", [[packageName, propertyName], [false]]);
 	}
+
+	public executeNpmCommand(npmCommandName: string, currentWorkingDirectory: string): IFuture<void> {
+		return (() => {
+			 this.$childProcess.exec(npmCommandName, { cwd: currentWorkingDirectory }).wait();
+		}).future<void>()();
```
Fatme: You don't need to wrap this to future. Just `return this.$childProcess.exec(npmCommandName, { cwd: currentWorkingDirectory })`

lib/services/project-service.ts
```diff
@@ -111,6 +112,7 @@ export class ProjectService implements IProjectService {
 
 			this.$projectDataService.initialize(projectDir);
 			this.$projectDataService.setValue("id", projectId).wait();
+			this.$npm.executeNpmCommand("npm install " + constants.TNS_CORE_MODULES_NAME + " --save --save-exact", projectDir).wait();
```
Fatme: Is it possible to specify version of `tns-core-modules`?

teobugslayer: Is it needed to specify both `--save` and `--save-exact`?

test/project-service.ts
```diff
@@ -2,6 +2,7 @@
 "use strict";
 
 import yok = require('../lib/common/yok');
+import ChildProcessLib = require("../lib/common/child-process");
```
teobugslayer: Consider using ES6 style imports

lib/services/project-service.ts
e2l3n: I think yes, if we want to specify the exact number and prevent change on `npm update`. Perhaps it's a good idea to add an optional parameter on `create` so that the user can pass the version. In both cases they will be able to edit the package.json and change the version.

lib/services/project-service.ts
```diff
@@ -111,6 +112,14 @@ export class ProjectService implements IProjectService {
 
 			this.$projectDataService.initialize(projectDir);
 			this.$projectDataService.setValue("id", projectId).wait();
+			
+			let tnsModulesVersion: string = this.$options.tnsModulesVersion;
+			let packageName: string  = constants.TNS_CORE_MODULES_NAME;
+			if (tnsModulesVersion) {
+				packageName = packageName.concat("@" + tnsModulesVersion);
```
Fatme: You can use `packageName = ${packageName}@${tnsModulesVersion}`

lib/services/project-service.ts
```diff
@@ -111,6 +112,14 @@ export class ProjectService implements IProjectService {
...
+			}
+
+			this.$npm.executeNpmCommand("npm install " + packageName + " --save --save-exact", projectDir).wait();
```
Fatme: You can use `npm install ${packageName} --save --save-exact`

test/project-service.ts
```diff
@@ -82,6 +87,9 @@ class ProjectIntegrationTest {
 			var expectedAppId = appId;
 			assert.equal(actualAppId, expectedAppId);
 
+			let tnsCoreModulesRecord = fs.readJson(tnsProjectFilePath).wait()["dependencies"][constants.TNS_CORE_MODULES_NAME];
+			assert.isTrue(tnsCoreModulesRecord != null);
```
Fatme: Use `!==`

lib/node-package-manager.ts
```diff
@@ -53,6 +54,12 @@ export class NodePackageManager implements INodePackageManager {
 	public view(packageName: string, propertyName: string): IFuture<any> {
 		return this.loadAndExecute("view", [[packageName, propertyName], [false]]);
 	}
+
+	public executeNpmCommand(npmCommandName: string, currentWorkingDirectory: string): IFuture<any> {
+		return (() => {
```
Fatme: `return this.$childProcess.exec(npmCommandName, { cwd: currentWorkingDirectory })`

lib/tools/broccoli/node-modules-dest-copy.ts
```diff
@@ -75,6 +75,12 @@ export class DestCopy implements IBroccoliPlugin {
 		if(isPlugin) {
 			this.$pluginsService.prepare(dependency).wait();
 		}
+
+		if (dependency.name === constants.TNS_CORE_MODULES_NAME) {
+			let tnsCoreModulesResourcePath = path.join(this.outputRoot, constants.TNS_CORE_MODULES_NAME);
+			shelljs.cp("-Rf", path.join(tnsCoreModulesResourcePath, "*"), this.outputRoot);
+			this.$fs.deleteDirectory(tnsCoreModulesResourcePath).wait();
```
teobugslayer: What happens when shelljs.cp throws (i.e. I/O error)? Should we deleteDirectory in this case?

test/project-service.ts
teobugslayer: Or, use `assert.isObject(tnsCoreModulesRecord)`

lib/tools/broccoli/node-modules-dest-copy.ts
e2l3n: You are right, we have to handle this case properly. Take a look at the next commit where the whole function prepareNodeModules is put in a `try-catch` block. Deleting the `app` folder in the `catch` will guarantee the process being triggered on next `prepare <platform>` command.

lib/services/platform-service.ts
```diff
@@ -175,9 +175,15 @@ export class PlatformService implements IPlatformService {
...
-			this.$pluginsService.ensureAllDependenciesAreInstalled().wait();
-			let tnsModulesDestinationPath = path.join(platformData.appDestinationDirectoryPath, constants.APP_FOLDER_NAME, PlatformService.TNS_MODULES_FOLDER_NAME);
-			this.$broccoliBuilder.prepareNodeModules(tnsModulesDestinationPath, this.$projectData.projectDir, platform, lastModifiedTime).wait();
+			try {
+				this.$pluginsService.ensureAllDependenciesAreInstalled().wait();
+				let tnsModulesDestinationPath = path.join(platformData.appDestinationDirectoryPath, constants.APP_FOLDER_NAME, PlatformService.TNS_MODULES_FOLDER_NAME);
+				this.$broccoliBuilder.prepareNodeModules(tnsModulesDestinationPath, this.$projectData.projectDir, platform, lastModifiedTime).wait();
+			} catch(error) {
+				this.$logger.debug(error);
+				this.$logger.error("Processing tns core modules failed.");
```
teobugslayer: This is user visible error. Consider at least showing the users the actual error. This way, they have a chance to understand what went wrong.

teobugslayer:
:+1: after addressing my small remark and a green build.

lib/services/platform-service.ts
Fatme: The error might be produced while processing some other module or plugin. It is not obligatory to be produced while "Processing tns core modules"

lib/services/project-service.ts
```diff
@@ -111,6 +112,14 @@ export class ProjectService implements IProjectService {
 
 			this.$projectDataService.initialize(projectDir);
 			this.$projectDataService.setValue("id", projectId).wait();
+			
+			let tnsModulesVersion: string = this.$options.tnsModulesVersion;
+			let packageName: string  = constants.TNS_CORE_MODULES_NAME;
```
Fatme: You do not need to specify the variable type.

Issue #819, opened 2015-08-26
Platform android exists after failed attempt to add it


teobugslayer:
$ tns platform add android --framework-path ~/.npm/tns-android/1.3.0/package.tgz 
Copying template files...
Installing  .npm/tns-android/1.3.0/package.tgz
AssertionError: package metadata must include name
    at getCacheRoot (nativescript-cli/node_modules/npm/lib/cache/cached-package-root.js:10:3)
    at addPlacedTarball_ (nativescript-cli/node_modules/npm/lib/cache/add-local-tarball.js:64:26)
    at nativescript-cli/node_modules/npm/lib/cache/add-local-tarball.js:59:12
    at nativescript-cli/node_modules/npm/lib/cache/get-stat.js:19:12
    at FSReqWrap.oncomplete (evalmachine.<anonymous>:95:15)
$ cp ~/.npm/tns-android/1.3.0/package.tgz android-1.3.0.tgz
$ ../../bin/tns platform add android --framework-path ./android-1.3.0.tgz 
Platform android already added

Note: the full fix require cooperation from npm developers.

PR #820, opened 2015-08-26
Guard against errors in npm

teobugslayer:
See https://github.com/NativeScript/nativescript-cli/issues/819

Note: npm actually `assert` their parameters instead of calling the provided callback with an error.
In our code, this means that an exception is being thrown in async code which was running without being guarded in a fiber. The end result is process tear down without any chance for us to intervene.

We need to file a issue to npm to fix their code to call the callback with an error in this case.


rosen-vladimirov:
:+1:  after green build

PR #821, opened 2015-08-27
Fix tns doctor command when javac is not added to PATH

Fatme:
Should be merged after this PR https://github.com/telerik/mobile-cli-lib/pull/445


dtopuzov:
:+1:

dtopuzov:
:+1: from @teobugslayer

PR #822, opened 2015-08-27
Respect settings.gradle file from template

Fatme:
Android runtime introduced gradle.settings file in order to produce apk named <projectName>-debug.apk or <projectName>-release.apk.
Currently the produced .apk files are named android-debug.apk or android-release.apk.


teobugslayer:
:+1:

Issue #772, opened 2015-08-11
Platform update ios command does not update the iOS runtime


vchimev:
It seems like `tns platform update ios` does not download the iOS runtime requested version to the npm cache, therefore the project could not be updated to it.

Output:
```
vchimevmac:issueApp vchimev$ tns platform update ios
ENOENT, open '/Users/vchimev/.npm/tns-ios/1.2.1/package/framework/__PROJECT_NAME__.xcodeproj/project.pbxproj'
```

Workaround:
```
tns platform remove ios
tns platform add ios
```

NickBranstein:
Related to this - I've updated to NativeScript 1.2.2 

```
tns --version
1.2.2
```

After removing and adding the ios platform

```
tns platform remove ios
tns platform add ios
```

My package.json now reports the tns-ios package is 1.2.1 - is this normal - I would expect it to say 1.2.2?

```
{
	"nativescript": {
		"id": "org.nativescript.navigation",
		"tns-ios": {
			"version": "1.2.1"
		}
	}
}
```

Fatme:
@NickBranstein this is expected behavior.
`tns platform add ios` downloads the latest available version. Currently the latest available version for `tns-ios` is `1.2.1`

maknz:
I had this issue after upgrading the CLI to 1.2.3 today, and with a new 1.2.2 release (we were on 1.2.1 prior). `tns platform update ios` failed with the above error. Removing the platform, bumping the package.json version manually and `tns install` worked.

vchimev:
Output of replication with `tns platform update ios --log trace`:
https://gist.github.com/vchimev/31d3fe6ba028ff5ca57b

vchimev:
I tried to update ios runtime from 1.2.2 to forthcoming 1.3.0.

The platform update was successful itself, but after that I failed to build the app.
Output: https://gist.github.com/vchimev/e054b8d10139cfc8c2c8

PR #823, opened 2015-08-27
Fix platform update

Fatme:
If the new version of runtime is not cached, tns platform update throws an error
Fixes https://github.com/NativeScript/nativescript-cli/issues/772


teobugslayer:
Cool! :+1:

PR #828, opened 2015-08-28
Do not embed static frameworks

jasssonpet:
An example for such is the Facebook SDK.

ping @fealebenpae


fealebenpae:
Looks good to me. @Fatme should weigh in as well.

Fatme:
:+1:  Looks ok to me too

PR #843, opened 2015-09-01
Implement command line option to run static bindings generator

teobugslayer:
Part of https://github.com/NativeScript/android-runtime/issues/103

### DIFF
--- a/lib/commands/build.ts
+++ b/lib/commands/build.ts
@@ -4,8 +4,8 @@
 export class BuildCommandBase {
 	constructor(private $platformService: IPlatformService) { }
 
-	executeCore(args: string[]): IFuture<void> {
-		return this.$platformService.buildPlatform(args[0]);
+	executeCore(args: string[], buildConfig?: IBuildConfig): IFuture<void> {
+		return this.$platformService.buildPlatform(args[0], buildConfig);
 	}
 }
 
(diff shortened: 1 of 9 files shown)


Plamen5kov:
:+1:

lib/definitions/project.d.ts
```diff
@@ -26,13 +26,17 @@ interface IPlatformProjectServiceBase {
 	getPluginPlatformsFolderPath(pluginData: IPluginData, platform: string): string;
 }
 
+interface IBuildConfig {
+	runSbGenerator?: boolean;
```
rosen-vladimirov: why `runSbGenerator` is not mandatory?

lib/commands/build.ts
```diff
@@ -25,14 +25,16 @@ $injector.registerCommand("build|ios", BuildIosCommand);
...
+				private $platformsData: IPlatformsData,
+				private $options: IOptions) {
 		super($platformService);
 	}
 
-	public allowedParameters: ICommandParameter[] = [];
-
 	public execute(args: string[]): IFuture<void> {
-		return this.executeCore([this.$platformsData.availablePlatforms.Android]);
+		let config = this.$options.staticBindings ? { runSbGenerator: true } : undefined;
```
rosen-vladimirov:
```TypeScript
let config: IBuildConfig = {runSbGenerator: this.$options.staticBindings};
```

lib/definitions/project.d.ts
teobugslayer: Because it is still experimental option. Once we prove it is stable, we will switch it to be on by default - possibly for {N} 1.4.

lib/commands/build.ts
teobugslayer: I prefer not to move around empty objects unless really needed.

rosen-vladimirov:
:+1:  and :+1:  from @Mitko-Kerezov

Issue #839, opened 2015-09-01
No warning/error when build android in release with out signing options


dtopuzov:
As of 1.3.0 {N} will not support android builds in release configuration with out specified signing options.

Current behavior:
tns build android --release
=>
Build start and after 1 min you have some error message that is not clear enough and you should run ---log trace in order to get the real reason for the failure - no signing options specified.

We should validate the command and provide appropriate error message.

PR #845, opened 2015-09-01
Warn users that android related command need all --key-store- paramet…

teobugslayer:
…ers when using --release

Fixes https://github.com/NativeScript/nativescript-cli/issues/839

### DIFF
--- a/lib/commands/build.ts
+++ b/lib/commands/build.ts
@@ -26,7 +26,8 @@ $injector.registerCommand("build|ios", BuildIosCommand);
 export class BuildAndroidCommand extends BuildCommandBase implements  ICommand {
 	constructor($platformService: IPlatformService,
 				private $platformsData: IPlatformsData,
-				private $options: IOptions) {
+				private $options: IOptions,
+				private $errors: IErrors) {
 		super($platformService);
 	}
 
@@ -36,5 +37,14 @@ export class BuildAndroidCommand extends BuildCommandBase implements  ICommand {
 	}
 
 	public allowedParameters: ICommandParameter[] = [];
+
+	public canExecute(args: string[]): IFuture<boolean> {
+		return (() => {
+			if (this.$options.release && (!this.$options.keyStorePath || !this.$options.keyStorePassword || !this.$options.keyStoreAlias || !this.$options.keyStoreAliasPassword)) {
+				this.$errors.fail("When producing a release build, you need to specify all --key-store-* options.");
+			}
+			return args.length === 0;
+		}).future<boolean>()();
+	}
 }
 $injector.registerCommand("build|android", BuildAndroidCommand);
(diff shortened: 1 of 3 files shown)


teobugslayer:
Ping @ikoevska for the message

lib/commands/build.ts
```diff
@@ -36,5 +37,14 @@ export class BuildAndroidCommand extends BuildCommandBase implements  ICommand {
 	}
 
 	public allowedParameters: ICommandParameter[] = [];
+
+	public canExecute(args: string[]): IFuture<boolean> {
+		return (() => {
+			if (this.$options.release && (!this.$options.keyStorePath || !this.$options.keyStorePassword || !this.$options.keyStoreAlias || !this.$options.keyStoreAliasPassword)) {
```
rosen-vladimirov: can you add a check that args.length is 0.
With the code in master, if you try to execute this command with parameters, you receive: "This command does not accept parameters.". But after adding canExecute method, the command service will not check the number of passed parameters and you'll be able to call ```tns build android goshko peshko```

lib/commands/deploy.ts
```diff
@@ -3,12 +3,23 @@
...
+				private $options: IOptions,
+				private $errors: IErrors) { }
 
 	execute(args: string[]): IFuture<void> {
 		return this.$platformService.deployOnDevice(args[0]);
 	}
 
+	public canExecute(args: string[]): IFuture<boolean> {
+		return (() => {
+			if (this.$options.release && (!this.$options.keyStorePath || !this.$options.keyStorePassword || !this.$options.keyStoreAlias || !this.$options.keyStoreAliasPassword)) {
```
rosen-vladimirov: same as the other command - we need check for args.length

lib/commands/deploy.ts
```diff
@@ -3,12 +3,23 @@
...
+				this.$errors.fail("When producing a release build, you need to specify all --key-store-* options.");
+			}
+			return true;
+		}).future<boolean>()();
+	}
+
 	allowedParameters = [this.$platformCommandParameter];
```
rosen-vladimirov: you should change allowedParameters to be:
```TypeScript
allowedParameters = [];
```
in order to avoid confusion, as it will never be used.

lib/commands/run.ts
```diff
@@ -34,5 +36,14 @@ export class RunAndroidCommand extends RunCommandBase implements ICommand {
 	public execute(args: string[]): IFuture<void> {
 		return this.executeCore([this.$platformsData.availablePlatforms.Android]);
 	}
+
+	public canExecute(args: string[]): IFuture<boolean> {
+		return (() => {
+			if (this.$options.release && (!this.$options.keyStorePath || !this.$options.keyStorePassword || !this.$options.keyStoreAlias || !this.$options.keyStoreAliasPassword)) {
```
rosen-vladimirov: we need check for args.length here as well :)

ikoevska:
:+1: for the msgs

rosen-vladimirov:
:+1:

Issue #831, opened 2015-08-28
Fail to build app with cocoapods


vchimev:
Steps to reproduce:
```
tns create TNS_App
cd TNS_App
tns plugin add <plugin containing Podfile>
tns platform add ios
tns build ios
```
Output with TNS_App project:
https://gist.github.com/vchimev/b004ab09dc188be2b712
Output with podPlatformPlugin project:
https://gist.github.com/vchimev/ef6a16886853f085defd

PR #848, opened 2015-09-02
Create main target scheme before installing pods.

e2l3n:
<b>Changes:</b> Create `.xcscheme` for the main target before installing pods.
<b>Reason:</b> `xcodebuild` fails to build workspaces without scheme in the main target.
Addresses <b>issue</b> #831.

### DIFF
--- a/lib/services/ios-project-service.ts
+++ b/lib/services/ios-project-service.ts
@@ -347,7 +347,14 @@ export class IOSProjectService extends projectServiceBaseLib.PlatformProjectServ
 				if(firstPostInstallIndex !== -1 && firstPostInstallIndex !== projectPodfileContent.lastIndexOf(IOSProjectService.PODFILE_POST_INSTALL_SECTION_NAME)) {
 					this.$logger.warn(`Podfile contains more than one post_install sections. You need to open ${this.projectPodFilePath} file and manually resolve this issue.`);
 				}
-
+				
+				let pbxprojFilePath = path.join(this.platformData.projectRoot, this.$projectData.projectName + IOSProjectService.XCODE_PROJECT_EXT_NAME, "xcuserdata");
+				if(!this.$fs.exists(pbxprojFilePath).wait()) {
+					this.$logger.info("Creating project scheme...");
+					let createScheme_rb = `echo \"require 'xcodeproj'; xcproj = Xcodeproj::Project.open('${this.$projectData.projectName}.xcodeproj'); xcproj.recreate_user_schemes; xcproj.save\" | ruby`;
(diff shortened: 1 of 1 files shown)


lib/services/ios-project-service.ts
```diff
@@ -347,7 +347,14 @@ export class IOSProjectService extends projectServiceBaseLib.PlatformProjectServ
 				if(firstPostInstallIndex !== -1 && firstPostInstallIndex !== projectPodfileContent.lastIndexOf(IOSProjectService.PODFILE_POST_INSTALL_SECTION_NAME)) {
 					this.$logger.warn(`Podfile contains more than one post_install sections. You need to open ${this.projectPodFilePath} file and manually resolve this issue.`);
 				}
-
+				
+				let pbxprojFilePath = path.join(this.platformData.projectRoot, this.$projectData.projectName + IOSProjectService.XCODE_PROJECT_EXT_NAME, "xcuserdata");
+				if(!this.$fs.exists(pbxprojFilePath).wait()) {
+					this.$logger.info("Creating project scheme...");
+					let createScheme_rb = `echo \"require 'xcodeproj'; xcproj = Xcodeproj::Project.open('${this.$projectData.projectName}.xcodeproj'); xcproj.recreate_user_schemes; xcproj.save\" | ruby`;
```
teobugslayer: We use only camelCase for variables. Consider naming it something like `createSchemeRubyScript`

teobugslayer: I've tested on a clean machine with only Xcode 6.4, node and nativescript installed. In my case, the following error was shown:
```
require "xcodeproj"
LoadError: cannot load such file -- xcodeproj
```

Do we need to install some additional dependencies?

teobugslayer: If we go ruby way, you can consider using the one-liners syntax: `ruby -e "our code here"`, which will save us some nasty escaping.

e2l3n: The ruby gem `cocoapods` depends on `xcodeproj` so it is automatically installed. Do you have the latest `cocoapods v.0.38.2` ? Users can manually uninstall gem `xcodeproj`. However doing so will cause a few lines down the execution of `pod install` command to fail as well although there would be installed `cocoapods` (with missing `xcodeproj`). Here is a snippet of the error message:
`/Library/Ruby/Site/2.0.0/rubygems/dependency.rb:315:in ``to_specs': Could not find 'xcodeproj' (~> 0.26.3) among 35 total gem(s) (Gem::LoadError)``

Perhaps there should be a check for installed `xcodeproj` gem as well. I've added it in the latest commit.

teobugslayer:
:+1:

PR #849, opened 2015-09-02
Fix ios commands when platforms dir is removed

rosen-vladimirov:
When you run `tns run ios`, nativescript-cli will add the ios platform in case you have not added it before that. But in case you have added the platform before that, but you have removed the  platforms dir, the call fails.
Make sure the platforms/ios directory exists before executing any action of copying the files.


teobugslayer:
:+1:

PR #853, opened 2015-09-02
Add --static-bindings to run and deploy commands

teobugslayer:
Part of https://github.com/NativeScript/android-runtime/issues/103


Fatme:
Shouldn't `runSbGenerator` option be passed to `gradle.build`?

teobugslayer:
This happens already: see https://github.com/NativeScript/nativescript-cli/blob/master/lib/services/android-project-service.ts#L200

Fatme:
:+1:

PR #855, opened 2015-09-02
Initialize project Podfile

### LINKED ISSUE #852: CocoaPods need modulemaps
We need the CocoaPods tooling to generate a framework target for each pod, as opposed to a static library target, so that it can have a modulemap generated.

The NativeScript for iOS runtime depends on modulemaps to discover native APIs and make them available to JavaScript.

fealebenpae:
Make sure the Podfile contains the "use_frameworks" directive by default.

Fixes #852


Fatme:
:+1:

PR #857, opened 2015-09-02
Clean res dir of Android project

rosen-vladimirov:
Currently when creating Android project, we copy the full src directory from android framework. After that we execute `copyResValues` which is totally not needed as we've already copied everything. Instead we have to clean the copied values as currently if you target android sdk 19 for example, we'll still copy res/values-v21 directory in the project and the build will fail.
Check all `values-v<SDK>` directories and remove the ones that have higher version than currently targeted one.


Fatme:
:+1:

Issue #854, opened 2015-09-02
CLI commands copy {N} modules to the native project from app/tns_modules folder if exists


vchimev:
Commands: prepare, build, run, emulate.

First time when executing the commands above they copy {N} modules to the native project from `node_modules/tns-core-modules` folder as described dependency in the package.json for {N} project created with CLI 1.3.0.

Every next time the commands above are executed they copy {N} modules to the native project from `app/tns_modules` folder if exists which is incorrect.

PR #858, opened 2015-09-02
Do not copy tns_modules from app dir

rosen-vladimirov:
In case there's tns_modules in app directory, they are copied on each prepare. Skip tns_modules as they are now installed as npm dependency. Warn the user that the tns_modules from app will not be used.
Fixes https://github.com/NativeScript/nativescript-cli/issues/854

### DIFF
--- a/lib/services/platform-service.ts
+++ b/lib/services/platform-service.ts
@@ -160,7 +160,16 @@ export class PlatformService implements IPlatformService {
 				.filter(directoryName => directoryName !== "tns_modules")
 				.each(directoryName => this.$fs.deleteDirectory(path.join(appDestinationDirectoryPath, directoryName)).wait())
 				.value();
-			shell.cp("-Rf", appSourceDirectoryPath, platformData.appDestinationDirectoryPath);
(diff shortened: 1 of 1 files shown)


lib/services/platform-service.ts
```diff
@@ -160,7 +160,16 @@ export class PlatformService implements IPlatformService {
 				.filter(directoryName => directoryName !== "tns_modules")
 				.each(directoryName => this.$fs.deleteDirectory(path.join(appDestinationDirectoryPath, directoryName)).wait())
 				.value();
-			shell.cp("-Rf", appSourceDirectoryPath, platformData.appDestinationDirectoryPath);
+
+			// Copy all files from app dir, but make sure to exclude tns_modules
+			let sourceFiles = this.$fs.readDirectory(appSourceDirectoryPath).wait();
+			if(_.contains(sourceFiles, "tns_modules")) {
```
Fatme: I think we have a constant for `tns_modules`

Fatme:
:+1:

Issue #838, opened 2015-09-01
No warning/error when add library with eclipse project


dtopuzov:
As of 1.3.0 {N} will support only gradle native projects, so we need some kind of warning/error when you try to add library/plugin with eclipse native project.

Current behavior:
tns library add android QA-TestApps/external-lib/AndroidAppProject --path TNS_App
=>
Library c:\Git\nativescript-cli-tests\QA-TestApps\external-lib\AndroidAppProject was successfully added for android platform.

But it does not work because it is eclipse project.

PR #859, opened 2015-09-03
Throw an error when add library command is executed with eclipse project

Fatme:
Fixes https://github.com/NativeScript/nativescript-cli/issues/838


teobugslayer:
:+1:

Issue #811, opened 2015-08-21
LiveSync reports an error and stops when editing CSS on windows


EdCharbeneau:
System Specs:
- Windows 8.1 pro
- {N} 1.2.3
- Device LG G4 (5.1)

When using LiveSync with `tns livesync android --watch` editing CSS files reports and error and stops LiveSync. LiveSync can be restarted again successfully until another edit is saved. **This only happens with platform specific CSS files ex: `platform.android.css`**

I tried this with both VS Code and notepad with the same result.

A video documenting the error can be seen here: http://1drv.ms/1Jp1M5s

tjvantoll:
This issue appears to be with `*.android.css` and `*.ios.css` files, and I can also recreate the problem on OS X.

For me changing `*.android.css` files crashes the `livesync` process. Changing `.ios.css` files doesn't crash `livesync`, but I just don't see the changes I made in my `*.ios.css` file actually being applied to my app.

teobugslayer:
@EdCharbeneau , @tjvantoll 

Thanks for reporting this issue. We reproduced it and created a fix which hopefully will be included in 1.3.

PR #865, opened 2015-09-03
Fix livesync with platform specific files

teobugslayer:
See https://github.com/NativeScript/nativescript-cli/issues/811


Fatme:
:+1:

Issue #866, opened 2015-09-03
Not detecting java on my system


anshulguleria:
My OS details are:
```
Distributor ID:	Ubuntu
Description:	Ubuntu 15.04
Release:	15.04
Codename:	vivid
```
and java details are: 
```
Picked up JAVA_TOOL_OPTIONS: -javaagent:/usr/share/java/jayatanaag.jar 
openjdk version "1.8.0_45-internal"
OpenJDK Runtime Environment (build 1.8.0_45-internal-b14)
OpenJDK 64-Bit Server VM (build 25.45-b02, mixed mode)
```

My environment variables are also set. I can access java command

but when ever i run `tns doctor` it says java not available or not setup properly

teobugslayer:
Hi @anshulguleria,

I noticed that you are using OpenJDK. We recommend using to the official Oracle JDK 8 as described in our [docs](http://docs.nativescript.org/setup/ns-cli-setup/ns-setup-linux.html) but you should be OK.

Also, please verify that:
 * your JAVA_HOME is set up correctly to point to your JDK root, using `echo $JAVA_HOME`. On my machine the path is /usr/lib/jvm/java-8-oracle but it will probably be different on yours.
 * the *full* path to java executable is correctly set in your PATH environment variable. Check that the path returned by `which java` exists in your PATH.

If you continue having problems, please run `tns doctor --log trace > log.txt` and send us the log.txt file. You can create a gist with it.

teobugslayer:
Ah, I saw you've issued a PR with the fix. Thank you!

anshulguleria:
Hi @teobugslayer 
My PATHS are all fine. After some digging up I found out that your sysinfo.js in mobile-cli-api project was using regex for only `java` not `openjdk`. I submitted PR on it.

Thanks for quick reply though.

teobugslayer:
@anshulguleria, I've merged your change in {N} master and it will be included in our next public release. Thank you again!

PR #869, opened 2015-09-04
Fix Java detection for OpenJDK

teobugslayer:
See https://github.com/NativeScript/nativescript-cli/issues/866


teobugslayer:
:+1:

Issue #827, opened 2015-08-28
Adding platform before adding plugin fails to install cocoapods


vchimev:
Steps to reproduce:
```
tns create testApp
cd testApp
tns platform add ios
tns plugin add <plugin containing Podfile>
tns prepare ios
```
It fails to install pods and produce `.xcworkspace` file when plugin containing Podfile is added to the project.

The workflow below works as expected:
```
tns create testApp
cd testApp
tns plugin add <plugin containing Podfile>
tns platform add ios
tns prepare ios
```
It install pods successfully and produces `.xcworkspace` when plugin containing Podfile is added to the project.

Fatme:
The issues above are not related.

fealebenpae:
This is the structure of the plugin I'm testing with:
```
├── package.json
└── platforms
    └── ios
        └── Podfile
```
The Podfile looks like this:
```
source 'https://github.com/CocoaPods/Specs.git'
platform :ios, '8.1'
pod 'GoogleMaps'
```

As far as I can tell, `tns prepare ios` should pick up the plugin, read the Podfile, and fire up the CocoaPods tooling. However, `PlatformService.preparePlatform` tries to see if the mtime of any file inside `node_modules/**` is newer than mtime of the `platforms/ios/<project-name>/app` folder and only prepares those plugins that pass this check. Since the above plugin never ends up in the `platforms/ios/<project-name>/app/tns_modules` folder because it has no JavaScript files, it will never pass the mtime check and will never be prepared.

It seems to me that the mtime check should be extended to cover things outside the `app` folder, such as the Xcode project file, which will actually get modified by this plugin. The question is, how do I extend the check in `PlatformService.preparePlatform` so that it can cover both the `app` folder and the Xcode project file? Also, how would this work for the same plugin if it had a couple of `.jar`s in `platforms/android` and nothing else?

fealebenpae:
@teobugslayer and I talked IRL about this, and it turns out this is far too big a change to be done this late in the release cycle. @teobugslayer also pointed out a neat workaround - add an empty .js file in the plugin to force the prepare mechanism to install the pods in the plugin.

PR #870, opened 2015-09-04
Execute pod install when a plugin containing only Podfile is added after the ios platform

Fatme:
Fixes https://github.com/NativeScript/nativescript-cli/issues/827


lib/definitions/project.d.ts
```diff
@@ -43,7 +43,7 @@ interface IPlatformProjectService {
 	addLibrary(libraryPath: string): IFuture<void>;
 	canUpdatePlatform(currentVersion: string, newVersion: string): IFuture<boolean>;
 	updatePlatform(currentVersion: string, newVersion: string): IFuture<void>;
-	preparePluginNativeCode(pluginData: IPluginData): IFuture<void>;
+	preparePluginNativeCode(pluginData: IPluginData, options?: any): IFuture<void>;
```
teobugslayer: Maybe rename options to opts? In implementations this parameter is called opts.

teobugslayer:
:+1:

PR #880, opened 2015-09-07
Fix errors when ANDROID_HOME is not set

rosen-vladimirov:
In case ANDROID_HOME environment variable is not set, we show `Arguments to path.join must be strings` error. This is because we collect the information first and show the real errors after that, but we fail during information gathering.
Check androidHome before using it in path.join.


Fatme:
:+1:

teobugslayer:
:+1:

PR #884, opened 2015-09-07
Add code coverage as a dev dependency

teobugslayer:
Use istanbul as a code coverage tool.

Code coverage is run on every build and one can check its output on the build console log.


rosen-vladimirov:
shouldn't we use it by default in `package.json` when running the tests?

rosen-vladimirov:
:+1:

PR #886, opened 2015-09-08
Update correctly ant template to gradle

Fatme:


lib/services/android-project-service.ts
```diff
@@ -153,11 +153,21 @@ class AndroidProjectService extends projectServiceBaseLib.PlatformProjectService
 	}
 
 	public canUpdatePlatform(currentVersion: string, newVersion: string): IFuture<boolean> {
-		return Future.fromResult<boolean>(true);
+		return (() => {
```
teobugslayer: perhaps `return Future.fromResult(true)`?

lib/services/android-project-service.ts
```diff
@@ -153,11 +153,21 @@ class AndroidProjectService extends projectServiceBaseLib.PlatformProjectService
...
+			return true;
+		}).future<boolean>()();
 	}
 
-	public updatePlatform(currentVersion: string, newVersion: string): IFuture<void> {
-		return Future.fromResult();
+	public updatePlatform(currentVersion: string, newVersion: string, canUpdate: boolean): IFuture<boolean> {
```
teobugslayer: I'd really like to see documented what return values `true` and `false` signify.

lib/services/ios-project-service.ts
```diff
@@ -240,24 +241,33 @@ export class IOSProjectService extends projectServiceBaseLib.PlatformProjectServ
...
-			this.$logger.info("Copied from %s at %s.", cachedPackagePath, this.platformData.projectRoot);
-
-			let pbxprojFilePath = path.join(this.platformData.projectRoot, this.$projectData.projectName + IOSProjectService.XCODE_PROJECT_EXT_NAME, "project.pbxproj");
-			this.replaceFileContent(pbxprojFilePath).wait();
-		}).future<void>()();
+			if(!canUpdate) {
+				let isUpdateConfirmed = this.$prompter.confirm(`We need to override xcodeproj file. The old one will be saved at ${this.$options.profileDir}. Are you sure?`, () => true).wait();
+				if(isUpdateConfirmed) {
+					// Copy old file to options["profile-dir"]
+					let sourceFile = path.join(this.platformData.projectRoot, util.format("%s.xcodeproj", this.$projectData.projectName));
```
teobugslayer: You can use templated strings here, but not a merge stopper.

lib/services/ios-project-service.ts
```diff
@@ -240,24 +241,33 @@ export class IOSProjectService extends projectServiceBaseLib.PlatformProjectServ
...
+					let destinationFile = path.join(this.$options.profileDir, "xcodeproj");
+					this.$fs.deleteDirectory(destinationFile).wait();
+					shell.cp("-R", path.join(sourceFile, "*"), destinationFile);
+					this.$logger.info("Backup file %s at location %s", sourceFile, destinationFile);
+					this.$fs.deleteDirectory(path.join(this.platformData.projectRoot, util.format("%s.xcodeproj", this.$projectData.projectName))).wait();
```
teobugslayer: Isn't this `deleteDirectory(sourceFile)`?

lib/services/ios-project-service.ts
```diff
@@ -240,24 +241,33 @@ export class IOSProjectService extends projectServiceBaseLib.PlatformProjectServ
...
+
+					// Copy xcodeProject file
+					let cachedPackagePath = path.join(this.$npmInstallationManager.getCachedPackagePath(this.platformData.frameworkPackageName, newVersion), constants.PROJECT_FRAMEWORK_FOLDER_NAME, util.format("%s.xcodeproj", IOSProjectService.IOS_PROJECT_NAME_PLACEHOLDER));
```
teobugslayer: Again, you can use template string, but not a merge stopper.

lib/services/ios-project-service.ts
```diff
@@ -240,24 +241,33 @@ export class IOSProjectService extends projectServiceBaseLib.PlatformProjectServ
...
+					shell.cp("-R", path.join(cachedPackagePath, "*"), path.join(this.platformData.projectRoot, util.format("%s.xcodeproj", this.$projectData.projectName)));
```
teobugslayer: again, can you use `sourceFile` here?

lib/services/android-project-service.ts
```diff
@@ -152,11 +153,19 @@ class AndroidProjectService extends projectServiceBaseLib.PlatformProjectService
...
-		return Future.fromResult<boolean>(true);
+		return Future.fromResult(true);
 	}
 
-	public updatePlatform(currentVersion: string, newVersion: string): IFuture<void> {
-		return Future.fromResult();
+	public updatePlatform(currentVersion: string, newVersion: string, canUpdate: boolean): IFuture<boolean> {
+		return (() => {
+			if(semver.eq(newVersion, AndroidProjectService.MIN_RUNTIME_VERSION_WITH_GRADLE)) {
+				this.$commandsService.tryExecuteCommand(process.argv[2], ["remove"].concat(process.argv.slice(4))).wait();
```
teobugslayer: Imho, this is a highly suspicious code - our project service depends on the UI code, which means that a lower-layered code depends on the higher level one. Can we approach this in another way?

lib/services/android-project-service.ts
Fatme: Done, you can check the last commit :)

lib/services/android-project-service.ts
Fatme: The another approach is to use `childProcess.spawn` but I'm not sure if this is a better choice.  I'm opened to any suggestions :)

teobugslayer:
Squash and go, :+1:

Issue #875, opened 2015-09-07
Init command to configure tns-core-modules version


vchimev:
Init command to configure tns-core-modules version in package.json of the project.

PR #890, opened 2015-09-08
Configure tns-core-modules version when init command is executed

Fatme:
Fixes https://github.com/NativeScript/nativescript-cli/issues/875

### DIFF
--- a/lib/services/init-service.ts
+++ b/lib/services/init-service.ts
@@ -9,7 +9,8 @@ import semver = require("semver");
 export class InitService implements IInitService {
 	private static MIN_SUPPORTED_FRAMEWORK_VERSIONS: IStringDictionary = {
 		"tns-ios": "1.1.0",
-		"tns-android": "1.1.0"
+		"tns-android": "1.1.0",
+		"tns-core-modules": "1.2.0"
 	};
 
 	private _projectFilePath: string;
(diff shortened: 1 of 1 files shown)


lib/services/init-service.ts
```diff
@@ -56,6 +57,12 @@ export class InitService implements IInitService {
 					});
 				}
 
+				let dependencies = projectData["dependencies"];
+				if(!dependencies) {
+					projectData["dependencies"] = Object.create(null);
```
teobugslayer: Perhaps `projectData.dependencies = {}`? You can use {} here, because we just serialize the object immediately.

lib/services/init-service.ts
```diff
@@ -56,6 +57,12 @@ export class InitService implements IInitService {
...
+				}
+				projectData["dependencies"][constants.TNS_CORE_MODULES_NAME] = this.getVersionData(constants.TNS_CORE_MODULES_NAME).wait()["version"];
```
teobugslayer: Perhaps `projectData.dependencies[constants.TNS_CORE_MODULES_NAME] = this.getVersionData(constants.TNS_CORE_MODULES_NAME).wait().version`

teobugslayer:
:+1: but you may consider using the simplified syntax I proposed.

PR #902, opened 2015-09-10
Fix undefined is not a function error

Fatme:
If android is not set to PATH and tns build android command is executed, undefined is not a function errror is thrown


teobugslayer:
:+1:

Issue #900, opened 2015-09-10
Inaccurate error message when Android SDK Build tools are not installed


siyka-marinova:
http://screencast.com/t/WaIK7DzQ

The error message in case Android SDK Build tools are not installed says: "You need to have the Android SDK Build-tools installed on your system. You can install any version in the following range: '>=22 <=22'."
The "range" part of the message could be improved.

rosen-vladimirov:
This is a valid range in fact, but as you are the second person that points this as error (I still think it is not :smile_cat:  ) I'll try to show another message :)

siyka-marinova:
Agree it's a valid range, but the error message could be improved (maybe remove the range and pick one side of it) :)

PR #903, opened 2015-09-11
Improve android tools warnings

rosen-vladimirov:
* Fix incorrect message for  selected target SDK (Bulgarian symbols were used)
* Improve message for build tools when the prefix and suffix have the same version (>=22 <=22) - https://github.com/NativeScript/nativescript-cli/issues/900
* Rename supportLibraryVersion to supportRepositoryVersion and improve messaging as Gradle does not work with Android Support Library - instead it works with Android Support Repository.


Fatme:
:+1:

PR #906, opened 2015-09-11
Fix doctor command output

rosen-vladimirov:
In case gradle is not setup correctly or some of the required Android Tools are missing, `tns doctor` command show the warnings, but at the end a "No issues detected" message is shown. Make sure to prevent such message when there are issues by modifying validateInfo's return type to boolean and use it in doctor service.


teobugslayer:
:+1:

PR #910, opened 2015-09-12
Support xcconfig file from plugin - close #883

### LINKED ISSUE #883: Support xcconfig file from plugin
The `.xcconfig` file should be in `platforms/ios` folder of the plugin and will be included into native project's xcconfig file.

jasssonpet:
Set the xcconfig to the project template, so that building from the Xcode project is the same as from the CLI: https://github.com/NativeScript/ios-runtime/pull/302/

If the added plugin has a `build.xcconfig` file in `platforms/ios`, an include directive gets appended to the `build.xcconfig` in the project template.

Example (incomplete) Facebook plugin:

```
package.json
platforms/ios/FBSDKCoreKit.framework
platforms/ios/build.xcconfig
```

FBSDKCoreKit is a static framework and has a dependency to CoreGraphics. So we add it In the build.xcconfig: `OTHER_LDFLAGS = $(inherited) -framework CoreGraphics`.

Ping @fealebenpae

### DIFF
--- a/lib/services/ios-project-service.ts
+++ b/lib/services/ios-project-service.ts
@@ -333,6 +333,10 @@ export class IOSProjectService extends projectServiceBaseLib.PlatformProjectServ
 		return path.join(this.platformData.projectRoot, "Podfile");
 	}
 
+	private get projectXcconfigFilePath(): string {
+		return path.join(this.platformData.appDestinationDirectoryPath, "build.xcconfig");
+	}
+
 	private replace(name: string): string {
 		if(_.startsWith(name, '"')) {
 			name = name.substr(1, name.length-2);
@@ -369,6 +373,7 @@ export class IOSProjectService extends projectServiceBaseLib.PlatformProjectServ
 			this.prepareFrameworks(pluginPlatformsFolderPath, pluginData).wait();
 			this.prepareStaticLibs(pluginPlatformsFolderPath, pluginData).wait();
 			this.prepareCocoapods(pluginPlatformsFolderPath).wait();
+			this.prepareXcconfigFile(pluginPlatformsFolderPath).wait();
 		}).future<void>()();
 	}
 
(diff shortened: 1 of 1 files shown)


lib/services/ios-project-service.ts
```diff
@@ -575,6 +604,10 @@ export class IOSProjectService extends projectServiceBaseLib.PlatformProjectServ
 		let modulemap = `module ${libraryName} { explicit module ${libraryName} { ${headers.join(" ")} } }`;
 		this.$fs.writeFile(path.join(headersFolderPath, "module.modulemap"), modulemap).wait();
 	}
+
+	private buildXcconfigContent(xcconfigFilePath: string, xcconfigFileContent: string): string {
+		return `${os.EOL}// Begin xcconfig - ${xcconfigFilePath} ${os.EOL} ${xcconfigFileContent} ${os.EOL} // End xcconfig ${os.EOL}`;
```
jasssonpet: I'll change this to #include directive instead of copy-paste.

jasssonpet:
ping @Fatme 

I think that this PR is ready to be merged.

lib/services/ios-project-service.ts
```diff
@@ -509,6 +515,16 @@ export class IOSProjectService extends projectServiceBaseLib.PlatformProjectServ
 		}).future<void>()();
 	}
 
+	private prepareXcconfigFile(pluginPlatformsFolderPath: string): IFuture<void> {
+		return (() => {
+			let pluginXcconfigFilePath = path.join(pluginPlatformsFolderPath, "build.xcconfig");
+			if(this.$fs.exists(pluginXcconfigFilePath).wait()) {
+				let contentToWrite = this.buildXcconfigContent(pluginXcconfigFilePath);
+				this.$fs.appendFile(this.projectXcconfigFilePath, contentToWrite).wait();
```
Fatme: Can we import the file instead append?

jasssonpet: We append an `#include "pluginFileXcconfig"`, is that what you mean?

teobugslayer:
:+1:

PR #932, opened 2015-09-17
Remove Inspector WebSocket

KristinaKoeva:
Inspector websocket crashes when the message is too big so use a tcp socket instead

The corresponding PR in ios-runtime is https://github.com/NativeScript/ios-runtime/pull/317

### DIFF
--- a/lib/services/ios-debug-service.ts
+++ b/lib/services/ios-debug-service.ts
@@ -119,13 +119,13 @@ class IOSDebugService implements IDebugService {
             let emulatorPackage = this.$platformService.getLatestApplicationPackageForEmulator(platformData).wait();
 
             this.$iOSEmulatorServices.startEmulator(emulatorPackage.packageName, { args: "--nativescript-debug-brk" }).wait();
-            this.wireDebuggerClinet(() => net.connect(InspectorBackendPort)).wait();
+            this.wireDebuggerClient(() => net.connect(InspectorBackendPort)).wait();
         }).future<void>()();
     }
 
     private emulatorStart(): IFuture<void> {
         return (() => {
-            this.wireDebuggerClinet(() => net.connect(InspectorBackendPort)).wait();
+            this.wireDebuggerClient(() => net.connect(InspectorBackendPort)).wait();
 
             let projectId = this.$projectData.projectId;
             let attachRequestMessage = notification.attachRequest(projectId);
@@ -160,7 +160,7 @@ class IOSDebugService implements IDebugService {
                     this.$errors.failWithoutHelp("Timeout waiting for NativeScript debugger.");
                 }
 
-                this.wireDebuggerClinet(() => iosDevice.connectToPort(InspectorBackendPort)).wait();
+                this.wireDebuggerClient(() => iosDevice.connectToPort(InspectorBackendPort)).wait();
                 deploy.wait();
             }).future<void>()()).wait();
         }).future<void>()();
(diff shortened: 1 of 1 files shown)


lib/services/ios-debug-service.ts
```diff
@@ -277,7 +291,41 @@ class IOSDebugService implements IDebugService {
...
 
-function createWebSocketProxy($logger: ILogger, socketFactory: (handler: (_socket: net.Socket) => void) => void): ws.Server {
+function createTcpSocketProxy(socketFileLocation: string, $logger: ILogger, socketFactory: (handler: (socket: net.Socket) => void) => void): net.Server {
+    $logger.info("\nSetting up debugger proxy...\nPress Ctrl + C to terminate, or disconnect.\n");
+    
+    var server = net.createServer({
+        allowHalfOpen: true
+    });
+    
+    server.listen(socketFileLocation);
```
teobugslayer: I think this code has a potential for race condition. The server starts listening before the callback for on connection i set. Should we move this call after the server.on calls?

KristinaKoeva: Will move it :)

lib/services/ios-debug-service.ts
```diff
@@ -202,24 +201,34 @@ class IOSDebugService implements IDebugService {
...
                     case readyForAttach:
-                        this.readyForAttachAction(iosDevice).wait();
+                        this.wireDebuggerClinet(() => iosDevice.connectToPort(InspectorBackendPort)).wait();
                         break;
                 }
             }).future<void>()()).wait();
         }).future<void>()();
     }
+    
+    private wireDebuggerClinet(factory: () => net.Socket): IFuture<void> {
```
ligaz: Should be `Client`.

lib/services/ios-debug-service.ts
```diff
@@ -227,12 +236,9 @@ class IOSDebugService implements IDebugService {
...
 
-    private openDebuggingClient(): IFuture<void> {
+    private openDebuggingClient(fileDescriptor: string): IFuture<void> {
         return (() => {
-            this.$projectDataService.initialize(this.$projectData.projectDir);
-            let platformData = this.$platformsData.getPlatformData(this.platform);
-            let frameworkVersion = this.$projectDataService.getValue(platformData.frameworkPackageName).wait().version;
-
+            let frameworkVersion = this.getProjectFrameworkVersion().wait();            
             let inspectorPath = this.getInspectorPath(frameworkVersion).wait();
```
ligaz: This can be moved in the `else` body because it is not used in the legacy scenario when we open Safari.

lib/services/ios-debug-service.ts
```diff
@@ -202,24 +201,34 @@ class IOSDebugService implements IDebugService {
...
         }).future<void>()();
     }
+    
+    private wireDebuggerClinet(factory: () => net.Socket): IFuture<void> {
+        return (() => {
+            let frameworkVersion = this.getProjectFrameworkVersion().wait();
+            let socketFileLocation = "";
+            if (semver.gte(frameworkVersion, "1.4.0")) {
+                socketFileLocation = temp.path({ suffix: ".sock" });
+                createTcpSocketProxy(socketFileLocation, this.$logger, (callback) => connectEventually(factory, callback));
```
ligaz: Isn't it better if we return the socket file descriptor from this method instead of passing it as first parameter?

lib/services/ios-debug-service.ts
KristinaKoeva: The ```inspectorPath``` is used on the next line where we create the location of the Main.html file
``` let inspectorSourceLocation = path.join(inspectorPath, "Safari/Main.html");```
If you prefer I could rename it to ```inspectorDirectoryPath``` or something similar so that it could be more descriptive.

KristinaKoeva:
Your comments are addressed 
ping @teobugslayer @ligaz

rosen-vladimirov:
:+1:

lib/services/ios-debug-service.ts
ligaz: Sorry, my mistake here. I need to bring my :eyeglasses:

ligaz:
:+1:

rosen-vladimirov:
I'm merging this as it has been approved and the respective change in ios runtime is already merged

Issue #936, opened 2015-09-17
javac version is not getting properly extracted in case of open jdk systems


anshulguleria:
`tns --version` = 1.3.0

In case of openjdk java `javac -version` command has different output.

```
Picked up JAVA_TOOL_OPTIONS: -javaagent:/usr/share/java/jayatanaag.jar 
javac 1.8.0_45-internal
```

thus `/^javac (.*)/i` regex won't work.
And it was causing error in `tns doctor` command.

Log for `tns doctor --log trace`:
```
Exec uname -a 
 stdout: Linux developer 3.19.0-28-generic #30-Ubuntu SMP Mon Aug 31 15:52:51 UTC 2015 x86_64 x86_64 x86_64 GNU/Linux
 
 stderr: 
Exec npm -v 
 stdout: 2.11.3
 
 stderr: 
spawn: java -version
Exec ant -version 
 stdout: Apache Ant(TM) version 1.9.4 compiled on January 27 2015
 
 stderr: Picked up JAVA_TOOL_OPTIONS: -javaagent:/usr/share/java/jayatanaag.jar 

Exec node-gyp -v 
 stdout:  
 stderr: /bin/sh: 1: node-gyp: not found

Exec adb version 
 stdout: Android Debug Bridge version 1.0.31
 
 stderr: 
Exec android -h 
 stdout: 
       Usage:
       android [global options] action [action options]
       Global options:
  -s --silent     : Silent mode, shows errors only.
  -v --verbose    : Verbose mode, shows errors, warnings and all messages.
     --clear-cache: Clear the SDK Manager repository manifest cache.
  -h --help       : Help on a specific command.

                                                                    Valid
                                                                    actions
                                                                    are
                                                                    composed
                                                                    of a verb
                                                                    and an
                                                                    optional
                                                                    direct
                                                                    object:
-    sdk              : Displays the SDK Manager window.
-    avd              : Displays the AVD Manager window.
-   list              : Lists existing targets or virtual devices.
-   list avd          : Lists existing Android Virtual Devices.
-   list target       : Lists existing targets.
-   list device       : Lists existing devices.
-   list sdk          : Lists remote SDK repository.
- create avd          : Creates a new Android Virtual Device.
-   move avd          : Moves or renames an Android Virtual Device.
- delete avd          : Deletes an Android Virtual Device.
- update avd          : Updates an Android Virtual Device to match the folders
                        of a new SDK.
- create project      : Creates a new Android project.
[...]

PR #940, opened 2015-09-17
Sync with common to get correct java detection code

teobugslayer:
See https://github.com/NativeScript/nativescript-cli/issues/936


rosen-vladimirov:
:+1:

PR #951, opened 2015-09-23
Add internal folder to ios framework directories

jasssonpet:
Related to https://github.com/NativeScript/ios-runtime/pull/325 and https://github.com/NativeScript/ios-runtime/issues/308.


rosen-vladimirov:
:+1:

PR #954, opened 2015-09-23
Fix lint errors

rosen-vladimirov:
Remove trailing spaces and rename variable in order to prevent
```>> lib/services/ios-debug-service.ts[343, 28]: shadowed variable: 'socket'```


Mitko-Kerezov:
:+1:

Issue #961, opened 2015-09-24
When cli is installed localy for a nativescript project build crashes


Plamen5kov:
When there is a nativescript project and cli is installed in the node modules folder rather than globally, the gradle build crashes with "UNEXPECTED TOP-LEVEL EXCEPTION:" due to the fact cli caries obsolete, unnecessary jars in the resource/tools folder.

PR #962, opened 2015-09-24
fix for issue #961

Plamen5kov:
@teobugslayer


ligaz:
:+1: Finally we are removing those :smiley:

Issue #941, opened 2015-09-17
tns doctor command reports Ant is not installed


vchimev:
When Gradle not installed, the `tns doctor` command reports that both Ant and Gradle are not installed.

PR #965, opened 2015-09-25
Remove check for ant version

rosen-vladimirov:
We do not use ant anymore, so we should remove the check.
Fixes https://github.com/NativeScript/nativescript-cli/issues/941


teobugslayer:
:+1:

PR #974, opened 2015-09-28
Depend on node.js 0.12

teobugslayer:
Due to c++ dependencies, we do not always work with the latest major version of node.js

Therefore, we do not use the automatic dependency resolution of homebrew.
Instead, manually install well-known version of node.js distribution from the external tap 'versions'


dtopuzov:
:+1:

rosen-vladimirov:
:+1:

PR #975, opened 2015-09-28
Do not install Android SDK 17

teobugslayer:
Since NativeScript 1.3, we do not hard-code android-17. Stop installing it as dependency


dtopuzov:
:+1:

Fatme:
:+1:

Issue #882, opened 2015-09-07
Plugin variables


Fatme:
## myPlugin's package.json: 
```json
"nativescript": {
    "platforms": {
      "android": "1.1.3",
      "ios": "1.2.3"
    },
    "variables": {
        "APP_ID": {
             "defaultValue": "myAppId"
        },
        "APP_NAME": {
        }
    }
  }
```

# Plugin add
{N} CLI will store all plugin variables and their values into project's package.json file:

```json
"nativescript": {
    "id": "org.nativescript.androidTest3",
        "tns-android": {
	    "version": "1.3.0"
	},
        "myPlugin-variables": {
            "APP_ID": "myAppId",
            "APP_NAME": "myAppName"
        }
}
```

## Valid plugin commands:

* `tns plugin add myPlugin --var.APP_ID myNewAppId  --var.APP_NAME myAppName
* `tns plugin add myPlugin --var.APP_NAME myAppName`

## Invalid plugin commands:
* `tns plugin add myPlugin`
* `tns plugin add myPlugin --var.APP_ID myNewAppId`

## Prepare
On prepare  {N} CLI will replace all references of plugin variable with the specified value by plugin add command. 
It seems that it will be enough to respect only `AndroidManifest.xml` and `Info.plist` file. But maybe in some cases {N} CLI should respect .java files as well.

PanayotCankov:
We have a plugin that needs the app's id.

Fatme:
@PanayotCankov I've logged your case as separate issue https://github.com/NativeScript/nativescript-cli/issues/1026 for next release.

hdeshev:
This feature has so much potential :grin: I hope I'm not too late to the party!

What if:

1. We could either replace variable occurrences in the source code, or provide an API to get variable values at run time? This could be used by JS/TS code to read usernames, keys, all sorts of configuration data and modify behavior according to some flag. That makes [feature flags](https://code.flickr.net/2009/12/02/flipping-out/) easy too.
2. Why store the variables in the app package.json file? This limits us to having a single "environment". What if we design this similarly to [Rails environments](http://teotti.com/use-of-rails-environments/) and store variables in different files, possibly named according to some convention. This way we could allow for easy switching between a dev/qa build of an app and a real production one. Users could use that to hit different API endpoints, enable instrumentation, detect when running in a simulator. With some build time tricks, you can even use this to ship a different version of an app to different customers and implement a multi-tenant solution.

PR #981, opened 2015-09-29
Plugin variables

Fatme:
https://github.com/NativeScript/nativescript-cli/issues/882

### DIFF
--- a/lib/definitions/plugins.d.ts
+++ b/lib/definitions/plugins.d.ts
@@ -9,7 +9,9 @@ interface IPluginsService {
 
 interface IPluginData extends INodeModuleData {
 	platformsData: IPluginPlatformsData;
-	pluginPlatformsFolderPath(platform: string): string; 
+	/* Gets all plugin variables from plugin */
+	pluginVariables: IDictionary<IPluginVariableData>;
+	pluginPlatformsFolderPath(platform: string): string;
 }
 
 interface INodeModuleData {
@@ -23,4 +25,38 @@ interface INodeModuleData {
 interface IPluginPlatformsData {
 	ios: string;
 	android: string;
+}
+
+interface IPluginVariablesService {
+	/**
+	 * Saves plugin variables in project package.json file.
+	 * @param  {IPluginData}		pluginData for the plugin.
+	 * @return {IFuture<void>}
+	 */
+	savePluginVariablesInProjectFile(pluginData: IPluginData): IFuture<void>;
+	/**
+	 * Removes plugin variables from project package.json file.
+	 * @param  {IPluginData}		pluginData for the plugin.
+	 * @return {IFuture<void>}
+	 */
+	removePluginVariablesFromProjectFile(pluginData: IPluginData): IFuture<void>;
+	/**
+	 * Replaces all plugin variables with their corresponding values.
+	 * @param {IPluginData}		pluginData for the plugin.
+	 * @param {pluginConfigurationFileContent}		pluginConfigurationFileContent for the plugin.
+	 * @return {IFuture<string>}		returns the changed plugin configuration file content.
+	 */
+	interpolatePluginVariables(pluginData: IPluginData, pluginConfigurationFileContent: string): IFuture<string>;
+	/**
+	 * Returns the
+	 * @param {IPluginData}		pluginData for the plugin.
+	 * @return {IFuture<string>}		returns the changed plugin configuration file content.
+	 */
+	getPluginVariablePropertyName(pluginData: IPluginData): string;
+}
+
+interface IPluginVariableData {
+	defaultValue?: string;
+	name?: string;
+	value?: string;
 }
--- a/lib/services/plugin-variables-service.ts
+++ b/lib/services/plugin-variables-service.ts
@@ -1,0 +1,103 @@
+///<reference path="../.d.ts"/>
+"use strict";
+
+import * as helpers from "./../common/helpers";
+
+export class PluginVariablesService implements IPluginVariablesService {
+	private static PLUGIN_VARIABLES_KEY = "variables";
+
+	constructor(private $errors: IErrors,
+		private $pluginVariablesHelper: IPluginVariablesHelper,
+		private $projectData: IProjectData,
+		private $projectDataService: IProjectDataService,
+		private $prompter: IPrompter) { }
+
+	public getPluginVariablePropertyName(pluginData: IPluginData): string {
+		return `${pluginData.name}-${PluginVariablesService.PLUGIN_VARIABLES_KEY}`;
+	}
+
+	public savePluginVariablesInProjectFile(pluginData: IPluginData): IFuture<void> {
+		return (() => {
+			let values = Object.create(null);
+			this.executeForAllPluginVariables(pluginData, (pluginVariableData: IPluginVariableData) =>
+				(() => {
+					let pluginVariableValue = this.getPluginVariableValue(pluginVariableData).wait();
+					this.ensurePluginVariableValue(pluginVariableValue, `Unable to find value for ${pluginVariableData.name} plugin variable from ${pluginData.name} plugin. Ensure the --var option is specified or the plugin variable has default value.`);
+					values[pluginVariableData.name] = pluginVariableValue;
+				}).future<void>()()).wait();
+
+			this.$projectDataService.initialize(this.$projectData.projectDir);
+			this.$projectDataService.setValue(this.getPluginVariablePropertyName(pluginData), values).wait();
+		}).future<void>()();
+	}
+
+	public removePluginVariablesFromProjectFile(pluginData: IPluginData): IFuture<void> {
+		this.$projectDataService.initialize(this.$projectData.projectDir);
+		return this.$projectDataService.removeProperty(this.getPluginVariablePropertyName(pluginData));
+	}
+
+	public interpolatePluginVariables(pluginData: IPluginData, pluginConfigurationFileContent: string): IFuture<string> {
+		return (() => {
+			this.executeForAllPluginVariables(pluginData, (pluginVariableData: IPluginVariableData) =>
+				(() => {
+					this.ensurePluginVariableValue(pluginVariableData.value, `Unable to find the value for ${pluginVariableData.name} plugin variable into project package.json file. Verify that your package.json file is correct and try again.`);
+					pluginConfigurationFileContent = pluginConfigurationFileContent.replace(new RegExp(`{${pluginVariableData.name}}`, "gi"), pluginVariableData.value);
+				}).future<void>()()).wait();
+			return pluginConfigurationFileContent;
+		}).future<string>()();
+	}
+
+	private ensurePluginVariableValue(pluginVariableValue: string, errorMessage: string): void {
+		if(!pluginVariableValue) {
+			this.$errors.failWithoutHelp(errorMessage);
+		}
+	}
+
+	private getPluginVariableValue(pluginVariableData: IPluginVariableData): IFuture<string> {
+		return (() => {
+			let pluginVariableName = pluginVariableData.name;
+			let value = this.$pluginVariablesHelper.getPluginVariableFromVarOption(pluginVariableName);
+			if(value) {
+				value = value[pluginVariableName];
+			} else {
+				value = pluginVariableData.defaultValue;
+				if(!value && helpers.isInteractive() ) {
+					let promptSchema = {
+						name: pluginVariableName,
+						type: "input",
+						message: `Enter value for ${pluginVariableName} variable:`,
+						validate: (val: string) => !!val ? true : 'Please enter a value!'
+					};
+					let promptData = this.$prompter.get([promptSchema]).wait();
+					value = promptData[pluginVariableName];
+				}
+			}
+
+			return value;
+		}).future<string>()();
+	}
+
+	private executeForAllPluginVariables(pluginData: IPluginData, action: (pluginVariableData: IPluginVariableData) => IFuture<void>): IFuture<void> {
+		return (() => {
+			let pluginVariables = pluginData.pluginVariables;
+			let pluginVariablesNames = _.keys(pluginVariables);
+			_.each(pluginVariablesNames, pluginVariableName => action(this.createPluginVariableData(pluginData, pluginVariableName).wait()).wait());
+		}).future<void>()();
+	}
+
+	private createPluginVariableData(pluginData: IPluginData, pluginVariableName: string): IFuture<IPluginVariableData> {
+		return (() => {
+			let variableData = pluginData.pluginVariables[pluginVariableName];
+
+			variableData.name = pluginVariableName;
+
+			this.$projectDataService.initialize(this.$projectData.projectDir);
+			let pluginVariableValues = this.$projectDataService.getValue(this.getPluginVariablePropertyName(pluginData)).wait();
+			variableData.value = pluginVariableValues ? pluginVariableValues[pluginVariableName] : undefined;
+
+			return variableData;
+		}).future<IPluginVariableData>()();
+	}
+}
+$injector.register("pluginVariablesService", PluginVariablesService);
+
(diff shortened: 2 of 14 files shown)


lib/definitions/plugins.d.ts
```diff
@@ -23,4 +24,16 @@ interface INodeModuleData {
...
+}
+
+interface IPluginVariablesService {
+	savePluginVariablesInProjectFile(pluginData: IPluginData): IFuture<void>;
+	removePluginVariablesFromProjectFile(pluginData: IPluginData): IFuture<void>;
+	interpolatePluginVariables(pluginData: IPluginData, pluginConfigurationFileContent: string): IFuture<string>;
+}
+
+interface IPluginVariableData {
+	default?: string;
```
ligaz: Should this be called `defaultValue`?

.vscode/launch.json
```diff
@@ -0,0 +1,42 @@
...
+			// Name of configuration; appears in the launch configuration drop down menu.
+			"name": "Launch ./lib/nativescript-cli.js",
+			// Type of configuration. Possible values: "node", "mono".
+			"type": "node",
+			// Workspace relative or absolute path to the program.
+			"program": "./lib/nativescript-cli.js",
+			// Automatically stop program after launch.
+			"stopOnEntry": false,
+			// Command line arguments passed to the program.
+			"args": ["plugin", "add", "/Users/havaluova/Work/plugin-vars", "--path", "/Users/havaluova/Work/nativescript-cli/scratch/examplePlVars"],
```
ligaz: Maybe you should remove this configuration :smile:

lib/services/project-data-service.ts
```diff
@@ -48,7 +49,8 @@ export class ProjectDataService implements IProjectDataService {
 
 	private loadProjectFile(): IFuture<void> {
 		return (() => {
-			assert.ok(this.projectFilePath, "Initialize method of projectDataService is not called");
+			//this.initialize(this.$injector.resolve("projectData").projectDir);
```
ligaz: Do we need this commented line?

lib/definitions/plugins.d.ts
```diff
@@ -9,7 +9,8 @@ interface IPluginsService {
 
 interface IPluginData extends INodeModuleData {
 	platformsData: IPluginPlatformsData;
-	pluginPlatformsFolderPath(platform: string): string; 
+	pluginVariables: IDictionary<IPluginVariableData>;
```
teobugslayer: Can we document these members with jsdoc?

lib/definitions/plugins.d.ts
```diff
@@ -23,4 +24,16 @@ interface INodeModuleData {
 interface IPluginPlatformsData {
 	ios: string;
 	android: string;
+}
+
+interface IPluginVariablesService {
+	savePluginVariablesInProjectFile(pluginData: IPluginData): IFuture<void>;
```
teobugslayer: Ditto, consider documenting this interface

lib/services/plugin-variables-service.ts
```diff
@@ -0,0 +1,103 @@
+///<reference path="../.d.ts"/>
+"use strict";
+
+import helpers = require("./../common/helpers");
```
teobugslayer: Consider ES6 syntax here

lib/tools/broccoli/builder.ts
```diff
@@ -10,78 +11,86 @@ let vinylFilterSince = require("vinyl-filter-since");
...
+                        nodeModules[rootModuleFullPath] = rootModuleFullPath;
+                    }
+
+                    cb(null);
+                }))
+                .pipe(gulp.dest(absoluteOutputPath));
+
+                let future = new Future<void>();
+
+                pipeline.on('end', (err: any, data: any) => {
```
teobugslayer: Can err be typed as Error?

lib/services/plugin-variables-service.ts
```diff
@@ -0,0 +1,103 @@
+///<reference path="../.d.ts"/>
+"use strict";
+
+import * as helpers from "./../common/helpers";
+
+export class PluginVariablesService implements IPluginVariablesService {
+	private static PLUGIN_VARIABLES_KEY = "variables";
+
+	constructor(private $errors: IErrors,
+		private $fs: IFileSystem,
```
Mitko-Kerezov: Unused dependency?

lib/services/plugin-variables-service.ts
```diff
@@ -0,0 +1,103 @@
...
+		private $pluginVariablesHelper: IPluginVariablesHelper,
+		private $projectData: IProjectData,
+		private $projectDataService: IProjectDataService,
+		private $prompter: IPrompter,
+		private $staticConfig: IStaticConfig) { }
```
Mitko-Kerezov: Another unused dependency I believe

lib/services/plugin-variables-service.ts
```diff
@@ -0,0 +1,103 @@
...
+			let value = this.$pluginVariablesHelper.getPluginVariableFromVarOption(pluginVariableName);
+			if(value) {
+				value = value[pluginVariableName];
+			} else {
+				value = pluginVariableData.defaultValue;
+				if(!value && helpers.isInteractive()) {
+					let promptSchema = {
+						name: pluginVariableName,
+						type: "input",
+						message: `Enter value for ${pluginVariableName} variable: `
```
Mitko-Kerezov: `$prompter` adds a white space at the end of every prompt so the white space at the end here seems redundant

Mitko-Kerezov: As a side note, if the user presses sets a false-like value for a plugin variable we will fail in `savePluginVariablesInProjectFile` function up above and the process will end which is no great UX.
Instead we can use `$prompter`'s  `validate` property like so 
```Typescript
validate: (value: string) => !!value ? true : 'Please enter a value!'
```

lib/services/project-data-service.ts
```diff
@@ -11,7 +11,8 @@ export class ProjectDataService implements IProjectDataService {
 	constructor(private $fs: IFileSystem,
 		private $staticConfig: IStaticConfig,
 		private $errors: IErrors,
-		private $logger: ILogger) {
+		private $logger: ILogger,
+		private $injector: IInjector) {
```
Mitko-Kerezov: Unnecessary dependency

lib/services/plugins-service.ts
```diff
@@ -24,15 +25,21 @@ export class PluginsService implements IPluginsService {
...
 
 	public add(plugin: string): IFuture<void> {
 		return (() => {
 			this.ensure().wait();
 			let dependencyData = this.$npm.cache(plugin, undefined, PluginsService.NPM_CONFIG).wait();
 			if(dependencyData.nativescript) {
+				let pluginData = this.convertToPluginData(dependencyData);
+				this.validate(pluginData).wait();
+
+				this.$pluginVariablesService.savePluginVariablesInProjectFile(pluginData).wait();
```
Mitko-Kerezov: In case this line of code or the one below it fails the project will be left in an unstable state. In order to avoid this a `try ... catch` statement with a logic for reversing changes inside the `catch` statement can be introduced

lib/services/plugins-service.ts
```diff
@@ -43,14 +50,62 @@ export class PluginsService implements IPluginsService {
...
+									if(index === 0) {
+										this.$projectDataService.initialize(this.$projectData.projectDir);
+										let frameworkVersion = this.$projectDataService.getValue(platformData.frameworkPackageName).wait().version;
+										this.$npm.cache(platformData.frameworkPackageName, frameworkVersion).wait();
+
+										let relativeConfigurationFilePath = path.relative(platformData.projectRoot, platformData.configurationFilePath);
+										let cachedPackagePath = this.$injector.resolve("npmInstallationManager").getCachedPackagePath(platformData.frameworkPackageName, frameworkVersion);
+										let cachedConfigurationFilePath = path.join(cachedPackagePath, constants.PROJECT_FRAMEWORK_FOLDER_NAME, relativeConfigurationFilePath);
+
+										shelljs.rm("-rf", platformData.configurationFilePath);
```
Mitko-Kerezov: Removing the old configuration file seems redundant. If it is not removed it'll be replaced by the force `shell.cp` below

Mitko-Kerezov:
:+1: After rebasing, getting a green build, merging pull request in common, updating SHA and squashing
:smiley_cat:

lib/services/plugin-variables-service.ts
```diff
@@ -0,0 +1,102 @@
...
+		return `${pluginData.name}-${PluginVariablesService.PLUGIN_VARIABLES_KEY}`;
+	}
+
+	public savePluginVariablesInProjectFile(pluginData: IPluginData): IFuture<void> {
+		return (() => {
+			let values = Object.create(null);
+			this.executeForAllPluginVariables(pluginData, (pluginVariableData: IPluginVariableData) =>
+				(() => {
+					let pluginVariableValue = this.getPluginVariableValue(pluginVariableData).wait();
+					if(!pluginVariableValue) {
```
teobugslayer: This could be extracted into a dedicated function `ensurePluginvarValue`, the same code is used in other places as well.

lib/services/plugin-variables-service.ts
```diff
@@ -0,0 +1,102 @@
...
+
+	private getPluginVariableValue(pluginVariableData: IPluginVariableData): IFuture<string> {
+		return (() => {
+			let pluginVariableName = pluginVariableData.name;
+			let value = this.$pluginVariablesHelper.getPluginVariableFromVarOption(pluginVariableName);
+			if(value) {
+				value = value[pluginVariableName];
+			} else {
+				value = pluginVariableData.defaultValue;
+				if(!value && helpers.isInteractive()) {
```
teobugslayer: If not interactive, we should give at least a warning.

lib/services/plugin-variables-service.ts
```diff
@@ -0,0 +1,102 @@
...
+		}).future<void>()();
+	}
+
+	private createPluginVariableData(pluginData: IPluginData, pluginVariableName: string): IFuture<IPluginVariableData> {
+		return (() => {
+			let variableData = pluginData.pluginVariables[pluginVariableName];
+
+			variableData.name = pluginVariableName;
+
+			this.$projectDataService.initialize(this.$projectData.projectDir);
```
teobugslayer: Is this code really needed to be executed for every plugin variable?

Fatme: We have caching inside `projectDataService` so the logic will be executed only the first time.

lib/services/plugin-variables-service.ts
Fatme: Why?

lib/services/plugins-service.ts
```diff
@@ -43,14 +59,64 @@ export class PluginsService implements IPluginsService {
...
+								if(this.$fs.exists(this.getPluginConfigurationFilePath(data, platformData)).wait()) {
+									executeUninstallCommand().wait();
+
+									if(index === 0) {
+										this.$projectDataService.initialize(this.$projectData.projectDir);
+										let frameworkVersion = this.$projectDataService.getValue(platformData.frameworkPackageName).wait().version;
+										this.$npm.cache(platformData.frameworkPackageName, frameworkVersion).wait();
+
+										let relativeConfigurationFilePath = path.relative(platformData.projectRoot, platformData.configurationFilePath);
+										let cachedPackagePath = this.$injector.resolve("npmInstallationManager").getCachedPackagePath(platformData.frameworkPackageName, frameworkVersion);
```
teobugslayer: Can we get this from constructor injection?

lib/services/plugins-service.ts
```diff
@@ -43,14 +59,64 @@ export class PluginsService implements IPluginsService {
...
+						let nodeModules = this.$broccoliBuilder.getChangedNodeModules(tnsModulesDestinationPath, platform).wait();
+						let plugins = _.map(_.keys(nodeModules), nodeModule => this.convertToPluginData(this.getNodeModuleData(pluginName).wait()));
+
+						_.forEach(plugins, (data, index) => {
+							if(data.isPlugin) {
+								if(this.$fs.exists(this.getPluginConfigurationFilePath(data, platformData)).wait()) {
+									executeUninstallCommand().wait();
+
+									if(index === 0) {
+										this.$projectDataService.initialize(this.$projectData.projectDir);
```
teobugslayer: I think the entire code  in the body of the if should be moved into a dedicated function - `initializeFromCache` (name could be improved, of course)

lib/services/plugins-service.ts
Fatme: Yes, we can :smile_cat:

lib/tools/broccoli/builder.ts
```diff
@@ -10,78 +11,86 @@ let vinylFilterSince = require("vinyl-filter-since");
...
+    private $projectData: IProjectData,
     private $projectDataService: IProjectDataService,
     private $injector: IInjector,
     private $logger: ILogger) { }
 
-  public prepareNodeModules(absoluteOutputPath: string, projectDir: string, platform: string, lastModifiedTime?: Date): IFuture<void> {
+    public getChangedNodeModules(absoluteOutputPath: string, platform: string, lastModifiedTime?: Date): IFuture<any> {
+        return (() => {
+            let projectDir = this.$projectData.projectDir;
+            let isNodeModulesModified = false;
```
teobugslayer: Is this the correct place for this var? Shouldn't it be local to the pipe callback where it is used?

lib/tools/broccoli/builder.ts
```diff
@@ -10,78 +11,86 @@ let vinylFilterSince = require("vinyl-filter-since");
...
+                    }
+
+                    cb(null);
+                }))
+                .pipe(gulp.dest(absoluteOutputPath));
+
+                let future = new Future<void>();
+
+                pipeline.on('end', (err: Error, data: any) => {
+                if(err) {
```
teobugslayer: This should be indented once on the right

lib/services/plugin-variables-service.ts
teobugslayer: This is implementation detail. The code simply does not belong to the callback. In case we decide to refactor projectDataService  we need to remember to change this code as well.

lib/services/plugin-variables-service.ts
teobugslayer: We talked in person and agreed that the code is correct

lib/services/plugins-service.ts
```diff
@@ -43,14 +59,54 @@ export class PluginsService implements IPluginsService {
...
+					let pluginData = this.convertToPluginData(this.getNodeModuleData(pluginName).wait());
+
+					platformData.platformProjectService.removePluginNativeCode(pluginData).wait();
+
+					// Remove the plugin and call merge for another plugins that have configuration file
+					let pluginConfigurationFilePath = this.getPluginConfigurationFilePath(pluginData, platformData);
+					if(this.$fs.exists(pluginConfigurationFilePath).wait()) {
+						let tnsModulesDestinationPath = path.join(platformData.appDestinationDirectoryPath, constants.APP_FOLDER_NAME, constants.TNS_MODULES_FOLDER_NAME);
+						let nodeModules = this.$broccoliBuilder.getChangedNodeModules(tnsModulesDestinationPath, platform).wait();
+						let plugins = _.map(_.keys(nodeModules), nodeModule => this.convertToPluginData(this.getNodeModuleData(pluginName).wait()));
```
teobugslayer: This is only a suggestion and not a merge stopper:

```
_.(nodeModules)
.map(nodeModule => this.getNodeModuleData(pluginName).wait())
.map(nodeModuleData => this.convertToPluginData(nodeModuleData))
.filter(data => data.isPlugin && this.$fs.exists(this.getPluginConfigurationFilePath(data, platformData)).wait())
.forEach((data, index) => {
	executeUninstallCommand().wait();

	if(index === 0) {
		this.initializeConfigurationFileFromCache(platformData).wait();
	}

	if(data.name !== pluginName) {
		this.merge(data, platformData).wait();
	}
})
.value();
```

I find this code to be more linear but the original is correct and I do not insist on changing it.

Issue #944, opened 2015-09-18
Using CocoaPods should update the app deployment target to iOS 8


jasssonpet:
CocoaPods uses dynamic frameworks at the moment, that are not supported on iOS 7.

PR #984, opened 2015-09-30
Update app deployment target to 8.0

Fatme:
Fixes https://github.com/NativeScript/nativescript-cli/issues/944

### DIFF
--- a/lib/services/ios-project-service.ts
+++ b/lib/services/ios-project-service.ts
@@ -519,6 +519,11 @@ export class IOSProjectService extends projectServiceBaseLib.PlatformProjectServ
 			if(opts && opts.executePodInstall && this.$fs.exists(pluginPodFilePath).wait()) {
 				this.executePodInstall().wait();
 			}
+
+			let project = this.createPbxProj();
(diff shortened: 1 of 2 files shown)


Fatme:
@fealebenpae @jasssonpet

fealebenpae:
I suggest that we add a log statement to indicate we've changed the Deployment Target build setting, like we do for adding dynamic frameworks.

Fatme:
@fealebenpae I added log statement.

fealebenpae:
Looks good to me.

dtopuzov:
:+1:

lib/services/ios-project-service.ts
```diff
@@ -519,6 +519,11 @@ export class IOSProjectService extends projectServiceBaseLib.PlatformProjectServ
 			if(opts && opts.executePodInstall && this.$fs.exists(pluginPodFilePath).wait()) {
 				this.executePodInstall().wait();
 			}
+
+			let project = this.createPbxProj();
```
teobugslayer: You can move this code in a dedicated function

teobugslayer:
:+1: after addressing a small, non-functional remark

PR #987, opened 2015-09-30
Fix check for Gradle version

rosen-vladimirov:
In case Gradle's version is with three numbers ("2.2.3" for example), our
current check is failing and makes the doctor command unusable. Compare
current gradle version with min required one only after they are in the
same format.

Fixes https://github.com/NativeScript/nativescript-cli/issues/989


Fatme:
:+1:

ligaz:
:+1:

Issue #943, opened 2015-09-18
tns now seems to be caching things where it shouldn't


sitefinitysteve:
I've been struggling all night with my 1.3 app (was fine on 1.2)

Seems that when I change things like adding\removing a style.xml file in App_Resources the only way to get NS to update is to platform remove\add android back in.

Worse was upgrading from 1.2 to 1.3 the entire app, all my custom scripts and views were all cached and loading old versions.  I could change style colors in styles.xml, they get reflected, but any of my views or scripts all ran\debugged as if I changed nothing.  However if I formed invalid xml, well then that would crash the thing, but any other changes wouldn't show up.  tns debug even showed me I was CLEARLY on the old scripts as well.

**CONFUSED**

sitefinitysteve:
Example 
![image](https://cloud.githubusercontent.com/assets/1542376/9951043/be46aa22-5d8f-11e5-8ba4-2a5f36fd52f1.png)

Added the debugger line because there's clearly a bug in there
tns debug android starts up the debugger, but as you can see the debug script does not have any of my changes.

rosen-vladimirov:
Hi @sitefinitysteve, 
May I ask you for the versions of the ios and android runtimes that you are using? In version 1.2.x of CLI and runtimes there are several issues with application's reloading after livesync. If you have used the livesync functionality and you have not updated the tns-ios and tns-android versions, you'll see the last livesynced application. 
Can you try using latest android and ios versions and see if it works for you?

sitefinitysteve:
I'm on {N} 1.3, I have done tns remove\add android multiple times to try to fix it.

So I encountered this this morning as well, and this is what I noticed
1) Livesync always has the latest code (ctrl-c to cancel)
2) Added a debugger line in code
3) tns debug android
4) App fired back up, no debugger line in the code
5) Started livesync BACK up, wait for app to load, kill livesync
6) tns debug android, debugger line is now there and breaking fine

Now I also noticed that my segmentedbar selectedIndexChanged event is also not firing (one of the reasons I needed to debug)

tns doctor reports
C:\Dropbox\Projects\Leap\OutbreakHelp\Mobile\Native\OutbreakHelp>tns doctor
WARNING: You can work with iOS only on Mac OS X systems.
To be able to work with iOS devices and projects, you need Mac OS X Mavericks or later.

Version strings are not in the same format
# doctor
┌─────────┬──────────────┐
│ Usage   │ Synopsis     │
[...]

PR #990, opened 2015-09-30
Fix run/debug after livesync

Fatme:
Due to some policy restrictions on devices with apilevel 21+, android runtime is not able to delete sync directories. This way if some files are changed after livesync, changes are not applied on the device.

Fixes https://github.com/NativeScript/nativescript-cli/issues/943

### DIFF
--- a/lib/services/android-project-service.ts
+++ b/lib/services/android-project-service.ts
@@ -28,7 +28,8 @@ class AndroidProjectService extends projectServiceBaseLib.PlatformProjectService
 		private $projectData: IProjectData,
 		private $projectDataService: IProjectDataService,
 		private $propertiesParser: IPropertiesParser,
-		private $sysInfo: ISysInfo) {
+		private $sysInfo: ISysInfo,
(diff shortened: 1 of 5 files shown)


lib/services/android-project-service.ts
```diff
@@ -285,6 +286,15 @@ class AndroidProjectService extends projectServiceBaseLib.PlatformProjectService
 		return Future.fromResult();
 	}
 
+	public deploy(device: Mobile.IAndroidDevice, appIdentifier: string): IFuture<void> {
+		return (() => {
+			let deviceRootPath = `/data/local/tmp/${appIdentifier}`;
+			device.adb.executeShellCommand(["rm", "-rf", this.$mobileHelper.buildDevicePath(deviceRootPath, "fullsync")]).wait();
```
rosen-vladimirov: can we execute `rm -rf <dir1> <dir2> <dir3>`

rosen-vladimirov:
:+1:

Issue #942, opened 2015-09-18
Updating App_Resources styles.xml kills livesync


sitefinitysteve:
If I have livesync enabled and change any of the xml styles it crashes with

ENOENT, stat 'c:\Dropbox\Projects\Leap\OutbreakHelp\Mobile\Native\OutbreakHelp\platforms\android\src\main\assets\app\App_Resources\Android\values-v21\colors.xml'

rosen-vladimirov:
Thanks for reporting this @sitefinitysteve 
It looks like changing anything inside `app/AppResources/<platformName>` is causing such ENOENT errors as we are searching for the files from AppResources on incorrect place.
Steps to reproduce:
1) Create new project (`tns create app1`)
2) Add platform, for example Android (`tns platform add android`)
3) Add testFile.xml inside app/AppResources/Android/
4) Deploy your application on the device (`tns run android`)
5) Start livesync: `tns livesync android --watch`
6) Modify testFile.xml created in step 3.

PR #994, opened 2015-10-01
Sync files that are in App_Resources

Fatme:
Fixes https://github.com/NativeScript/nativescript-cli/issues/942


teobugslayer:
:+1:

Issue #988, opened 2015-09-30
OSX Setup script issues


nmladenov:
1. Very hard Xcode install. Which version with which is compatible (This is Apple's fault, but may be it is good to require Xcode to be installed prior running the script)
2. Asking for passwords and continues when password is wrong, may be script must stop when some of the steps is failing
3. No progress indicator, not clear what is happening
4. Despite Xcode user agreement is accepted, script is saying it is not, and offering to accept it again.
- You have not agreed to the Xcode license agreements, please run 'xcodebuild -license' (for user-level acceptance) or 'sudo xcodebuild -license' (for system-wide acceptance) from within a Terminal window to review and agree to the Xcode license agreements.

5. Homebrew install is hanging 
- Installing Homebrew... (You might need to provide your password.) (Left it for 30 minutes, and nothing happened) ( I run the installation separately, and it is asking to press Return to continue, but this is not showing up during installation. Separately installation passed - *****ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"****   )

6.Java install hang: ==> Downloading http://download.oracle.com/otn-pub/java/jdk/8u60-b27/jdk-8u60-macosx-x64.dmg
######################################################################## 100.0%
==> Running installer for java; your password may be necessary.
==> Package installers may write to any location; options such as --appdir are ignored.

P.S. extremely slow - around 10 minutes and more .... succeeded at the end.

7. ==> Installing node012 from homebrew/homebrew-versions
==> Downloading https://homebrew.bintray.com/bottles-versions/node012-0.12.7.yosemite.bottle.ta
######################################################################## 100.0%
==> Pouring node012-0.12.7.yosemite.bottle.tar.gz
==> Caveats
Bash completion has been installed to:
  /usr/local/etc/bash_completion.d
==> Summary
🍺  /usr/local/Cellar/node012/0.12.7: 2698 files, 30M
Creating Homebrew formula for NativeScript.
-e:35:in `block in <main>': uninitialized constant DATA (NameError)
	from -e:34:in `open'
	from -e:34:in `<main>'

vchimev:
Later updates here: https://github.com/NativeScript/NativeScript/issues/1556.

From the list above:
- [x] 1. asked if Xcode is installed prior executing the script.
- [ ] 2. continues when password is wrong is still an issue

enchev:
Hey @vchimev, 

Can we verify if this is still valid?

[...]

PR #995, opened 2015-10-01
Fix mac setup

teobugslayer:
Make Xcode installation entirely user driven.
Introduce simple y/n dialogs with the user.
Check that Xcode and Homebrew have been successfully installed.
Print output from called programs immediately.
Add warning about long JDK install

See https://github.com/NativeScript/nativescript-cli/issues/988


ikoevska:
:+1:  for the messaging

rosen-vladimirov:
:+1:

Issue #1009, opened 2015-10-05
Bash autocompletion fails with sudo


tjvantoll:
This is an odd one @sfeather found. If you run `sudo npm install -g nativescript`, enabling command-line completion does not work—specifically I get this error:

```
If you are using bash or zsh, you can enable command-line completion.
? Do you want to enable it now? Yes
Unable to update /Users/tj/.bash_profile. Command-line completion might not work.
Unable to update /Users/tj/.bashrc. Command-line completion might not work.
Unable to update /Users/tj/.zshrc. Command-line completion might not work.
```

What's weird is that `npm install -g nativescript` (without using `sudo`) can update those files without any issues—the problem only occurs with `sudo`. (On a side note `tns autocomplete enable` works as expected.)

rosen-vladimirov:
Hi @tjvantoll ,
The issue is caused by `npm` specific behavior when executing postinstall scripts. When you use `sudo npm install -g nativescript`, the postinstall script is executed with specific user called `nobody`. This user has limited permissions and cannot update user profiles. We have check in our code for this situation and in case it happens, we should print message `To enable command-line completion, run '$ tns autocomplete enable'`. In your case this message is not shown as on Mac the error code is EACCES, while in our code we were expecting EPERM. I've fixed the check, so the correct message will be shown to the user from now on. 
Unfortunately we cannot do anything in order to fix this behavior. The only workaround that I could think about is to use:
```
sudo npm install -g nativescript --unsafe-perm
```
This way the scripts will be executed with root user and the user will be able to enable autocompletion.

Hope this helps. Feel free to contact us in case you have any problems or concerns!

tjvantoll:
Thanks @rosen-vladimirov. My biggest concern is that we gave an error message that said something failed without explaining why. The new message sounds great.

PR #1024, opened 2015-10-07
Show message when sudo cannot update profiles

rosen-vladimirov:
When user installs nativescript-cli with sudo (`sudo npm i -g nativescript`), npm uses `nobody` user for installation.
This user cannot update .bash_profile, .bashrc and .zshrc files as it has limited permissions.
Show correct message to the user when this happens. In some cases the error is EACCES, in other - EPERM.

Improves error message, related to https://github.com/NativeScript/nativescript-cli/issues/1009


teobugslayer:
:+1: after green build

PR #1028, opened 2015-10-08
Remove xcconfig flag for xcodebuild for tns-ios 1.4+

jasssonpet:


lib/services/android-project-service.ts
```diff
@@ -52,6 +52,7 @@ class AndroidProjectService extends projectServiceBaseLib.PlatformProjectService
 					`${this.$projectData.projectName}-release.apk`
 				],
 				frameworkFilesExtensions: [".jar", ".dat", ".so"],
+				frameworkVersion: require(this.$projectData.projectFilePath).nativescript["tns-ios"].version,
```
rosen-vladimirov: shouldn't this be `tns-android`

rosen-vladimirov: also consider using `this.$fs.readJson` instead of require, as the require is caching the results
one more thing - won't this fail in case `tns-ios` does not exist in nativescript key?

lib/services/ios-project-service.ts
```diff
@@ -57,6 +57,7 @@ export class IOSProjectService extends projectServiceBaseLib.PlatformProjectServ
 			frameworkFilesExtensions: [".a", ".framework", ".bin"],
 			frameworkDirectoriesExtensions: [".framework"],
 			frameworkDirectoriesNames: ["Metadata", "metadataGenerator", "NativeScript", "internal"],
+			frameworkVersion: require(this.$projectData.projectFilePath).nativescript["tns-ios"].version,
```
rosen-vladimirov: same concerns as above:
also consider using `this.$fs.readJson` instead of require, as the require is caching the results
one more thing - won't this fail in case `tns-ios` does not exist in nativescript key?

lib/services/android-project-service.ts
jasssonpet: yep, it should be.

test/ios-project-service.ts
```diff
@@ -120,7 +137,7 @@ describe("Cocoapods support", () => {
 				"version": "0.1.0",
 				"nativescript": {
 					"id": "org.nativescript.myProject2",
-					"tns-android": {
+					"tns-ios": {
```
rosen-vladimirov: maybe this change could be reverted after you fix the `tns-android` issue in `android-project-service.ts`

lib/services/android-project-service.ts
```diff
@@ -52,6 +52,7 @@ class AndroidProjectService extends projectServiceBaseLib.PlatformProjectService
 					`${this.$projectData.projectName}-release.apk`
 				],
 				frameworkFilesExtensions: [".jar", ".dat", ".so"],
+				frameworkVersion: require(this.$projectData.projectFilePath).nativescript["tns-android"].version,
```
rosen-vladimirov: consider getting the frameworkVersion in a separate variable, something like
```TypeScript
let frameworkVersion: string;
let jsonData = this.$fs.readJson(this.$projectData.projectFilePath);
if (jsonData && jsonData.nativescript && jsonData.nativescript["tns-android"]) { 
    frameworkVersion =  jsonData.nativescript["tns-android"].version
}

...
frameworkVersion: frameworkVersion
```

lib/services/ios-project-service.ts
rosen-vladimirov: consider getting the frameworkVersion in a separate variable, something like
```TypeScript
let frameworkVersion: string;
let jsonData = this.$fs.readJson(this.$projectData.projectFilePath);
if (jsonData && jsonData.nativescript && jsonData.nativescript["tns-ios"]) { 
    frameworkVersion =  jsonData.nativescript["tns-ios"].version
}

...
frameworkVersion: frameworkVersion
```

test/ios-project-service.ts
jasssonpet: Cocoapods is available for iOS only, so this should be right.

lib/services/ios-project-service.ts
```diff
@@ -157,10 +158,14 @@ export class IOSProjectService extends projectServiceBaseLib.PlatformProjectServ
 				basicArgs.push("-target", this.$projectData.projectName);
 			}
 
+			// Starting from tns-ios 1.4 the xcconfig file is referenced in the project template
+			if (semver.lt(this.platformData.frameworkVersion, "1.4.0")) {
```
rosen-vladimirov: is it possible to get here in case the framework version is undefined? For example:
```
tns create app1
cd app1
tns build ios
```
In this case I'm not sure that the moment when we set frameworkVersion property, it will have any value.
Maybe just for security we can change the if:
```TypeScript
if (this.platformData.frameworkVersion && semver.lt(this.platformData.frameworkVersion, "1.4.0")) 
```

rosen-vladimirov: okay, I'm not correct, sorry

rosen-vladimirov:
:+1:

PR #1036, opened 2015-10-09
Fix plugin remove if the plugin has Podfile

Fatme:
1. Add plugin that has Podfile
2. Try to remove the plugin

Note: Don't prepare or add platform between step1 and step2


rosen-vladimirov:
:+1:

teobugslayer:
:+1:

PR #1037, opened 2015-10-09
Don't save empty object into package.json file

Fatme:
1. Add plugin that doesn't have plugin variables
2. Open pacakge.json
3. We've added:  "[plugin-name]-variables": {}


dtopuzov:
:+1:

Issue #1044, opened 2015-10-12
iOS debug break on Simulator causes app crash when the debugger is paused on the first line for a long time


KristinaKoeva:
Execute ```tns debug ios --emulator``` then leave the Inspector application as is without any interaction. After a few seconds the application running on the Simulator would crash and the inspector application would detach itself.

vchimev:
This fix will be released with version 1.5.0.

PR #1047, opened 2015-10-12
Pass wait for debugger option. Handle simulator output so that we cou…

KristinaKoeva:
…ld retrieve the PID of the running application.

If you break on a line before executing UIApplicationMain, the Simulator crashes the app after a few seconds. To workaround that behavior we attach lldb which must be killed after the debugging session ends.

Part of the fix for NativeScript/nativescript-cli#1044


Fatme:
:+1:

Issue #1041, opened 2015-10-09
Unable to run the application when project root directory is renamed


vchimev:
Steps to reproduce:
```
tns create appTest
mv appTest applicationTest
cd applicationTest
tns run android
```
Output:
```No .apk found in /Work/release/applicationTest/platforms/android/build/outputs/apk directory```

PR #1053, opened 2015-10-13
Use correct .apk name

rosen-vladimirov:
Gradle uses the third part of the identifier for apk name. Currently we are
checking for projectName (created from dir name). Check both of them when
searching for apk.

Fixes https://github.com/NativeScript/nativescript-cli/issues/1041

### DIFF
--- a/lib/services/android-project-service.ts
+++ b/lib/services/android-project-service.ts
@@ -39,7 +39,7 @@ class AndroidProjectService extends projectServiceBaseLib.PlatformProjectService
 	public get platformData(): IPlatformData {
 		if (!this._platformData) {
 			let projectRoot = path.join(this.$projectData.platformsDir, "android");
-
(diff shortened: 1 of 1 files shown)


dtopuzov:
:100:

lib/services/android-project-service.ts
```diff
@@ -39,7 +39,7 @@ class AndroidProjectService extends projectServiceBaseLib.PlatformProjectService
 	public get platformData(): IPlatformData {
 		if (!this._platformData) {
 			let projectRoot = path.join(this.$projectData.platformsDir, "android");
-
+			let packageName = this.getProjectNameFromId(this.$projectData.projectId);
```
Fatme: You can use `this.$projectData.projectId` inside `getProjectNameFromId` function

Fatme:
:+1:  after addressing my small suggestion

PR #1057, opened 2015-10-13
Restructure error messaging

Mitko-Kerezov:
Detailed explanation can be found here: https://github.com/telerik/mobile-cli-lib/pull/484

### DIFF
--- a/lib/messages.d.ts
+++ b/lib/messages.d.ts
@@ -1,0 +1,13 @@
+///<reference path=".d.ts"/>
+//
+// automatically generated code; do not edit manually!
+//
+interface IMessages{
+			Devices : {
+			NotFoundDeviceByIdentifierErrorMessage: string;
+			NotFoundDeviceByIdentifierErrorMessageWithIdentifier: string;
+			NotFoundDeviceByIndexErrorMessage: string;
+		};
+
+}
+
--- a/lib/nativescript-cli.ts
+++ b/lib/nativescript-cli.ts
@@ -17,6 +17,9 @@ fiber(() => {
 
 	let commandDispatcher: ICommandDispatcher = $injector.resolve("commandDispatcher");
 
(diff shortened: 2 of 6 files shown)


lib/nativescript-cli.ts
```diff
@@ -14,6 +14,9 @@ fiber(() => {
 
 	let commandDispatcher: ICommandDispatcher = $injector.resolve("commandDispatcher");
 
+	let messages = <IMessagesService>$injector.resolve("$messagesService");
```
teobugslayer: You may optionally use `let messages: IMessagesService = $injector.resolve... ` for less code clutter.

lib/messages.d.ts
```diff
@@ -0,0 +1,13 @@
+//
+// automatically generated code; do not edit manually!
+//
+///<reference path=".d.ts"/>
+interface IMessages{
```
Fatme: We've generated the same exact interface in common-lib - https://github.com/telerik/mobile-cli-lib/pull/484/files#diff-2f056d897a840743988e1666c5f9b1baR5. Can we reuse it?

Mitko-Kerezov: Unfortunately I was not able to reuse any parts of the interface due to the following scenario:
These interfaces are generated based upon json files - one default one in the common lib and client-specific ones inside each client respectively. This means that a following case is possible:
Common json file:
```JavaScript
{ 
    MessagesObject: {
        Message1: "Some message",
        Message2: "Some other message"
    }
}
```

Client json file:
```JavaScript
{ 
    MessagesObject: {
        Message1: "Some overriden message"
    }
}
```

This would mean that the interface, generated from the client json file won't be able to extend the interface, generated from the common json because it wouldn't compile.

Personally I don't feel like this is a major issue, because all of this code is generated anyways.

rosen-vladimirov:
:+1:

PR #1061, opened 2015-10-13
Fix tns init command

rosen-vladimirov:
When tns init command is called in empty directory, there's no project data
and getting projectName from projectId is failing.


Fatme:
:+1:

PR #1068, opened 2015-10-15
Node modules NPM scope support

hdeshev:
- Based this on the release branch, but I can rebase it on master too. I just needed the 1.4.2 `apk` fix in order to get this working with my Angular project.
- I have no idea how to test this. All suggestions are welcome.
- Sorry about the ugly diffs -- I decided it's best if I fix some of the space-based indentation too.

----

Copy node modules even if they are below a NPM scope subdir. …

- Needed mostly by Angular's @reactivex/rxjs module.
- Change the broccoli builder node_modules child discovery to
recursively walk NPM scope subdirs (any subdir starting with a @ char).
- Change node-modules-dest-copy to copy @myscope/mymodule dependencies to
the correct subdir.

PR #1072, opened 2015-10-15
Npm scope support

hdeshev:
Rebased the [last PR](https://github.com/NativeScript/nativescript-cli/pull/1068) from master.

### DIFF
--- a/test/npm-support.ts
+++ b/test/npm-support.ts
@@ -24,6 +24,7 @@ import Future = require("fibers/future");
 
 import path = require("path");
 import temp = require("temp");
+import shelljs = require("shelljs");
 temp.track();
 
 let assert = require("chai").assert;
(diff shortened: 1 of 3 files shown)


test/npm-support.ts
```diff
@@ -185,6 +186,27 @@ describe("Npm support tests", () => {
...
 		assert.isTrue(fs.exists(bplistParserFolderPath).wait());
 	});
+	it("Ensures that scoped dependencies are prepared correctly", () => {
+		// Setup
+		let scopedName = "@reactivex/rxjs";
+		let scopedModule = path.join(projectFolder, "node_modules", "@reactivex/rxjs");
+		let scopedPackageJson = path.join(scopedModule, "package.json");
+		addDependencies(testInjector, projectFolder, {scopedName:  "0.0.0-prealpha.3"}).wait();
+		//create module dir, and add a package.json
+		shelljs.mkdir('-p', scopedModule);
```
Fatme: Why do you need to create this folder and add `package.json`? This should be handled by `preparePlatform`.

hdeshev: I needed it because the `rebuildChangedDirectory` routine adds dependencies only if there is at least one `package.json` file:

https://github.com/NativeScript/nativescript-cli/blob/master/lib/tools/broccoli/node-modules-dest-copy.ts#L39

teobugslayer:
Looks good to me. Squash the commits and merge the PR. :+1:

Issue #1054, opened 2015-10-13
Can no longer 'tns platform add ios', ENOENT.


emiloberg:
All of a sudden, when adding iOS platform (`tns platform add ios`), I get:
```
ENOENT, open '/Users/emiloberg/code/slask/cleanproj/clean/platforms/ios/__PROJECT_NAME__/main.m
```

As far as I know, nothing has changed on my machine, no updates, no nothing.

How to repeat:

```
~/code/slask
▶ npm cache clean

~/code/slask
▶ node -v
v0.10.33

~/code/slask
▶ tns --version
1.3.0

~/code/slask
▶ mkdir cleanproj

~/code/slask
▶ cd cleanproj

code/slask/cleanproj
▶ tns create clean
Project clean was successfully created

code/slask/cleanproj
▶ cd clean

slask/cleanproj/clean
▶ tns platform add ios
zsh: correct 'platform' to 'platforms' [nyae]? n
Copying template files...
ENOENT, open '/Users/emiloberg/code/slask/cleanproj/clean/platforms/ios/__PROJECT_NAME__/main.m'
# platform add
┌────────────────────────┬──────────────────────────────────────────────────────────────────────────────────────────────┐
│ Usage                  │ Synopsis                                                                                     │
│ Android latest runtime │ $ tns platform add android [--frameworkPath <File Path>] [--symlink] [--sdk <API Level>]     │
│ Android selected       │ $ tns platform add android[@<Version>] [--frameworkPath <File Path>] [--symlink] [--sdk <API │
│ runtime                │ Level>]                                                                                      │
│ iOS latest runtime     │ $ tns platform add ios [--frameworkPath <File Path>] [--symlink]                             │
│ iOS selected runtime   │ $ tns platform add ios[@<Version>] [--frameworkPath <File Path>] [--symlink]                 │
└────────────────────────┴──────────────────────────────────────────────────────────────────────────────────────────────┘

Configures the current project to target the selected platform.

### Options

    * --frameworkPath - Sets the path to a NativeScript runtime for the specified platform that you want to use instead of the default runtime. If --symlink is specified, <File Path> must point to directory in which the runtime is already extracted. If --symlink is not specified, <File Path> must point to a valid npm package.
[...]

PR #1073, opened 2015-10-15
Fix ENOENT error when add ios platform

Fatme:
Fixes https://github.com/NativeScript/nativescript-cli/issues/1054

### DIFF
--- a/lib/npm-installation-manager.ts
+++ b/lib/npm-installation-manager.ts
@@ -6,6 +6,12 @@ import semver = require("semver");
 import * as npm from "npm";
 import * as constants from "./constants";
 
+interface IVersionData {
+	major: string;
+	minor: string;
+	patch: string;
+}
+
 export class NpmInstallationManager implements INpmInstallationManager {
 	private static NPM_LOAD_FAILED = "Failed to retrieve data from npm. Please try again a little bit later.";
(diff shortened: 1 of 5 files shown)


teobugslayer:
You need to fix the linter errors.

lib/npm-installation-manager.ts
```diff
@@ -6,6 +6,12 @@ import semver = require("semver");
 import * as npm from "npm";
 import * as constants from "./constants";
 
+interface IVersionData {
```
teobugslayer: Some JSDocs would be much welcome

lib/npm-installation-manager.ts
```diff
@@ -163,5 +191,14 @@ export class NpmInstallationManager implements INpmInstallationManager {
...
 	}
+	
+	private getVersionData(version: string): IVersionData {
+		let versionParts = version.split(".");
+		return {
+			major: versionParts[0],
+			minor: versionParts[1],
+			patch: versionParts[2]
+		};
+	}
```
teobugslayer: The code is correct, but could be made more concise:

```
let [ major, minor, patch ] = version.split(".")
return { major, minor, patch }
```

teobugslayer:
:+1: after making the linter happy. My remarks are purely cosmetic.

PR #1079, opened 2015-10-19
Fix symlink of android framework

rosen-vladimirov:
When you use --symlink for android runtime, CLI's modifying the original AndroidManifest.xml and res dirs.
So we should not symlink src directory and build.gradle, settings.gradle.


Fatme:
:+1:  after green build

PR #1087, opened 2015-10-20
Always call npm install when preparing project

rosen-vladimirov:
Preparing project should always call `npm install` just to make sure all dependencies
are installed. Currently we are not calling it when node_modules dir exists.
This was due to requirement that when you have node_modules and change something in
node_modules/tns-core-modules directory, `npm install` will override your changes.
The fact is that this is totally incorrect behavior on our side and also making changes
in `node_modules/<package_name>` is forbidden. Anyway, in case you make changes there,
calling `npm install` will not override them.

Remove the incorrect check and always call `npm install`.

Fix unit test which was passing due to missing npm install before. The problem with it now is that 
```
{scopedName:  "0.0.0-prealpha.3"}
```
Will create an object with property "scopedName" instead of the required one `"@reactivex/rxjs"`.

### DIFF
--- a/lib/services/plugins-service.ts
+++ b/lib/services/plugins-service.ts
@@ -189,13 +189,11 @@ export class PluginsService implements IPluginsService {
 
 	public ensureAllDependenciesAreInstalled(): IFuture<void> {
 		return (() => {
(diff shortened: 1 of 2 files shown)


rosen-vladimirov:
run ci

teobugslayer:
:+1:

PR #1094, opened 2015-10-22
Fix EISDIR error when the plugin is specified with path to local folder

Fatme:
Fixes the issue1 - https://github.com/NativeScript/nativescript-cli/issues/883


dtopuzov:
:+1:

teobugslayer:
:+1:

PR #1095, opened 2015-10-22
Fix usage of --frameworkPath <directory> with --symlink

rosen-vladimirov:
When `--frameworkPath` is used, we should follow this behavior:
Always install to `node_modules` under `platforms/<platform_name>/node_modules`

options used   | is frameworkPath File | is frameworkPath Directory
---------|--------|------------
frameworkPath + symlink | symlink to node_modules. **DO NOT DELETE node_modules**. | symlink to **frameworkPath dir**. Delete node_modules.
frameworkPath | Delete node_modules. | Delete node_modules.


dtopuzov:
:+1:

teobugslayer:
:+1:

Fatme:
:+1:  after green build

PR #1102, opened 2015-10-26
Do not call npm install if everything is installed

rosen-vladimirov:
In case there's node_modules dir and there's directory for each dependency defined
in package.json of the project, we should not call `npm install`. In case `node_modules` dir
does not exist and the project has dependencies or any of the dependencies is not installed,
we'll call `npm install`. Another case when we'll call `npm install` is when `--force` option is used.

This change is in order to allow modifications of modules inside `node_modules` directory.

## Second commit
Install latest available version if no matching version is found

In case CLI is version 1.4.x, on platform add we are trying to install latest available
1.4.x runtime. In case there's no 1.4.x version, the code fails. 
Instead we should install latest available version in this case.

### DIFF
--- a/lib/npm-installation-manager.ts
+++ b/lib/npm-installation-manager.ts
@@ -84,7 +84,7 @@ export class NpmInstallationManager implements INpmInstallationManager {
 				.sortBy(verData => verData.patch)
 				.value();
 
-			let result = _.last(compatibleVersions);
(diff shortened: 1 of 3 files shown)


ligaz:
:+1: adding a unit test will be a nice addition.

dtopuzov:
:100:

PR #1104, opened 2015-10-26
Check cocoapods availability before executing pod install

Fatme:


vchimev:
:+1:

lib/services/ios-project-service.ts
```diff
@@ -499,9 +491,19 @@ export class IOSProjectService extends projectServiceBaseLib.PlatformProjectServ
...
 
 	private executePodInstall(): IFuture<any> {
-		this.$logger.info("Installing pods...");
-		let podTool = this.$config.USE_POD_SANDBOX ? "sandbox-pod" : "pod";
-		return this.$childProcess.spawnFromEvent(podTool,  ["install"], "close", { cwd: this.platformData.projectRoot, stdio: 'inherit' });
+		return (() => {
+			// Check availability
+			try {
+				this.$childProcess.exec("gem which cocoapods").wait();
+				this.$childProcess.exec("gem which xcodeproj").wait();
```
ligaz: Do we need to check for xcodeproj? It is a dependency of cocoapods.

Fatme: We need this in order to ensure that this script will be executed correctly - https://github.com/NativeScript/nativescript-cli/blob/master/lib/services/ios-project-service.ts#L441

PR #1108, opened 2015-10-27
Fix fs copyFile when source and target are the same

rosen-vladimirov:
When source and target file are the same, copyFile produces empty file.
Add check and if they are the same, do not execute any action.

Move fs tests from appbuilder-cli to common. Add tests for copyFile.


teobugslayer:
:+1:

PR #1126, opened 2015-10-30
Fix  ENOENT, open '~/myApp3/node_modules/.bin/package.json/'

Fatme:
Steps to reproduce:
1. Create {N} project
2. tns plugin add nativescript-unit-test-runner
3. tns prepare ios


rosen-vladimirov:
:+1:

PR #1133, opened 2015-11-02
Use gradle wrapper instead of global gradle

teobugslayer:
See https://github.com/NativeScript/NativeScript/issues/1024


teobugslayer:
Note, this must be merged AFTER corresponding Android runtime PR

lib/services/android-project-service.ts
```diff
@@ -109,11 +106,21 @@ export class AndroidProjectService extends projectServiceBaseLib.PlatformProject
...
+				this.copy(this.platformData.projectRoot, frameworkDir, "gradle", "-R");
+				this.copy(this.platformData.projectRoot, frameworkDir, "gradlew gradlew.bat", "-f");
+			}
+
 			this.cleanResValues(targetSdkVersion, frameworkVersion).wait();
 
 		}).future<any>()();
 	}
 
+	private useGradleWrapper(frameworkDir: string): boolean {
```
rosen-vladimirov: I would name this `canUseGradleWrapper`, but this is just an opinion

PR #1134, opened 2015-11-02
Fail if sandbox install command fails

Fatme:
sandbox-pod install blocks arbitrary post-install scripts but returns with an exit code of 0. This is why we have to check the stderr for any data and fail if it is present.
Also help should not be displayed in case such a fail occurs.

### DIFF
--- a/lib/services/ios-project-service.ts
+++ b/lib/services/ios-project-service.ts
@@ -514,7 +514,24 @@ export class IOSProjectService extends projectServiceBaseLib.PlatformProjectServ
 
 			this.$logger.info("Installing pods...");
 			let podTool = this.$config.USE_POD_SANDBOX ? "sandbox-pod" : "pod";
-			return this.$childProcess.spawnFromEvent(podTool,  ["install"], "close", { cwd: this.platformData.projectRoot, stdio: 'inherit' }).wait();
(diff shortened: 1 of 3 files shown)


teobugslayer:
:

lib/services/ios-project-service.ts
```diff
@@ -510,7 +510,15 @@ export class IOSProjectService extends projectServiceBaseLib.PlatformProjectServ
 
 			this.$logger.info("Installing pods...");
 			let podTool = this.$config.USE_POD_SANDBOX ? "sandbox-pod" : "pod";
-			return this.$childProcess.spawnFromEvent(podTool,  ["install"], "close", { cwd: this.platformData.projectRoot, stdio: 'inherit' }).wait();
+			let childProcess = this.$childProcess.spawnFromEvent(podTool,  ["install"], "close", { cwd: this.platformData.projectRoot, stdio: ['pipe', process.stdout, 'pipe'] }).wait();
+			if (childProcess.stderr) {
+				let errors = childProcess.stderr.replace(/(\u001b\[(?:\d*;){0,5}\d*m)[\s\S]+?(\u001b\[(?:\d*;){0,5}\d*m)/mg, "").trim();
```
rosen-vladimirov: please use updated regex from @Mitko-Kerezov 
Also let's print the warnings with `$logger.warn` instead of just removing them.

Mitko-Kerezov:
Merge after https://github.com/telerik/mobile-cli-lib/pull/519
Ping @rosen-vladimirov and @teobugslayer for a second review

rosen-vladimirov:
run ci

Issue #594, opened 2015-06-26
Implement a hard dependency on the node.js version and exit if not satisfied


valentinstoychev:
We are currently logging a WARN when we run in an older version of Node.JS, but I think we need to stop the execution of the command if there is an older version of Node.js installed, otherwise some of the commands simply do not work. I hit an issues with the platform add command under Windows which prevented adding the Android platform at all. @rosen-vladimirov  is aware of this.

dtopuzov:
With latest behavior we do WARN on each command if Node version is not in supported versions.

PR #1136, opened 2015-11-03
Verify node version and warn the user when it is not supported

teobugslayer:
Update node versions in package.json

Implements https://github.com/NativeScript/nativescript-cli/issues/594


rosen-vladimirov:
:+1:

ErjanGavalji:
:+1:

Issue #493, opened 2015-05-21
Enable transpilers support in NativeScript projects


ligaz:
A common request we have received in the last couple of months is the ability to enable (or start with) JavaScript transpilers for a given NativeScript project. The usual suspects are TypeScript and Babel which recently became quite popular with the approaching ratification of ES6.

### Guiding principles 
There a couple of guiding principles we want to follow when we enable this functionality.

#### Making it generic
We want to design this feature so we can accommodate any transpiler, not just ones for JavaScript. What if you want to use LESS or Sass instead of CSS? Fan of Jade instead of XML we should have you covered. 
 
#### Per project configuration
This feature should allow usage of different version of the transpiler in different projects. For example you can use TypeScript 1.4 for your production app but switch to the cutting edge TypeScript 1.5 in your new project.

#### Ability to to start with transpiler or enable it on the go
We believe that the barrier to entry for NativeScript should be as low as the one of the Web. We shouldn't force the users to learn all the cool ES6 features or TypeScript's types system. Users should be able to enable a transpiler for an already existing project or start a new project with given transpiler  if this is what they want.

### The Command
The support will boil down to a single command that users will have to execute:
```sh
$ tns install typescript
```
or
```sh
$ tns install babel
``` 
~~As you can see we have not settled on the command name :smile: We did a little brainstorming sessions internally and we come up with `enable` (ex. `$ tns enable babel`) however this sounds too broad and we are not totally sold on it. If you have ideas how we can better name this command please file them in the comments.~~

### Implementation details
Here are some internal and more technical implementation details:
* We plan to add the transpilation code itself in a form of project level hooks. The given transpiler hooks will come in the form of a separate npm package that will be "hooked" in the current NativeScript project. 
* All transpilation tools will be added as `devDependencies` to the project's `package.json`.

# Work breakdown

#### TypeScript support
- [x] `tns install typescript` command. The compiler should be added as `devDependency`.
- [x] Consume TypeScript declarations from the `tns-core-modules` package.
- [x] Generate `tsconfig.json` if not present.
- [x] Add a project level hook that will do the actual transpilation.

#### TypeScript support for NativeScript plugins

tjvantoll:
Hey @ligaz,

I like what you have here. I agree that each transpiler should be opt in and as generic as possible. I have two questions:

[...]

PR #1138, opened 2015-11-03
remove environment vars from `tns install`

tailsu:
If your modules depend on these vars, then they can't be reinstalled with `npm install`


tailsu:
Part of #493

tailsu:
:+1:

Issue #1123, opened 2015-10-30
[Unit Testing] First run of `tns test ios --emulator` fails


ikoevska:
Steps to reproduce.
1. Create a new project and switch to its dir.
1. Run `tns test init --framework jasmine`.
1. Run `rm -rf node_modules/.bin`. (You only need to run this if we still don't have the code which ignores .bin during prepare.)
1. Run `tns test ios --emulator`.

After the project preparing and building successfully, I get this in the log and my tests never run.

```
** BUILD SUCCEEDED **

Project successfully built
Using  /Volumes/HDD/cli-projects/jasmine2/platforms/ios/build/emulator/jasmine2.app
Cleaning up before starting the iOS Simulator
Starting iOS Simulator
projectFilesPath = /Volumes/HDD/cli-projects/jasmine2/platforms/ios/jasmine2/app applicationPath = null
shell.js: internal error
TypeError: Arguments to path.join must be strings
    at Object.posix.join (path.js:488:13)
    at /usr/local/lib/node_modules/nativescript/node_modules/shelljs/src/cp.js:171:28
    at Array.forEach (native)
    at Object._cp (/usr/local/lib/node_modules/nativescript/node_modules/shelljs/src/cp.js:157:11)
    at Object.cp (/usr/local/lib/node_modules/nativescript/node_modules/shelljs/src/common.js:186:23)
    at syncAction (/usr/local/lib/node_modules/nativescript/lib/common/mobile/ios/ios-emulator-services.js:57:26)
    at /usr/local/lib/node_modules/nativescript/lib/common/mobile/ios/ios-emulator-services.js:196:13
    at Function.settle (/usr/local/lib/node_modules/nativescript/node_modules/fibers/future.js:224:26)
    at FiberFuture.Future.wait (/usr/local/lib/node_modules/nativescript/node_modules/fibers/future.js:509:10)
    at /usr/local/lib/node_modules/nativescript/lib/common/services/usb-livesync-service-base.js:75:112
    at Function.settle (/usr/local/lib/node_modules/nativescript/node_modules/fibers/future.js:224:26)
    at FiberFuture.Future.wait (/usr/local/lib/node_modules/nativescript/node_modules/fibers/future.js:509:10)
    at /usr/local/lib/node_modules/nativescript/lib/services/test-execution-service.js:68:368
    at Function.settle (/usr/local/lib/node_modules/nativescript/node_modules/fibers/future.js:224:26)
[...]

PR #1140, opened 2015-11-04
Fixes https://github.com/NativeScript/nativescript-cli/issues/1123

Fatme:


lib/services/usb-livesync-service.ts
```diff
@@ -67,8 +67,11 @@ export class UsbLiveSyncService extends usbLivesyncServiceBaseLib.UsbLiveSyncSer
 				}).future<boolean>()();
 			};
 
-			let notRunningiOSSimulatorAction = (): IFuture<void> => {
-				return this.$platformService.deployOnEmulator(this.$devicePlatformsConstants.iOS.toLowerCase());
+			let notRunningiOSSimulatorAction = (): IFuture<boolean> => {
+				return (() => {
+					 this.$platformService.deployOnEmulator(this.$devicePlatformsConstants.iOS.toLowerCase()).wait();
+					 return false;
```
rosen-vladimirov: what's the point of returning false every time?

Fatme: https://github.com/telerik/mobile-cli-lib/pull/506/files#diff-c8066f499a9cad66c8bae208dd04b18cR145. The problem is that `livesync` command doesn't need to sync the application if the simulator is not running, but the `test` command needs to perform livesync after the simulator get started.

teobugslayer: I suggest to add a comment in the code explaining this.

rosen-vladimirov:
:+1:  after green build

Issue #1021, opened 2015-10-07
tns build android --compileSdk 17 returns bad error message


dtopuzov:
```
tns build android --compileSdk 17 
```
returns bad error message:

```
[31;1mYou need to have the latest Android Support Repository installed on your system.[0m
```
This error can not be solved by installing latest Android Support Repository, here is the full log:
```
Versions found in /Users/topuzov/Tools/adt-bundle/sdk/build-tools are 19.1.0, 21.1.2, 22.0.1, 23.0.0
Selected targetSdk is: 17
Versions found in /Users/topuzov/Tools/adt-bundle/sdk/extras/android/m2repository/com/android/support/appcompat-v7 are 18.0.0, 19.0.0, 19.0.1, 19.1.0, 20.0.0, 21.0.0, 21.0.2, 21.0.3, 22.0.0, 22.1.0, 22.1.1, 22.2.0, 22.2.1, 23.0.0
Selected AppCompat version is: null
[31;1mYou need to have the latest Android Support Repository installed on your system.[0m
```

dtopuzov:
Not fixed. Current master:

Directories found in /Users/topuzov/Tools/adt-bundle/sdk/build-tools are 19.1.0, 21.1.2, 22.0.1, 23.0.0, android-4.4W
Versions found in /Users/topuzov/Tools/adt-bundle/sdk/build-tools are 19.1.0, 21.1.2, 22.0.1, 23.0.0
Selected targetSdk is: 17
Directories found in /Users/topuzov/Tools/adt-bundle/sdk/extras/android/m2repository/com/android/support/appcompat-v7 are 18.0.0, 19.0.0, 19.0.1, 19.1.0, 20.0.0, 21.0.0, 21.0.2, 21.0.3, 22.0.0, 22.1.0, 22.1.1, 22.2.0, 22.2.1, 23.0.0, 23.0.1, 23.1.0, maven-metadata.xml, maven-metadata.xml.md5, maven-metadata.xml.sha1
Versions found in /Users/topuzov/Tools/adt-bundle/sdk/extras/android/m2repository/com/android/support/appcompat-v7 are 18.0.0, 19.0.0, 19.0.1, 19.1.0, 20.0.0, 21.0.0, 21.0.2, 21.0.3, 22.0.0, 22.1.0, 22.1.1, 22.2.0, 22.2.1, 23.0.0, 23.0.1, 23.1.0
Selected AppCompat version is: undefined
[...]

PR #1154, opened 2015-11-06
Fix warning when attempting to use too old version of Android SDK

teobugslayer:
See https://github.com/NativeScript/nativescript-cli/issues/1021

### DIFF
--- a/lib/android-tools-info.ts
+++ b/lib/android-tools-info.ts
@@ -71,8 +71,8 @@ export class AndroidToolsInfo implements IAndroidToolsInfo {
 			}
 
 			if(!toolsInfoData.supportRepositoryVersion) {
-				this.printMessage(`You need to have the latest Android Support Repository installed on your system.`,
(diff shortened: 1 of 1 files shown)


lib/android-tools-info.ts
```diff
@@ -71,8 +71,8 @@ export class AndroidToolsInfo implements IAndroidToolsInfo {
 			}
 
 			if(!toolsInfoData.supportRepositoryVersion) {
-				this.printMessage(`You need to have the latest Android Support Repository installed on your system.`,
-					'Run `$ android`  to manage the Android Support Repository.');
+				this.printMessage("You need to have Android SDK 22 or later and the latest Android Support Repository installed on your system.",
```
rosen-vladimirov: instead of Android SDK 22 you should use `MIN_REQUIRED_COMPILE_TARGET`

rosen-vladimirov:
:+1:

Issue #1086, opened 2015-10-20
--key-store-path option to look for the keystore file path relative to the app root


vchimev:
Currently the `<File Path>` value of `--key-store-path` option used with `tns build/run android` needs to be relative to the *platforms/android* folder. It would be more convenient to be relative to the app root folder.

Actual: `$ tns run android --keyStorePath ../../../../keystore/Telerik.keystore`
Expected: `$ tns run android --keyStorePath ../keystore/Telerik.keystore`

```
vchimev@VCHIMEV /c/Work/release
$ tns --help run android
# run android
┌─────────────────────────────────┬─────────────────────────────────────────────────
│ Usage                           │ Synopsis
│ Run on all connected devices    │ $ tns run android [--key-store-path <File Path> ...
```

rosen-vladimirov:
It should be relative to the directory where the command is executed, definitely a bug.

PR #1157, opened 2015-11-06
Fix --key-store-path path

teobugslayer:
See https://github.com/NativeScript/nativescript-cli/issues/1086


rosen-vladimirov:
:+1:

PR #1159, opened 2015-11-08
Improve Getting-started for Android

rosen-vladimirov:
* Remove requirement to have `<android-sdk>/tools` and `<android-sdk>/platform-tools` directories to your PATH - use them based on ANDROID_HOME variable.
* Based on the above point - replace all calls to `android` directly with correct path to the `android` executable.
* Remove requirement for JAVA added in your PATH - validate it by using JAVA_HOME.
* Improve error messages when ANDROID_HOME is not set - for example warning about missing compile SDK will inform you in case ANDROID_HOME is not set.
* Improve ANDROID_HOME checks - validate that directories like "extras", "tools", "platform-tools" exist (at least on of them).
* Validate JAVA version in doctor command.
* Sometimes `build-tools` directories are not called with version(`22.0.1`), but have the name: `build-tools-22.0.1` - make sure we will be able to use such dirs
* Skip cocoapods warnings on Linux and Windows (for `tns doctor` command).

### DIFF
--- a/lib/android-tools-info.ts
+++ b/lib/android-tools-info.ts
@@ -3,24 +3,78 @@
 
 import * as path from "path";
 import * as semver from "semver";
+import {EOL} from "os";
 
 export class AndroidToolsInfo implements IAndroidToolsInfo {
 	private static ANDROID_TARGET_PREFIX = "android";
 	private static SUPPORTED_TARGETS = ["android-17", "android-18", "android-19", "android-21", "android-22", "android-23"];
 	private static MIN_REQUIRED_COMPILE_TARGET = 22;
 	private static REQUIRED_BUILD_TOOLS_RANGE_PREFIX = ">=22";
-	private static VERSION_REGEX = /^(\d+\.){2}\d+$/;
+	private static VERSION_REGEX = /((\d+\.){2}\d+)/;
+	private static MIN_JAVA_VERSION = "1.7.0";
+
 	private showWarningsAsErrors: boolean;
 	private toolsInfo: IAndroidToolsInfoData;
 	private selectedCompileSdk: number;
 	private installedTargetsCache: string[] = null;
 	private androidHome = process.env["ANDROID_HOME"];
+	private pathToAndroidExecutable: string;
+	private _androidExecutableName: string;
+	private get androidExecutableName(): string {
+		if(!this._androidExecutableName) {
+			this._androidExecutableName = "android";
+			if(this.$hostInfo.isWindows) {
+				this._androidExecutableName += ".bat";
+			}
+		}
+
+		return this._androidExecutableName;
+	}
 
 	constructor(private $childProcess: IChildProcess,
 		private $errors: IErrors,
 		private $fs: IFileSystem,
+		private $hostInfo: IHostInfo,
 		private $logger: ILogger,
 		private $options: IOptions) {}
+
+	public getPathToAndroidExecutable(): IFuture<string> {
+		return ((): string => {
+			if (!this.pathToAndroidExecutable) {
+				if(this.validateAndroidHomeEnvVariable(this.androidHome).wait()) {
+					let androidPath = path.join(this.androidHome, "tools", this.androidExecutableName);
+					if(!this.trySetAndroidPath(androidPath).wait() && !this.trySetAndroidPath(this.androidExecutableName).wait()) {
+						this.$errors.failWithoutHelp(`Unable to find "${this.androidExecutableName}" executable file. Make sure you have set ANDROID_HOME environment variable correctly.`);
+					}
+				} else {
+					this.$errors.failWithoutHelp("ANDROID_HOME environment variable is not set correctly.");
+				}
+			}
+
+			return this.pathToAndroidExecutable;
+		}).future<string>()();
+	}
+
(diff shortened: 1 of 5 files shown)


rosen-vladimirov:
run ci

lib/android-tools-info.ts
```diff
@@ -3,25 +3,63 @@
...
+				} else {
+					this.$errors.failWithoutHelp("ANDROID_HOME environment variable is not set.");
+				}
+			}
+
+			return this.pathToAndroidExecutable;
+		}).future<string>()();
+	}
+
+	private trySpawnAndroidHelp(androidPath: string): IFuture<boolean>{
```
ligaz: Isn't it better to name this method `trySetAndroidPath`? It is unclear when reading `getPathToAndroidExecutable` that `trySpawnAndroidHelp` actually sets the path to the executable.

rosen-vladimirov: yes, I've changed the logic to set the path in this method, but I forgot to rename it :smiley_cat:

lib/android-tools-info.ts
```diff
@@ -235,5 +318,24 @@ export class AndroidToolsInfo implements IAndroidToolsInfo {
 	private getMaxSupportedVersion(): number {
 		return this.parseAndroidSdkString(_.last(AndroidToolsInfo.SUPPORTED_TARGETS.sort()));
 	}
+
+	private validateAndroidHomeEnvVariable(androidHomeEnvVar: string): IFuture<boolean> {
```
ligaz: Usually validate methods return `true` when validation is successful and `false` otherwise.

rosen-vladimirov: fixed :)

rosen-vladimirov:
run ci

rosen-vladimirov:
run ci

lib/android-tools-info.ts
```diff
@@ -92,7 +149,31 @@ export class AndroidToolsInfo implements IAndroidToolsInfo {
...
+				" described in https://github.com/NativeScript/nativescript-cli#system-requirements.";
+			let matchingVersion = (installedJavaVersion || "").match(AndroidToolsInfo.VERSION_REGEX);
+			if(matchingVersion && matchingVersion[1]) {
+				if(semver.lt(matchingVersion[1], AndroidToolsInfo.MIN_JAVA_VERSION)) {
+					hasProblemWithJavaVersion = true;
+					this.printMessage(`Javac version ${installedJavaVersion} is not supported. You have to install at least ${AndroidToolsInfo.MIN_JAVA_VERSION}.`, additionalMessage);
+				}
+			} else {
+				hasProblemWithJavaVersion = true;
+				this.printMessage("Error executing command 'javac'. Make sure you have java installed and set JAVA_HOME environment variable.", additionalMessage);
```
teobugslayer: It is not enough to install java (the JRE). you need to install JDK. THis should be reflected in the message.

lib/services/doctor-service.ts
```diff
@@ -36,17 +36,34 @@ class DoctorService implements IDoctorService {
...
+
+			if(!sysInfo.cocoapodVer ) {
+				this.$logger.warn("WARNING: CocoaPod is not installed or is not configured properly.");
+				this.$logger.out("You will not be able to build your projects for iOS if they contain plugin with CocoaPod file." + EOL
+					+ "To be able to build such projects, verify that you have installed CocoaPod.");
+				result = true;
+			}
+
+			if(sysInfo.cocoapodVer && semver.lt(sysInfo.cocoapodVer, DoctorService.MIN_SUPPORTED_POD_VERSION)) {
+				this.$logger.warn(`WARNING: CocoaPod version is lower than ${DoctorService.MIN_SUPPORTED_POD_VERSION}`);
```
teobugslayer: a typo - it shopuld be CocoaPods (plural)

lib/android-tools-info.ts
```diff
@@ -235,5 +337,24 @@ export class AndroidToolsInfo implements IAndroidToolsInfo {
...
+	private validateAndroidHomeEnvVariable(androidHomeEnvVar: string): IFuture<boolean> {
+		return ((): boolean => {
+			let isAndroidHomeCorrect = true;
+			let expectedDirectoriesInAndroidHome = ["build-tools", "tools", "platform-tools", "extras"];
+			if(!androidHomeEnvVar || !this.$fs.exists(androidHomeEnvVar).wait()) {
+				this.printMessage("The ANDROID_HOME environment variable is not set or it points to a non-existent directory. You will not be able to perform any build-related operations for Android.",
+					"To be able to perform Android build-related operations, set the `ANDROID_HOME` variable to point to the root of your Android SDK installation directory.");
+				isAndroidHomeCorrect = false;
+			} else if(!_.any(expectedDirectoriesInAndroidHome.map(dir => this.$fs.exists(path.join(androidHomeEnvVar, dir)).wait()))) {
+				this.printMessage("The ANDROID_HOME environment variable points to incorrect directory. You will not be able to perform any build-related operations for Android.",
```
teobugslayer: This message does not reflect the check being made. For example, if I forget to install Android extras, but correctly set ANDROID_HOME, I will still get this message.

teobugslayer:
:+1:

Issue #945, opened 2015-09-18
tns build --release should add --for-device


tjvantoll:
I ran `tns build ios --release` for the first time today, and I was really surprised to find my .ipa in the `platforms/ios/build/emulator` folder.

If I'm doing a release build, shouldn't it be assumed that I'm building for a device and not emulators?

tjvantoll:
On top of this, currently, `tns build ios --release` does not show me where my .ipa was placed. However, `tns build ios --release --for-device` includes this very helpful message:

```
Results at '/Users/tj/Desktop/Groceries/platforms/ios/build/device/Groceries.ipa'
Project successfully built
```

rosen-vladimirov:
Hi @tjvantoll 

Thanks for the feedback. I totally agree with you that release builds must target devices and we'll have to fix this behavior. I'll have to check why you do not see the path to the .ipa when producing release build. I consider both questions as issues in current implementation, so I've marked this as bug.

Thanks again!

tjvantoll:
Hey @rosen-vladimirov,

Thanks for the response. Just to note I do see the feedback when running `tns build ios --release --for-device`; I only do *not* see the path when I omit `--for-device`—aka  `tns build ios --release`.

Mitko-Kerezov:
Hi @tjvantoll 
We've decided against limiting users and instead sticking with the current behavior where any combination of `--release`, `--debug`, `--for-device` and `--emulator` is possible. Instead we've decided to implement a brand new command for performing publish builds which most probably generate an .xcarchive file and/or and .ipa file for ios. 
Having that in mind the fix was reverted with this https://github.com/NativeScript/nativescript-cli/pull/1166

tjvantoll:
Thanks for the update @Mitko-Kerezov. I like the idea of having a new command because I think it’ll be easier for users to find, so :+1:

teobugslayer:
This will be addressed as part of https://github.com/NativeScript/nativescript-cli/issues/1172 , closing this issue.

PR #1165, opened 2015-11-09
Always build --for-device with --release

Mitko-Kerezov:
References https://github.com/NativeScript/nativescript-cli/issues/945

Ping @Fatme @teobugslayer for a quick review

### DIFF
--- a/lib/services/ios-project-service.ts
+++ b/lib/services/ios-project-service.ts
@@ -172,7 +172,7 @@ export class IOSProjectService extends projectServiceBaseLib.PlatformProjectServ
 			}
 
 			let args: string[] = [];
-			if(this.$options.forDevice) {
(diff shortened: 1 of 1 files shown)


lib/services/ios-project-service.ts
```diff
@@ -172,7 +172,7 @@ export class IOSProjectService extends projectServiceBaseLib.PlatformProjectServ
 			}
 
 			let args: string[] = [];
-			if(this.$options.forDevice) {
+			if(this.$options.forDevice || this.$options.release) {
```
rosen-vladimirov: You can extract this check to a separate variable and use it here and in the next if :smile_cat: 
But definitely not a merge stopper

teobugslayer:
:+1:

PR #1170, opened 2015-11-10
Fix doctor command for android and adb

rosen-vladimirov:
Fix doctor command, which checks adb and android incorrectly. Pass correct paths for verification.
Also make sure "No issues were detected" is printed when there are no issues.

### DIFF
--- a/lib/android-tools-info.ts
+++ b/lib/android-tools-info.ts
@@ -180,7 +180,7 @@ export class AndroidToolsInfo implements IAndroidToolsInfo {
 	public getPathToAdbFromAndroidHome(): IFuture<string> {
 		return (() => {
 			if(this.androidHome) {
-				let pathToAdb = path.join(this.androidHome, "platform-tools", "adb") + (this.$hostInfo.isWindows ? ".exe" : "");
+				let pathToAdb = path.join(this.androidHome, "platform-tools", "adb");
 				try {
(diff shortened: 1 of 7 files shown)


lib/android-tools-info.ts
```diff
@@ -177,6 +177,24 @@ export class AndroidToolsInfo implements IAndroidToolsInfo {
 		}).future<boolean>()();
 	}
 
+	public getPathToAdbFromAndroidHome(): IFuture<string> {
+		return (() => {
+			if(this.androidHome) {
+				let pathToAdb = path.join(this.androidHome, "platform-tools", "adb") + (this.$hostInfo.isWindows ? ".exe" : "");
```
teobugslayer: Do you really need to take care of the .exe suffix? It should work without it.

dtopuzov:
:100:

lib/android-tools-info.ts
dtopuzov: Not sure if it works without, it but it works with .exe and CI is green :) :100:

rosen-vladimirov: yep, I do not need this, I'll remove it (I've added it for debugging purposes and I forgot to remove it).

Mitko-Kerezov:
:+1:

teobugslayer:
Do not forget to squash, :+1:

PR #1174, opened 2015-11-11
Do not copy .ts and .js.map files in real app

rosen-vladimirov:
Currently all `.ts` and `.js.map` files are copied to `platforms/<platform>/...` etc.
We do not need them in the real application for the moment. So ignore them.
Also fix problem with ignoring `tests` directory during release build - we were preparing collection of files to be copied, but finally we copy the whole app dir.
From now on we'll respect the filtered collection.


rosen-vladimirov:
run ci

teobugslayer:
:+1:

PR #1176, opened 2015-11-12
Fix run on multiple devices

rosen-vladimirov:
When multiple devices with same platform are attached, `tns run <platform>` command should start the app on all of them.
Currently the application is started, but the call to commandsService (device|run) command is failing as more than one device is attached.
Fix this by calling the correct method from the devicesService.

Fixes https://github.com/NativeScript/nativescript-cli/issues/1177


Fatme:
:+1:

teobugslayer:
:+1:

PR #1179, opened 2015-11-12
Fail in case node version is blacklisted

rosen-vladimirov:
There are some node versions that cannot be used with our CLIs.
In case user is trying to use CLI with them, inform him that it is not supported and exit the process.


teobugslayer:
:+1: after green build

PR #1183, opened 2015-11-13
Install devDependencies before prepare

rosen-vladimirov:
As our hooks (like typescript) are called before executing prepare method, we must be sure we have
all devDependencies installed in order to be able to execute before-prepare hooks.
This is mandatory for projects where typescript is enabled, but you do not have node_modules directory yet.

Steps to reproduce the bug:
1) `tns create app1`
2) `cd app1`
3) `tns install typescript`
4) `rm -rf node_modules`
5) `tns build android`
You'll see error:
```
Error: Cannot find module 'nativescript-dev-typescript/lib/before-prepare.js'
```

### DIFF
--- a/lib/services/plugins-service.ts
+++ b/lib/services/plugins-service.ts
@@ -185,6 +185,22 @@ export class PluginsService implements IPluginsService {
 		return this.executeForAllInstalledPlatforms(action);
 	}
 
+	public installDevDependencies(): IFuture<void> {
+		return (() => {
+			let installedDependencies = this.$fs.exists(this.nodeModulesPath).wait() ? this.$fs.readDirectory(this.nodeModulesPath).wait() : [];
+			let packageJsonContent = this.$fs.readJson(this.getPackageJsonFilePath()).wait();
+			let devDependencies = _.keys(packageJsonContent.devDependencies);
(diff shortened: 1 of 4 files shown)


teobugslayer:
:+1:

lib/services/plugins-service.ts
```diff
@@ -185,6 +185,22 @@ export class PluginsService implements IPluginsService {
 		return this.executeForAllInstalledPlatforms(action);
 	}
 
+	public installDevDependencies(): IFuture<void> {
+		return (() => {
+			let installedDependencies = this.$fs.exists(this.nodeModulesPath).wait() ? this.$fs.readDirectory(this.nodeModulesPath).wait() : [];
+			let packageJsonContent = this.$fs.readJson(this.getPackageJsonFilePath()).wait();
+			let devDependencies = _.keys(packageJsonContent.devDependencies);
+			if(devDependencies.length && (this.$options.force || _.difference(devDependencies, installedDependencies).length)) {
+				let command = `npm install ${devDependencies.join(" ")}`;
```
Fatme: Can we use npm install --development?

rosen-vladimirov: nope, I want only devDependencies to be installed, while `npm i --development` is installing everything **including** devDependencies.

PR #1186, opened 2015-11-13
Install ALL dependencies and devDependencies before prepare

rosen-vladimirov:
The old fix for installing only devDependencies before preparing the project has a problem:
`npm install <devPackagesNames>` does not respect the versions from the devDependencies section and always installs the latest versions.

We cannot use `npm install <devPackageName>@<version>` as we do not know if they are real versions or paths or even "frog legs".

There's a magical flag `--dev` which should allow you to do:
`npm install <devPackagesNames> --dev` and it will respect the versions from devDependencies. Unfortunately this flag forces recursive installation of devDependencies all down the tree of packages.

Of course npm has another rabbit in the hat - `--only=dev`. However I couldn't get it work at all.

So the only solution I could think about is to install all dependencies before preparing the project.


teobugslayer:
:+1:

Mitko-Kerezov:
:+1:

Fatme:
:+1:

PR #1188, opened 2015-11-13
Collect all warnings upon pod installing

rosen-vladimirov:
Leftover warnings printed on the stderr are considered errors and as a result fail the `$ tns prepare ios` command. Make a more general regex for catching warnings.
Ping @Mitko-Kerezov @Fatme


lib/services/ios-project-service.ts
```diff
@@ -524,7 +524,7 @@ export class IOSProjectService extends projectServiceBaseLib.PlatformProjectServ
 				// HACK for silencing irrelevant linking warnings when pod installing on
```
rosen-vladimirov: you do not need this comment anymore

Fatme:
:+1:

rosen-vladimirov:
:+1:  after green build

PR #1189, opened 2015-11-13
Remove .ts from tns-core-modules package when moving it to platforms dir

rosen-vladimirov:
The new tns-core-modules package has .d.ts files. We should skip them when moving tns-core-modules to platforms dir.
The current logic is - copy everything from node_modules to tns_modules dir.
After that move everything from tns_modules/tns-core-modules dir to tns_modules and delete tns-core-modules.

Add logic to delete .ts files BEFORE moving the files from tns_modules/tns-core-modules dir to tns_modules.


vchimev:
:+1:

Fatme:
:+1:

PR #1193, opened 2015-11-16
Show message after successful deploy on android

rosen-vladimirov:
Get back the message that reports successful deploy on android device. We've deleted it a couple of commits ago.


dtopuzov:
:+1:

Mitko-Kerezov:
:+1:

PR #1196, opened 2015-11-16
Release the console after livesync command and fix a typo.

Fatme:
The console is not release if `tns livesync ios` command is executed


rosen-vladimirov:
:+1:

PR #1202, opened 2015-11-16
Fix getting of cocoapod version

rosen-vladimirov:
`pod --version` command, which is used to determine the version of cocoapods, might return some warnings.
Here's a sample output:
```
WARNING: CocoaPods requires your terminal to be using UTF-8 encoding.
Consider adding the following to ~/.profile:

export LANG=en_US.UTF-8

0.39.0
```
Make sure to get the real version in this output (0.39.0).


vchimev:
:+1:

Fatme:
:+1:

Issue #1199, opened 2015-11-16
"tns test" fails with Unknown command 'dev-test|android'.


dtopuzov:
C:\Git>rm -rf TestApp

C:\Git>tns create TestApp
Project TestApp was successfully created

C:\Git>tns test init --path TestApp
? Select testing framework: jasmine
npm WARN package.json @ No description
npm WARN package.json @ No repository field.
npm WARN package.json @ No README data
karma@0.13.15 TestApp\node_modules\karma
├── batch@0.5.3
├── di@0.0.1
├── rimraf@2.4.3
├── graceful-fs@4.1.2
├── mime@1.3.4
├── colors@1.1.2
├── http-proxy@1.12.0 (eventemitter3@1.1.1, requires-port@0.0.1)
├── minimatch@3.0.0 (brace-expansion@1.1.1)
├── glob@5.0.15 (path-is-absolute@1.0.0, inherits@2.0.1, inflight@1.0.4, once@1.3.2)
├── dom-serialize@2.2.1 (custom-event@1.0.0, void-elements@2.0.1, extend@3.0.0, ent@2.2.0)
├── optimist@0.6.1 (wordwrap@0.0.3, minimist@0.0.10)
├── connect@3.4.0 (utils-merge@1.0.0, parseurl@1.3.0, debug@2.2.0, finalhandler@0.4.0)
├── useragent@2.1.7 (lru-cache@2.2.4)
├── expand-braces@0.1.2 (array-unique@0.2.1, array-slice@0.2.3, braces@0.1.5)
├── bluebird@2.10.2
├── body-parser@1.14.1 (bytes@2.1.0, content-type@1.0.1, depd@1.1.0, qs@5.1.0, http-errors@1.3.1, on-finished@2.3.0, raw-body@2.1.4, debug@2.2.0, iconv-lite@0.4.12, type-is@1.6.9)
├── source-map@0.5.3
├── chokidar@1.2.0 (arrify@1.0.0, path-is-absolute@1.0.0, glob-parent@2.0.0, async-each@0.1.6, is-binary-path@1.0.1, is-glob@2.0.1, lodash.flatten@3.0.2, readdirp@2.0.0, anymatch@1.3.0)
├── log4js@0.6.27 (async@0.2.10, underscore@1.8.2, readable-stream@1.0.33, semver@4.3.6)
├── lodash@3.10.1
├── core-js@1.2.6
[...]

PR #1207, opened 2015-11-17
Fix unit test runner on Windows

rosen-vladimirov:
## Use fork instead of spawn for startTestRunner

In `karma-nativescript-launcher` we call startTestRunner through spawn.
Communication between karma launcher and tns is through stdin.
This is causing exception in buildMetadata step when gradle wrapper is used.
For some reason it throws error in java code.
So, replace the spawn with fork in the launcher. Modify `startTestRunner` to wait for message from parent process and send "ready" state to the parent.
After that the parent (karma-nativescript-launcher) sends configuration data and the `startTestRunner` will execute the actions.

Fixes https://github.com/NativeScript/nativescript-cli/issues/1199

## Add help for dev-test-* commands

As the current workflow of unit-testing is:
1) user calls `tns test android`
2) CLI starts a new process in which `tns dev-test android` is called
3) In case the dev-test command fails, the error is caught in the child process and command help is printed on the terminal directly in it.
4) User receives message: `Unable to find command dev-test|android` (from the child process) as we do not have help for this commands.

Copy-paste the help fo test-android and test-ios commands and change the filenames so when the child process fails, it will print correct help.
Parent process will not fail as the error is caught inside the child, so there's no way to print the help twice.

Update common-lib to search dev-commands as well when searching for help pages.

## Make sure all dependencies are installed before starting karma
In case node_modules dir is deleted, `tns test <platform>` command will fail.
So ensure all dependencies are installed before starting karma server.

### Should be merged after:
* https://github.com/telerik/mobile-cli-lib/pull/529 - fix in common-lib
* https://github.com/NativeScript/karma-nativescript-launcher/pull/21 - fix in karma launcher - publish new version before merging this PR

### DIFF
--- a/lib/services/test-execution-service.ts
+++ b/lib/services/test-execution-service.ts
@@ -32,7 +32,8 @@ class TestExecutionService implements ITestExecutionService {
 		private $config: IConfiguration,
 		private $logger: ILogger,
 		private $fs: IFileSystem,
-		private $options: IOptions) {
+		private $options: IOptions,
+		private $pluginsService: IPluginsService) {
 	}
 
 	public startTestRunner(platform: string) : IFuture<void> {
@@ -133,6 +134,7 @@ class TestExecutionService implements ITestExecutionService {
 	public startKarmaServer(platform: string): IFuture<void> {
 		return (() => {
 			platform = platform.toLowerCase();
+			this.$pluginsService.ensureAllDependenciesAreInstalled().wait();
 			let pathToKarma = path.join(this.$projectData.projectDir, 'node_modules/karma');
 			let KarmaServer = require(path.join(pathToKarma, 'lib/server'));
 			if (platform === 'ios' && this.$options.emulator) {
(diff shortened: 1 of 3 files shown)


Fatme:
:+1:

lib/services/test-execution-service.ts
```diff
@@ -31,93 +32,109 @@ class TestExecutionService implements ITestExecutionService {
...
+						let platformSpecificLiveSyncServices: IDictionary<any> = {
+							android: (_device: Mobile.IDevice, $injector: IInjector): IPlatformSpecificLiveSyncService => {
+								return $injector.resolve(this.$androidUsbLiveSyncServiceLocator.factory, {_device: _device});
+							},
+							ios: (_device: Mobile.IDevice, $injector: IInjector) => {
+								return $injector.resolve(this.$iosUsbLiveSyncServiceLocator.factory, {_device: _device});
+							}
+						};
+
+						let notInstalledAppOnDeviceAction = (device: Mobile.IDevice): IFuture<void> => {
```
teobugslayer: I think we can move these callbacks outside the body of startTestRunner.

PR #1209, opened 2015-11-18
Release the console after `tns run ios` command

Fatme:
If `tns run ios` command is executed, the console is not released.


teobugslayer:
:+1:

PR #1213, opened 2015-11-18
Add instructions for use to setup scripts

teobugslayer:
Rename the windows setup script to ps1, so it can be directly executed if the user has chocolatey installed already

### DIFF
--- a/setup/native-script.ps1
+++ b/setup/native-script.ps1
@@ -1,5 +1,9 @@
 
 # A Boxstarter script to set up Windows machine for NativeScript development
+# To run it against RELEASE branch (recommended) use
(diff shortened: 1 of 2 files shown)


setup/native-script.ps1
```diff
@@ -1,5 +1,9 @@
 
 # A Boxstarter script to set up Windows machine for NativeScript development
+# To run it against RELEASE branch (recommended) use
```
rosen-vladimirov: maybe it is worth using the 'production' branch (that we've discussed for our docs) here as well

rosen-vladimirov:
:+1:

setup/native-script.ps1
teobugslayer: is this branch ready yet?

PR #1217, opened 2015-11-19
Create configurations and src folders only for plugins that have native files

Fatme:


lib/services/android-project-service.ts
```diff
@@ -296,18 +296,20 @@ export class AndroidProjectService extends projectServiceBaseLib.PlatformProject
...
+				this.$fs.ensureDirectoryExists(pluginConfigurationDirectoryPath).wait();
+
+				// Copy all resources from plugin
+				let resourcesDestinationDirectoryPath = path.join(this.platformData.projectRoot, "src", pluginName);
+				this.$fs.ensureDirectoryExists(resourcesDestinationDirectoryPath).wait();
+				shell.cp("-Rf", path.join(pluginPlatformsFolderPath, "*"), resourcesDestinationDirectoryPath);
+			}
 
 			// Copy include.gradle file
 			let includeGradleFilePath = path.join(pluginPlatformsFolderPath, "include.gradle");
```
rosen-vladimirov: should we copy this file in case the plugin does not have native code?

rosen-vladimirov:
:+1:

lib/services/android-project-service.ts
Fatme: No, we copy it only when the plugin has platforms/android folder

Plamen5kov:
:+1:

vchimev:
:+1:

Issue #1130, opened 2015-11-02
Livesync with console.log


N3ll:
_From @prijindal on November 1, 2015 18:26_

Currently you can't log anything with livesync
Any plans on upgrading livesync to show log output?

_Copied from original issue: NativeScript/NativeScript#1009_

N3ll:
_From @x4080 on November 2, 2015 2:18_

+1

N3ll:
_From @manijak on November 2, 2015 9:58_

+1

emiloberg:
One way of solving it would be to make it easier for the developer to start view the log files, as described in this issue https://github.com/NativeScript/NativeScript/issues/1009#issuecomment-152999397

x4080:
I found that livesync ios don't update changed .ios the Android side works though

vchimev:
@x4080,

I tried what you reported with the latest {N} CLI 1.5 Release Candidate and experienced only this issue: https://github.com/NativeScript/nativescript-cli/issues/1234.

Please, give a try to {N} 1.5 when released and let us know if you still experience such problem by opening a separate issue.

vchimev:
There is an issue using `console.log` during LiveSync that on Android after making a change to a *.js* file (respectively the app restarts) than the `console.log` messages duplicate. /CC: @Fatme

vchimev:
Sum up:
- `consoloe.log()` will be supported with LiveSync in {N} CLI 1.5.0 for Android device/emulator and iOS device
- `consoloe.log()` support with LiveSync for iOS simulators implementation could be followed here: https://github.com/NativeScript/nativescript-cli/issues/1262
- the issue described above for duplicate `console.log` messages during LiveSync on Android will be resolved in {N} CLI 1.5.1

PR #1219, opened 2015-11-19
Show application output from livesync command

Fatme:
https://github.com/NativeScript/nativescript-cli/issues/1130


teobugslayer:
:+1:

PR #1220, opened 2015-11-19
Silent shelljs by default

rosen-vladimirov:
Shelljs likes to report some warnings, which we do not want to show to our users.
So set it to silent mode by default. In case for any shelljs command you want to show the warnings, just set silent to false before executing it.

This will also fix some warnings that are currently shown when preparing plugins. (`cp - unable to find file...`).


teobugslayer:
:+1:

PR #1226, opened 2015-11-20
Prevent Future resolved more than once errors

rosen-vladimirov:
Different node versions have different behaviors of timers. In some cases, on some machines we throw error that future is resolved more than once.
Prevent it.


teobugslayer:
:+1:

vchimev:
:+1:

PR #1227, opened 2015-11-20
Android debugger unixsockets

blagoev:
This pull request is releated to the changes here #273. A help is needed to merge this into master and then investigate what issues there might be in app builder CLI. 

The long story short. The android debugger now uses named sockets instead of ports hence the code rewrite in adndroid-debug-service to accommodate for that.
Contact me directly in order to merge this successfully into the respected main branches.

### DIFF
--- a/lib/services/android-debug-service.ts
+++ b/lib/services/android-debug-service.ts
@@ -3,6 +3,8 @@
 import * as helpers from "../common/helpers";
 import * as path from "path";
 import * as util from "util";
+import * as net from "net";
+import Future = require("fibers/future");
 
 class AndroidDebugService implements IDebugService {
     private static ENV_DEBUG_IN_FILENAME = "envDebug.in";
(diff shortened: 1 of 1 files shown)


lib/services/android-debug-service.ts
```diff
@@ -43,14 +45,64 @@ class AndroidDebugService implements IDebugService {
...
+            var server = net.createServer();
+            server.listen(candidatetPort, function (err: any) {
+                server.once('close', function () {
+                    future.return(true);
+                })
+                server.close();
+            });
+            
+            server.on('error', function (err: any) {
+               future.return(false);
```
Fatme: maybe `future.throw`?

lib/services/android-debug-service.ts
```diff
@@ -43,14 +45,64 @@ class AndroidDebugService implements IDebugService {
...
+            });
+            
+            return future;
+		}).future<boolean>()();
```
Fatme: You don't need this line.

blagoev:
@Fatme : Please fix all in place. This commit should be regarded as a reference for the necessary fixes that correspond to the related one in the engine.

teobugslayer:
reimplemented in https://github.com/NativeScript/nativescript-cli/pull/1373

PR #1228, opened 2015-11-22
Add support for node 5.1.x

rosen-vladimirov:
Add support for node 5.1.x by updating references to native dependencies.
Add versions between 5.1.x and 6.0.0 as validated (we will not show warning for them).


Mitko-Kerezov:
:+1:

PR #1232, opened 2015-11-23
Fix device log for API 23

rosen-vladimirov:
When the Android is using API Level 23, the device log is in different format and our current code fails.
Use new regex for API Level 23 and fix the incorrect code that was failing.


dtopuzov:
:+1:

vchimev:
:+1:

Mitko-Kerezov:
:+1:

PR #1248, opened 2015-11-25
Fix cocoapods check and add unit tests for doctor service

rosen-vladimirov:
When cocoapods are not installed on Mac machine, doctor is failing.
Fix this by skipping match check in case pods are not installed.
Add basic unit tests for doctor service.

Add unit test for the failing cocoapods check.


teobugslayer:
:+1:

PR #1250, opened 2015-11-25
Mac setup improvements

teobugslayer:
- rewritten as pure Ruby script, no longer tries to be a brew formula
- update installed node to 4 LTS

### DIFF
--- a/setup/native-script.rb
+++ b/setup/native-script.rb
@@ -42,41 +42,19 @@ system('brew install caskroom/cask/brew-cask')
 
 puts "Installing the Java SE Development Kit... This might take some time, please, be patient. (You might need to provide your password.)"
 system('brew cask install java')
-system('echo "export JAVA_HOME=$(/usr/libexec/java_home)" >> ~/.bash_profile')
-system('echo "export ANDROID_HOME=/usr/local/opt/android-sdk" >> ~/.bash_profile')
+system('echo "export JAVA_HOME=$(/usr/libexec/java_home)" >> ~/.profile')
 
-puts "Installing node.js 0.12"
-system('brew install homebrew/versions/node012')
+puts "Installing Android SDK"
+system('brew install android-sdk')
+system('echo "export ANDROID_HOME=/usr/local/opt/android-sdk" >> ~/.profile')
 
-puts "Creating Homebrew formula for NativeScript."
-File.open("/usr/local/Library/Formula/native-script.rb", "w:utf-8") do |f|
-  f.write DATA.read
-end
+puts "Configuring your system for Android development... This might take some time, please, be patient."
+system "echo yes | /usr/local/opt/android-sdk/tools/android update sdk --filter tools,platform-tools,android-23,build-tools-23.0.2,extra-android-m2repository --all --no-ui"
 
-puts "Installing NativeScript formula... This might take some time, please, be patient."
-system('brew install native-script')
+puts "Installing Node.js 4"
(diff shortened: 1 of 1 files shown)


teobugslayer:
ping @ErjanGavalji @dtopuzov

setup/native-script.rb
```diff
@@ -43,40 +43,18 @@
...
-
-class NativeScript < Formula
-  desc "NativeScript"
-  homepage "https://www.nativescript.org"
-  version "1.3.0"
-  url "https://raw.githubusercontent.com/NativeScript/nativescript-cli/brew/setup/empty.tar.gz"
-  sha256 "813e1b809c094d29255191c14892a32a498e2ca298abbf5ce5cb4081faa4e88f"
+puts "Installing Android SDK"
+system('brew install android-sdk')
+system('echo "export ANDROID_HOME=/usr/local/opt/android-sdk" >> ~/.bash_profile')
```
ligaz: This will not work with custom shells like zsh.

setup/native-script.rb
```diff
@@ -43,40 +43,18 @@
...
-  depends_on "pkg-config" => :build
-#  depends_on "node" # currently we do not work with latest node, and we manually install 0.12 (see above)
-  depends_on "android-sdk"
+puts "Configuring your system for Android development... This might take some time, please, be patient."
+system "echo yes | /usr/local/opt/android-sdk/tools/android update sdk --filter tools,platform-tools,android-23,build-tools-23.0.2,extra-android-m2repository --all --no-ui"
 
-  def install
-    ohai "Installing NativeScript CLI..."
-    system "/usr/local/bin/npm install -g nativescript"
+puts "Installing node.js 4"
```
ligaz: The correct spelling is Node.js

setup/native-script.rb
ErjanGavalji: .profile might be the better option, for [it gets considered by both zsh and bash](https://en.wikipedia.org/wiki/Unix_shell). It will still not work in fancier shells like fish, but I suggest that bash and zsh are a good coverage.

teobugslayer: Great suggestion, I will change the file.

teobugslayer:
All comments are addressed

rosen-vladimirov:
:+1:

setup/native-script.rb
```diff
@@ -42,41 +42,19 @@
...
 system('brew cask install java')
-system('echo "export JAVA_HOME=$(/usr/libexec/java_home)" >> ~/.bash_profile')
-system('echo "export ANDROID_HOME=/usr/local/opt/android-sdk" >> ~/.bash_profile')
+system('echo "export JAVA_HOME=$(/usr/libexec/java_home)" >> ~/.profile')
 
-puts "Installing node.js 0.12"
-system('brew install homebrew/versions/node012')
+puts "Installing Android SDK"
+system('brew install android-sdk')
+system('echo "export ANDROID_HOME=/usr/local/opt/android-sdk" >> ~/.profile')
```
ligaz: I'm not sure that zsh loads `.profile`: http://superuser.com/questions/187639/zsh-not-hitting-profile

ErjanGavalji: so sad...
still, let's use the .profile file instead of sticking to the bash-specific .bash_profile :)

ErjanGavalji:
:+1:

ligaz:
We already have zsh support for our completion so I'm proposing that we add it here as well :smile:

Issue #1242, opened 2015-11-24
"ANDROID_HOME environment variable is not set correctly" error is thrown then `tns run ios --log trace` command is executed


Fatme:
The ANDROID_HOME environment variable is not set or it points to a non-existent directory. You will not be able to perform any build-related operations for Android.
To be able to perform Android build-related operations, set the ANDROID_HOME variable to point to the root of your Android SDK installation directory.

ANDROID_HOME environment variable is not set correctly.
error is thrown when ios related command is executed with --log trace option.

vchimev:
```
VCHIMEVMAC:nativescript-cli-tests vchimev$ node_modules/.bin/tns doctor
The ANDROID_HOME environment variable is not set or it points to a non-existent directory. You will not be able to perform any build-related operations for Android.
To be able to perform Android build-related operations, set the ANDROID_HOME variable to point to the root of your Android SDK installation directory.

ANDROID_HOME environment variable is not set correctly.
WARNING: adb from the Android SDK is not installed or is not configured properly.
For Android-related operations, the NativeScript CLI will use a built-in version of adb.
To avoid possible issues with the native Android emulator, Genymotion or connected
Android devices, verify that you have installed the latest Android SDK and
its dependencies as described in http://developer.android.com/sdk/index.html#Requirements

TIP: To avoid setting up the necessary environment variables, you can use the Homebrew package manager to install the Android SDK and its dependencies.

WARNING: The Android SDK is not installed or is not configured properly.
You will not be able to build your projects for Android and run them in the native emulator.
To be able to build for Android and run apps in the native emulator, verify that you have
installed the latest Android SDK and its dependencies as described in http://developer.android.com/sdk/index.html#Requirements

TIP: To avoid setting up the necessary environment variables, you can use the Homebrew package manager to install the Android SDK and its dependencies.

WARNING: CocoaPod is not installed or is not configured properly.
You will not be able to build your projects for iOS if they contain plugin with CocoaPod file.
To be able to build such projects, verify that you have installed CocoaPod.
ANDROID_HOME environment variable is not set correctly.
ANDROID_HOME environment variable is not set correctly.
ANDROID_HOME environment variable is not set correctly.
Cannot find a compatible Android SDK for compilation. To be able to build for Android, install Android SDK 22 or later.
Run $ android to manage your Android SDK versions.

You need to have the Android SDK Build-tools installed on your system. You can install any version in the following range: '>=22 <=23'.
[...]

PR #1255, opened 2015-11-26
Fix "ANDROID_HOME environment variable is not set correctly"  error

Fatme:
Fixes https://github.com/NativeScript/nativescript-cli/issues/1242


rosen-vladimirov:
:+1:

PR #1271, opened 2015-11-27
Check XML files for well-formedness

teobugslayer:
Check XML files for well-formedness. No validation is performed. The check should happen at least on build and livesync.

Whenever an invalid XML file is found:

  - for livesync --watch operation, warn the user, but do not break the CLI
  - for every other operation, abort it

Implementation detail: it happens actually on prepare.
See https://github.com/NativeScript/nativescript-cli/issues/1272

### DIFF
--- a/lib/services/platform-service.ts
+++ b/lib/services/platform-service.ts
@@ -154,7 +154,7 @@ export class PlatformService implements IPlatformService {
 		}).future<string[]>()();
 	}
 
-	public preparePlatform(platform: string): IFuture<void> {
+	public preparePlatform(platform: string): IFuture<boolean> {
 		return (() => {
 			this.validatePlatform(platform);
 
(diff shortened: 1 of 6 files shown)


lib/services/platform-service.ts
```diff
@@ -166,12 +166,36 @@ export class PlatformService implements IPlatformService {
...
+			return this.preparePlatformCore(platform).wait();
+		}).future<boolean>()();
+	}
+
+	private checkXmlFiles(sourceFiles: string[]): IFuture<boolean> {
+		return (() => {
+			let xmlHasErrors = false;
+			let DomParser = require("xmldom").DOMParser;
+			sourceFiles
+				.filter(file => _.endsWith(file, ".xml"))
```
rosen-vladimirov: Maybe path.ext

rosen-vladimirov:
Nice one! :+1:

PR #1277, opened 2015-11-30
Fix prepare command

rosen-vladimirov:
Add missing wait in prepare command.
Make sure the whole message for invalid xml is red.


Fatme:
:+1:

Issue #1268, opened 2015-11-27
Hook failures treated as bad user input


hdeshev:
A failing hook correctly fails the build (`tns prepare`, `tns run`, etc), but that case gets treated as a user input error and the user sees the help page containing command invocation details and parameter lists.

This is mostly a problem with the typescript compile hook since the `tns run` help page is pretty big and pushes the compile errors reported by the compiler off the screen.

```
Executing before-prepare hook from /home/local/TELERIK/hdeshev/w/template-hello-world-ng/hooks/before-prepare/20-nativescript-dev-typescript.js
Found peer TypeScript 1.6.2
/home/local/TELERIK/hdeshev/w/template-hello-world-ng/app/app.ts(6,1): error TS2304: Cannot find name 'kuuuuuuuuuuuuuuu__'.
TypeScript compiler failed with exit code 1
# prepare
┌─────────┬───────────────────────┐
│ Usage   │ Synopsis              │
│ General │ $ tns prepare android │
└─────────┴───────────────────────┘

Copies common and relevant platform-specific content from the app directory to the subdirectory for the selected target platform
in the platforms directory. This lets you build the project with the SDK for the selected platform. 

```

vchimev:
Tested with 1.5.1 RC.

If the hook fails due to typescript errors the executed command does not interrupt.

```
VCHIMEVMAC:nativescript-cli-tests vchimev$ node_modules/.bin/tns run  android --emulator --path tsApp
Executing before-prepare hook from /Work/GitHub/nativescript-cli-tests/tsApp/hooks/before-prepare/nativescript-dev-typescript.js
Found peer TypeScript 1.7.3
/Work/GitHub/nativescript-cli-tests/tsApp/app/main-view-model.ts(6,34): error TS2339: Property 'counter' does not exist on type 'HelloWorldModel'.
/Work/GitHub/nativescript-cli-tests/tsApp/app/main-view-model.ts(10,14): error TS2339: Property 'counter' does not exist on type 'HelloWorldModel'.
/Work/GitHub/nativescript-cli-tests/tsApp/app/main-view-model.ts(11,18): error TS2339: Property 'counter' does not exist on type 'HelloWorldModel'.
/Work/GitHub/nativescript-cli-tests/tsApp/app/main-view-model.ts(15,38): error TS2339: Property 'counter' does not exist on type 'HelloWorldModel'.
Project successfully prepared

:config phase:  createDefaultIncludeFiles

[...]

PR #1279, opened 2015-11-30
Do not show help when hook fails

rosen-vladimirov:
When any of the hooks fail, we show help of the command.
Hooks failure is not bad user input, so we should not show help.

Fixes https://github.com/NativeScript/nativescript-cli/issues/1268


teobugslayer:
:+1:

PR #1281, opened 2015-12-01
Do not delete app's package.json

rosen-vladimirov:
App's package.json contains information about application's entry point, so we should not delete it.


teobugslayer:
:+1: but we really need to clear the user confusion with multiple package.json files

ErjanGavalji:
Well, searching for a package.json file in a folder to find the main executable is a part of the philosophy of the runtimes. We'd better make that clear enough everywhere possible so that the users don't get confused. By the way, examples of confusion could help clear out the next steps...

ErjanGavalji:
:+1:

PR #1289, opened 2015-12-01
Update npm version to latest 2.x

rosen-vladimirov:
Update npm version to latest 2.x as the old one (2.6.1) has some bugs that prevent us from using git urls, installing some plugins, etc.


vchimev:
:+1:

teobugslayer:
:+1:

Issue #1272, opened 2015-11-27
Check XML files for well-formedness


teobugslayer:
Check XML files for well-formedness. No validation is performed. The check should happen at least on build and livesync.

Whenever an invalid XML file is found:

  - for livesync --watch operation, warn the user, but do not break the CLI
  - for every other operation, ~~abort it~~

Implementation detail: it happens actually on prepare.

vchimev:
**Important note** - in {N} CLI 1.5.1 the xml files check for well-formedness will only warn the user for syntax errors, it will not abort any CLI operations.

Implementation detail - it happens on prepare.

PR #1291, opened 2015-12-02
Only warn about XML parsing errors

teobugslayer:
See https://github.com/NativeScript/nativescript-cli/issues/1272


rosen-vladimirov:
:+1:

PR #1298, opened 2015-12-03
Fix tns doctor command

rosen-vladimirov:
When ANDROID_HOME environment variable is not set, we call childProcess exec and spawn with undefined command.
This leads to "futures left behind" errors.
Make sure the commands will be executed only in case they are not undefined.
Also make sure that `tns prepare android` will fail immediately in case ANDROID_HOME is not set (currently it shows warning for ANDROID_HOME and shows error for compile SDK).

Related issue: https://github.com/NativeScript/nativescript-cli/issues/1242


teobugslayer:
:+1:

vchimev:
:+1:

PR #1305, opened 2015-12-03
Add baseConfig option

rosen-vladimirov:
Add baseConfig option that can be used to replace the default AndroidManifest.xml / Info.plist that is used for merge.
It should be passed to all build related commands. It can be passed to platform add command as well.

Required for AppBuilder integration.


teobugslayer:
:+1:

Issue #1247, opened 2015-11-25
don't kill adb server if possible


hshristov:
When uninstalling nativescript adb is killed. This requires to restart emulators. If it possible remove this kill.

PR #1319, opened 2015-12-07
Don't stop adb uninstall

teobugslayer:
See https://github.com/NativeScript/nativescript-cli/issues/1247


rosen-vladimirov:
:+1:  after green build :)

PR #1344, opened 2015-12-12
Fix geny 2.6

Fatme:
On some devices ro.product.manufacturer returns unknown. This leads to endless loop for CLI.

Since the underlying emulation engine for Genymotion is VirtualBox and that's not going to change any time soon, I've added a check for ro.build.product property.


vchimev:
:+1:

PR #1352, opened 2015-12-15
Merge release in master

rosen-vladimirov:
Merge release in master.

Conflicts:
	PLUGINS.md
	lib/common
	package.json

### Require logger in CLI, not in common lib
`logger` has been required in common lib, but due to some Proton requirements, it should be required from each CLI.
Fix this by adding it to bootstrap and update to latest common lib.


teobugslayer:
:+1:

PR #1356, opened 2015-12-16
Init command should not force tns-core-modules version

rosen-vladimirov:
Currently `tns init` command forces the application to use the default version of tns-core-modules.
For example in case your CLI is version 1.4.3, `tns init` in non-interactive terminal will set version 1.4.0 (latest 1.4.x) of tns-core-modules even in case you already have defined your version in your package.json
Use --tnsCoreModules option by default - if it is specified, it has highest priority.
In case the console is not interactive, check if package.json has entry in dependencies section for tns-core-modules and use it.
In case the console is interactive, show all available versions to the user and let him pick up the one he wants.
In case the console is not interactive and the package.json does not have entry for `tns-core-modules`, use the default logic.

Targeting release branch as this feature is required for AppBuilder integration.

### DIFF
--- a/lib/services/init-service.ts
+++ b/lib/services/init-service.ts
@@ -60,7 +60,9 @@ export class InitService implements IInitService {
 				if(!dependencies) {
 					projectData.dependencies = Object.create(null);
 				}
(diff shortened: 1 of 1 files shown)


dtopuzov:
:+1:

Mitko-Kerezov:
:+1:

PR #1372, opened 2016-01-05
Do not consider App_Resources/Android as plugin

rosen-vladimirov:
Currently when `app/App_Resources/Android` contains AndroidManifest.xml or include.gradle, we treat the folder in a specail way and copy it on two places.
This will be handled from the runtime itself, so no need to do it anymore.

Related to https://github.com/NativeScript/android-runtime/pull/309


Mitko-Kerezov:
:+1:

Issue #1302, opened 2015-12-03
`--nobrowser` option for the `tns help` command (command to generate html help)


ErjanGavalji:
Introduce an option for the tns help command, so that it prepares the documentation, but does not launch a browser. This way the documentation will get collectible by other tools.

rosen-vladimirov:
What's the use-case here?
Currently we are showing browser help in case you use
`tns help <command name>`
But in case you need command line help you can use `tns <command name> -h `

teobugslayer:
This is needed to generate CLI documentation for publishing on docs.nativescript.org.
I think, a dev- command is more appropriate here.

teobugslayer:
Actually, I've implemented the `dev-generate-help` command. As any dev- command it lacks help.

ErjanGavalji:
:+1:

dtopuzov:
Html is generated at node_modules\nativescript\docs\html

PR #1374, opened 2016-01-05
Add command to generate HTML help

teobugslayer:
Implements https://github.com/NativeScript/nativescript-cli/issues/1302

### DIFF
--- a/lib/commands/generate-help.ts
+++ b/lib/commands/generate-help.ts
@@ -1,0 +1,14 @@
+///<reference path="../.d.ts"/>
+"use strict";
+
+export class GenerateHelpCommand implements ICommand {
+	constructor(private $htmlHelpService: IHtmlHelpService) {
+	}
+
+	public allowedParameters: ICommandParameter[] = [];
+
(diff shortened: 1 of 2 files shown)


teobugslayer:
ping @ErjanGavalji

lib/commands/generate-help.ts
```diff
@@ -0,0 +1,19 @@
...
+	}
+
+	public allowedParameters: ICommandParameter[] = [];
+
+	public execute(args: string[]): IFuture<void> {
+		return this.$htmlHelpService.generateHtmlPages();
+	}
+
+	public canExecute(args: string[]): IFuture<boolean> {
+		return Future.fromResult(true);
```
rosen-vladimirov: Why do you need `canExecute` method? The idea is to use it when you cannot specify exactly the number (and type) of the allowed parameter. In this case the command will not accept any parameters, so I suggest to remove the `canExecute` implementation, or I'm missing something here?

rosen-vladimirov:
:+1:

Issue #1375, opened 2016-01-06
Improve tns debug android --get-port


teobugslayer:
It is not deterministic (especially on node 0.12). This is needed for integration with VS Code.

blagoev:
Parsing the output on something like "port 40001" in the output will be more deterministic and will safeguard us for future warnings messages that precede this text on this command

PR #1378, opened 2016-01-07
make getport command deterministic

blagoev:
fix output when multiple devices are available


blagoev:
issue https://github.com/NativeScript/nativescript-cli/issues/1375

PR #1383, opened 2016-01-08
Fix baseConfig option for iOS

rosen-vladimirov:
When `--baseConfig` option is used for `tns platform add ios`, the command fails with ENOTEMPTY.
The reason is that we copy the baseConfig file before creating correct directories and when we try to rename `__PROJECT_NAME__` dir to real project name, it fails as we have already created this dir.


nadyaA:
:+1:

Mitko-Kerezov:
:+1:

Issue #1294, opened 2015-12-02
TNS Doctor should check CocoaPods/Ruby version


burkeholland:
I noticed that Ruby can wreak absolute havoc on a project when using CocoaPods. The wrong version (< 2.0.0) will throw cryptic errors that make people feel sad.

It would be nice if TNS doctor could...

1) Detect existence of CocoaPods (Mac only of course)
2) Detect version of Ruby and warn if the version is known to be incompatible

burkeholland:
Why can't I add a label to this issue? GitHubbing is hard.

teobugslayer:
@burkeholland,

We generally avoid to deal with dependencies of dependencies, because once you start this way, there's no end. CocoaPods site clearly states that `CocoaPods is built with Ruby and is installable with the default Ruby available on OS X. We recommend you use the default ruby.` We cannot guarantee that we or our dependencies work on broken machines.

Otherwise, we check for CocoaPods existence: https://github.com/NativeScript/nativescript-cli/blob/master/lib/services/doctor-service.ts#L49

burkeholland:
Understood. However, let me provide the feedback that as a developer, I would have bailed on this endeavor and would have mentally blamed {N} out of ignorance. I'm convinced that our ability to smooth our the developer experience is paramount. Throwing cryptic errors about `sandbox-pod` and then expecting the developer to figure it out on their own is going to cause abandonment.

Maybe we could provide better error handling in the build where pods are added?

hshristov:
`tns doctor` did not show me that I was missing pod. It throws exception that ANDROID_HOME is not set correctly although I was using `tns emulate ios`

teobugslayer:
I agree that we must check whether CocoaPods works robustly. I disagree that checking for particular version of Ruby brings value, because there can be millions of others issues which prevents CocoaPods from working. Imho, a better approach is to add a dummy pod into a dummy project and see if everything works. But this may not be easily doable.

teobugslayer:
@hshristov, this was fixed in https://github.com/NativeScript/nativescript-cli/commit/4834e800809e11ddc8d7c372f6974528d2ad2c41 and is included in the upcoming CLI 1.5.2 release.

hshristov:
@teobugslayer, Great.

Thanks

dtopuzov:
Now doctor get iOS runtime and build a project to check cocoapods.

PR #1385, opened 2016-01-08
Verify CocoaPods

teobugslayer:
See https://github.com/NativeScript/nativescript-cli/issues/1294

How it works:

- install iOS runtime into a temporary directory
- create a simple Podfile into the template project
- run (sandobox-)pod
- verify nothing breaks

### DIFF
--- a/lib/services/doctor-service.ts
+++ b/lib/services/doctor-service.ts
@@ -10,7 +10,12 @@ class DoctorService implements IDoctorService {
 	constructor(private $androidToolsInfo: IAndroidToolsInfo,
 		private $hostInfo: IHostInfo,
 		private $logger: ILogger,
-		private $sysInfo: ISysInfo) {	}
+		private $progressIndicator: IProgressIndicator,
+		private $sysInfo: ISysInfo,
+		private $childProcess: IChildProcess,
+		private $config: IConfiguration,
+		private $npm: INodePackageManager,
+		private $fs: IFileSystem) {	}
 
 	public printWarnings(): boolean {
 		let result = false;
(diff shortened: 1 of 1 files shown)


rosen-vladimirov:
hm, why should we create new project, add platform and verify `pod install` command after that. Do we want to validate that project with pods can be created or that pods are useful on the machine?
I prefer to create Podfile in a temp dir and call `pod install`. Could this be done or we need some files from the runtime?

teobugslayer:
Well, CocoaPods need a Xcode project to work on and that was the easiest way I could think of creating one.

rosen-vladimirov:
We can print dots while waiting for pod install (we have a method in helpers that does this :) )

teobugslayer:
@rosen-vladimirov, without existing {N} project, CocoaPods complains with `[!] Could not automatically select an Xcode project.`

rosen-vladimirov:
Can we mock it - for example we can add one default proj file in our resources, copy it to temp folder, create Podfile there and call `pod install`
There are many, many reasons why `tns create app1 && tns platform add ios` may fail and it will break doctor command. Do we want this?

lib/services/doctor-service.ts
```diff
@@ -83,5 +97,35 @@ class DoctorService implements IDoctorService {
...
 		}
 	}
+
+	private verifyCocoaPods(): boolean {
+		this.$logger.out("Verifying CocoaPods. This may take more than a minute, please be patient.");
+
+		let temp = require("temp");
+		temp.track();
+		let projDir = temp.mkdirSync("nativescript-check-cocoapods");
+		this.$npm.install("tns-ios", projDir).wait();
```
rosen-vladimirov: maybe we should try-catch this one as well. As we know - npm fails sometimes and in this case our doctor command will fail

rosen-vladimirov:
:+1:

PR #1394, opened 2016-01-14
wait for the debugger to settle before starting a client and output t…

blagoev:
…hat the debugger has started

### DIFF
--- a/lib/services/android-debug-service.ts
+++ b/lib/services/android-debug-service.ts
@@ -3,6 +3,7 @@
 import * as path from "path";
 import * as net from "net";
 import Future = require("fibers/future");
+import { sleep } from "../common/helpers";
 
 class AndroidDebugService implements IDebugService {
 	private static DEFAULT_NODE_INSPECTOR_URL = "http://127.0.0.1:8080/debug";
@@ -21,7 +22,7 @@ class AndroidDebugService implements IDebugService {
 		private $opener: IOpener,
 		private $config: IConfiguration) { }
 
-	private get platform() { return "android"; }
+	public get platform() { return "android"; }
 
 	private get device(): Mobile.IAndroidDevice {
 		return this._device;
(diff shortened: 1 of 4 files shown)


blagoev:
This removes race condition debugger/client. Also outputs a message that the debugger has indeed started, which is required by vs code extension to attach correctly

lib/services/android-debug-service.ts
```diff
@@ -186,7 +193,28 @@ class AndroidDebugService implements IDebugService {
 
 			this.device.applicationManager.stopApplication(packageName).wait();
 			this.device.applicationManager.startApplication(packageName).wait();
-
+			
+
```
rosen-vladimirov: multiple blank lines - this breaks the linter

lib/services/android-debug-service.ts
```diff
@@ -186,7 +193,28 @@ class AndroidDebugService implements IDebugService {
...
+				let forwardsResult = this.device.adb.executeShellCommand([`ls -s /data/local/tmp/${packageName}-debugbreak`]).wait();
+				maxWait--;
+				debugerStarted = forwardsResult.indexOf(waitText) === -1;
+				if (!debugerStarted) {
+					this.sleep(500);
+				}	
+			}
+			
+			if (debugerStarted) {
+				console.log("# NativeScript Debugger started #");
```
rosen-vladimirov: is there a particular reason not to use `this.$logger.info`

lib/services/android-debug-service.ts
```diff
@@ -186,7 +193,28 @@ class AndroidDebugService implements IDebugService {
...
+			}
+			else {
+				console.log("# NativeScript Debugger did not start in time #");
```
rosen-vladimirov: same as above

lib/services/android-debug-service.ts
```diff
@@ -177,6 +178,12 @@ class AndroidDebugService implements IDebugService {
 			this.$devicesService.execute(action).wait();
 		}).future<void>()();
 	}
+	
+	private sleep(ms: number): void {
+		let fiber = Fiber.current;
```
rosen-vladimirov: hm, is there a way to get here without fiber? In this case Fiber.current will be undefined

lib/services/android-debug-service.ts
blagoev: I don't want to. This console.log is intentionally used so the vs code extension should be able to pars it not matter what. this.logger may be filtered now or in the future.

lib/services/android-debug-service.ts
```diff
@@ -187,6 +194,25 @@ class AndroidDebugService implements IDebugService {
...
+				debugerStarted = forwardsResult.indexOf(waitText) === -1;
+				if (!debugerStarted) {
+					this.sleep(500);
+				}
+			}
+
+			if (debugerStarted) {
+				console.log("# NativeScript Debugger started #");
+			}
+			else {
```
rosen-vladimirov: this else should be on the line above :)

lib/services/android-debug-service.ts
teobugslayer: We use `$logger` so that we can write unit tests, please use it to aid us when we later add tests.

If the text printed with `console.log` is designed to be consumed by a tool (VS Code) and not a human user, I suggest to hide it behind a command-line flag so that it does not complicate the user's work flow.

Filtering out $logger in production code (not unit tests) would completely break the CLI and I do not see how could this happen.

lib/services/android-debug-service.ts
```diff
@@ -178,6 +179,12 @@ class AndroidDebugService implements IDebugService {
 		}).future<void>()();
 	}
 
+	private sleep(ms: number): void {
```
teobugslayer: You can use the existing function:

```
import { sleep } from "../common.helpers";
```

lib/services/android-debug-service.ts
blagoev: lets not over engineer a simple console output, exactly how is "NativeScritp Debugger started" message complicating user's workflow.

otherwise updated the code as per your suggestions

Issue #1362, opened 2015-12-22
Crash when launching applications on iPhone 6 in debug mode


jasssonpet:
_From @manijak on December 6, 2015 21:38_

I just updated my iPhone 6 to iOS 9.1, and I am having issues in running the app from device.

If I use `tns deploy ios` or `tns run ios` or `tns debug ios` I get the app deployed successfully message, but the app crashes as soon as it starts. No error or exception in the terminal... 

If I deploy the app directly from xcode 7.1.1, everything runs fine. App does not crash, no errors. 

I managed to find the crashlog on the device, here is the exception part, but it doesn't say much to me. 
```
Date/Time:           2015-12-06 22:04:43.43 +0100
Launch Time:         2015-12-06 22:04:42.42 +0100
OS Version:          iOS 9.1 (13B143)
Report Version:      105

Exception Type:  EXC_BAD_ACCESS (SIGBUS)
Exception Subtype: KERN_PROTECTION_FAILURE at 0x0000000015d0e070
Triggered by Thread:  0

Filtered syslog:
None found

Thread 0 name:  Dispatch queue: com.apple.main-thread
Thread 0 Crashed:
0   ???                           	0x15d0e070 0 + 366010480
1   CoreFoundation                	0x2759096b 0x274e7000 + 694635
2   CoreFoundation                	0x27590741 0x274e7000 + 694081
3   CoreFoundation                	0x275e4ab3 0x274e7000 + 1039027
4   CoreFoundation                	0x274f1f3b 0x274e7000 + 44859
5   Foundation                    	0x282dce27 0x282d7000 + 24103
6   UIKit                         	0x2b9ac579 0x2b6fd000 + 2815353
7   UIKit                         	0x2b9b0285 0x2b6fd000 + 2830981
8   UIKit                         	0x2b9c3b55 0x2b6fd000 + 2911061
9   UIKit                         	0x2b9ad7a7 0x2b6fd000 + 2820007
```

_Copied from original issue: NativeScript/ios-runtime#431_

jasssonpet:
_From @fealebenpae on December 6, 2015 21:39_

Hi @manijak,

Please try to find the device log as well (`tns device log`).

jasssonpet:
_From @manijak on December 7, 2015 7:13_

Here you go.
[Device log](https://dl.dropboxusercontent.com/u/33168380/iOS_Console_iPhone-Ned_2015-12-0708.09.44.log)

[...]

PR #1396, opened 2016-01-15
Fix crash when launching applications on iOS device in debug mode

Fatme:
Fixes https://github.com/NativeScript/nativescript-cli/issues/1362


teobugslayer:
:+1:

Issue #1377, opened 2016-01-06
Do not override the forward but create new one when debugger port is taken 


blagoev:
The tns android debug should create a new override for the debugged package if the port is already taken. It means that some other package has been debuged on that device recently or is currently being debugged.
Incrementally increase the ports from 40001 until one free is found.

Do not create a new forward if the forward exists for the package

PR #1402, opened 2016-01-18
Fix debugging of two apps at the same time

teobugslayer:
See https://github.com/NativeScript/nativescript-cli/issues/1377


ErjanGavalji:
:+1:

Issue #1026, opened 2015-10-08
Out of the box support for application id in plugins


Fatme:
We have a plugin that needs the app's id.

Currently its manifest contains something like:
``` XML
<Manifest>
	<Service id="SERVICE_NAME" />
</Manifest>
```
Registering an intent service exposed through JavaScript API.

If the service name is provided as plug-in variable, the app package.json will look like:
``` JSON
{
	"nativescript": {
		"id": "com.nativescript.MyPlayground",
		"UploadPlugin": {
			"SERVICE_NAME": "com.nativescript.MyPlayground.UploadPluginService"
		}
	}
}
```

Currently the plug-in is so designed that the service name is always expected to be the app id suffixed with the plugin name. It would be nice if we could simply use an expression like `{{nativescript.id}}.UploadPluginService` in the plug-in's manifest.

PR #1408, opened 2016-01-19
Out of the box support for application id in plugins

Fatme:
Implements https://github.com/NativeScript/nativescript-cli/issues/1026
Removes the unneeded logic that merges configuration files


test/plugins-service.ts
```diff
@@ -547,94 +545,5 @@ describe("Plugins service", () => {
 			mockBeginCommand(testInjector, expectedErrorMessage);
 			pluginsService.prepare(pluginJsonData, "android").wait();
 		});
-		it("merges AndroidManifest.xml and produces correct xml", () => {
```
Fatme: Maybe it is better this test to be rewritten for ios platform

teobugslayer:
This change should be mentioned in the docs. Consider creating a PR against the docs repo,

rosen-vladimirov:
:+1:

Issue #1403, opened 2016-01-19
tns deploy ios --release command prints incorrect message.


Fatme:
The printed message is `When producing a release build, you need to specify all --key-store-* options.`. The --key-store options are specific only for android platform.

PR #1410, opened 2016-01-20
Don't print incorrect message when `tns deploy ios --release` command is executed

Fatme:
FIxes https://github.com/NativeScript/nativescript-cli/issues/1403


teobugslayer:
:+1:

Issue #1412, opened 2016-01-20
Store settings in the Roaming Windows profile


teobugslayer:
Currently the settings are lost when a user roams among machines in the same Windows domain.

PR #1413, opened 2016-01-20
Move settings into Roaming Windows profile

teobugslayer:
See https://github.com/NativeScript/nativescript-cli/issues/1412

 - Change the default settings dir to be inside Roaming Windows profile
 - Move the existing file to its new location, if applicable
 - Remove the source comment which indicates that separate {N} provacy policy is needed - it is not


rosen-vladimirov:
:+1:

PR #1415, opened 2016-01-20
Set this._isInitialized after preparePlatform in order before-prepare hooks to work properly

Fatme:
…


rosen-vladimirov:
:dancer: :+1:

PR #1423, opened 2016-01-22
Separate disableCommandHooks from DISABLE_HOOKS

Mitko-Kerezov:
The latter is a mechanism to disable all hooks, while the former only disables hooks not invoked through method decoration.
Ping @rosen-vladimirov


rosen-vladimirov:
:+1:  after green build

Fatme:
:+1:

Issue #1329, opened 2015-12-09
Unpaired but connected Android device does not return any devices info on tns device


nikolay-diyanov:
If you connect an Android device to a Mac, but you do no tap the Authorize button, adb devices returns an unauthorized device, while tns device returns a record for the device in the table, but there is no device identifier, nor you are told that the device is unauthorized. It will be better, if we return that info.

Here is the log of that situation:

execFile: /usr/local/Cellar/android-sdk/24.3.2/platform-tools/adb "help"
spawn: /usr/local/Cellar/android-sdk/24.3.2/tools/android "--help"

       Usage:
       android [global options] action [action options]
       Global options:
  -h --help       : Help on a specific command.
  -v --verbose    : Verbose mode, shows errors, warnings and all messages.
     --clear-cache: Clear the SDK Manager repository manifest cache.
  -s --silent     : Silent mode, shows errors only.

                                                                    Valid
                                                                    actions
                                                                    are
                                                                    composed
                                                                    of a verb
                                                                    and an
                                                                    optional
                                                                    direct
                                                                    object:
-    sdk              : Displays the SDK Manager window.
-    avd              : Displays the AVD Manager window.
-   list              : Lists existing targets or virtual devices.
-   list avd          : Lists existing Android Virtual Devices.
-   list target       : Lists existing targets.
-   list device       : Lists existing devices.
-   list sdk          : Lists remote SDK repository.
- create avd          : Creates a new Android Virtual Device.
-   move avd          : Moves or renames an Android Virtual Device.
- delete avd          : Deletes an Android Virtual Device.
- update avd          : Updates an Android Virtual Device to match the folders
                        of a new SDK.
- create project      : Creates a new Android project.
- update project      : Updates an Android project (must already have an
                        AndroidManifest.xml).
- create test-project : Creates a new Android project for a test package.
- update test-project : Updates the Android project for a test package (must
                        already have an AndroidManifest.xml).
- create lib-project  : Creates a new Android library project.
- update lib-project  : Updates an Android library project (must already have
                        an AndroidManifest.xml).
- create uitest-project: Creates a new UI test project.
- update adb          : Updates adb to support the USB devices declared in the
                        SDK add-ons.
- update sdk          : Updates the SDK by suggesting new platforms to install
                        if available.

Exec uname -a 
[...]

PR #1426, opened 2016-01-26
Show device status when listing available devices

rosen-vladimirov:
Add new column - Status.
The values in it will be "Connected" when the device is reachable and "Unreachable" in case it is locked, not trusted, etc.

Implements https://github.com/NativeScript/nativescript-cli/issues/1329


teobugslayer:
:+1:

rosen-vladimirov:
run ci

vchimev:
run ci

Issue #1429, opened 2016-01-27
Setup script issues


vchimev:
- [x] the script installs *nodejs* version 5.1.0. It would be the best to install *nodejs* LTS version.
- [x] the script fails to install android build tools.
Output: `Error: Ignoring unknown package filter 'build-tools-23.0.2'`
- [x] the command `npm install -g nativescript` fails due to *npm* not found.

vchimev:
https://gist.github.com/vchimev/e90a0f9a3ab1f7aa54a4

vchimev:
![nodejs](https://cloud.githubusercontent.com/assets/12251337/12647634/c80cb0da-c5de-11e5-933e-e62767dbba83.png)

vchimev:
If I have already installed jdk and android sdk they are not detected. The script install them again.

Log: https://gist.github.com/vchimev/c3381a754fb059596e1c

Screenshot:
![jdk](https://cloud.githubusercontent.com/assets/12251337/12648956/1595cb10-c5e5-11e5-8969-bffc60d329b2.png)

vchimev:
The NPM warnings of the NativeScript CLI installation are colored in red, they can be mistaken for errors and make harder to see the real errors if any.

![tns](https://cloud.githubusercontent.com/assets/12251337/12649292/a5ba2d7a-c5e6-11e5-93e3-1ca78f3c05e9.png)

teobugslayer:
About the red text, it is beyond my control.
About failing re-installations, I'll fix them.
About the first reported fail, I cannot reproduce it. I'll try couple more times.

dtopuzov:
The script does not install "Local Maven repository for Support Libraries" which is required by tns (and doctor report it)

enchev:
Already done.

PR #1432, opened 2016-01-27
Install node.js 4.2.5 - most stable at the moment

teobugslayer:
Fix Android SDK installation
Fix nativescript cli installation

See https://github.com/NativeScript/nativescript-cli/issues/1429


setup/native-script.ps1
```diff
@@ -34,6 +37,12 @@ if (!$env:JAVA_HOME) {
 
 # install NativeScript CLI
 write-host -BackgroundColor Black -ForegroundColor Yellow "Installing NativeScript CLI"
-npm install -g nativescript
+
+$oldPathUser = [Environment]::GetEnvironmentVariable("PATH", "User")
+$pathMachine = [Environment]::GetEnvironmentVariable("PATH", "Machine")
+$myPath = [Environment]::GetEnvironmentVariable("PATH")
+[Environment]::SetEnvironmentVariable("PATH", "$myPath;$oldPathUser;$pathMachine;$env:ProgramFiles\nodejs")
```
rosen-vladimirov: I've not tested it, but according to [this forum](http://www.vistax64.com/powershell/239765-how-set-systems-environment-variable-permanently.html#post1101062) the PATH will not be "set" in the current Shell until you restart:
```
To set it PERMANENTLY you use SetEnvironmentVariable, but to have it
take effect NOW you need to use the Env: drive.
```

rosen-vladimirov: According to MSDN the value will not be set until PowerShell is restarted:
```
Or at least it didn’t show up until we restarted PowerShell. (Or started up a new instance of PowerShell.)
```
Maybe we have to use full path to npm (from node installation). 
[source](https://technet.microsoft.com/en-us/library/ff730964)

teobugslayer: I am modifying the environment of the *currently running* PowerShell process, and not the User or System ones. So the forum post and the MSDN article are correct, but not applicable to the code. I can use the shorter syntax though, I did not know it.

We cannot use the full path to npm. This was my first approach and it failed - something (npm, dependencies, or our code) executes `node` without full path and the installation of the CLI fails because it cannot find node.exe.

rosen-vladimirov: Ok, thanks for the information :)

rosen-vladimirov:
:+1:

PR #1441, opened 2016-02-02
Fix exception tracking message

rosen-vladimirov:
Remove the message of the exception as it is already printed in most of the cases.


rosen-vladimirov:
run ci

PR #1442, opened 2016-02-02
Use Debugger UI from a dedicated npm package

teobugslayer:
See https://github.com/nativescript/ios-runtime/issues/474

### DIFF
--- a/lib/services/ios-debug-service.ts
+++ b/lib/services/ios-debug-service.ts
@@ -2,17 +2,19 @@
 "use strict";
 
 import * as iOSDevice from "../common/mobile/ios/device/ios-device";
-
 import * as net from "net";
 import * as path from "path";
 import * as semver from "semver";
 import byline = require("byline");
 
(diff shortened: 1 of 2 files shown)


lib/services/ios-debug-service.ts
```diff
@@ -2,17 +2,19 @@
...
 import * as iOSDevice from "../common/mobile/ios/device/ios-device";
-
 import * as net from "net";
 import * as path from "path";
 import * as semver from "semver";
 import byline = require("byline");
 
-let InspectorBackendPort = 18181;
+const InspectorBackendPort = 18181;
+const inspectorAppName = "NativeScript Inspector.app";
```
rosen-vladimirov: why the first constant name `InspectorBackendPort` starts with capital letter and the others do not

teobugslayer: Because it was old code and I missed it. Will fix it.

rosen-vladimirov:
:+1:

Issue #1332, opened 2015-12-10
Application entry point file not found... when using livesync (android)


x4080:
TNS 1.5, my steps :
1. tns run android
2. multi tasking button, close application (to force cold boot)
3. cold boot, works ok
4. tns livesync android
5. multi tasking button, close application (to force cold boot)
6. Above error (...bootstrap js etc)

I think in 1.4 it works ok

Is it only me ? Thanks

EDIT : in 4.4.4 (nexus 7), but in 6.0 (nexus 5) no error

x4080:
Found the problem :

@app.js, application.cssFile = "app.css"; // was using Nathanael Livesync plugin

should be application.cssFile = "./app.css";

x4080:
I tried using tns build android and the error come back on android 4.4.4, I check the package.json in the app directory and it is already correct :

{
    "name": "tns-template-blank-ts",
    "main": "app.js",
    "version": "1.2.0",
    "author": "Telerik <support@telerik.com>",
    "description": "Nativescript blank-ts project template",
    "license": "BSD",
    "keywords": [
        "telerik",
        "mobile",
        "nativescript",
        "{N}",
        "tns",
        "appbuilder",
        "template"
    ],
    "repository": {
        "url": "https://github.com/NativeScript/NativeScript/commit/06ddbf133a1c022a9ec61dc442d7f7ce0841c915"
    }
}

rosen-vladimirov:
Hi @x4080 ,
Please excuse us for somehow missing this case. Do you still have the problem? Can you send your application that reproduces the issue?

x4080:
Yes sir, tns livesync android on 4.4, nexus 7 and genymotion.. Everything good for lollipop and marshmallow.

Strangely the work around is using --watch :)

x4080:
Oh forgot, it happened when app open then we press the home button, then app error happened

rosen-vladimirov:
Ok, I've reproduced the issue. Here's the description of the problem (writing it down so I'm sure I'll not forget it):
When using `tns livesync android` CLI moves all files inside fullsync directory (in `/data/loca/tmp/<app_id>`) and changes permissions of all files to 777 so the runtime will be able to work with them.
[...]

PR #1443, opened 2016-02-02
Fix livesync on Android devices and Genymotion emulator

rosen-vladimirov:
The following fix has been applied in common lib:

On some Android devices the `android-runtime` cannot delete the files in `/data/local/tmp/<app id>/fullsync` directory.
The next time when the application is started, runtime reads the fullsync directory and replaces the real application with content of fullsync dir.
So the app does not start as it is not full app.
Fix two issues - first one is missing `chmod` command in the file that we are generating.
As we cannot spawn chmod of all files directly, due to some command line limitations, we are writing the command in a file, transfer the file on the device and execute it.
BUT, we've lost the command in this commit https://github.com/telerik/mobile-cli-lib/commit/6d52c6ab9b51aaac7b5fae7e4412c0f58737fcf1
In fact our file contains only list of files and when trying to execute it on the device, we receive error, but we have added parameter to `spawnFromEvent` not to throw. So the error is hidden.
Add the missing chmod, so the command can be executed successfully.

However this fixes the issues for some files, but when we have directories, we do not write them in the file, so android runtime does not have permissions to delete them and they remain in fullsync directory.
Fix this by including directories in the list of application files (not sure if this will not break something else).

Fixes https://github.com/NativeScript/nativescript-cli/issues/1332


teobugslayer:
:+1:

PR #1444, opened 2016-02-02
Update .npmignore file

rosen-vladimirov:
Update the file with new directories and files to exclude.

I'm not sure about this file `.travis.yml`


teobugslayer:
we control the travis ci builds with travis.tml, it definitely must be npm ignored.

:+1:

PR #1445, opened 2016-02-03
Log socket path on running 'tns debug ios'

ivanbuhov:
This change is necessary for NativeScript VSCode plugin.

### DIFF
--- a/lib/device-sockets/ios/socket-proxy-factory.ts
+++ b/lib/device-sockets/ios/socket-proxy-factory.ts
@@ -11,7 +11,8 @@ import * as helpers from "../../common/helpers";
 export class SocketProxyFactory implements ISocketProxyFactory {
 	constructor(private $logger: ILogger,
 		private $projectData: IProjectData,
-		private $projectDataService: IProjectDataService) { }
+		private $projectDataService: IProjectDataService,
+		private $options: IOptions) { }
 
 	public createSocketProxy(factory: () => net.Socket): IFuture<any> {
 		return (() => {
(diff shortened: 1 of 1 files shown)


ivanbuhov:
ping @rosen-vladimirov

lib/device-sockets/ios/socket-proxy-factory.ts
```diff
@@ -119,6 +119,7 @@ export class SocketProxyFactory implements ISocketProxyFactory {
 
 		let socketFileLocation = temp.path({ suffix: ".sock" });
 		server.listen(socketFileLocation);
+		this.$logger.info("socket-file-location: " + socketFileLocation);
```
rosen-vladimirov: as far as I understand, we need this only when `this.$options.client` is false (`--no-client` is passed)
When started from CLI, the --no-client will not be passed and users should not see this row?

If my understanding is correct, can you add the if, something like:
```TypeScript
if(!this.$options.client)  {
    this.$logger.info("socket-file-location: " + socketFileLocation);
}
```
and in the constructor:
```TypeScript
constructor(private $logger: ILogger,
 		private $projectData: IProjectData,
 		private $options: IOptions) { }
```

ivanbuhov:
You are right, the extension always runs `tns debug` with `--no-client`. I will fix it ;)

lib/device-sockets/ios/socket-proxy-factory.ts
ivanbuhov: You are right, the extension always runs tns debug with --no-client. I will fix it ;)

rosen-vladimirov:
:+1:  after green build

PR #1449, opened 2016-02-05
Exclude some dirs in tsconfig.json

rosen-vladimirov:
In `tsconfig.json` we have to exclude `node_modules`, `coverage` and `scratch` dirs.
When they are huge (in my case scratch dir was 1.5 GB), TypeScript is not working at all in VS Code.


Mitko-Kerezov:
:+1:

Issue #1027, opened 2015-10-08
Prepare android/ios command prepares both platforms when added


vchimev:
Log:
```
VCHIMEVMAC:release vchimev$ node_modules/.bin/tns build ios --path appTest
The iOS Deployment Target is now 8.0 in order to support Cocoa Touch Frameworks in CocoaPods.
Successfully prepared plugin nativescript-iqkeyboardmanager for ios.
nativescript-iqkeyboardmanager is not supported for android.
Successfully prepared plugin nativescript-social-share for ios.
Creating project scheme...
Installing pods...
Updating local specs repositories
...
...
...
VCHIMEVMAC:release vchimev$ node_modules/.bin/tns build android --path appTest
The iOS Deployment Target is now 8.0 in order to support Cocoa Touch Frameworks in CocoaPods.
Successfully prepared plugin nativescript-iqkeyboardmanager for ios.
nativescript-iqkeyboardmanager is not supported for android.
Successfully prepared plugin nativescript-social-share for ios.
Successfully prepared plugin nativescript-social-share for android.

```

Fatme:
Steps to reproduce:

1. Add ios and android platforms.
2. Open `package.json` file and add 
```
"email-validator": "^1.0.2",
"nativescript-iqkeyboardmanager": "^1.0.0",
"nativescript-social-share": "^1.1.0"
```  
as dependencies
3. Prepare the project for android platform.

dtopuzov:
Still not fixed. Tested with Fatme's steps and here is the result:

node_modules/.bin/tns prepare android --path TNS_App
Copying template files...
Project successfully created.
nativescript-iqkeyboardmanager is not supported for android.
Successfully prepared plugin nativescript-social-share for android.
Installing pods...
Updating local specs repositories
Analyzing dependencies
Downloading dependencies
Using IQKeyboardManager (3.3.4)
Generating Pods project
Integrating client project
Sending stats
Sending stats
Pod installation complete! There is 1 dependency from the Podfile and 1 total pod installed.
Project successfully prepared

We still do some work for nativescript-iqkeyboardmanager even if does not target android.

kvelikov:
Hi guys, 
This issue can be blocking when building on windows. If a plugin fails to prepare the CLI won't prepare if there are other plugins and the project won't run.
I'm getting the following error:
Processing node_modules failed. Error:Error: Command failed: C:\WINDOWS\system32
[...]

PR #1453, opened 2016-02-05
Call afterPrepareAllPlugins for the specific platform only

Mitko-Kerezov:
Instead of calling `afterPrepareAllPlugins` for all installed platforms.
Fixes https://github.com/NativeScript/nativescript-cli/issues/1027
Ping @rosen-vladimirov and @teobugslayer for a quick review.


teobugslayer:
:+1:

PR #1456, opened 2016-02-07
Speed-up grunt lint

rosen-vladimirov:
Currently `grunt lint` might take more than 2-3 minutes. The problem is in the src files passed.
In my case the scratch directory was around 2.5 GB and this was slowing down the linting, as we have included `!*.d.ts` files.
Make sure only lib and test directories are used when linting as our source code and tests are there.

Fixed mixed tabs and spaces in Gruntfile

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -59,15 +59,15 @@ module.exports = function(grunt) {
 		},
 
 		tslint: {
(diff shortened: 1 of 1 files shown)


teobugslayer:
:+1:

PR #1457, opened 2016-02-08
Revert parallel execution of action on multiple devices

rosen-vladimirov:
Revert change applied in this commit - https://github.com/telerik/mobile-cli-lib/commit/065512bd01ccae6dfd4648244c0ee21bcf5abf19#diff-3afe2bc0eef11fb206a6cfb650f4057fL213
The idea was to speed up actions executed on multiple devices. However in our current code is not able to work this way.
For example `tns run android` is trying to execute multiple simultaneous builds and it fails. Same will happen with livesync in some cases.
The problem is with the action that's passed for each device. Until we fix the other code, let's keep the old behavior.


teobugslayer:
:+1:

Issue #1455, opened 2016-02-06
CLI not supporting a project path having a spaces


nassimus26:
The `-c otool -Vh /Volumes/VMware Shared Folders/dev/nativescript/gregories/` should be 
`-c otool -Vh /Volumes/VMware\ Shared\ Folders/dev/nativescript/gregories/`

```
NMac:gregories nassimus$ tns livesync ios --emulator
nativescript-floatingactionbutton is not supported for ios.
The iOS Deployment Target is now 8.0 in order to support Cocoa Touch Frameworks in CocoaPods.
Successfully prepared plugin nativescript-iqkeyboardmanager for ios.
Successfully prepared plugin nativescript-social-share for ios.
Processing node_modules failed. Error: Command failed: /bin/sh -c otool -Vh /Volumes/VMware Shared Folders/dev/nativescript/gregories/node_modules/nativescript-telerik-ui/platforms/ios/TelerikUI.framework/TelerikUI
error: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/otool: can't open file: /Volumes/VMware (No such file or directory)
error: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/otool: can't open file: Shared (No such file or directory)
error: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/otool: can't open file: Folders/dev/nativescript/gregories/node_modules/nativescript-telerik-ui/platforms/ios/TelerikUI.framework/TelerikUI (No such file or directory)

```

nassimus26:
projectFilePath = projectFilePath.replace(/ /g, '\\ '); must be added to project-helper.js

nassimus26:
this fix my issue
```
if (this.$fs.exists(projectFilePath).wait() && this.isProjectFileCorrect(projectFilePath)) {
		    ->>> projectDir = projectDir.replace(/ /g, "\ ");
                    this.$logger.debug("Project directory is '%s'.", projectDir);
                    this.cachedProjectDir = projectDir;
                    break;
                }
```

teobugslayer:
I prefer quoting the strings where needed. At any rate,  we will fix it. Thank you!

teobugslayer:
@vchimev,  can we add integration tests with projects with spaces in key places?

nassimus26:
[...]

PR #1458, opened 2016-02-08
Use spawn wherever possible

Mitko-Kerezov:
Wherever possible use `spawn` instead of `exec`.
This minifies the chance of something going wrong wherever there are spaces, quotation marks and whatnots in the project's name.
Addresses #1455
Ping @rosen-vladimirov @teobugslayer

### DIFF
--- a/lib/services/ios-project-service.ts
+++ b/lib/services/ios-project-service.ts
@@ -268,7 +268,7 @@ export class IOSProjectService extends projectServiceBaseLib.PlatformProjectServ
 			let project = this.createPbxProj();
 			let frameworkName = path.basename(frameworkPath, path.extname(frameworkPath));
 			let frameworkBinaryPath = path.join(frameworkPath, frameworkName);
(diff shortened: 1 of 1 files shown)


lib/services/ios-project-service.ts
```diff
@@ -268,7 +268,7 @@ export class IOSProjectService extends projectServiceBaseLib.PlatformProjectServ
 			let project = this.createPbxProj();
 			let frameworkName = path.basename(frameworkPath, path.extname(frameworkPath));
 			let frameworkBinaryPath = path.join(frameworkPath, frameworkName);
-			let isDynamic = _.contains(this.$childProcess.exec(`otool -Vh ${frameworkBinaryPath}`).wait(), " DYLIB ");
+			let isDynamic = _.contains(this.$childProcess.exec(`otool -Vh "${frameworkBinaryPath}"`).wait(), " DYLIB ");
```
rosen-vladimirov: what if it already has `"`

rosen-vladimirov:
:+1:  great

PR #1459, opened 2016-02-08
More setup fixes for Windows

teobugslayer:
- No longer uses BoxStarter, instead it is a plain PowerShell script
 - Force re-installation of dependencies
 - Can self-elevate when needed but in this case only PRODUCTION branch can be used

See https://github.com/NativeScript/nativescript-cli/issues/1429

### DIFF
--- a/setup/native-script.ps1
+++ b/setup/native-script.ps1
@@ -1,23 +1,33 @@
 
-# A Boxstarter script to set up Windows machine for NativeScript development
-# To run it against RELEASE branch (recommended) use
-# http://boxstarter.org/package/nr/url?https://raw.githubusercontent.com/NativeScript/nativescript-cli/release/setup/native-script.ps1
-# To run it against MASTER branch (usually only developers of NativeScript need to) use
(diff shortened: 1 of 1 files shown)


rosen-vladimirov:
I've tried the script, and I've received two errors:

### When installing JDK:
Installing the following packages:
jdk8
By installing you accept licenses for the packages.
jdk8 not installed. The package was not found with the source(s) listed.
 If you specified a particular version and are receiving this message, it is pos
sible that the package name exists but the version does not.
 Version: ""
 Source(s): "https://chocolatey.org/api/v2/"

Chocolatey installed 0/1 package(s). 1 package(s) failed.
 See the log for details (C:\ProgramData\chocolatey\logs\chocolatey.log).
Failures:
 - jdk8
Installing Android SDK

### Final pause statement:
The term 'pause' is not recognized as the name of a cmdlet, function, script fi
le, or operable program. Check the spelling of the name, or if a path was inclu
ded, verify that the path is correct and try again.
At line:61 char:6
+ pause <<<<
    + CategoryInfo          : ObjectNotFound: (pause:String) [], CommandNotFou
   ndException
    + FullyQualifiedErrorId : CommandNotFoundException

rosen-vladimirov:
:+1:

PR #1464, opened 2016-02-09
Fix detection of changed shasums

rosen-vladimirov:
Currently we are validating only the shasum of the files in order to detect which one are changed and sync only them.
However this is not correct as the file may be on a different location, but without changing its size.
Also in case you copy some of your files, their shasums might be the same as the original file.
In case the original file is already synced, its hash size will be found in the "oldShasums" and the new file will not be copied.
Validate by using both path and hash.

Fix unit tests and add some new ones.


rosen-vladimirov:
run ci

vchimev:
run ci

Issue #1466, opened 2016-02-10
(Android) Fail to run app


dtopuzov:
Steps to reproduce:
tns create tns-app --copy-from template-hello-world (latest template from master)
tns platform add android --framework-path tns-android.tgz (latest from master) --path tns-app
tns run android --path tns-app --log trace

Result:
```
Result when throw error is false:
{ stdout: '',
  stderr: '1348 KB/s (41485 bytes in 0.030s)\r\n',
  exitCode: 0 }
spawn: D:\adt-bundle\sdk\platform-tools\adb "-s" "emulator-5554" "shell" "pm" "uninstall" "org.nativescript.tnsapp"
Result when throw error is false:
{ stdout: 'WARNING: linker: libdvm.so has text relocations. This is wasting memory and is a security risk. Please fix.\r\r\nFailure\r\r\n',
  stderr: '',
  exitCode: 0 }
spawn: D:\adt-bundle\sdk\platform-tools\adb "-s" "emulator-5554" "install" "-r" "c:\Git\nativescript-cli-tests\tns-app\platforms\android\build\outputs\apk\tnsapp-debug.apk"
Result when throw error is false:
{ stdout: 'WARNING: linker: libdvm.so has text relocations. This is wasting memory and is a security risk. Please fix.\r\r\n\tpkg: /data/local/tmp/tnsapp-debug.apk\r\r\nFailure [INSTALL_FAILED_INVALID_APK]\r\r\n',
  stderr: '1871 KB/s (12461263 bytes in 6.501s)\r\n',
  exitCode: 0 }
Successfully deployed on device with identifier 'emulator-5554'.
spawn: D:\adt-bundle\sdk\platform-tools\adb "-s" "emulator-5554" "shell" "am" "start" "-a" "android.intent.action.MAIN" "-n" "org.nativescript.tnsapp/com.tns.NativeScriptActivity" "-c" "android.intent.category.LAUNCHER"
Result when throw error is false:
{ stdout: 'WARNING: linker: libdvm.so has text relocations. This is wasting memory and is a security risk. Please fix.\r\r\nStarting: Intent { act=android.intent.action.MAIN cat=[android.intent.category.LAUNCHER] cmp=org.nativescript.tnsapp/com.tns.N
ativeScriptActivity }\r\r\nError type 3\r\r\nError: Activity class {org.nativescript.tnsapp/com.tns.NativeScriptActivity} does not exist.\r\r\n',
  stderr: '',
[...]

PR #1467, opened 2016-02-10
Interpolate data in config files from App_Resources on prepare

rosen-vladimirov:
Interpolate data in config files from App_Resources on prepare so `__PACKAGE__` and other placeholders will be replaced.
Ensure the file in the App_Resources is correct in the beginning of the prepare method and make sure all of the placeholders are interpolated after finishing the processing. The interpolating process will replace the placeholders in `platforms/<platform>` dir


rosen-vladimirov:
should fix https://github.com/NativeScript/nativescript-cli/issues/1466

vchimev:
:+1:

PR #1468, opened 2016-02-11
Some improvements for AndroidManifest in App_Resources

rosen-vladimirov:
* Correct the code, so AndroidManifest will not be backed up in case we are unable to find the manifest in the template (current code was always creating the backup and reverting it in case it fails on the extraction from template).
* Fix backup name - use guid in case AndroidManifest.backup exists.


dtopuzov:
:+1:

teobugslayer:
:+1:

PR #1470, opened 2016-02-11
Remove unused dependency

teobugslayer:


rosen-vladimirov:
:+1:

losttime:
Any thoughts on when we'll see a release with this change?

teobugslayer:
the plan is to release later this month.

PR #1472, opened 2016-02-11
Print iOS Simulator device logs

rosen-vladimirov:
When using `tns run ios` and `tns livesync ios` for iOS Simulator, we exit the process immediately instead of printing device logs.


teobugslayer:
:+1:

Issue #1437, opened 2016-02-01
Debugger does not stop at the app initial breakpoint


vchimev:
Tested with {N} CLI 1.6.0-524 and tns-android 1.6.0-2016.1.29.1.

`tns debug android --emulator` starts successfully the node inspector, but it does not stop at the initial breakpoint set in prepareExtend.js. The same command works correctly if `--emulator` is not specified.

bfattori:
Is there a workaround for this problem? Sometimes this works, most times it runs the app and doesn't launch node-inspector for Android.  I have tried:

`tns debug android --debug-brk --emulator`

That has only managed to stop _once_ in the inspector. Every other time it launches my app, then sits forever doing nothing. My login dialog (which should show) isn't... But when the debug worked _that one time_ it showed up when I let the process keep running from node-inspector.

vchimev:
Can you try to leave the emulator running and execute: `tns debug android`? @bfattori

bfattori:
It runs extremely slowly, but it seems to work. What happened was:

  # I launched the emulator
  # Ran `tns debug android` and it started the app, shut it down, then started it again. The first time, the login window showed. Second time it didn't. The inspector came up.
  # Closed my app with the "home" button
  # Re-ran my app from the application drawer. It came up, loaded the dialog, then stopped at the breakpoint I set when i clicked the "Login/Ok" button

I can use this workaround, but hoping you all can figure this out soon.

Thanks,
- Brett

bfattori:
Here's an interesting note...

I created a Chromium Remote Debug config for IntelliJ Idea to run the command `tns debug android --emulator --debug-brk --debug-port 5858` which runs and starts the Node Inspector, as expected. This works every time.

Just an FYI

vchimev:
@rosen-vladimirov,

I gave a try to the 1.6.0 release candidates and here is what I got out of `tns debug android --emulator`:
- on Windows it works as expected,
- on OS X it works the first time executed (cleaned Chrome browsing data), but every next time the initial breakpoint is skipped and the app loads. Also it happened to open the node inspector twice.

`tns debug android` works as expected.

enchev:
We changed the default behavior and debug will not longer stop on first code statement. I'm closing this.

PR #1474, opened 2016-02-12
Fix debug android --emulator

rosen-vladimirov:
Fix `tns debug android --emulator`. There were two issues:
* With latest changes in master/release branch, when there's no android device detected and user executes `tns debug android --emulator` an error is raised - "Cannot find connected devices...". Fix this by returning true in command's canExecute method when `--emulator` is passed.
* When there's no running emulator, CLI tries to start it (android-debug-service, debugOnEmulator method). The emulator is started and after that we try to execute action on the started device.
However devicesService (in this case the androidDeviceDiscovery) had not reported yet the started emulator as a new device. So when device's service execute method is called, it checks if there are devices (still thinks there aren't) and tries to start emulator again.
The final result is that application starts on the device, but there's no debugger attached.
Fix this by forcing android device detection in debugOnEmulator method.

https://github.com/NativeScript/nativescript-cli/issues/1437

### DIFF
--- a/lib/commands/debug.ts
+++ b/lib/commands/debug.ts
@@ -16,6 +16,10 @@ export class DebugPlatformCommand implements ICommand {
 	canExecute(args: string[]): IFuture<boolean> {
 		return ((): boolean => {
 			this.$devicesService.initialize({ platform: this.debugService.platform, deviceId: this.$options.device }).wait();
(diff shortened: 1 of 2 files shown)


teobugslayer:
:+1: nice catch!

PR #1478, opened 2016-02-12
Fix ENOENT, EISDIR, EEXIST errors in livesync

rosen-vladimirov:
During the last refactoring we've lost the queue of futures. So we had some racing issues when the same file is saved multiple times.
Fix this by adding back the queue and adding some try-catch statements.


teobugslayer:
:+1:

PR #1479, opened 2016-02-12
Remove rimraf as dependency

teobugslayer:


rosen-vladimirov:
:+1:  after rebase :)

PR #1487, opened 2016-02-15
Do not install the CLI

teobugslayer:
Focus these scripts on installing only the native platform dependencies.
The documentation will guide the user to install the CLI separately.

See https://github.com/NativeScript/NativeScript/issues/1556


rosen-vladimirov:
:+1:

PR #1494, opened 2016-02-16
Initialize devicesService on tns test

rosen-vladimirov:
We have to initialize devicesService when executing tns test command as currently it fails when calling livesyncBase.


Mitko-Kerezov:
:+1: better late than never

Issue #1499, opened 2016-02-17
Livesync --watch fails for TypeScript projects with sourceMaps enabled


rosen-vladimirov:
In case you have typescript project and you have enabled sourceMaps in your tsconfig.json, `tns livesync <platform> --watch` does not sync files and shows the following warning:
```
Unable to sync files. Error is: ENOENT: no such file or directory, stat 'D:\work\nativescript-cli\scratch\newTs\platforms\android\src\main\assets\app\main-page.js.map'
```
The issue has high impact as sourceMaps are true by default when using our typescript template or when adding nativescript-dev-typescript module (`tns install typescript`).

Steps to reproduce:
* `tns create myApp --template typescript`
* `tns livesync android --watch`
* after livesync is finished, change any .ts file in your app dir.

rosen-vladimirov:
Workaround: 
Open your tsconfig.json and set sourceMaps to false.

PR #1500, opened 2016-02-17
Fix livesync of TypeScript projects

rosen-vladimirov:
Pass .js.map and .ts as excluded patterns for livesync.
Fix livesync of TypeScript projects - make sure gaze does not trigger any livesync logic in case the file that's changed is excluded via any pattern from livesync data.
Currently when .ts file is change, gaze detects changes for .ts, .js and .js.map files and fails as only .js files exist in platforms dir.

Fixes https://github.com/NativeScript/nativescript-cli/issues/1499


dtopuzov:
:+1:

teobugslayer:
:+1:

ignaciofuentes:
:+1:

PR #1502, opened 2016-02-18
Fix livesync on multiple devices when app is not installed

rosen-vladimirov:
When you try to livesync on multiple devices and application is not installed on more than one, calling `tns livesync <platform>` installs the application only on one of them.
There's warning in the output that the application is not installed for each of the devices, but it is installed only on one of them.
The problem is the current logic that is executed for each device - as we want to build only one, in case the app is already built, we've skipped the install part for all devices except the first one.
So we have to build once, but install the built package on all devices where the app does not exist.

Fixes https://github.com/NativeScript/nativescript-cli/issues/1503


Mitko-Kerezov:
:+1:

PR #1525, opened 2016-02-22
Add support for custom platform templates

rosen-vladimirov:
Add new option when adding platform: `--platform-template`. When it is used, CLI will use the specified template instead of the default template from the runtime. The path to the specified template will be saved in project's package.json, so next time when platform is added, the same template will be used.
In case when `--platform-template` is not passed, CLI will check the package.json and if there's value for the template, it will be used. Otherwise the default template from runtime will be used.

Sample usage:
```
$ tns create app1
$ cd app1
$ tns run android --platform-template https://github.com/rosen-vladimirov/android-platform-template/tarball/master
```


rosen-vladimirov:
run ci

rosen-vladimirov:
run ci

lib/services/platform-service.ts
```diff
@@ -137,12 +146,42 @@ export class PlatformService implements IPlatformService {
...
+			if(specifiedTemplate) {
+				let tempDir = temp.mkdirSync("platform-template");
+				try {
+					let installedTemplatePath = this.$npm.install(specifiedTemplate, tempDir, {"ignore-scripts": true, "production": true}).wait()[0][1];
+					return {
+						selectedTemplate: specifiedTemplate,
+						pathToTemplate: installedTemplatePath
+					};
+				} catch(err) {
+					this.$logger.warn(`Unable to install platform template ${specifiedTemplate}. Default template will be used.`);
```
rosen-vladimirov: maybe we should throw correct error here instead of just printing warning

lib/services/android-project-service.ts
```diff
@@ -111,7 +111,13 @@ export class AndroidProjectService extends projectServiceBaseLib.PlatformProject
 			}
 
 			// These files and directories should not be symlinked as CLI is modifying them and we'll change the original values as well.
-			this.copy(this.platformData.projectRoot, frameworkDir, "src", "-R");
+			if(pathToTemplate) {
+				let mainPath = path.join(this.platformData.projectRoot, "src", "main");
+				this.$fs.createDirectory(mainPath).wait();
+				shell.cp("-R", path.join(path.resolve(pathToTemplate), "*"), mainPath);
```
teobugslayer: Do we care that shell.cp can fail? We need to call shell.error to check for failures.

lib/services/ios-project-service.ts
```diff
@@ -104,10 +104,16 @@ export class IOSProjectService extends projectServiceBaseLib.PlatformProjectServ
...
-	public createProject(frameworkDir: string, frameworkVersion: string): IFuture<void> {
+	public createProject(frameworkDir: string, frameworkVersion: string, pathToTemplate?: string): IFuture<void> {
 		return (() => {
 			this.$fs.ensureDirectoryExists(path.join(this.platformData.projectRoot, IOSProjectService.IOS_PROJECT_NAME_PLACEHOLDER)).wait();
-			if(this.$options.symlink) {
+			if(pathToTemplate) {
+				// Copy everything except the template from the runtime
+				this.$fs.readDirectory(frameworkDir).wait()
+					.filter(dirName => dirName.indexOf(IOSProjectService.IOS_PROJECT_NAME_PLACEHOLDER) === -1)
+					.forEach(dirName => shell.cp("-R", path.join(frameworkDir, dirName), this.platformData.projectRoot));
```
teobugslayer: Do we care that shell.cp can fail? We need to call shell.error to check for failures.

lib/services/platform-service.ts
teobugslayer: I second this.

lib/services/platform-service.ts
```diff
@@ -137,12 +146,42 @@ export class PlatformService implements IPlatformService {
...
+				// check the nativescript.tns-<platform>.template value
+				let nativescriptPlatformData = this.$projectDataService.getValue(frameworkPackageName).wait();
+				specifiedTemplate = nativescriptPlatformData && nativescriptPlatformData.template;
+			}
+
+			if(specifiedTemplate) {
+				let tempDir = temp.mkdirSync("platform-template");
+				try {
+					let installedTemplatePath = this.$npm.install(specifiedTemplate, tempDir, {"ignore-scripts": true, "production": true}).wait()[0][1];
+					return {
```
teobugslayer: If you rename installedTemplatePath, you can use the short syntax here. Not a stopper, though.

rosen-vladimirov:
Close and retarget to master

Issue #1513, opened 2016-02-19
`tns test <platform>` does not work in case the application is not installed on the device.


rosen-vladimirov:
In case application is not installed on the device, instead of running the tests, the original app is stared.

### Steps to reproduce
* Create new app - `tns create myApp`
* Navigate to the app - `cd myApp`
* Initialize test structure - `tns test init`
* Run the tests (make sure you do not have org.nativescript.myApp installed on the device): `tns test <platform>` 

### Workaround
Install the app first:
* `tns run <plartform> --justlaunch`

rosen-vladimirov:
Also we **must** prepare the project first and start Karma launcher after that - in case the project is TypeScript, we will not have the `.js` files and the tests will not run. Preparing the project will generate the .js files :)

rosen-vladimirov:
Scenarios that must be tested:
* Application is not installed on device:
  * `tns test <platform>` - should prepare the project, deploy it on device and start the tests
  * `tns test <platform> --watch` should prepare the project, deploy it on device and start the tests. When change is applied, test must be restarted. NOTE: Test must be restarted no matter if .xml, .css or .js is changed.
* Application is installed on device:
  * `tns test <platform>` - prepare the project and run the tests
  * `tns test <platform> --watch` should prepare the project, deploy it on device and start the tests. When change is applied, test must be restarted. NOTE: Test must be restarted no matter if .xml, .css or .js is changed.

* TypeScript application
  * `tns test <platform>` - make sure that you do not have `.js` files in your app folder before testing - this command should transpile the files, prepares the project and runs the tests
  * `tns test <platform> --watch` - make sure changing of `.ts` files does not lead to infinite loop

* Debug:
  * `tns test <platform> --debug-brk` should break and allow you to debug the application
  * `tns test <platform> --debug-brk --watch` should fail - these two flags cannot be used simultaneously.

* Additional tests
  * `tns test <platform>` and after that `tns run <platform>` - first call should run the tests, the second one should run the real application
  * `tns test <platform>` and after that `tns livesync <platform>` - first call should run the tests, the second one should deploy changed files on device and start the REAL application.

PR #1545, opened 2016-02-25
Fix unit test runner

rosen-vladimirov:
Fix the following issues in unit test runner:
* Tests cannot start when application is not installed on device
* Karma does not start when there are no .js files in app dir (initial state of TypeScript projects).
* `tns test <platform> --watch` is not working for TypeScript projects
* `tns test <platform> --debug-brk` does not debug.

### Implementation details
Change the way unit-tests are started. The current code was:
1) Start karma server
2) When start method of `karma-nativescript-launcher` is called, it spawns new CLI process (calls `tns dev-test` command with some arguments).
3) The new CLI process writes down some files in `node_modules/nativescript-unit-test-runner`
4) The new CLI process prepares the project.
5) The new CLI process changes the entry point of the application to point to nativescript-unit-test-runner's main-page.
6) The new CLI process calls livesync-base which should restart the application.
7) In case `--watch` option is used, karma launcher will listen for `file_list_modified` event and spawn new CLI process to run the tests.

Problems were in all the steps. New way:
1) When `tns test <platform>` is called, first step is to prepare the project.
2) Initialize devices service, so we are sure all devices are detected.
3) Prepare livesync data - here we set canExecuteFastSync to false, so any change will restart the application and tests will be started again.
4) Fork new process, which should start karma server.
5) When `karma-nativescript-launcher`'s start method is called in the forked process, it will send required information to current CLI process.
6) CLI process receives the data and writes the required files in `node_modules/nativescript-unit-test-runner`.
7) CLI process calls livesync. In case --debug-brk is specified, debugService is called instead.
8) karma-nativescript-launcher no longer listens for `file-list-modified` event. CLI is alreday doing this (livesync logic).
9) Entry point of application is change by `nativescript-unit-test-runner` via after-prepare hook.

The new behavior depends on the livesync - when app is not installed on the device, it will be installed. In case `--watch` is used, livesync will detect changes, prepare the project and restart the app - this will start the tests again.

### Problem
With this change `tns dev-test` command will stop working. As I consider it not usable, I've decided to skip its fixing for later.

Related PRs:
https://github.com/NativeScript/nativescript-unit-test-runner/pull/2
https://github.com/NativeScript/karma-nativescript-launcher/pull/23

Fixes https://github.com/NativeScript/nativescript-cli/issues/1513

### DIFF
--- a/lib/services/test-execution-service.ts
+++ b/lib/services/test-execution-service.ts
@@ -32,11 +32,17 @@ class TestExecutionService implements ITestExecutionService {
 		private $options: IOptions,
 		private $pluginsService: IPluginsService,
 		private $errors: IErrors,
-		private $devicesService: Mobile.IDevicesService) {
+		private $androidDebugService:IDebugService,
+		private $iOSDebugService: IDebugService,
+		private $devicesService: Mobile.IDevicesService,
+		private $childProcess: IChildProcess) {
 	}
+
+	public platform: string;
 
 	public startTestRunner(platform: string) : IFuture<void> {
 		return (() => {
+			this.platform = platform;
 			this.$options.justlaunch = true;
 			let blockingOperationFuture = new Future<void>();
 			process.on('message', (launcherConfig: any) => {
@@ -50,7 +56,7 @@ class TestExecutionService implements ITestExecutionService {
 						let configOptions: IKarmaConfigOptions = JSON.parse(launcherConfig);
 						this.$options.debugBrk = configOptions.debugBrk;
 						this.$options.debugTransport = configOptions.debugTransport;
-						let configJs = this.generateConfig(configOptions);
+						let configJs = this.generateConfig(this.$options.port.toString(), configOptions);
 						this.$fs.writeFile(path.join(projectDir, TestExecutionService.CONFIG_FILE_NAME), configJs).wait();
 
 						let socketIoJsUrl = `http://localhost:${this.$options.port}/socket.io/socket.io.js`;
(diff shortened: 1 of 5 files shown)


rosen-vladimirov:
ping @ligaz , @teobugslayer , @hdeshev , @Mitko-Kerezov  - I've retarget the PR to release branch, so we can release it in 1.6.2

lib/services/test-execution-service.ts
```diff
@@ -93,12 +98,30 @@ class TestExecutionService implements ITestExecutionService {
...
+			this.$devicesService.initialize({ platform: platform, deviceId: this.$options.device }).wait();
+			let projectFilesPath = path.join(platformData.appDestinationDirectoryPath, constants.APP_FOLDER_NAME);
+
+			let liveSyncData: ILiveSyncData = {
+				platform: platform,
+				appIdentifier: this.$projectData.projectId,
+				projectFilesPath: projectFilesPath,
+				syncWorkingDirectory: path.join(projectDir, constants.APP_FOLDER_NAME),
+				canExecuteFastSync: false, // Always restart the application when change is detected, so tests will be rerun.
+				excludedProjectDirsAndFiles: ["**/*.js.map", "**/*.ts"]
```
ligaz: We can extract this array in a constant because I believe it is used in another place as well.

lib/services/test-execution-service.ts
```diff
@@ -122,8 +145,36 @@ class TestExecutionService implements ITestExecutionService {
...
-			new KarmaServer(karmaConfig).start();
+
+			let karmaRunner = require("child_process").fork(path.join(__dirname, "karma-execution.js"));
+			karmaRunner.send({karmaConfig: karmaConfig});
+			karmaRunner.on("message", (karmaData: any) => {
+				fiberBootstrap.run(() => {
+					this.$logger.trace("## Unit-testing: Parent process received message", karmaData);
+					let port: string;
+					if(karmaData.url) {
+						port = karmaData.url.port;
```
ligaz: Consider using `hostname` here.

lib/services/test-execution-service.ts
rosen-vladimirov: oh, yes, I was thinking the same and forgot to do it. Thanks for pointing it :)

lib/services/test-execution-service.ts
```diff
@@ -122,8 +145,36 @@ class TestExecutionService implements ITestExecutionService {
...
+					}
+
+					if(karmaData.launcherConfig) {
+						let configOptions: IKarmaConfigOptions = JSON.parse(karmaData.launcherConfig);
+						let configJs = this.generateConfig(port, configOptions);
+						this.$fs.writeFile(path.join(projectDir, TestExecutionService.CONFIG_FILE_NAME), configJs).wait();
+					}
+
+					if(this.$options.debugBrk) {
+						this.getDebugService(platform).debug().wait();
```
ligaz: Why we do not need `LiveSync` in this case?

lib/services/test-execution-service.ts
```diff
@@ -154,5 +204,16 @@ class TestExecutionService implements ITestExecutionService {
 
 		return 'module.exports = ' + JSON.stringify(config);
 	}
+
+	private getDebugService(platform: string): IDebugService {
+		let lowerCasedPlatform = platform.toLowerCase();
+		if(platform.toLowerCase() === this.$devicePlatformsConstants.iOS.toLowerCase()) {
```
ligaz: `lowerCasedPlatform` maybe.

lib/services/test-execution-service.ts
rosen-vladimirov: I need the port for configuration below (next `if`). However, I've replaced the hard-coded  `localhost:${port}` with `host` which contains both `hostname` and `port` (check next line)

lib/services/test-execution-service.ts
rosen-vladimirov: I'm not sure I've understood your point here, but I've moved the generation of LiveSync data and calling LiveSync service in a separate method that is called only in the `else` statement.

ligaz:
:+1:

Issue #1313, opened 2015-12-06
tns livesync <platform> --watch is not working for changed platform specific files


x4080:
Hi,

When I change the content of .ios.js, the changes not reflected in the simulator (don't know about device), if I re run ins live sync iOS --watch again or without --watch, it changes (but not in the --watch state)

Fatme:
https://github.com/telerik/mobile-cli-lib/pull/562

dtopuzov:
*.ios.js files fail to sync in iOS simulator:

Unable to sync files. Error is: ENOENT: no such file or directory, stat '/Users/topuzov/Git/test/TestApp/platforms/ios/TestApp/app/details.ios.js'

m-abs:
I see a similar error as @dtopuzov with Android modules from a Linux PC:

"""
Unable to sync files. Error is: ENOENT: no such file or directory, stat '/home/m_abs/project/playerdemo/platforms/android/src/main/assets/app/shared/sound.android.js'
"""

NativeScript 1.6.1

rosen-vladimirov:
Yep, the livesync is not working when `--watch` option is used and platform specific file is modified, for example `main-page.android.js`. I've just merged a fix for this in the release branch :)

PR #1551, opened 2016-02-29
Fix livesync of platform/configuration specific files

rosen-vladimirov:
Fix LiveSync of platform specific / configuration specific files. Currently mappedFilePath does not respect the platform/configuration name.
Add base class that can parse the files and return correct on device filename.
Add unit tests for the provider.

Fixes https://github.com/NativeScript/nativescript-cli/issues/1313


ErjanGavalji:
:+1:

ligaz:
:+1:

PR #1552, opened 2016-02-29
Fix installation of latest compatible version

rosen-vladimirov:
CLI should install latest compatible version of runtimes, based on it's current version. For example in case CLI's version is "1.6.0", when `tns platform add android` is called, CLI should install latest 1.6.x version of android runtime, no matter if 1.7.x exists.
This was not working correctly as it was using beta versions, incorrect sorting, etc.
Fix it by using semver methods and always use the `@latest` tag of the package. In case it doesn't match CLI version, try finding latest matching tag with maxSatisfying method of semver, which dismisses beta versions.

Required for publishing `@next` in npm.


ErjanGavalji:
:+1:

Mitko-Kerezov:
:+1:

Issue #1416, opened 2016-01-20
Livesync does not work when /data/local/tmp/<app-identifier> file exists on Android device


dtopuzov:
Steps to reproduce:
tns run android --path TNS_App --justlaunch
<edit some xml file>
tns livesync android --path TNS_App --justlaunch
=> Still see old app on the phone.

Fatme:
This happens when a file with name `app-identifier` exists in `/data/local/tmp` directory on device. In this case CLI is not able to create directory.

teobugslayer:
Is this related to https://github.com/NativeScript/nativescript-cli/issues/1081 ?

Fatme:
Yep

teobugslayer:
`ls -l` -> `rm - rf`

dtopuzov:
Duplicates #1081

PR #1560, opened 2016-03-02
Delete /data/local/tmp/<app-identifier> file if exists on Android device

nadyaA:
Livesync does not work when /data/local/tmp/<app-identifier> file exists on Android device.
This happens when a file with name app-identifier exists in /data/local/tmp directory on device. In this case CLI is not able to create directory.

https://github.com/NativeScript/nativescript-cli/issues/1416

ping @rosen-vladimirov @Mitko-Kerezov

### DIFF
--- a/lib/services/livesync/android-livesync-service.ts
+++ b/lib/services/livesync/android-livesync-service.ts
@@ -35,7 +35,18 @@ class AndroidLiveSyncService extends liveSyncServiceBaseLib.LiveSyncServiceBase<
 
 	public beforeLiveSyncAction(deviceAppData: Mobile.IDeviceAppData): IFuture<void> {
 		return (() => {
-			let deviceRootPath = this.getDeviceRootPath(deviceAppData.appIdentifier);
+			let deviceRootPath = this.getDeviceRootPath(deviceAppData.appIdentifier),
+				deviceRootDir = path.dirname(deviceRootPath),
+				deviceRootBasename = path.basename(deviceRootPath),
+				listResult = this.device.adb.executeShellCommand(["ls", "-l", deviceRootDir]).wait(),
+				regex = new RegExp(`^-.*${deviceRootBasename}$`, "m"),
+				matchingFile = (listResult || "").match(regex);
+
+			// Check if there is already a file with deviceRootBasename. If so, delete it as it breaks LiveSyncing.
+			if(matchingFile && matchingFile[0] && _.startsWith(matchingFile[0], '-')){
+				this.device.adb.executeShellCommand(["rm", "-f", deviceRootPath]).wait();
+			}
+
 			this.device.adb.executeShellCommand(["rm", "-rf", this.$mobileHelper.buildDevicePath(deviceRootPath, "fullsync"),
 				this.$mobileHelper.buildDevicePath(deviceRootPath, "sync"),
 				this.$mobileHelper.buildDevicePath(deviceRootPath, "removedsync")]).wait();


lib/services/livesync/android-livesync-service.ts
```diff
@@ -35,7 +35,17 @@ class AndroidLiveSyncService extends liveSyncServiceBaseLib.LiveSyncServiceBase<
...
-			let deviceRootPath = this.getDeviceRootPath(deviceAppData.appIdentifier);
+			let deviceRootPath = this.getDeviceRootPath(deviceAppData.appIdentifier),
+				deviceRootDir = path.dirname(deviceRootPath),
+				deviceRootBasename = path.basename(deviceRootPath),
+				listResult = this.device.adb.executeShellCommand(["ls", "-l", deviceRootDir]).wait(),
+				regex = new RegExp(`^-.*${deviceRootBasename}$`, "m"),
+				matchingFile = (listResult || "").match(regex);
+
+			if(matchingFile && matchingFile[0] && _.startsWith(matchingFile[0], '-')){
+				this.device.adb.executeShellCommand(["rm", "-rf", deviceRootPath]).wait();
```
rosen-vladimirov: in case it's a file, you do not need `-r`

rosen-vladimirov: I would love to see some `log.trace` message here, informing that the file had been removed

lib/services/livesync/android-livesync-service.ts
```diff
@@ -35,7 +35,17 @@ class AndroidLiveSyncService extends liveSyncServiceBaseLib.LiveSyncServiceBase<
...
 		return (() => {
-			let deviceRootPath = this.getDeviceRootPath(deviceAppData.appIdentifier);
+			let deviceRootPath = this.getDeviceRootPath(deviceAppData.appIdentifier),
+				deviceRootDir = path.dirname(deviceRootPath),
+				deviceRootBasename = path.basename(deviceRootPath),
+				listResult = this.device.adb.executeShellCommand(["ls", "-l", deviceRootDir]).wait(),
+				regex = new RegExp(`^-.*${deviceRootBasename}$`, "m"),
+				matchingFile = (listResult || "").match(regex);
+
+			if(matchingFile && matchingFile[0] && _.startsWith(matchingFile[0], '-')){
```
rosen-vladimirov: can you add comment why the check is with `-` as after a week I'll not remember why do we need this and what's the reason to check if a string starts with `-`

rosen-vladimirov:
:+1:

lib/services/livesync/android-livesync-service.ts
```diff
@@ -35,7 +35,18 @@ class AndroidLiveSyncService extends liveSyncServiceBaseLib.LiveSyncServiceBase<
...
-			let deviceRootPath = this.getDeviceRootPath(deviceAppData.appIdentifier);
+			let deviceRootPath = this.getDeviceRootPath(deviceAppData.appIdentifier),
+				deviceRootDir = path.dirname(deviceRootPath),
+				deviceRootBasename = path.basename(deviceRootPath),
+				listResult = this.device.adb.executeShellCommand(["ls", "-l", deviceRootDir]).wait(),
+				regex = new RegExp(`^-.*${deviceRootBasename}$`, "m"),
+				matchingFile = (listResult || "").match(regex);
+
+			// Check if there is already a file with deviceRootBasename. If so, delete it as it breaks LiveSyncing.
+			if(matchingFile && matchingFile[0] && _.startsWith(matchingFile[0], '-')){
```
rosen-vladimirov: We do not need the last check! In case the symbol is not `-` there will be no match :)

Issue #1563, opened 2016-03-02
Install runtime by npm tagname


ErjanGavalji:
Running `tns platform add android@next` throws an error. Using the absolute version number works properly.

It would be best if tagged npm packages work seamlessly.

PR #1568, opened 2016-03-04
Fix installation of runtime with next tag

rosen-vladimirov:
Our npmInstallationManager is using the passed `next` string as a version and searches for directory called `next` in the npm cache.
Instead it should use the real version. Fix this by getting the result of npm call addToCache, which returns object with information about cached package.

With this change you can call:
```
tns platform add android@next
```

Fixes https://github.com/NativeScript/nativescript-cli/issues/1563

### DIFF
--- a/lib/npm-installation-manager.ts
+++ b/lib/npm-installation-manager.ts
@@ -42,19 +42,26 @@ export class NpmInstallationManager implements INpmInstallationManager {
 		return path.join(this.getCacheRootPath(), packageName, version, "package");
 	}
 
-	public addToCache(packageName: string, version: string): IFuture<void> {
+	public addToCache(packageName: string, version: string): IFuture<any> {
 		return (() => {
 			let cachedPackagePath = this.getCachedPackagePath(packageName, version);
+			let cachedPackageData: any;
 			if(!this.$fs.exists(cachedPackagePath).wait() || !this.$fs.exists(path.join(cachedPackagePath, "framework")).wait()) {
-				this.addToCacheCore(packageName, version).wait();
+				cachedPackageData = this.addToCacheCore(packageName, version).wait();
 			}
 
-			if(!this.isShasumOfPackageCorrect(packageName, version).wait()) {
+			// In case the version is tag (for example `next`), we need the real version number from the cache.
+			// In these cases the cachePackageData is populated when data is added to the cache.
+			// Also whenever the version is tag, we always get inside the above `if` and the cachedPackageData is populated.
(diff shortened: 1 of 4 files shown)


ErjanGavalji:
:+1:

lib/npm-installation-manager.ts
```diff
@@ -42,19 +42,22 @@ export class NpmInstallationManager implements INpmInstallationManager {
...
+				cachedPackageData = this.addToCacheCore(packageName, version).wait();
 			}
-
-			if(!this.isShasumOfPackageCorrect(packageName, version).wait()) {
+			let realVersion = (cachedPackageData && cachedPackageData.version) || version;
+			if(!this.isShasumOfPackageCorrect(packageName, realVersion).wait()) {
 				// In some cases the package is not fully downloaded and there are missing directories
 				// Try removing the old package and add the real one to cache again
-				this.addCleanCopyToCache(packageName, version).wait();
+				cachedPackageData = this.addCleanCopyToCache(packageName, version).wait();
```
Mitko-Kerezov: According to the interface, `addCleanCopyToCache(packageName: string, version: string): IFuture<void>;`, so this reassignment of cachedPackageData is pointless.
I'm guessing the interface is a bit off.

lib/npm-installation-manager.ts
```diff
@@ -42,19 +42,22 @@ export class NpmInstallationManager implements INpmInstallationManager {
...
 		return (() => {
 			let cachedPackagePath = this.getCachedPackagePath(packageName, version);
+			let cachedPackageData: any;
 			if(!this.$fs.exists(cachedPackagePath).wait() || !this.$fs.exists(path.join(cachedPackagePath, "framework")).wait()) {
-				this.addToCacheCore(packageName, version).wait();
+				cachedPackageData = this.addToCacheCore(packageName, version).wait();
 			}
-
-			if(!this.isShasumOfPackageCorrect(packageName, version).wait()) {
+			let realVersion = (cachedPackageData && cachedPackageData.version) || version;
```
Mitko-Kerezov: You can preferably add a comment here explaining why the `cachedPackageData`'s version prevails over `version` if both are present

Mitko-Kerezov:
:+1: after addressing comments

ikoevska:
Are we documenting this somewhere?

lib/npm-installation-manager.ts
rosen-vladimirov: Fixed :)

rosen-vladimirov:
@ikoevska no, it should work as expected. It's a bug that it didn't work :)

PR #1574, opened 2016-03-07
Add tests for installing beta versions

rosen-vladimirov:
In case CLI's version is marked as beta (has dash in it), CLI should consider installing beta versions of runtimes.
This is already working. Just add tests for this functionality.


teobugslayer:
:+1:

PR #1575, opened 2016-03-07
Gulp fails miserably with symlinks in node_modules, use glob and stat instead

PanayotCankov:
Now you can:
```
cd ~/GitRoot/NativeScript
tsc -p .
npm link
cd ~/Projects/MyApp
npm link tns-core-modules
code .
tns run ios
```

The gulp we used to detect changes in the installed modules used to fail internally due fs-ing `tns-core-modules` symlink.


package.json
```diff
@@ -40,6 +40,7 @@
     "fibers": "https://github.com/icenium/node-fibers/tarball/v1.0.6.3",
     "filesize": "3.1.2",
     "gaze": "0.5.2",
+    "glob": "^7.0.3",
     "gulp": "3.9.0",
```
rosen-vladimirov: do we need gulp now ?

lib/tools/broccoli/builder.ts
```diff
@@ -28,30 +29,42 @@ export class Builder implements IBroccoliBuilder {
...
+						if (!future.isResolved()) {
+							future.throw(er);
+						}
+						match.abort();
+						return;
 					}
-
-					if(!isNodeModulesModified) {
-						let rootModuleName = chunk.path.split(nodeModulesPath)[1].split(path.sep)[1];
+					for (let i = 0, l = files.length; i < l; i++) {
```
rosen-vladimirov: I prefer using `_.each(files, file => {.... })`

lib/tools/broccoli/builder.ts
```diff
@@ -9,6 +9,7 @@ import * as destCopyLib from "./node-modules-dest-copy";
 let gulp = require("gulp");
 let vinylFilterSince = require("vinyl-filter-since");
```
rosen-vladimirov: as it is not used anymore, can we remove this require.
Also please remove the `vinyl-filter-since` dependency from `package.json`

lib/tools/broccoli/builder.ts
```diff
@@ -9,6 +9,7 @@ import * as destCopyLib from "./node-modules-dest-copy";
 let gulp = require("gulp");
```
rosen-vladimirov: please remove `gulp`

lib/tools/broccoli/builder.ts
```diff
@@ -9,6 +9,7 @@ import * as destCopyLib from "./node-modules-dest-copy";
...
 let vinylFilterSince = require("vinyl-filter-since");
 let through = require("through2");
```
rosen-vladimirov: please remove through2 from here and from package.json as well

lib/tools/broccoli/builder.ts
```diff
@@ -28,30 +29,42 @@ export class Builder implements IBroccoliBuilder {
...
-
-					if(!isNodeModulesModified) {
-						let rootModuleName = chunk.path.split(nodeModulesPath)[1].split(path.sep)[1];
+					for (let i = 0, l = files.length; i < l; i++) {
+						let file = files[i];
+						let stat = match.statCache[path.resolve(file)];
+						if (stat.mtime <= lastModifiedTime) {
+							continue;
+						}
+						if (file === "node_modules") {
```
rosen-vladimirov: use
```TypeScript
if(file === constants.NODE_MODULES_FOLDER_NAME) {
```

lib/tools/broccoli/builder.ts
PanayotCankov: i prefer files.forEach(file => { })

PanayotCankov: Then return and continue get worse

package.json
PanayotCankov: I have absolutely no idea :D Lets do a search.
vinylFilterSince may be dropped as well.

lib/tools/broccoli/builder.ts
rosen-vladimirov: I do not like `files.forEach` as it fails when files is null
Anyway, the current code is simple enough and as you said, the return statements will be hard for maintenance, so you can keep it this way ;)

package.json
rosen-vladimirov: `gulp`, `vinyl-filter-since` and `through2` should be removed

PanayotCankov:
Applied most of the comments in the PR

lib/tools/broccoli/builder.ts
```diff
@@ -28,30 +27,50 @@ export class Builder implements IBroccoliBuilder {
...
+				}, (er: Error, files: string[]) => {
+					fiberBootstrap.run(() => {
+						if (er) {
+							if (!future.isResolved()) {
+								future.throw(er);
+							}
+							match.abort();
+							return;
+						}
+						for (let i = 0, l = files.length; i < l; i++) {
```
teobugslayer: Not opposed to the current code, but why not using _.each?

teobugslayer:
I am very happy with the simplified dependencies. :+1:

teobugslayer:
But, is this for master - 2.0 or release - 1.7? In latter case, the PR must be re-targeted.

rosen-vladimirov:
We've decided to use keep it against master branch :)

rosen-vladimirov:
run ci

rosen-vladimirov:
:+1:  tested on Linux as well, great work

PanayotCankov:
If there is a need I'll remove the glob dependency as it appears it has an issue or two when it comes to its cross platform features.

However it would be nice if we can use it to filter the files used to change detect and copy to the `platforms/ios` and `platoforms/android` based on `.npmignore`, `.gitignore` files or *package.json's files key*.

PR #1587, opened 2016-03-09
Verify JAVA_HOME is set correctly

rosen-vladimirov:
Our system requirements point that JAVA_HOME must be set, but `tns doctor` is not strict about this.
With version 1.7.0 of the runtime, `jarsigner` from $JAVA_HOME/bin/jarsigner will be called explicitly, so we must be sure JAVA_HOME is correct.
`jarsigner` does not exist in JRE, so add logic to call it in order to verify that the currently set JAVA_HOME is correct.
Validate the information in the following order:
1) Check if JAVA_HOME is set. In case it is not, you cannot build for Android (`tns build android` will fail before starting).
2) Check if JAVA_HOME is correct by executing JAVA_HOME/bin/jarsigner - in case it does not exist - fail
3) Check if the version of the JAVA is correct by calling JAVA_HOME/bin/javac -version and compare it with min required one. In case it's lower - fail.

`tns doctor` will print warning in case any of the above check fails. `tns build android` and any other Android related build operations will fail.

### DIFF
--- a/lib/android-tools-info.ts
+++ b/lib/android-tools-info.ts
@@ -156,28 +156,63 @@ export class AndroidToolsInfo implements IAndroidToolsInfo {
 		}).future<boolean>()();
 	}
 
-	public validateJavacVersion(installedJavaVersion: string, options?: {showWarningsAsErrors: boolean}): IFuture<boolean> {
+	public validateJava(javacVersion: string, options?: {showWarningsAsErrors: boolean}): IFuture<boolean> {
 		return ((): boolean => {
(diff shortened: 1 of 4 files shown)


rosen-vladimirov:
ping @Plamen5kov , @slavchev , @teobugslayer

slavchev:
:+1: 
related to https://github.com/NativeScript/android-runtime/commit/321c51a06d4c4d48ca43311abb0e34d9cc79783f

Plamen5kov:
:+1:

vchimev:
run ci

Issue #1576, opened 2016-03-08
tns run android just starts emulator but does not deploy app on it


DeepakDonde:
Hello @rosen-vladimirov ,
I have recently get rid of [this ](https://github.com/NativeScript/nativescript-cli/issues/1564) issue.

Now I am getting issue while running the app.
when I run "tns run android" from cmd, emulator gets started and nothing happens after it launches.
Earlier when I have connected android device and run the same command I was getting following error.
Processing node_modules failed. TypeError: Invalid Version: 1.?2.0 (in red color on cmd)
Sending exception report (press Ctrl+C to stop).....

Also please find attached log file for log trace.
[log.txt](https://github.com/NativeScript/nativescript-cli/files/162691/log.txt)

Please help.
Thanks in advance.

rosen-vladimirov:
I confirm this is a bug on our side and it should be fixed in our next release.
Meanwhile you can use `tns emulate android` instead :)

rosen-vladimirov:
Steps to reproduce:
1) Make sure you do not have any emulator started or any device attached.
2) Call `tns run android` inside project.

DeepakDonde:
@rosen-vladimirov ,
What am I supposed to do to reflect this change at my side ?
I am still facing same issue while running app. :(
please help.

rosen-vladimirov:
@DeepakDonde you'll have to wait for our next release in order to get it working. Meanwhile you can use the suggested workaround - start the emulator before calling `tns run android`.

DeepakDonde:
@rosen-vladimirov 

I have tried the workaround but I am getting following error - 
Command platforms\android\gradlew.bat failed with exit code 1
Attaching log file for your reference.
[log.txt](https://github.com/NativeScript/nativescript-cli/files/171776/log.txt)

rosen-vladimirov:
@DeepakDonde Please navigate to `<project dir>\platforms\android` and execute: `./gradelw clean`
After that try `tns run android` again (from your project dir). If it still fails, navigate back to `<project dir>\platforms\android` and execute `./gradelw build -S` and send us the output.

DeepakDonde:
@rosen-vladimirov 
Still the build failed.. 
>A problem occurred evaluating root project 'sampleGroceries'.
> Could not find method compile() for arguments [com.android.support:appcompat-v
7:22.2.0] on DefaultExternalModuleDependency{group='com.android.support', name='
support-v4', version='22.2.0', configuration='default'}.

[...]

PR #1589, opened 2016-03-09
Fix starting of emulator when no devices are attached

rosen-vladimirov:
When there are no devices attached, CLI should start emulator and execute the action on it.
For example calling `tns run android` when there are no devices attached, the devicesService should start emulator and run the application in it.

Currently this is not working as the startEmulator method is starting the instance and then calls method for device detection.
However the checkForDevices method is async and it does not guarantee when the devices will be detected.
Instead use startLookingForDevices which is synchronous.

Fixes https://github.com/NativeScript/nativescript-cli/issues/1576


teobugslayer:
:+1:

rosen-vladimirov:
run ci

rosen-vladimirov:
run ci

Issue #1541, opened 2016-02-25
Plugin remove fails if plugin version is not available in npm


dtopuzov:
package.json:
```
{
  "nativescript": {
    "id": "com.telerik.plugindemo.barcodescanner",
    "tns-ios": {
      "version": "1.7.0-2016.2.23.1"
    },
    "tns-android": {
      "version": "1.6.2-2016.2.22.2"
    }
  },
  "dependencies": {
    "nativescript-barcodescanner": "1.2.1",
    "tns-core-modules": "file:../tns-core-modules.tgz"
  },
  "devDependencies": {
    "nativescript-dev-typescript": "^0.3.1",
    "typescript": "^1.8.2"
  }
}
```
tns plugin remove nativescript-barcodescanner
npm WARN package.json @ No description
npm WARN package.json @ No repository field.
npm WARN package.json @ No README data
npm WARN package.json @ No license field.
version not found: nativescript-barcodescanner@1.2.1

=> and pacakge.json is still the same, plugin not removed.

Tested with CLI in master.

This bug makes building sample apps for verified plugins hard.

dtopuzov:
Same applies if I have reference to local file and this file does not exist.
```
"dependencies": {
        "nativescript-imagepicker": "file:../../dist/package",
        "tns-core-modules": "1.5.0"
}
```
npm WARN package.json ExampleImgPick@0.0.2 No license field.
npm ERR! addLocal Could not install /Users/nsbuilduser/dist/package
ENOENT: no such file or directory, open '/Users/nsbuilduser/dist/package'

PR #1590, opened 2016-03-09
Fix plugin remove command for invalid plugins

rosen-vladimirov:
When invalid plugin data is written in project's package.json, `tns plugin remove <name>` command fails.
In such cases `npm uninstall --save <name>` just removes the entry from package.json, so we should do the same.
By invalid data I mean invalid version (not existing version or path to non-existing file).

Plugin remove command's canExecute method is trying to install all plugins in order to get information about their native code, so it will be removed from `platforms` dir.
The installation fails when some plugin's data is invalid.
Catch the error and check package.json's dependencies instead.
Also make sure plugin variables data is removed from package.json when plugin is removed.
Currently it's been removed only when platform is added and plugin has native code.

Fixes https://github.com/NativeScript/nativescript-cli/issues/1541

### DIFF
--- a/lib/commands/plugin/remove-plugin.ts
+++ b/lib/commands/plugin/remove-plugin.ts
@@ -3,7 +3,9 @@
 
 export class RemovePluginCommand implements ICommand {
 	constructor(private $pluginsService: IPluginsService,
-		private $errors: IErrors) { }
+		private $errors: IErrors,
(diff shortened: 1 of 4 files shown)


rosen-vladimirov:
run ci

vchimev:
run ci

teobugslayer:
:+1:

PR #1591, opened 2016-03-10
Wait for sending usage report

rosen-vladimirov:
In case eqatecMonitor is not started it may take a while to send the usage report.
In case the process exits, the report is never sent.
So wait for sending it.


teobugslayer:
:+1:

PR #1592, opened 2016-03-10
Remove shelljs.config.fatal

rosen-vladimirov:
When adding platform, CLI tries to copy some files from runtime to the native project.
When some of the files do not exist and `shelljs.config.fatal` is true, shelljs exits the process without warn.
Remove the option for the moment and consider better usage for it.

### DIFF
--- a/lib/nativescript-cli.ts
+++ b/lib/nativescript-cli.ts
@@ -9,7 +9,6 @@ import * as fiber from "fibers";
 import Future = require("fibers/future");
 import * as shelljs from "shelljs";
 shelljs.config.silent = true;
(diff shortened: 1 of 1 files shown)


vchimev:
"CLI tries to copy some files from runtime to the native project. When some of the files do not exist" - how so they do not exist?

rosen-vladimirov:
@vchimev  we have list of files which are expected to exist in current runtime, but when you execute `tns platform add android@1.3.0` some of them do not exist

vchimev:
:+1:

rosen-vladimirov:
close and retarget to release

PR #1596, opened 2016-03-10
Track Feature usage in a separate Analytics project

rosen-vladimirov:
As we need to track new users separately from existing ones, we have to track AcceptFeatureUsageTracking in a separate project.
Also track `InstallEnvironmentSetup` with values 'correct' and 'incorrect' - it will be tracked only on post-install of CLI and will report if user's environment is correct.
Track `DoctorEnvironmentSetup` with values 'correct' and 'incorrect'. On post-install this feature will not be tracked.


teobugslayer:
:+1:

PR #1597, opened 2016-03-10
Revert strict check for JAVA_HOME

rosen-vladimirov:
Due to change in runtime, we do not need strict check for JAVA_HOME, so revert the logic introduced yesterday.


teobugslayer:
Well, :+1:

PR #1599, opened 2016-03-11
Fix livesync for iOS

rosen-vladimirov:
CLI is destroying the socket too early and so the runtime cannot handle the message that we've sent.
As a result livesync is not working sometimes.
When we sent message, the socket will always receive data as result. If it's error, the result will contain it.
In any case, we should destroy the socket when data is received.


Mitko-Kerezov:
:+1:

vchimev:
:+1:

Issue #1612, opened 2016-03-15
Unable to execute unit tests on iOS Simulator in some cases


rosen-vladimirov:
In some cases related to network configurations, unit-tests on ios-simulator are not run as CLI is filtering internal networks. We have to allow using `localhost` in such cases.

PR #1613, opened 2016-03-15
Enable using localhost as ip for tests

rosen-vladimirov:
When using iOS Simulator we need the localhost enabled as ip for karma. This does not break the Android tests, so enable it for all configurations.

Fixes https://github.com/NativeScript/nativescript-cli/issues/1612

> NOTE: Merge after 1.7.0 is released


Mitko-Kerezov:
:+1:

PR #1633, opened 2016-03-22
Replace plugin variables in all Android .xml files

rosen-vladimirov:
In case a plugin has plugin variable, during project preparation CLI replaces the usage of the variable in plugin's AndroidManifest with the value set in the current project's package.json.
This is not enough, we have to set the variables in all plugin's .xml files, so the value could be used in resource and later read from the JAVA code.
Enumarate plugin's files and replace plugin variables in all .xml files.

https://github.com/NativeScript/nativescript-cli/issues/1634

### DIFF
--- a/lib/services/android-project-service.ts
+++ b/lib/services/android-project-service.ts
@@ -327,10 +327,11 @@ export class AndroidProjectService extends projectServiceBaseLib.PlatformProject
 				this.$fs.ensureDirectoryExists(resourcesDestinationDirectoryPath).wait();
 				shell.cp("-Rf", path.join(pluginPlatformsFolderPath, "*"), resourcesDestinationDirectoryPath);
 
-				let pluginConfigurationFilePath = path.join(resourcesDestinationDirectoryPath, this.platformData.configurationFileName);
-				if (this.$fs.exists(pluginConfigurationFilePath).wait()) {
-					this.$pluginVariablesService.interpolate(pluginData, pluginConfigurationFilePath).wait();
-				}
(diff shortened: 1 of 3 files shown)


rosen-vladimirov:
run ci

lib/services/android-project-service.ts
```diff
@@ -327,10 +327,11 @@ export class AndroidProjectService extends projectServiceBaseLib.PlatformProject
 				this.$fs.ensureDirectoryExists(resourcesDestinationDirectoryPath).wait();
 				shell.cp("-Rf", path.join(pluginPlatformsFolderPath, "*"), resourcesDestinationDirectoryPath);
 
-				let pluginConfigurationFilePath = path.join(resourcesDestinationDirectoryPath, this.platformData.configurationFileName);
-				if (this.$fs.exists(pluginConfigurationFilePath).wait()) {
-					this.$pluginVariablesService.interpolate(pluginData, pluginConfigurationFilePath).wait();
-				}
+				let files = this.$fs.enumerateFilesInDirectorySync(resourcesDestinationDirectoryPath, file => this.$fs.getFsStats(file).wait().isDirectory() || path.extname(file) === ".xml");
```
Mitko-Kerezov: You can extract ".xml" as a constant as it's also used in `xml-validator`

lib/services/android-project-service.ts
```diff
@@ -327,10 +327,11 @@ export class AndroidProjectService extends projectServiceBaseLib.PlatformProject
...
+				_.each(files, file => {
```
Mitko-Kerezov: You can safely use `.forEach` instead of the `lodash` method here

teobugslayer: We are using lodash everywhere for consistency and because no one has to think "is it safe to use blah in here".

Mitko-Kerezov:
:+1:

rosen-vladimirov:
run ci

PR #1641, opened 2016-03-23
Respect --baseConfig on prepare

rosen-vladimirov:
When `--baseConfig` is passed to prepare (or any other build related command) it should be respected.
Currently the option is respected only on `platform add`.


Mitko-Kerezov:
:+1:

PR #1643, opened 2016-03-24
Fixed autocomplete

TsvetanMilanov:
When path to node has spaces (C:\Program Files) the child process exec cannot run the command. Need to place it in "" to run it without errors.
Update FileSystemStub with new method from interface.


TsvetanMilanov:
:+1:  from @rosen-vladimirov  after green build

PR #1644, opened 2016-03-24
Fix analytics tracking of users

rosen-vladimirov:
* Move tracking of environment setup at the end of post install. This way we'll first ask the users if they want to be tracked and after that track the InstallEnvironmentSetup only for the users who allowed us to track them.
* Explicitly call the check for analytics status on post-install. Currently the question was raised as we were calling autocomplete command. On each command we check if user had answered the question and if not - we ask him. Make the action more visible (in the code).
* When asking user if analytics tracking should be enabled, currently we start the analytics monitor before showing the question. This is incorrect as when the users does not answer immediately, we end up with very long sessions (for example user keeps the terminal opened and the question just stays there, the session never ends). Start the analytics monitor after user had answered.
* When the user answers the question, we track the answer and after that we persist the value in the user-settings. However when the sending takes more than expected, the user may press Ctrl + C, we may receive the data, but the setting will not be saved in the user settings. So next time when the user executes any command, we'll ask him again. Set the value in the user settings before tracking it. This way even if Ctrl + C is pressed, the value will be persisted.
* Add process.exit event handler which should stop the monitor. In many cases while tracking, we do not call `stop` of analytics monitor, so the sessions look like they are taking more than expected. Add event handler for process.exit and stop the monitor, in case it's been started.
* Fix unit tests to call the new method `tryStopEqatecMonitor` in order to remove process.exit event handlers attached by each test.
* Remove `disableAnalyticsConsentCheck` from ICommand. Looks like it's been used in a really strange way and I couldn't find it's real purpose. The property was used to prevent asking the user for confirmation when executing `help`, `usage-reporting` and `error-reporting` commands. This leads to incorrect behavior - when the user installs CLI with `--ignore-scripts` and executes `tns help` immediatley after that, CLI will track `help` command without asking.
* Track features and exceptions only when status of the properties is "enabled". Currently we track even if status is "not confirmed", but this is not correct. This way, CLI used for automation is always tracked.
* Disable error reporting when user says "no" for usage reporting. As the user does not want to be tracked, we should disable error reporting.
* Fix helpers method for converting string value to boolean. In case the string is "True" the method returned false, while it should be true.


Mitko-Kerezov:
:+1:

Issue #1610, opened 2016-03-15
tns plugin add <plugin> has issues with postinstall scripts


rosen-vladimirov:
`tns plugin add <plugin>` has some issues with stdin when postinstall is using inquirer.
For example create plugin that has the following postinstall script:
```
"postinstall": "npm install inquirer && node myScript.js"
```
where myScript.js is something like:
```JavaScript
var inquirer = require('inquirer');
var quest = {
    'name': 'myinq',
    'type': 'input',
    'message': 'Ask something: ',
    'default': undefined
};

function promptForEmail() {
    inquirer.prompt([quest],
        function(answers) {
            console.log("answers:", answers);
        });
}

promptForEmail();
```

In case you call `npm install <plugin>` the prompt is executed correctly and user can type.
In case you call `tns plugin add <plugin>` the user's input is slowed down, some symbols are not shown at all, etc.

Required for `nativescript-telerik-ui`.

rosen-vladimirov:
Reproducible without using inquirer. Set the postinstall script to be:
```JavaScript

const readline = require('readline');

const rl = readline.createInterface({
    input: process.stdin,
    output: process.stdout
});

rl.question('Ask something:? ', (answer) => {
    console.log('Thank you for your valuable feedback, which was:\n', answer);
    rl.close();
});
```

PR #1647, opened 2016-03-27
Unmute stdout after prompter execution

rosen-vladimirov:
In order to prevent double printing of stdout data, CLI mutes the stdout. However this is done in prompter constructor, so every class, that depends on the prompter, automatically mutes the stdout.
Basically CLI mutes the stdout on almost all commands. This causes issues when `tns plugin add` command is used and the plugin wants to prompt the user for something on postinstall.
Fix this by calling mute exactly before prompting the user and calling unmute immediately after user sends his answer.

Fixes https://github.com/NativeScript/nativescript-cli/issues/1610


Mitko-Kerezov:
Nice fix

PR #1653, opened 2016-03-28
Improve Setup Scripts' Android bits

Mitko-Kerezov:
Changes for *Windows* 
* In case `ANDROID_HOME` environment variable is not set, try to determine the correct path to `android-sdk` through `android.bat`'s location
* Use `android.bat` executable relative to `android-sdk`, set in `ANDROID_HOME` environment variable, for sdk updates
* Split android sdk updates into single command calls in order to avoid problems with license agreements

Changes for *Mac*:
* In case `ANDROID_HOME` environment variable is not set, try to determine the correct path to `android-sdk` - either use homebrew's symlink or through `android` executable's location
* Use `android` executable relative to `android-sdk`, set in `ANDROID_HOME` environment variable, for sdk updates
* Split android sdk updates into single command calls in order to avoid problems with license agreements

Ping @teobugslayer @rosen-vladimirov


rosen-vladimirov:
:+1:

PR #1655, opened 2016-03-29
Install platform-tools before tools in setup scripts

Mitko-Kerezov:
There are some cases where updating tools requires updating platform-tools and if the former is attempted before the latter the following error is raised:
`Skipping 'Android SKD Tools, revision 24.4.1'; it depends on 'Android SDK Platform-tools, revision 23.1 rc1' which was not installed`

Fix this by reversing order
Ping @rosen-vladimirov @teobugslayer


rosen-vladimirov:
:+1:

PR #1656, opened 2016-03-29
Fix --baseConfig option

rosen-vladimirov:
Apply --baseConfig before replacing placeholders in manifest. This way when there are placeholders in the file passed as `--baseConfig`, they will be replaced correctly.


nadyaA:
:+1:  Let us integrate some more of this juicy {N} :yum:

teobugslayer:
:+1:

PR #1660, opened 2016-03-31
AppBuilder fixes

rosen-vladimirov:
Use baseConfig when merging Info.plist files.

Disable analytics in AppBuilder.


Mitko-Kerezov:
Tested on SIT's build machine and appears to be working :+1:

Issue #1526, opened 2016-02-22
Platform add <platform> prints command help when platform is already added


rosen-vladimirov:
When platform is already added, calling `platform add` prints command help:
```
$ tns platform add android
Platform android already added
# platform add
┌──────────────────────────┬──────────────────────────────────────────────────────────────────────────────────────────────────────┐
│ Usage                    │ Synopsis                                                                                             │
│ Android latest runtime   │ $ tns platform add android [--frameworkPath <File Path>] [--symlink] [--sdk <API Level>]             │
│ Android selected runtime │ $ tns platform add android[@<Version>] [--frameworkPath <File Path>] [--symlink] [--sdk <API Level>] │
└──────────────────────────┴──────────────────────────────────────────────────────────────────────────────────────────────────────┘

Configures the current project to target the selected platform.

### Options

    * --frameworkPath - Sets the path to a NativeScript runtime for the specified platform that you want to use instead of the default runtime. If --symlink is specified, <File Path> must point to directory in which the runtime is already extracted. If --symlink is not specified, <File Path> must point to a valid npm package.
    * --symlink - Creates a symlink to a NativeScript runtime for the specified platform that you want to use instead of the default runtime. If --frameworkPath is specified, creates a symlink to the specified directory. If --frameworkPath is not specified, creates a symlink to platform runtime installed with your current version of NativeScript.
    * --sdk - Sets the target Android SDK for the project.

### Attributes

    * <API Level> is a valid Android API level. For example: 17, 19, MNC.
    * <File Path> is the complete path to a valid npm package or a directory that contains a NativeScript runtime for the selected platform.
    * <Version> is any available version of the respective platform runtime published in npm.
```

It should just inform you that the platform is already added.

PR #1662, opened 2016-03-31
Fix warning when adding existing platform

TsvetanMilanov:
Do not display help for platform add command when adding an existing platform to the project.

Fixes https://github.com/NativeScript/nativescript-cli/issues/1526


rosen-vladimirov:
:+1:

Issue #1532, opened 2016-02-23
"tns livesync ios --watch" doesn't sync js changes if pacakgeId and folder name does not match


xuezier:
tns livesync ios --emulator --watch, js file has changed,but not sync
xml and css are working good
os x 10.11.1

The real problem: 
When id of your app is "org.nativescript.TestApp" and it is placed in folder != TestApp (for example SampleApp) livesync --watch does not sync changes in js files.

Workaround: Make last part ot your app id == name of the folder where project is located.

vchimev:
@xuezier, could you please share:
- what is your Xcode version,
- does the app restart changing a *.js* file,

xuezier:
Shasum of file /Users/xuezhijing/NativeScriptProject/sample-Groceries/app/views/login/login.js is 404b7e2415abf25a7a05238f7d9f4e4a455ca6aa
Adding /Users/xuezhijing/NativeScriptProject/sample-Groceries/app/views/login/login.js file with 404b7e2415abf25a7a05238f7d9f4e4a455ca6aa hash.
Syncing filePath /Users/xuezhijing/NativeScriptProject/sample-Groceries/app/views/login/login.js, mappedFilePath is /Users/xuezhijing/NativeScriptProject/sample-Groceries/platforms/ios/sampleGroceries/app/views/login/login.js
Syncing /Users/xuezhijing/NativeScriptProject/sample-Groceries/platforms/ios/sampleGroceries/app/views/login/login.js
BeforeHookName for command prepare is before-prepare
Images from Xcode project
[ 'Default-568h@2x.png',
  'Default-Landscape.png',
  'Default-Landscape@2x.png',
  'Default-Portrait.png',
  'Default-Portrait@2x.png',
  'Default.png',
  'Default@2x.png',
  'Icon-40@3x.png',
  'Icon-60@3x.png',
  'Icon-Small-50.png',
  'Icon-Small-50@2x.png',
  'Icon-Small.png',
  'Icon-Small@2x.png',
  'Icon-Small@3x.png',
  'icon-40.png',
  'icon-40@2x.png',
  'icon-60.png',
  'icon-60@2x.png',
  'icon-72.png',
  'icon-72@2x.png',
  'icon-76.png',
  'icon-76@2x.png',
  'icon.png',
  'icon@2x.png',
  'logo.png',
  'logo@2x.png',
  'logo@3x.png' ]
Current images from App_Resources
[ 'Default-568h@2x.png',
[...]

PR #1663, opened 2016-04-01
Fix livesynnc on iOS simulator-livesync

TsvetanMilanov:
When restarting the application if pacakgeId and folder name does not match the name of the application should be the projectName.

Fixes https://github.com/NativeScript/nativescript-cli/issues/1532


rosen-vladimirov:
:+1:

Issue #1665, opened 2016-04-06
Add command for searching nativescript plugins


TsvetanMilanov:
The command should have flags to specify the count of the displayed plugins.
For example --count 10 should list 10 plugins per page, --all should list all plugins.

dtopuzov:
Not working, move back in progress.

rosen-vladimirov:
What's not working, currently I found only one issue - when there are no results, we print table with headers only.
Please note that the first time when the command is executed, it may take a while to build the local index. On my  Linux it took around 40 seconds, but based on your setup, machine, internet connection, etc. it may last even longer.
Please clarify what's "not working". @dtopuzov

dtopuzov:
I tough it will return search results immediately (as fast as search in npm website). Closing this issues since the feature works, however it is much slower compared to search in npm website.

PR #1670, opened 2016-04-07
Add message when no plugins found with tns plugin find

TsvetanMilanov:
When there are no plugins found with tns plugin find/search command the cli should show message for no plugins found instead of showing empty table with headers.

Fixes https://github.com/NativeScript/nativescript-cli/issues/1665#issuecomment-206720900


lib/commands/plugin/find-plugins.ts
```diff
@@ -54,6 +58,8 @@ export class FindPluginsCommand implements ICommand {
 			let pluginsToDisplay: string[] = allPluginsNames.splice(0, count);
 			let shouldDisplayMorePlugins: boolean = true;
 
+			this.$logger.out("Available NativeScript plugins");
```
rosen-vladimirov: you can add `:` at the end.

rosen-vladimirov:
:+1:  after green build

PR #1671, opened 2016-04-07
Add a --bundle option that disables node_modules -> tns_modules update.

hdeshev:
Delete `<platform/app>/tns_modules`, if present, and make sure we restore it
on subsequent runs without the `--bundle` option set.

Ping @rosen-vladimirov

### DIFF
--- a/lib/services/platform-service.ts
+++ b/lib/services/platform-service.ts
@@ -280,11 +280,16 @@ export class PlatformService implements IPlatformService {
 
 			platformData.platformProjectService.prepareProject().wait();
 
-			// Process node_modules folder
 			let appDir = path.join(platformData.appDestinationDirectoryPath, constants.APP_FOLDER_NAME);
 			try {
 				let tnsModulesDestinationPath = path.join(appDir, constants.TNS_MODULES_FOLDER_NAME);
-				this.$broccoliBuilder.prepareNodeModules(tnsModulesDestinationPath, platform, lastModifiedTime).wait();
+				if (!this.$options.bundle) {
+					// Process node_modules folder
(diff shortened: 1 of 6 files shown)


hdeshev:
Figured an easy way to add a test too! :balloon:

lib/services/platform-service.ts
```diff
@@ -230,6 +230,7 @@ export class PlatformService implements IPlatformService {
 			let lastModifiedTime = this.$fs.exists(appDestinationDirectoryPath).wait() ?
 				this.$fs.getFsStats(appDestinationDirectoryPath).wait().mtime : null;
 
+
```
rosen-vladimirov: Multiple blank lines are not allowed (this will fail the build)

lib/tools/broccoli/builder.ts
```diff
@@ -125,5 +131,11 @@ export class Builder implements IBroccoliBuilder {
 
 		}).future<void>()();
 	}
+
+	public cleanNodeModules(absoluteOutputPath: string, platform: string): IFuture<void> {
+		return (() => {
+			shelljs.rm("-rf", absoluteOutputPath);
```
rosen-vladimirov: This is not async operation, so you can change the return type of the method to be `void` instead of `IFuture<void>` and remove the future wrapper.

hdeshev:
Fixed both the whitespace and Future issues.

rosen-vladimirov:
:+1:

jasssonpet:
By the way, when running with the `bundle` option, the command does not process NativeScript plugins which can result to `Can't find variable: TNSLabel` when the tns-core-modules haven't been prepared yet.

Issue #979, opened 2015-09-29
Android APK Deployment: Storage Errors Unreported


jbristowe:
The NativeScript CLI fails to report a deployment failure via ADB if the Android emulator storage becomes full. This is non-optimal and creates a confusing situation where developers think the APK has been cached.

Request: Ensure that ADB deployment failures are reported through the CLI.

Fatme:
Related: https://github.com/NativeScript/nativescript-cli/issues/1060

Plamen5kov:
@teobugslayer, @rosen-vladimirov 
i think this can be closed.

rosen-vladimirov:
Nope, we've not implemented parsing of adb failures. It's in my TO-DO list, I'll raise the priority of this feature.

PR #1678, opened 2016-04-13
Add Android debug bridge result handler

TsvetanMilanov:
The android debug bridge result handler checks if there are any errors after executing adb command.
Errors are copied from https://github.com/android/platform_frameworks_base/blob/master/core/java/android/content/pm/PackageManager.java and the handler checks if the result contains any of the errors.

The non-device Android debug bridge must be separated from the device-specific Android debug bridge. Non-device Android debug bridge can be injected through constructor and the device-specific Android debug bridge can be resolved using the $injector. This way the device-specific Android debug bridge will be unique for each device.

The adb commands should be executed through the custom adb wrapper to use the error handling.

Implements https://github.com/NativeScript/nativescript-cli/issues/979


rosen-vladimirov:
:+1:

PR #1679, opened 2016-04-14
Introduce `hooks` option

rosen-vladimirov:
When `--no-hooks` is passed on command line, the hooks will not be executed.


Mitko-Kerezov:
LGMT :+1:

PR #1681, opened 2016-04-15
Add devDependencies when adding Android platform

rosen-vladimirov:
In order to support Static Binding Generation, we have to add some dev Dependencies to the project when platform Android is added.
As some of them do not work with Nodejs 0.x, we have to limit this functionality only to users who are using node 4.2.1 (the earliest version of 4.x that we support) or later.
Show warning in case the user has older node version.
Remove --staticBindings flag as it's turned on by default now.

### DIFF
--- a/lib/services/android-project-service.ts
+++ b/lib/services/android-project-service.ts
@@ -16,6 +16,18 @@ export class AndroidProjectService extends projectServiceBaseLib.PlatformProject
 	private static VALUES_VERSION_DIRNAME_PREFIX = AndroidProjectService.VALUES_DIRNAME + "-v";
 	private static ANDROID_PLATFORM_NAME = "android";
 	private static MIN_RUNTIME_VERSION_WITH_GRADLE = "1.3.0";
+	private static MIN_REQUIRED_NODEJS_VERSION_FOR_STATIC_BINDINGS = "4.2.1";
+	private static REQUIRED_DEV_DEPENDENCIES = [
+		{ name: "babel-traverse", version: "^6.4.5"},
+		{ name: "babel-types", version: "^6.4.5"},
+		{ name: "babylon", version: "^6.4.5"},
+		{ name: "filewalker", version: "^0.1.2"},
+		{ name: "lazy", version: "^1.0.11"}
+	];
+
+	private get sysInfoData(): ISysInfoData {
+		return this.$sysInfo.getSysInfo(path.join(__dirname, "..", "..", "package.json")).wait();
+	}
 
 	private _androidProjectPropertiesManagers: IDictionary<IAndroidProjectPropertiesManager>;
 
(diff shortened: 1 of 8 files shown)


rosen-vladimirov:
ping @Mitko-Kerezov , @TsvetanMilanov , @ikoevska

rosen-vladimirov:
ping @Plamen5kov

Plamen5kov:
:+1:

lib/services/android-project-service.ts
```diff
@@ -16,6 +16,23 @@ export class AndroidProjectService extends projectServiceBaseLib.PlatformProject
...
+		{ name: "babel-traverse", version: "^6.4.5"},
+		{ name: "babel-types", version: "^6.4.5"},
+		{ name: "babylon", version: "^6.4.5"},
+		{ name: "filewalker", version: "^0.1.2"},
+		{ name: "lazy", version: "^1.0.11"}
+	];
+
+	private _sysInfoData: ISysInfoData;
+	private get sysInfoData(): ISysInfoData {
+		if(!this._sysInfoData) {
```
rosen-vladimirov: remove this `if`, getSysInfo is caching

lib/services/android-project-service.ts
```diff
@@ -126,10 +144,28 @@ export class AndroidProjectService extends projectServiceBaseLib.PlatformProject
...
+					"save-exact": true,
+					"silent": true
+				};
+
+				_.each(AndroidProjectService.REQUIRED_DEV_DEPENDENCIES, (dependency: any) =>
+					this.$npm.install(`${dependency.name}@${dependency.version}`, this.$projectData.projectDir, npmConfig).wait()
+				);
+			} else {
+				this.$logger.printMarkdown(`As you are using Node.js \`${this.sysInfoData.nodeVer}\` Static Binding Generator will be turned off.
+				Upgrade your Node.js to ${AndroidProjectService.MIN_REQUIRED_NODEJS_VERSION_FOR_STATIC_BINDINGS} or later, so you can use this feature.`);
```
rosen-vladimirov: move this to the line above :)

rosen-vladimirov:
:+1: from @Mitko-Kerezov  after fixing comments

Mitko-Kerezov:
run ci

lib/services/android-project-service.ts
```diff
@@ -126,10 +139,28 @@ export class AndroidProjectService extends projectServiceBaseLib.PlatformProject
...
+					"save-dev": true,
+					"save-exact": true,
+					"silent": true
+				};
+
+				_.each(AndroidProjectService.REQUIRED_DEV_DEPENDENCIES, (dependency: any) =>
+					this.$npm.install(`${dependency.name}@${dependency.version}`, this.$projectData.projectDir, npmConfig).wait()
+				);
+			} else {
+				this.$logger.printMarkdown(` As you are using Node.js \`${this.sysInfoData.nodeVer}\` Static Binding Generator will be turned off.` +
```
ikoevska: Static Binding Generator will be turned off because you are using Node.js \`${this.sysInfoData.nodeVer}\` .

To use the Static Binding Generator, upgrade your Node.js to ${AndroidProjectService.MIN_REQUIRED_NODEJS_VERSION_FOR_STATIC_BINDINGS} or later.

PR #1657, opened 2016-03-30
Add warning for new versions of nativescript components

TsvetanMilanov:
When executing tns doctor outside of project if the nativescript cli needs update it will print information for the available update.
If tns doctor is executed in project it will check the installed runtimes versions and the nativescript core modules version and print information if an update is needed.
Add tns info command to display versions information for NativeScript CLI, core modules and runtimes.

Implements #1431

PR #1684, opened 2016-04-18
Add warning for new versions of nativescript components

TsvetanMilanov:
When executing tns doctor outside of project if the nativescript cli needs update it will print information for the available update.
If tns doctor is executed in project it will check the installed runtimes versions and the nativescript core modules version and print information if an update is needed.
Add tns info command to display versions information for NativeScript CLI, core modules and runtimes.


TsvetanMilanov:
Approved https://github.com/NativeScript/nativescript-cli/pull/1657

PR #1682, opened 2016-04-15
Disable fast livesync with hook

TsvetanMilanov:
Add before livesync hook to be able to control the livesync data. Added property in livesync service to force full sync which will be passed to the livesync data.

PR #1686, opened 2016-04-18
Disable fast livesync with hook

TsvetanMilanov:
Add before livesync hook to be able to control the livesync data. Added property in livesync service to force full sync which will be passed to the livesync data.


TsvetanMilanov:
Approved in https://github.com/NativeScript/nativescript-cli/pull/1682

PR #1692, opened 2016-04-19
Fix build for iOS

rosen-vladimirov:
Cocoapods have versions like 1.0.0.beta.8, which are not valid semver versions.
Use only the major.minor.patch part of the versions as all 1.0.0 betas work correctly with XCode 7.3

Revert special code for XCode 7.2 that adds `-destination` flag when building for iOS Simulator. Revert to original code, as it's working fine with all other XCode versions.

### DIFF
--- a/lib/services/ios-project-service.ts
+++ b/lib/services/ios-project-service.ts
@@ -677,6 +677,13 @@ export class IOSProjectService extends projectServiceBaseLib.PlatformProjectServ
 			let cocoapodsVer = this.$sysInfo.getSysInfo(this.$staticConfig.pathToPackageJson).wait().cocoapodVer,
 				xcodeVersion = this.$xcodeSelectService.getXcodeVersion().wait();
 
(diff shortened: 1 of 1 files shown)


lib/services/ios-project-service.ts
```diff
@@ -677,6 +677,13 @@ export class IOSProjectService extends projectServiceBaseLib.PlatformProjectServ
 			let cocoapodsVer = this.$sysInfo.getSysInfo(this.$staticConfig.pathToPackageJson).wait().cocoapodVer,
 				xcodeVersion = this.$xcodeSelectService.getXcodeVersion().wait();
 
+			if(cocoapodsVer && !semver.valid(cocoapodsVer)) {
```
Mitko-Kerezov: This could be shortened to `!semver.valid(cocoapodsVer)` only

Mitko-Kerezov:
:+1: after fixing minor comment

rosen-vladimirov:
run ci

Issue #1419, opened 2016-01-21
Cannot find module 'colors' during uninstall


vchimev:
I get this on `npm` 3.5.1 version, have not tried it on `npm` 2.*.

```
VCHIMEVMAC:mobile-master vchimev$ npm uninstall -g nativescript

> nativescript@1.5.2 preuninstall /usr/local/lib/node_modules/nativescript
> node preuninstall.js

Failed to complete all pre-uninstall steps. 
{ [Error: Command failed: /bin/sh -c node bin/nativescript.js dev-preuninstall
module.js:340
    throw err;
    ^

Error: Cannot find module 'colors'
    at Function.Module._resolveFilename (module.js:338:15)
    at Function.Module._load (module.js:289:25)
    at Module.require (module.js:366:17)
    at require (module.js:385:17)
    at Object.verifyNodeVersion (/usr/local/lib/node_modules/nativescript/lib/common/verify-node-version.js:6:5)
    at Object.<anonymous> (/usr/local/lib/node_modules/nativescript/lib/nativescript-cli.js:3:41)
    at Module._compile (module.js:425:26)
    at Object.Module._extensions..js (module.js:432:10)
    at Module.load (module.js:356:32)
    at Function.Module._load (module.js:313:12)
]
```

teobugslayer:
I can repro on Mac / El Capitan.

PR #1693, opened 2016-04-19
Hide npm preuninstall errors

TsvetanMilanov:
Due to issue in npm versions 3.0, 3.5.1, 3.7.3 (may be more 3.x versions) the NativeScript node_modules are removed before the preuninstall script is executed and it fails.
The issue is logged in the npm repository: https://github.com/npm/npm/issues/8806 but it is not fixed in version 3.1.1 as commented in the discusion.

Workaround for https://github.com/NativeScript/nativescript-cli/issues/1419

### DIFF
--- a/preuninstall.js
+++ b/preuninstall.js
@@ -2,8 +2,9 @@ var child_process = require("child_process");
 
 var child = child_process.exec("node bin/nativescript.js dev-preuninstall", function (error) {
 	if (error) {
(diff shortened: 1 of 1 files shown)


rosen-vladimirov:
:+1:  after updating common-lib sha

TsvetanMilanov:
run ci

rosen-vladimirov:
run ci

PR #1694, opened 2016-04-19
Do not symlink build-tools directory

rosen-vladimirov:
Due to latest changes in Android runtime, we can no longer symlink `build-tools` directory from the runtime when `--symlink` is passed.
There are two major reasons:
- There are `.js` files in the build-tools, which use `require(<smth>)`. This is failing as node's searching for the installed modules from the original location of the file.
- During build, new files are generated in the build-tools directory. As they are unique (each one is project specific), we cannot symlink them.


Mitko-Kerezov:
:+1:

vchimev:
:+1:

PR #1697, opened 2016-04-20
Track sessions count per analytics project

rosen-vladimirov:
Each sessions count should be tracked per project for which it is started. Currently we were using the same counter for all sessions.
Fix this by using separate counter for each project.
Remove the old key ("SESSIONS_STARTED") and use it as current value for the new key (`SESSIONS_STARTED_<analytics_project_key>`) in case it existed.


Mitko-Kerezov:
:+1:

PR #1699, opened 2016-04-20
Fix the updating of changed node modules in platform folder

TsvetanMilanov:
When getting the stats for file from node_modules in the broccoli builder the event loop gets messed up from the Glob fiber and the cli fiber. That's why we need lock in the Glob fiber.


rosen-vladimirov:
:+1:  after green build

PR #1701, opened 2016-04-20
Fix debugging for simulator when --emulator is not passed

rosen-vladimirov:
When --emulator is not passed, CLI will try to debug on iOS Simulator in case there's one started (or CLI will start it).
However the iOS debug service calls methods for debug on device and it fails.
Use correct methods when the detected devices is emulator.

Fixes `tns debug ios` when no devices are connected


TsvetanMilanov:
:+1:

rosen-vladimirov:
run ci

PR #1702, opened 2016-04-20
Fix livesync for ios devices

rosen-vladimirov:
Currently we are creating new instance of the AndroidLiveSyncService and IOSLiveSyncService. When `--watch` option is used, CLI gets new instance for each change of a file. This leads to opening new socket every single time. The second changed file leads to process exit as the inspector runtime logic is triggered many times.
Fix this by caching the instances per device.


Mitko-Kerezov:
:+1: after green builld

rosen-vladimirov:
run ci

PR #1703, opened 2016-04-21
Fix new Android emulator fail to start application

TsvetanMilanov:
After updating the Android default emulator the CLI starts it and cannot deploy the application on it. This is caused by the change in the OK message after the net.connect(port) client connects to the emulator.


rosen-vladimirov:
:+1:  after rebasing and updating sha of common lib

PR #1700, opened 2016-04-20
Include custom platform template when using tns init -f

TsvetanMilanov:
When the `tns-<platform>` key in the package.json of nativescript project has custom template the template key must be kept and the version key should be added.

PR #1704, opened 2016-04-21
Include custom platform template when using tns init -f

TsvetanMilanov:
When the tns-<platform> key in the package.json of nativescript project has custom template the template key must be kept and the version key should be added.

Approved in https://github.com/NativeScript/nativescript-cli/pull/1700


rosen-vladimirov:
:+1:  after green build

PR #1705, opened 2016-04-21
Add xcproj check to tns doctor

Mitko-Kerezov:
Add an entry in the docs as well.
In addition: hide warnings from CocoaPods whenever running `$ tns doctor`.
Ping @rosen-vladimirov for code review


rosen-vladimirov:
:+1:  after green build

PR #1706, opened 2016-04-21
Fix printing of simulator logs when livesyncing

rosen-vladimirov:
Currenly when `tns livesync ios --watch` is used for iOS Simulator, CLI starts the application on device and prints filtered logs.
The filter is based on the process id of the currently started application. During livesync, the application is restarted, which changes the PID of the running process.
In this case we should start filtering based on the new PID.


Mitko-Kerezov:
:+1:

PR #1707, opened 2016-04-21
Fix get current tns-core-modules version

TsvetanMilanov:
When getting the current version of the tns-core-modules the version in project's package.json can be `^<version>` or `*` so the version in the package.json of the installed tns-core-modules is the correct version.

### DIFF
--- a/lib/services/versions-service.ts
+++ b/lib/services/versions-service.ts
@@ -2,15 +2,21 @@
 "use strict";
 import * as constants from "../constants";
 import * as semver from "semver";
+import * as path from "path";
 import {createTable} from "../common/helpers";
 
 class VersionsService implements IVersionsService {
+	private static UP_TO_DATE_MESSAGE = "Up to date".green.toString();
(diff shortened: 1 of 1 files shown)


lib/services/versions-service.ts
```diff
@@ -119,7 +128,9 @@ class VersionsService implements IVersionsService {
 	public createTableWithVersionsInformation(versionsInformation: IVersionInformation[]): any {
 		let headers = ["Component", "Current version", "Latest version", "Information"];
 		let data: string[][] = [];
-		let upToDate: string = "Up to date".green.toString();
+		let upToDate = "Up to date".green.toString();
```
Mitko-Kerezov: These three variables could be made into static constants for this class

Mitko-Kerezov:
In addition, I don't see the need to default to an empty string [here](https://github.com/NativeScript/nativescript-cli/pull/1707/files#diff-253a90a336fbee7d28e4db2632cb91ecR138)

Mitko-Kerezov:
:+1: after addressing minor comments - great job!

PR #1708, opened 2016-04-22
Project creation improvements

rosen-vladimirov:
### Update License Copyright
Use correct copyright message.

### 	Improve project creation
Add required fields in `package.json` when project is created. This way there will be no warnings during `tns create <app>`.
Also improve the message (use markdown).

### 	Add message on postinstall
Add message on postinstall to show that NS CLI is installed successfully and where to follow us.


Mitko-Kerezov:
:+1:

PR #1710, opened 2016-04-22
Fix output parsing when connecting to started emulator

TsvetanMilanov:
When starting Android emulator the output from the `net` `data` event always contains \r\n and when using EOL to split it on Mac or Linux it contains \r and then the parsing fails.


rosen-vladimirov:
:+1:  after getting latest sha of common-lib

PR #1712, opened 2016-04-22
Fix livesync --watch for ios devices

rosen-vladimirov:
Do not try creating the socket again when `--watch` is used. It's already created buddy, keep using it.


Mitko-Kerezov:
:+1:

vchimev:
:+1:

PR #1715, opened 2016-04-25
Remove semver validation for CocoaPods version string

Mitko-Kerezov:
Latest betas of CocoaPods include 1.0.0.betaX which is not a valid semver string. These versions are the only ones however, which work with Xcode 7.3 out-of-the-box.

Ping @rosen-vladimirov @vchimev @dtopuzov


rosen-vladimirov:
:+1:

PR #1722, opened 2016-04-26
Fix iOS socket issues

rosen-vladimirov:
Sometimes when `--watch` is used, the socket is closed from the other side. We have to attach to close event of the socket and set our internal variable to null. This way next time when a file is changed, we'll create a new socket.
Sometimes on SIGINT, SIGTERM or process.exit we do not have `this.socket`, but we try to call `destroy` method of the socket, which fails.
Also make sure correct events are added when livesync to device.


rosen-vladimirov:
ping @Mitko-Kerezov , @TsvetanMilanov , @vchimev , @dtopuzov

Mitko-Kerezov:
:+1: after QA approval

dtopuzov:
:+1:

vchimev:
:+1:

Mitko-Kerezov:
run ci

vchimev:
run ci

vchimev:
run ci

PR #1730, opened 2016-04-28
Remove node_modules when preparing nativescript plugins

rosen-vladimirov:
When preparing nativescript plugins, we copy the whole directory to the native project.
This leads to a problem that node_modules dir of such plugins is also copied, while it should be flattened.
CLI had already flatten it, but the copy of node_modules is never deleted and we receive `.aar` files in iOS projects for example.


rosen-vladimirov:
ping @Mitko-Kerezov @TsvetanMilanov

dtopuzov:
:+1:

PR #1736, opened 2016-05-03
Check Android platform dependencies when adding platform

TsvetanMilanov:
If the current project already have some of the Android platform dependencies installed the installed version should be checked. If the installed version satisfies the Android platform the CLI should not install this dependency, if not the CLI should throw error to inform the user about the current version and the required version.


rosen-vladimirov:
:+1:

Issue #1097, opened 2015-10-22
The ANDROID_HOME environment variable is not set


nraboy:
I am receiving the following message when I try to add the Android platform:

```
The ANDROID_HOME environment variable is not set or it points to a non-existent directory. You will not be able to perform any build-related operations for Android.
```

When I run `echo $ANDROID_HOME` on my Mac, it prints out the following:

```
/Users/nraboy/Library/Android-SDK
```

I validate this path by executing `cd $ANDROID_HOME` followed by a `ls` and everything seems normal.

My Android installation is valid because it works fine with React Native, Ionic Framework, and native Android development.  

Here is the process I used for installing NativeScript on my machine:

```
sudo curl -o- https://raw.githubusercontent.com/creationix/nvm/v0.29.0/install.sh | bash
nvm install 0.12.7
nvm use 0.12.7
sudo npm install -g nativescript
```

Yes, I am using NVM to manage my Node.js versions.

Here are my full `--log trace` for Android:

```
$ tns platform add android --log trace
Exec uname -a 
 stdout: Darwin macbookpro 14.5.0 Darwin Kernel Version 14.5.0: Wed Jul 29 02:26:53 PDT 2015; root:xnu-2782.40.9~1/RELEASE_X86_64 x86_64
 
 stderr: 
Exec npm -v 
 stdout: 2.11.3
 
 stderr: 
spawn: java -version
Exec node-gyp -v 
 stdout:  
 stderr: /bin/sh: node-gyp: command not found

Exec xcodebuild -version 
 stdout: Xcode 7.0.1
Build version 7A1001
 
 stderr: 
Exec pod --version 
 stdout:  
 stderr: /bin/sh: pod: command not found

Exec adb version 
 stdout: Android Debug Bridge version 1.0.32
Revision eac51f2bb6a8-android
 
 stderr: 
Exec android -h 
 stdout: 
       Usage:
       android [global options] action [action options]
       Global options:
  -h --help       : Help on a specific command.
  -v --verbose    : Verbose mode, shows errors, warnings and all messages.
     --clear-cache: Clear the SDK Manager repository manifest cache.
  -s --silent     : Silent mode, shows errors only.

                                                                    Valid
                                                                    actions
                                                                    are
                                                                    composed
                                                                    of a verb
                                                                    and an
                                                                    optional
                                                                    direct
                                                                    object:
-    sdk              : Displays the SDK Manager window.
-    avd              : Displays the AVD Manager window.
[...]

PR #1744, opened 2016-05-08
Bash users with ~/.bash_profile don't pickup environment variable changes in ~/.profile

ejsuncy:
Per the man pages for bash, "it looks for `~/.bash_profile`, `~/.bash_login`, and `~/.profile`, in that order, and reads and executes commands from the **first** one that exists and is readable."

Since some users already have `~/.bash_profile` or `~/.bash_login`, the third file `~/.profile` is never picked up for them. The installation script writes the `ANDROID_HOME` and the `JAVA_HOME` environment variables to `~/.profile`, so the `bin/tns` and `bin/nativescript` executables should explicitly pull in those variables.

This may resolve issues related to #1097.

### DIFF
--- a/bin/nativescript
+++ b/bin/nativescript
@@ -1,4 +1,5 @@
 #!/bin/sh
 
+source ~/.profile
 AB_DIR="`dirname \"$0\"`"
 node "$AB_DIR/nativescript.js" "$@"
--- a/bin/tns
+++ b/bin/tns
@@ -1,4 +1,5 @@
 #!/bin/sh
 
+source ~/.profile
 AB_DIR="`dirname \"$0\"`"
 node "$AB_DIR/nativescript.js" "$@"


rosen-vladimirov:
Hi @ejsuncy ,
I'm not really sure that's the correct fix. In fact the `~/.bash_profile` should call `source ~/.profile`.
We do not want CLI to mess the loaded environment files, it's up to the user to decide which one are required. So I'm gonna close this PR, but thank you for the really great description of the issue and submitting this PR.

ejsuncy:
In that case, maybe it would be helpful to end users to modify the message at the end of the installation script, something to the effect of:

 The ANDROID_HOME and JAVA_HOME environment variables have been added to your ~/.profile. If you use a different profile, be sure to add `source ~/.profile` to it. Restart the terminal to use them.`

majodi:
Took a while to find out what was going on here but/so like suggested by ejsuncy I also advise to get this info more prominent somewhere.

So when getting errors/warnings/messages about The ANDROID_HOME/JAVA_HOME environment variable not set or it points to a non-existent directory:

it could be that Terminal/bash is getting its settings from "~/.bash_profile" only and ignoring what was set in  "~/.profile". To fix this add a "source ~/.profile" line to the "~/.bash_profile". Then it will include the "~/.profile" and set the necessary environment variables. (restart terminal)

Granight:
Thanks, this fixed my issue too!

dragogio:
My issue too! Thanks.

jgodson:
Fixed my issue as well!

serdem420:
@majodi +1

PR #1749, opened 2016-05-11
Fix build for Android with custom platform template

rosen-vladimirov:
When you use custom platform template, we still need build-tools dir from the runtime. Copy it no matter of the platform template.


TsvetanMilanov:
:+1:

PR #1753, opened 2016-05-12
Fix xcproj touch

rosen-vladimirov:
Fix using xcproj for projects with space in the name. Currently this fails.


Mitko-Kerezov:
:+1:

PR #1754, opened 2016-05-12
Fix Segmentation fault

TsvetanMilanov:
When getting the adb file path need to use childProcess not our adb wrapper becaulse it will fail with `Segmentation fault node "<file-path>" "$@"`.


TsvetanMilanov:
run ci

vchimev:
run ci

Mitko-Kerezov:
👍

PR #1755, opened 2016-05-12
Change adb result handling in adb wrapper

TsvetanMilanov:
Some adb commands like adb start-server return array of strings not object with stdout need to check if stdout is undefined or null because if it is empty string it will return incorrect result.


TsvetanMilanov:
run ci

PR #1757, opened 2016-05-13
Fixed a path not found issue where the variable was not being wrapped in quotes

NickBranstein:
Fixed a path not found issue where the variable was not being wrapped
in quotes

### DIFF
--- a/setup/native-script.ps1
+++ b/setup/native-script.ps1
@@ -99,11 +99,11 @@ if (!$env:JAVA_HOME) {
 # following commands are separated in case of having to answer to license agreements
 # the android tool will introduce a --accept-license option in subsequent releases
 $androidExecutable = [io.path]::combine($env:ANDROID_HOME, "tools", "android")
-echo y | cmd /c "$androidExecutable" update sdk --filter "platform-tools" --all --no-ui
-echo y | cmd /c "$androidExecutable" update sdk --filter "tools" --all --no-ui
-echo y | cmd /c "$androidExecutable" update sdk --filter "android-23" --all --no-ui
-echo y | cmd /c "$androidExecutable" update sdk --filter "build-tools-23.0.2" --all --no-ui
-echo y | cmd /c "$androidExecutable" update sdk --filter "extra-android-m2repository" --all --no-ui
+echo y | cmd /c "($androidExecutable)" update sdk --filter "platform-tools" --all --no-ui
+echo y | cmd /c "($androidExecutable)" update sdk --filter "tools" --all --no-ui
+echo y | cmd /c "($androidExecutable)" update sdk --filter "android-23" --all --no-ui
+echo y | cmd /c "($androidExecutable)" update sdk --filter "build-tools-23.0.2" --all --no-ui
+echo y | cmd /c "($androidExecutable)" update sdk --filter "extra-android-m2repository" --all --no-ui
 
 
 Write-Host -ForegroundColor Green "This script has modified your environment. You need to log off and log back on for the changes to take effect."


rosen-vladimirov:
Hi @NickBranstein 
We've tried your fix on our side, but it looks like the quotes do not work for us. The original code works even when there's space in the directory name. May I ask you for more details on your case, steps to reproduce the unexpected behavior will be a great help for us.
Also what's your OS version and PowerShell version?

NickBranstein:
Thanks @rosen-vladimirov - I'm running Windows 10 x64. Powershell version 5.

When I ran the following from (which runs the install script), everything works until the end where I get Path not found errors.

`start-process -FilePath PowerShell.exe -Verb Runas -Wait -ArgumentList "-NoProfile -ExecutionPolicy Bypass -Command iex ((new-object net.webclient).DownloadString('https://raw.githubusercontent.com/NativeScript/nativescript-cli/production/setup/native-script.ps1'))"`

Strangely enough I tried this on several other Windows 10 machines with the same problem. I also had to manually fix my ANDROID_HOME and JAVA_HOME variables because the script had pointed them to the wrong spot (different problem but not sure what is causing that).

Seems like after I fixed those variables and I am rerunning the script today I no longer need to add the parens. After a little more investigation I think this is because it somehow found the wrong path. I ran this on a new VM i spun up on Azure and it had another version of the JDK already installed in C:\Program Files (x86)\Java - I'm going to try and uninstall the versions and give the script a shot again. Sorry for any confusion!

NickBranstein:
No such luck still fails for me. Perhaps this should be an issue now because I think this fix won't appear to work in all cases. Here is the error I get on a new Windows 10x64 install. Powershell version 5.

error

Mitko-Kerezov:
@NickBranstein
We can't seem to reproduce the issue on our side - tried it in. According to [ss64](http://ss64.com/ps/syntax-operators.html) parentheses are used for grouping expressions only. Meaning that `"$androidExecutable"` will expand the variable and the result will be the variable's value, while `"($androidExecutable)"` results in the variable's value inside parentheses.

You said your `ANDROID_HOME` and `JAVA_HOME` variables were wrong - could you tell us what their values were?

My guess is that you had an `android` executable, but it was not added to your `PATH` environment variable. In this case, the script defaults to `"$env:localappdata\Android\android-sdk"`

NickBranstein:
`ANDROID_HOME` did point to the default user profile location.
`JAVA_HOME` pointed to an older version of the java framework in `c:\program files (x86)\java\` 

Strangely enough the android sdk was installed but it was installed to `c:\program files\android\` (which is where I had to update the path to.

Mitko-Kerezov:
Looks like the following happened:
 1. You installed the android sdk `c:\program files\android\` but didn't add `c:\program files\android\tools` in your `PATH` environment variable.
 2. The script detected you didn't have the `ANDROID_HOME` environment variable set and tried to set it for you.
  1. The script tries to locate the android sdk but fails because it hasn't been added to the `PATH` (hence, it can't be executed unless a full path is provided) in which case it defaults to `"$env:localappdata\Android\android-sdk"`
  2. Your `ANDROID_HOME` environment variable is set to `"$env:localappdata\Android\android-sdk"`

Then later on when the script tries to run the executable `android`, using the `ANDROID_HOME` variable it fails because the variable's value is plain wrong.

The proper fix for this problem (and all other related future problems) would be to verify that the `android` executable can be run and that it is where `ANDROID_HOME` says it is.

If that seems reasonable to you I'll address it in a separate pull request and we can close this one?

NickBranstein:
Yes I think it should be a separate pull request thanks. I think the fix is to verify the location of the android executable properly and create the variable like you said.  Thanks!

PR #1760, opened 2016-05-14
Fix add android platform dependencies

TsvetanMilanov:
When checking if platform dependency is installed in the current project before adding the platform need to check if the current version is valid or check in the node_modules directory for the installed plugin version.

### DIFF
--- a/lib/services/android-project-service.ts
+++ b/lib/services/android-project-service.ts
@@ -149,12 +149,23 @@ export class AndroidProjectService extends projectServiceBaseLib.PlatformProject
 				let projectPackageJson: any = this.$fs.readJson(this.$projectData.projectFilePath).wait();
 
 				_.each(AndroidProjectService.REQUIRED_DEV_DEPENDENCIES, (dependency: any) => {
-					let dependencyVersionInProject = projectPackageJson.dependencies[dependency.name] || projectPackageJson.devDependencies[dependency.name];
+					let dependencyVersionInProject = (projectPackageJson.dependencies && projectPackageJson.dependencies[dependency.name]) ||
+						(projectPackageJson.devDependencies && projectPackageJson.devDependencies[dependency.name]);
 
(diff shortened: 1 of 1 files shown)


lib/services/android-project-service.ts
```diff
@@ -149,12 +149,23 @@ export class AndroidProjectService extends projectServiceBaseLib.PlatformProject
...
 					if (!dependencyVersionInProject) {
 						this.$npm.install(`${dependency.name}@${dependency.version}`, this.$projectData.projectDir, npmConfig).wait();
-					} else if (!semver.satisfies(dependencyVersionInProject, dependency.version)) {
-						this.$errors.failWithoutHelp(`Your project have installed ${dependency.name} version ${dependencyVersionInProject} but Android platform requires version ${dependency.version}.`);
+					} else {
+						let cleanedVerson = semver.clean(dependencyVersionInProject);
+
+						// The plugin version is not valid. Check node_modules for the valid version.
+						if (!cleanedVerson) {
+							dependencyVersionInProject = this.$fs.exists(path.join(this.$projectData.projectDir, "node_modules", dependency.name, "package.json")).wait() &&
```
rosen-vladimirov: "node_modules" -> `constants.NODE_MODULES_FOLDER_NAME`
"package.json" -> `constants.PACKAGE_JSON_FILE_NAME`

lib/services/android-project-service.ts
```diff
@@ -149,12 +149,23 @@ export class AndroidProjectService extends projectServiceBaseLib.PlatformProject
...
+								this.$fs.readJson(path.join(this.$projectData.projectDir, "node_modules", dependency.name, "package.json")).wait().version;
```
rosen-vladimirov: Extract this to variable: `path.join(this.$projectData.projectDir, "node_modules", dependency.name, "package.json")`

rosen-vladimirov:
:+1:

PR #1772, opened 2016-05-20
Shorten setup scripts links

Mitko-Kerezov:
Where necessary, shorten:
* https://raw.githubusercontent.com/NativeScript/nativescript-cli/production/setup/native-script.rb to https://www.nativescript.org/setup/mac
* https://raw.githubusercontent.com/NativeScript/nativescript-cli/production/setup/native-script.ps1 to https://www.nativescript.org/setup/win

The shorthands actually redirect to the longer urls.

Only the urls which are used for starting the scripts have been shortened. I prefer we keep the urls in the code unchanged.


rosen-vladimirov:
:+1:

PR #1779, opened 2016-05-26
package.json indent preserved

### LINKED ISSUE #1752: Use spaces instead of tabs in package.json for consistency with npm
npm maintains packages.json files with spaces for indentation instead of tabs, tns uses tabs instead of spaces.

This causes entire file to change if you use both tools to manipulate it, which then leads to hard to read diffs.

Propose that tns use spaces for consistency with npm, with the justification that npm is the de facto tool for manipulating package.json files.

enchev:
Fixed #1752


rosen-vladimirov:
:+1:

Issue #1777, opened 2016-05-25
Invalid XML kills livesync


bradmartin:
Using livesync, if you bork the XML and save changes. The app crashes, which is fine, but then after correcting the invalid XML you have to do a new build for Android to get the app to livesync again. Big time killer. It's come up on the slack channel a couple of time and confuses people who don't know how to get around it.

enchev:
Hey @bradmartin,

Maybe we should stop deploying broken XML at all? Right now you will get only warning from the CLI if the XML is invalid however we can change that to real error which will stop the current task (livesync, run, etc.).

Hey @atanasovg @dtopuzov @hdeshev @hshristov @PanayotCankov @rosen-vladimirov what do you think?

bradmartin:
Anything to prevent the need to rebuild and start livesync over I'm all
for. Sometimes I even forget to do that and get frustrated and I've been
around for awhile so I just don't want newer NativeScript devs to get
frustrated and give up :smile: I trust you guys to implement a good
solution. Most of the time for me my IDE might throw in an extra double
quote when im making quick attribute property changes. I know I can change
the IDE auto completion settings but that's besides the issue. So again
thanks for getting a discussion going so this can be resolved. Its my only
livesync complaint to be honest.

rosen-vladimirov:
Hi guys,
I believe we should **not stop** the livesync when there's invalid XML, but we should not livesync the file as well. My idea is to validate the file and tell the user: "Hey, this file's incorrect, fix it, save it and we'll do the job for you". So the user will not restart the operation, we'll check the file and if we consider it as valid, we'll send it to device.

enchev:
We have [similar issue](https://github.com/NativeScript/nativescript-cli/issues/1524) for TypeScript compilation. In both cases (XML is not well formed and TypeScript cannot be compiled) in my opinion we should stop the transfer (not the watch itself). When you fix all errors your files will be transferred automatically again.

hdeshev:
I also think we should alert the user by changing the app behavior when we prevent a sync due to an error. Here's how RN does it for app crashes:

![](https://i.stack.imgur.com/Q0WNU.png)

We should use that for compile errors, xml validation failures, etc.

enchev:
Hey,

I've played a bit with various approaches and here is what I've found:

### Approach A (Stop livesync transfer in case of error)
+ (pros): will notify the developer in the livesync console about errors in any file in the application
[...]

PR #1780, opened 2016-05-27
XML validation fixed

enchev:
Related to #1777


rosen-vladimirov:
Maybe you should update this test as it expects to return false when there's invalid `.xml`, but with current changes the code will throw error
https://github.com/NativeScript/nativescript-cli/blob/master/test/platform-service.ts#L340

enchev:
run ci

enchev:
@rosen-vladimirov done!

rosen-vladimirov:
:+1:

PR #1781, opened 2016-05-27
Flatten scoped dependencies of dependency

TsvetanMilanov:
When some dependency of the project has scoped dependency we need to enumerate all the subdirectories to flatten all dependencies correctly.

Fixes https://github.com/NativeScript/nativescript-cli/issues/1783


rosen-vladimirov:
Great work, can you please add a new issue with short description, so the QAs can test it.
:+1:

PR #1789, opened 2016-05-31
Flatten scoped dependencies of dependency

rosen-vladimirov:
When some dependency of the project has scoped dependency we need to enumerate all the subdirectories to flatten all dependencies correctly.


Mitko-Kerezov:
👍

PR #1792, opened 2016-05-31
Allow the copy-to to copy .app folder for iOS

PanayotCankov:


rosen-vladimirov:
Linter fails:
```
>> lib/services/platform-service.ts[403, 1]: trailing whitespace
>> lib/services/platform-service.ts[375, 5]: forbidden var keyword
```

lib/services/platform-service.ts
```diff
@@ -368,15 +368,56 @@ export class PlatformService implements IPlatformService {
 				this.$errors.failWithoutHelp("Unable to find built application. Try 'tns build %s'.", platform);
 			}
 
-			this.$fs.ensureDirectoryExists(path.dirname(targetPath)).wait();
+			function ext(file: string): string {
+				return path.extname(file).toLowerCase();
+			}
+			function targetExt(file: string): string {
+				let e = ext(file);
+				return (e === ".ipa" || e === ".app") ? e : ".";
```
rosen-vladimirov: this `.` confuses me, I've started thinking that you expect to pass `--copy-to .`, while in fact the dot means everything that's not `.ipa` or `.app`
Can we change it to something else (not a merge stopper, I'm just wondering if we can find a better name)

lib/services/platform-service.ts
```diff
@@ -368,15 +368,56 @@ export class PlatformService implements IPlatformService {
...
+					this.$logger.trace(`Copying directory '${packageFile}' to '${targetPath}'...`);
+					shell.cp("-Rf", packageFile + "/*", targetPath);
+					this.$logger.info(`Copied directory '${packageFile}' to '${targetPath}'.`);
+					break;
+
+				case ".ipa->.":
+				case ".app->.":
+					this.$fs.ensureDirectoryExists(path.dirname(targetPath)).wait();
+					if (!this.$fs.exists(targetPath).wait() || !this.$fs.getFsStats(targetPath).wait().isDirectory()) {
+						this.$logger.trace(`Expected directory for --copy-to at ${targetPath}`);
```
rosen-vladimirov: in case I write:
```
$ tns build ios --emulator --copy-to /Users/vladimirov/test/myFile.txt`
```
this message will not be shown, but I'll also not receive my moved file. Consider using 
```TypeScript
this.$logger.warn("...");
```
or even
```TypeScript
this.$errors.failWithoutHelp("...");
```

lib/services/platform-service.ts
```diff
@@ -368,15 +368,56 @@ export class PlatformService implements IPlatformService {
...
+					this.$logger.trace(`Specified target path: '${targetPath}' is directory. Same filename will be used: '${sourceFileName}'.`);
+					targetPath = path.join(targetPath, sourceFileName);
+
+					this.$logger.trace(`Copying '${packageFile}' to '${targetPath}'...`);
+					shell.cp("-Rf", packageFile, targetPath);
+					this.$logger.info(`Copied '${packageFile}' to '${targetPath}'.`);
+					break;
+
+				case ".ipa->.app":
+					this.$logger.error(`Extension mismatch when --copy-to tries to copy '${packageFile}' into '${targetPath}'. To ensure the build produces an .app provide the --emulator flag.`);
```
rosen-vladimirov: this will print red message, but it will not throw real error, is this intended ?

lib/services/platform-service.ts
```diff
@@ -368,15 +368,56 @@ export class PlatformService implements IPlatformService {
...
+					break;
+
+				case ".app->.ipa":
+					this.$logger.error(`Extension mismatch when --copy-to tries to copy '${packageFile}' into '${targetPath}'. To ensure the build produces an .ipa provide the --for-device or --emulator flag.`);
```
rosen-vladimirov: same as above:
this will print red message, but it will not throw real error, is this intended ?

lib/services/platform-service.ts
```diff
@@ -368,15 +368,56 @@ export class PlatformService implements IPlatformService {
...
+					break;
 
-			if (this.$fs.exists(targetPath).wait() && this.$fs.getFsStats(targetPath).wait().isDirectory()) {
-				let sourceFileName = path.basename(packageFile);
-				this.$logger.trace(`Specified target path: '${targetPath}' is directory. Same filename will be used: '${sourceFileName}'.`);
-				targetPath = path.join(targetPath, sourceFileName);
+				default:
+					this.$logger.error(`Unsupported package for --copy-to at '${packageFile}'.`);
```
rosen-vladimirov: and here
this will print red message, but it will not throw real error, is this intended ?

lib/services/platform-service.ts
PanayotCankov: If it is not ".ipa" or ".app" it will be treated as directory path. I do not like the `.` either but couldn't figure out better while I was writing it. Other options are `*`, `?`, `<dir>`, `(dir)` etc. Would you prefer one of these?

lib/services/platform-service.ts
PanayotCankov: Nice catch. It should be error.

lib/services/platform-service.ts
rosen-vladimirov: ok, I do not have better ideas as well, so let's keep it simple (use `.`) :smiley_cat:

lib/services/platform-service.ts
PanayotCankov: Will this.$logger.fatal throw real error?

rosen-vladimirov: errors are thrown with `this.$errors.failWithoutHelp` or `this.$errors.fail` (the second one will print the help as well, which I believe is not what we need).

PanayotCankov: I think this.$errors.failWithoutHelp("..."); will suffice then :)

lib/services/platform-service.ts
PanayotCankov: I can refactor it to use enum?

lib/services/platform-service.ts
```diff
@@ -368,15 +368,56 @@ export class PlatformService implements IPlatformService {
...
+			switch(ext(packageFile) + "->" + targetExt(targetPath)) {
+				case ".ipa->.ipa":
+					this.$logger.trace(`Copying file '${packageFile}' to '${targetPath}'...`);
+					this.$fs.copyFile(packageFile, targetPath).wait();
+					this.$logger.info(`Copied file '${packageFile}' to '${targetPath}'.`);
+					break;
+
+				case ".app->.app":
+					this.$logger.trace(`Copying directory '${packageFile}' to '${targetPath}'...`);
+					shell.cp("-Rf", packageFile + "/*", targetPath);
```
PanayotCankov: And I will probably remove the "f" flag. I've played a little with `xcodebuild` and it throws if you specify output path that already exists so we should probably stick to that behavior as you initially suggested.

hshristov:
👍

PR #1793, opened 2016-06-01
Add warning for Node.js 0.10.x deprecation

rosen-vladimirov:
Warn users that support from version 0.10.x will be removed in version 2.2.0.


Mitko-Kerezov:
👍

Issue #1783, opened 2016-05-27
Not flattening scoped dependencies of  project dependency


TsvetanMilanov:
When the project has dependency **A** and **A** has scoped dependency **B**, `tns prepare <platform>` is not flattening **B**.

For example the **nativescript-angular** has dependency **@angular/core** and after `tns prepare <platform>` in the tns_modules folder there is no folder @angular with subfolder core.

rosen-vladimirov:
The problem occurs when in your project you have dependency (for example "nativescript-angular") and it has scopede dependencies as its own dependencies (in it's package.json some dependencies are defined as `@angular/<smth>`). In order to make all dependencies usable in NativeScript project, CLI must flatten the dependencies and send them to correct folder in platforms directory. For example for Android this is `platforms/android/src/main/assets/app/tns_modules/` - there you should find `@angular` directory in which you'll find all scoped packages.
However this was not working in CLI.

Steps to reproduce:
1) `$ tns create myApp --ng`
2) `$ cd myApp`
3) Edit project's package.json and in dependencies section keep ONLY `tns-core-modules` and `nativescript-angular`.
4) Try running the app on device or emulator - it will fail at runtime as it cannot find some modules. Now if you check `platforms/android/src/main/assets/app/tns_modules/` you'll see there's no `@angular` dir at all

PR #1795, opened 2016-06-01
Flatten only dependencies with valid package.json

vchimev:
Related to: https://github.com/NativeScript/nativescript-cli/issues/1783


rosen-vladimirov:
:+1:  after green build

PR #1796, opened 2016-06-01
Flatten only dependencies with valid package.json

rosen-vladimirov:
Flatten only dependencies with valid package.json


TsvetanMilanov:
:+1:

PR #1797, opened 2016-06-01
Include linting of unit tests

rosen-vladimirov:
Allow linter for unit tests and fix all errors.

Add public method in liveSyncService that should be used to delete files when LiveSync operation is in progress.
This way, when user deletes file in Proton, the application will call `liveSyncService.deleteFiles` method for it and it will be removed from device.


TsvetanMilanov:
:+1:

PR #1800, opened 2016-06-02
Do not use $projectConstants as it's not available in {N}

rosen-vladimirov:
Do not use $projectConstants in AndroidApplicationManager as it's not available in {N}.


rosen-vladimirov:
run ci

Issue #1376, opened 2016-01-06
Consider removing --debug-brk parameter


blagoev:
Neither --debug-brk nor --start option was specified. Defaulting to --debug-brk.

Since the android debug command is defaulting always to --debug-brk why do we have this parameter at all. Lets just remove this and have one thing less to worry about and show warning of.

sitefinitysteve:
Agree
https://github.com/NativeScript/nativescript-cli/issues/1314

teobugslayer:
I started implementing this but I am blocked by a bug in the Karma launcher and cannot verify that my change does not break the unit testing. Once this is resolved, I'll create a PR with the fix.

blagoev:
On the other hand if we want the functionality to be able to run an app in debug without braking lets leave the parameter but remove the default setting. 
ie tns debug android will run the app in debug without braking on the first available statement.

PR #1801, opened 2016-06-03
No debug break by default implementation added for iOS and Android

enchev:
Related to https://github.com/NativeScript/nativescript-cli/issues/1376

When you call `tns debug ios|android` (no `--debug-brk`) now we will build, deploy and start the application and attach the debugger. You can debug normally by placing breakpoint(s) in desired file(s). If you want to break somewhere initially you can use `debugger` keyword in your code.

The `--debug-brk` switch is still working as before in case you want to use it (not sure for what reason).

### DIFF
--- a/docs/man_pages/project/testing/debug-android.md
+++ b/docs/man_pages/project/testing/debug-android.md
@@ -1,6 +1,9 @@
 debug android
 ==========
 
+All you need to do to start debugging your app is to execute `tns debug android`. The NativeScript CLI will build, deploy, start your app, start Chrome DevTools and attach the debugger.
+* You cannot debug if there are multiple devices available (emulators and/or real devices). You need to have started only one device or emulator.
+
 Usage | Synopsis
 ---|---
 Deploy on device, run the app and stop at the first breakpoint | `$ tns debug android --debug-brk [--device <Device ID>] [--debug-port <port>] [--timeout <timeout>]`
--- a/docs/man_pages/project/testing/debug-ios.md
+++ b/docs/man_pages/project/testing/debug-ios.md
@@ -1,5 +1,7 @@
 debug ios
 ==========
+All you need to do to start debugging your app is to execute `tns debug ios`. The NativeScript CLI will build, deploy, start your app, start Chrome DevTools and attach the debugger.
+* You cannot debug if there are multiple devices available (emulators and/or real devices). You need to have started only one device or emulator.
 
 Usage | Synopsis
 ---|---
--- a/lib/services/android-debug-service.ts
+++ b/lib/services/android-debug-service.ts
@@ -106,12 +106,7 @@ class AndroidDebugService implements IDebugService {
 		return (() => {
 			let packageFile = "";
 
-			if (!this.$options.debugBrk && !this.$options.start && !this.$options.getPort && !this.$options.stop) {
-				this.$logger.warn("Neither --debug-brk nor --start option was specified. Defaulting to --debug-brk.");
-				this.$options.debugBrk = true;
-			}
-
-			if (this.$options.debugBrk && !this.$options.emulator) {
+			if (!this.$options.start && !this.$options.emulator) {
 				let cachedDeviceOption = this.$options.forDevice;
 				this.$options.forDevice = true;
 				this.$platformService.buildPlatform(this.platform).wait();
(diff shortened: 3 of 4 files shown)


rosen-vladimirov:
You should update help content and documentation as well

lib/services/android-debug-service.ts
```diff
@@ -141,6 +136,11 @@ class AndroidDebugService implements IDebugService {
 				this.detachDebugger(packageName).wait();
 			} else if (this.$options.debugBrk) {
 				this.startAppWithDebugger(packageFile, packageName).wait();
+			} else {
+				this.startAppWithDebugger(packageFile, packageName).wait();
+				//TODO: Find different way to make sure that the app is started.
+				sleep(500);
+				this.attachDebugger(device.deviceInfo.identifier, packageName).wait();
```
rosen-vladimirov: maybe I'm totally missing something here, but shouldn't `startAppWithDebugger` do this... Based on it's name it looks like that's what it should do and we do not have to attach debugger again?
What's the case here?

lib/services/ios-debug-service.ts
```diff
@@ -72,14 +74,15 @@ class IOSDebugService implements IDebugService {
         }).future<void>()();
     }
 
-    private emulatorDebugBrk(): IFuture<void> {
+    private emulatorDebugBrk(shouldBreak? : boolean): IFuture<void> {
```
rosen-vladimirov: when we have to pass boolean options to method, we prefer using objects and pass a property of them:
```TypeScript
 private emulatorDebugBrk(debugOptions?: {shouldBreak : boolean}): IFuture<void> {
```
So the usage will be:
```TypeScript
this.emulatorDebugBrk().wait()
this.emulatorDebugBrk({shouldBreak: true}).wait();
```
This way it's pretty easy to understand the purpose of the passed value. (passing only `true` does not give you this information)

And in the method you'll make the check:
```TypeScript
let args = debugOptions && debugOptions.shouldBreak ? "--nativescript-debug-brk" : "--nativescript-debug-start";
```

lib/services/android-debug-service.ts
enchev: dunno, you tell me?

lib/services/ios-debug-service.ts
rosen-vladimirov: Forgot to mention that this is not a merge stopper, it's up to you to decide which is the approach you prefer :) :)

enchev: I don't see many other methods with such signature. Can you point some? In fact most methods are using signature similar to what I've made here.

rosen-vladimirov: Of course,
https://github.com/telerik/mobile-cli-lib/blob/master/declarations.d.ts#L133
https://github.com/telerik/mobile-cli-lib/blob/master/declarations.d.ts#L413
https://github.com/telerik/mobile-cli-lib/blob/master/declarations.d.ts#L431
https://github.com/NativeScript/nativescript-cli/blob/master/lib/declarations.ts#L170
https://github.com/NativeScript/nativescript-cli/blob/master/lib/declarations.ts#L320

But anyway, it depends on you to decide which approach to take. As I've said, this is not a merge stopper :)

enchev: Again, these in my opinion are more like exceptions not common rule. Look at other methods signature.

enchev:
Hey @vchimev @dtopuzov can you test the CLI behavior in this pull request and like if this is the expected?

PanayotCankov:
:+1: for the iOS changes, you may ping @slavchev to review the Android part.

enchev:
Hey @blagoev @slavchev do you have any comments about these changes?

lib/services/ios-debug-service.ts
slavchev: >we prefer using objects and pass a property of them

It is not a matter of what is preferred but what works. Different teams have different cultures. Thanks for suggestion but please try to focus on arguments that both teams consider as constructive.

slavchev:
:+1:

docs/man_pages/project/testing/debug-android.md
```diff
@@ -1,6 +1,9 @@
 debug android
 ==========
 
+All you need to do to start debugging your app is to execute `tns debug android`. The NativeScript CLI will build, deploy, start your app, start Chrome DevTools and attach the debugger.
```
rosen-vladimirov: This should be under the `Usage|Synopsis` part of the help (here -> https://github.com/NativeScript/nativescript-cli/pull/1801/files#diff-e7b2eaa70b7b2394cb03e78ab260308aR18 ). We print the table of usage first in all of the help content.

docs/man_pages/project/testing/debug-ios.md
```diff
@@ -1,5 +1,7 @@
 debug ios
 ==========
+All you need to do to start debugging your app is to execute `tns debug ios`. The NativeScript CLI will build, deploy, start your app, start Chrome DevTools and attach the debugger.
```
rosen-vladimirov: Do we start Chrome DevTools for iOS?
Also this should be under the `Usage|Synopsis` part of the help. We print the table of usage first in all of the help content.

Issue #1734, opened 2016-05-01
ios builds with 2.0.0: Processing node_modules failed. TypeError: Cannot read property 'split' of null


georgeedwards:
I have just restarted my mac for the first time since I updated to the CLI 2.0.0. My project which was working before now won't build, citing error:

>Processing node_modules failed. TypeError: Cannot read property 'split' of null

I am running OSX El Capitan with XCode 7.3

Even if I run `npm install nativescript -g` I get the same error:

```
Unknown-88-63-df-a3-2e-e9:demo georgeedwards$ npm install nativescript -g
npm WARN excluding symbolic link docs/stylesheets/hightlight.css -> ../../node_modules/highlight.js/src/styles/solarized_light.css
npm WARN deprecated lodash-node@2.4.1: This package has been discontinued in favor of lodash@^4.0.0.
npm WARN deprecated lodash@1.0.2: lodash@<3.0.0 is no longer maintained. Upgrade to lodash@^4.0.0.
npm WARN deprecated graceful-fs@1.2.3: graceful-fs v3.0.0 and before will fail on node releases >= v6.0. Please update to graceful-fs@^4.0.0 as soon as possible. Use 'npm ls graceful-fs' to find it in the tree.
npm WARN excluding symbolic link docs/assets/ir_black.css -> ../../node_modules/highlight.js/src/styles/ir_black.css
npm WARN excluding symbolic link examples/TestFramework/Test Framework.framework/Resources -> Versions/Current/Resources
npm WARN excluding symbolic link examples/TestFramework/Test Framework.framework/Test Framework -> Versions/Current/Test Framework
npm WARN excluding symbolic link examples/TestFramework/Test Framework.framework/Versions/Current -> A
/Users/georgeedwards/.npm-packages/bin/tns -> /Users/georgeedwards/.npm-packages/lib/node_modules/nativescript/bin/nativescript.js
/Users/georgeedwards/.npm-packages/bin/nativescript -> /Users/georgeedwards/.npm-packages/lib/node_modules/nativescript/bin/nativescript.js

> nativescript@2.0.0 postinstall /Users/georgeedwards/.npm-packages/lib/node_modules/nativescript
> node postinstall.js

WARNING: CocoaPods is not installed or is not configured properly.
You will not be able to build your projects for iOS if they contain plugin with CocoaPod file.
To be able to build such projects, verify that you have installed CocoaPods.
Cannot read property 'split' of null
[...]

PR #1804, opened 2016-06-06
null/undefined check added

enchev:
Related to #https://github.com/NativeScript/nativescript-cli/issues/1734

### DIFF
--- a/lib/services/xcproj-service.ts
+++ b/lib/services/xcproj-service.ts
@@ -39,7 +39,7 @@ class XcprojService implements IXcprojService {
 				let cocoapodVer = this.$sysInfo.getSysInfo(this.$staticConfig.pathToPackageJson).wait().cocoapodVer,
 					xcodeVersion = this.$xcodeSelectService.getXcodeVersion().wait();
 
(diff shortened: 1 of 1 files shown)


rosen-vladimirov:
In case cocoadpodVer is null/undefined, the following line will fail as well:
https://github.com/NativeScript/nativescript-cli/pull/1804/files#diff-054de9ebd89900812d117a4a095f7e5fR53
Can you add the check there as well:
```TypeScript
let shouldUseXcproj = cocoapodVer && !!(semver.lt(cocoapodVer, "1.0.0") && ~helpers.versionCompare(xcodeVersion, "7.3.0")),
```

vakrilov:
💯

hshristov:
I though we cannot install CLI if there is no CocoaPod installed. Right?

rosen-vladimirov:
@hshristov 
Cocoapods are not mandatory (or at least by 2.0.1 they have not been mandatory for NativeScript CLI).

rosen-vladimirov:
:+1: after green build

hshristov:
👍

PR #1815, opened 2016-06-08
emulate --availableDevices fixed

enchev:


lib/services/platform-service.ts
```diff
@@ -482,8 +482,34 @@ export class PlatformService implements IPlatformService {
 	}
 
 	public deployOnEmulator(platform: string, buildConfig?: IBuildConfig): IFuture<void> {
-		this.$options.emulator = true;
-		return this.deployOnDevice(platform, buildConfig);
+		platform = platform.toLowerCase();
+
+		if (this.$options.availableDevices) {
+			return (() => {
+				let exec = require("child_process").exec;
```
hshristov: Will it be faster if we require this once (e.g. check if undefined and then require)?

PanayotCankov: Wasn't there some service that handles the execution of child processes?

lib/services/platform-service.ts
```diff
@@ -482,8 +482,34 @@ export class PlatformService implements IPlatformService {
...
+
+				if (this.$mobileHelper.isiOSPlatform(platform)) {
```
hshristov: This whole code looks the same.
It seems safer to check outside if this is iOS or Android and exec(command, ...) because the function after that call are same.

hshristov:
👍 After modification in comments

lib/services/platform-service.ts
PanayotCankov: Instead of branching here you can use something like:
```
let platformData = this.$platformsData.getPlatformData(platform); 
platformData.platformProjectService.listDevices().wait();
```
And add the listDevices to the project interface and implement as necessary for the iOS and Android project services.

PR #1818, opened 2016-06-08
Exit with non-zero code on test failures.

hdeshev:
Propagate the Karma server exit code and correctly fail the `tns test` command.


rosen-vladimirov:
:+1:

PR #1820, opened 2016-06-09
Min supported Java version check added

enchev:
Related to https://github.com/NativeScript/nativescript-cli/issues/1729

### DIFF
--- a/lib/android-tools-info.ts
+++ b/lib/android-tools-info.ts
@@ -159,6 +159,13 @@ export class AndroidToolsInfo implements IAndroidToolsInfo {
 
 	public validateJavacVersion(installedJavaVersion: string, options?: { showWarningsAsErrors: boolean }): IFuture<boolean> {
 		return ((): boolean => {
+			// Check for min supported Java version
+			let minJavaVersion = "1.8";
+			if (parseInt(installedJavaVersion.substr(0, 3).replace(".", "")) < parseInt(minJavaVersion.replace(".", ""))) {
+				this.printMessage(`Java version ${installedJavaVersion} is not supported. You have to install at least ${minJavaVersion}.`);
+				return false;
+			}
+
 			let hasProblemWithJavaVersion = false;
 			if (options) {
 				this.showWarningsAsErrors = options.showWarningsAsErrors;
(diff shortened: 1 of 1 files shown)


lib/android-tools-info.ts
```diff
@@ -159,6 +159,13 @@ export class AndroidToolsInfo implements IAndroidToolsInfo {
 
 	public validateJavacVersion(installedJavaVersion: string, options?: { showWarningsAsErrors: boolean }): IFuture<boolean> {
 		return ((): boolean => {
+			// Check for min supported Java version
+			let minJavaVersion = "1.8";
+			if (parseInt(installedJavaVersion.substr(0, 3).replace(".", "")) < parseInt(minJavaVersion.replace(".", ""))) {
```
rosen-vladimirov: This could fail on several places, installedJavaVersion could be null or undefined.
parseInt could return NaN

lib/android-tools-info.ts
```diff
@@ -159,6 +159,13 @@ export class AndroidToolsInfo implements IAndroidToolsInfo {
...
+				this.printMessage(`Java version ${installedJavaVersion} is not supported. You have to install at least ${minJavaVersion}.`);
+				return false;
```
rosen-vladimirov: when there's an error you should return true, check the code below

lib/android-tools-info.ts
```diff
@@ -159,6 +159,13 @@ export class AndroidToolsInfo implements IAndroidToolsInfo {
 
 	public validateJavacVersion(installedJavaVersion: string, options?: { showWarningsAsErrors: boolean }): IFuture<boolean> {
 		return ((): boolean => {
+			// Check for min supported Java version
+			let minJavaVersion = "1.8";
```
rosen-vladimirov: is there any reason to add this code, several lines below there's a check for currently installed version and verification against AndroidToolsInfo.MIN_JAVA_VERSION
Can't you just set MIN_JAVA_VERSION to 1.8.0 and do not change anything in this method?

enchev:
MIN_JAVA_VERSION! I totally missed this!

Issue #1296, opened 2015-12-02
tns build breaks permanately if you have drawable-invalid name


NathanaelA:
I created a drawable-xdpi & drawable-xxxdpi  (notice the missing H)
Well tns aborted the build and reported the first one was bad, and I realized I missed the h in both of them.  So I added the missing h to both folder..   Tried to build again, still fails with the same bad folder name.  But my copy isn't bad.   However, tns still has the BAD folder names in the platform folder
platforms/android/src/main/res/ that you have to go manually delete before you can build again.

dtopuzov:
Still valid in {N} 2.0.1.
Workaround: Delete your platform folder.

enchev:
Hey @NathanaelA,

What we can do in case of gradle build error is to suggest `platform remove/add`. Will this work for you? I've made pull request with this implemented.

enchev:
I've made second attempt to fix this. We will validate App_Resources drawable resources and we will stop the build in case of invalid.

NathanaelA:
I know about **platforms remove/add** -- however, I think this is something that maybe should be handled by the new **tns clean** command (not sure the status but I thought it was coming in the next version), and if a gradle image error occurs, maybe a clean is auto-ran and then the rebuild is restarted, so the end user actually is not impacted..

One note:  I personally feel remove/add really shouldn't be needed for normal things, tns is in control of the platforms folder, it should maintain it properly.  :grinning:

enchev:
Hey @NathanaelA,

In my opinion it will be better if we prevent the error at first place by validating resources. `tns clean` is another story and we will do our best to deliver it asap.

NathanaelA:
@enchev I TOTALLY agree; eliminating any errors a end user sees is the BEST policy.  :grinning:   

I just was trying to throw out suggestions in case verifying the images turned into a problem...

enchev:
Hey @NathanaelA,

Unfortunately we found that cannot handle this problem easily and I've removed the code that validates drawables. The other option will be to use our new clean command in such cases:
`tns platform clean android`

NathanaelA:
Bummer on the validation.   As long as the clean works that is acceptable to me.   To me their needs to be a easy way to recover from this...

In the event any error occurs during building; can we make sure that the "tns platform clean" command is displayed as part of the suggestions for how to recover?

PR #1828, opened 2016-06-10
Related to #1296

enchev:
try/catch gradle spawn and suggest platform add/remove in case of error

### DIFF
--- a/lib/services/android-project-service.ts
+++ b/lib/services/android-project-service.ts
@@ -256,6 +256,7 @@ export class AndroidProjectService extends projectServiceBaseLib.PlatformProject
 	}
 
 	public buildProject(projectRoot: string, buildConfig?: IBuildConfig): IFuture<void> {
+		let platformRemoveMessage = "Run `tns platform remove android && tns platform add android` and try again.";
 		return (() => {
 			if (this.canUseGradle().wait()) {
 				this.$androidToolsInfo.validateInfo({ showWarningsAsErrors: true, validateTargetSdk: true }).wait();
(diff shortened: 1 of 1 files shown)


lib/services/android-project-service.ts
```diff
@@ -287,10 +288,14 @@ export class AndroidProjectService extends projectServiceBaseLib.PlatformProject
 				if (this.$hostInfo.isWindows) {
 					gradleBin += ".bat"; // cmd command line parsing rules are weird. Avoid issues with quotes. See https://github.com/apache/cordova-android/blob/master/bin/templates/cordova/lib/builders/GradleBuilder.js for another approach
 				}
-				this.spawn(gradleBin, buildOptions, { stdio: "inherit", cwd: this.platformData.projectRoot }).wait();
+
+				try {
+					this.spawn(gradleBin, buildOptions, { stdio: "inherit", cwd: this.platformData.projectRoot }).wait();
+				} catch (ex) {
+					this.$errors.failWithoutHelp("Gradle build failed." + EOL + platformRemoveMessage);
```
rosen-vladimirov: this will hide the real error from us and from the user. At least add
```TypeScript
this.$logger.trace("Error while executing gradle build: ", ex);
```
before the `this.$errors.fail...`
As debugging this will be painful without real error.

rosen-vladimirov:
Do you think suggesting **platform remove** and **platform add** on every single build error is useful? Most of the times the build fails due to missing files, incorrect user input or whatever else, which will not be fixed in case you remove and add the platform. The related issue is not solved with this PR, it's just suggesting the workaround of the issue to the user, but this will also be applied for any build failures.

I do not like this approach of general error handling, but it's up to you to decide if you want to merge this PR.

:+1: for the code after adding `$logger.trace`
:-1:  for general error handling

PR #1842, opened 2016-06-13
cli cleans android project when plugins are updated in anyway

Plamen5kov:
This PR does a gradle clean each time the plugins are updated in any way, for example on plugin remove / add. This prevents problems concerning the android build. This is done automatically so the user doesn't have to take care of cleaning the native project manually when there is a change that will affect the build.

related to: https://github.com/NativeScript/android-runtime/issues/459


enchev:
:+1:

atanasovg:
:+1:

PR #1848, opened 2016-06-15
Move TARGET_FRAMEWORK_IDENTIFIERS to constants

rosen-vladimirov:
Update to latest common lib where the following change is applied:
Move TARGET_FRAMEWORK_IDENTIFIERS from $projectConstants (available in AppBuilder CLI only) to constants.
This way we can use it in common lib and all CLIs directly.


enchev:
I'm my opinion we should not have Cordova and NativeScript in the common code. In this way the NativeScript part will have Cordova constant and the Cordova part will have NativeScript constant. It will be better if we can have method in the common code that will throw "Not implemented" and override it in the specific implementations to return "Cordova" or "NativeScript"

PanayotCankov:
I too think that in general the common should not be aware of the {N}, AB CLIs or the proton, that includes constants. It should be lib that helps you move stuff to and from a device, get logs, deploy apps etc in a framework independent manner.

rosen-vladimirov:
If you take a closer look at the "problematic" code [here](https://github.com/telerik/mobile-cli-lib/pull/714/files#diff-37d731e13e5eb55322cb42204a53828dL96) the method is used to get all debuggable applications from the device. As the Cordova and NativeScript applications are different and this method should return all debuggable apps, no matter if they are NativeScript or Cordova based, we need the frameworks. 
I agree that in general the common lib should help you to move stuff to and from device, but it should also help you to get required information - in this case - which app from this device can be debugged.

So I'll stick with my approach. Also note - at the moment the constants are not used in NativeScript CLI, I've raised this PR in order to prevent errors from `Unable to resolve projectConstants`.

PanayotCankov:
:+1:

PR #1854, opened 2016-06-16
Add support for Node.js 6

rosen-vladimirov:
Add support for Node.js 6 by:
* Updating references of our forked modules. In each of the new tags there are prebuilt binaries for Node.js 6.x.x. The binaries are prebuilt for Mac OS X (x64), Windows (x64 and x86) and Linux (x64 and x86).
* Updating reference to ios-sim-portable to the latest version where the binaries for Node.js 6.x.x are included.
* Updating npm dependency to it's latest 2.x.x version in order to prevent the error: `(node:13828) fs: re-evaluating native module sources is not supported. If you are using the graceful-fs module, please update it to a more recent vers
ion.` (it's been shown on every command with Node.js 6.x.x and the old version of npm).
* Update some of the dev-dependencies in order to clean the usage of graceful-fs module.


enchev:
:+1:

PanayotCankov:
:+1:

PR #1856, opened 2016-06-17
Do not display all files when livesync is executed

TsvetanMilanov:
When the livesync command is executed we do not need to log info for each file that is transfered. The information should be displayed when the --watch flag is set and some file is changed.


rosen-vladimirov:
:+1:  after upadting SHA and green build

PR #1870, opened 2016-06-22
Execute batch sync when you live sync application with the --watch option

KristinaKoeva:
Remove the data.canExecuteFastSync. The check is made in the respective {N} service instead of in the base implementation


KristinaKoeva:
Common lib PR https://github.com/telerik/mobile-cli-lib/pull/725

rosen-vladimirov:
👍

PR #1873, opened 2016-06-23
Add interpolation for app.gradle applicationId value in android projects

Pip3r4o:
Change `applicationId` field in `app.gradle` for android projects 
[Need to add a Grade Android.defaultConfig #454](https://github.com/NativeScript/android-runtime/issues/454)

ping: @rosen-vladimirov @enchev @Plamen5kov

### DIFF
--- a/lib/services/android-project-service.ts
+++ b/lib/services/android-project-service.ts
@@ -211,6 +211,10 @@ export class AndroidProjectService extends projectServiceBaseLib.PlatformProject
 
 			let gradleSettingsFilePath = path.join(this.platformData.projectRoot, "settings.gradle");
 			shell.sed('-i', /__PROJECT_NAME__/, this.getProjectNameFromId(), gradleSettingsFilePath);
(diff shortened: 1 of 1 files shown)


enchev:
:+1:

vbresults:
@Pip3r4o There's a big problem with this; it interpolates in the source folder instead of the platforms folder so it's not reusable..

Pip3r4o:
@vbresults one might call that a problem, another - a feature and intended behavior. 

If you believe this to be a problem, please open an issue, stating why you think so, and how you think it should be implemented instead.

As of the current NativeScript version, the app.gradle is interpolated inside the App_Resources, and will not be copied over to platforms/android for additional transformations and interpolations. The file is left entirely in the application author's control beyond the initial interpolation to set a default value for the package name. The team is working on additional features and ways to change that value using json config file. pinging @KristianDD as he might be able to share more about that.

PR #1874, opened 2016-06-23
HTML files fast synced as well (Angular)

enchev:
Fast livesync without app restart in Angular apps if you use `templateUrl` instead `template` in your component. For example:
### app.component.ts
```TypeScript
import {Component} from "@angular/core";

@Component({
    selector: "my-app",
    templateUrl: "component.html",
})
export class AppComponent {
    public counter: number = 16;

    public get message(): string {
        if (this.counter > 0) {
            return this.counter + " taps left";
        } else {
            return "Hoorraaay! \nYou are ready to start building!";
        }
    }
    
    public onTap() {
        this.counter--;
    }
}
```
### component.html
```HTML
<StackLayout>
    <Label text="Tap the button" class="title"></Label>
    <Button text="Tap" (tap)="onTap()"></Button>
    <Label [text]="message" class="message" textWrap="true"></Label>
</StackLayout>
```

If you change something in the HTML only the page will be reloaded without application restart.

TODO: We may need to change the default NG app template to use  `templateUrl` instead `template`. We need to check also for possible memory leaks, ping @hdeshev.

### DIFF
--- a/lib/providers/livesync-provider.ts
+++ b/lib/providers/livesync-provider.ts
@@ -9,7 +9,7 @@ export class LiveSyncProvider implements ILiveSyncProvider {
 		private $logger: ILogger,
 		private $childProcess: IChildProcess) { }
 
-	private static FAST_SYNC_FILE_EXTENSIONS = [".css", ".xml"];
+	private static FAST_SYNC_FILE_EXTENSIONS = [".css", ".xml" ,".html"];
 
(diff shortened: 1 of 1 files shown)


hdeshev:
I'm almost 100% certain the old Angular app needs to be disposed on "page" livesync since we bootstrap a new app there.

Anyway, this is probably a good interim solution (given we don't leak old app objects) until we figure out a way to reload individual components in Angular apps.

hdeshev:
:+1:

PR #1879, opened 2016-06-27
Fixed: the app is recompiled always when first time livesync 

tzraikov:
When the application is installed on device for the first time with the livesync command, it always recompiles the app e.g:

tns livesync ios --watch


rosen-vladimirov:
:+1:

tzraikov:
run ci

Issue #1725, opened 2016-04-27
Ctrl + C shortcut does not stop the livesync ios --watch command


vchimev:
In NativeScript CLI 2.0.0 the shortcut Ctrl + C does not stop the `tns livesync ios --watch` command in Terminal of OS X.

Workaround:
Press Ctrl + Z.

vchimev:
Reopen this issue.

I am able to replicate it on Simulator iPhone 6 iOS9.3 with the default **hello-world** application - after the first change in the *.xml* file I cannot stop the command by pressing Ctrl + C.

PR #1884, opened 2016-06-27
Fixed Ctrl+C not working when live syncing

tzraikov:
#1725


alereisan:
Still an Issue on Windows when running "tns debug android"

PR #1886, opened 2016-06-28
support for node 0.12

Plamen5kov:
update cli to exclude special logic for older versions of node
> filewalker is removed because it's no longer necessary

ping @rosen-vladimirov @enchev @hshristov


PanayotCankov:
:+1:

PR #1892, opened 2016-06-30
Deprecate iOS deployment target check for frameworks

jasssonpet:
NativeScript for iOS 2.1 has a minimum one of 8.0.

ping @ivanbuhov


ivanbuhov:
Why not remove the whole deployment target check? The only reason I can think of is to support old runtime/project template version with the latest CLI, but is this something we need to support?

jasssonpet:
I think it's best to deprecate it now and remove it in a future release. This way we will not break currently existing applications.

jasssonpet:
run ci

jasssonpet:
run ci

enchev:
run ci

enchev:
run ci

PR #1911, opened 2016-07-08
Improve check for sending analytics data

rosen-vladimirov:
Currently when we check if analytics is sending information we always wait at least one second. In fact we do not have to wait when analytics is not sending data at the moment.
Improve this check and also speed up the unit tests of analytics service a lot.


tzraikov:
👍

PR #1914, opened 2016-07-08
Move analytics service registration

TsvetanMilanov:
We need to create analytics service in the NS CLI to implement the NS specific logic for sending analytics data.

### DIFF
--- a/lib/services/analytics-service.ts
+++ b/lib/services/analytics-service.ts
@@ -1,0 +1,28 @@
+import {AnalyticsServiceBase} from "../common/services/analytics-service-base";
+
+export class AnalyticsService extends AnalyticsServiceBase implements IAnalyticsService {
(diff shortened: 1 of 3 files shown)


lib/services/analytics-service.ts
```diff
@@ -0,0 +1,28 @@
+import {AnalyticsServiceBase} from "../common/services/analytics-service-base";
+
+export class AnalyticsService extends AnalyticsServiceBase implements IAnalyticsService {
+	private static ANALYTICS_FEATURE_USAGE_TRACKING_API_KEY = "9912cff308334c6d9ad9c33f76a983e3";
+
+	constructor(protected $logger: ILogger,
+		protected $options: ICommonOptions,
```
rosen-vladimirov: `IOptions` maybe?

Issue #1912, opened 2016-07-08
"tns device run <bundle> --device <device>" noes no work


dtopuzov:
tns device run org.nativescript.TNSApp --device YT91194WVS
=> unable to resolve projectConstants

dtopuzov:
Fixed in master

PR #1918, opened 2016-07-11
Fix device run command

rosen-vladimirov:
Fix `tns device run` command by removing `$projectConstants` dependency, which is not available in NativeScirpt CLI. It has not been used in the command, only declared in the constructor.

Fixes https://github.com/NativeScript/nativescript-cli/issues/1912


enchev:
:+1:

dtopuzov:
:100:

PR #1926, opened 2016-07-14
Remove some of the AppBuilder related logic

KristinaKoeva:


lib/services/livesync/livesync-service-base.ts
```diff
@@ -176,53 +175,45 @@ class LiveSyncServiceBase implements ILiveSyncServiceBase {
 			return (() => {
 				let shouldRefreshApplication = true;
 				let deviceAppData = this.$deviceAppDataFactory.create(appIdentifier, this.$mobileHelper.normalizePlatformName(platform), device, liveSyncOptions);
-				if (deviceAppData.isLiveSyncSupported().wait()) {
```
KristinaKoeva: isLiveSyncSupported is implemented only in AppBuilder so that's why I've removed it

lib/services/livesync/livesync-service-base.ts
```diff
@@ -176,53 +175,45 @@ class LiveSyncServiceBase implements ILiveSyncServiceBase {
...
-						device.applicationManager.installApplication(packageFilePath).wait();
+				let wasInstalled = true;
+				if (!device.applicationManager.isApplicationInstalled(appIdentifier).wait()) {
+					this.$logger.warn(`The application with id "${appIdentifier}" is not installed on device with identifier ${device.deviceInfo.identifier}.`);
+					if (!packageFilePath) {
+						packageFilePath = this.$liveSyncProvider.buildForDevice(device).wait();
+					}
+					device.applicationManager.installApplication(packageFilePath).wait();
 
-						if (platformLiveSyncService.afterInstallApplicationAction) {
```
KristinaKoeva: afterInstallApplicationAction is implemented in both platforms so this condition should always be true. So I've removed the else part. ping @tzraikov if I am missing something

lib/services/livesync/livesync-service-base.ts
```diff
@@ -235,7 +226,7 @@ class LiveSyncServiceBase implements ILiveSyncServiceBase {
 			for (let dataItem of data) {
 				let appIdentifier = dataItem.appIdentifier;
 				let platform = dataItem.platform;
-				let canExecute = this.getCanExecuteAction(platform, appIdentifier, dataItem.canExecute);
```
KristinaKoeva: dataItem.canExecute is implemented in AppBuilder only so remove it 🌷

tzraikov:
👍  - after green build

PR #1939, opened 2016-07-21
Fixed first time debug with livesync

tzraikov:


lib/services/livesync/ios-livesync-service.ts
```diff
@@ -33,10 +42,13 @@ class IOSLiveSyncService extends PlatformLiveSyncServiceBase<Mobile.IiOSDevice>
 	}
 
 	protected restartApplication(deviceAppData: Mobile.IDeviceAppData): IFuture<void> {
+		let projectData: IProjectData = this.$injector.resolve("projectData");
 		if (this.$options.debug) {
+			this.device.applicationManager.restartApplication(deviceAppData.appIdentifier, projectData.projectName).wait();
+			let appId = this.device.isEmulator ? projectData.projectName : deviceAppData.appIdentifier;
+			this.device.applicationManager.stopApplication(appId).wait();
```
KristinaKoeva: Why do you need to first restart the application and the stop it ?

lib/services/platform-service.ts
```diff
@@ -461,7 +461,9 @@ export class PlatformService implements IPlatformService {
 
 	public deployOnDevice(platform: string, buildConfig?: IBuildConfig): IFuture<void> {
 		return (() => {
-			this.installOnDevice(platform, buildConfig).wait();
+			if (!this.$options.debug) {
+				this.installOnDevice(platform, buildConfig).wait();
```
KristinaKoeva: Currently you have a method that is called deployOnDevice but in some special situations it doesn't deploy anything, it just starts the application. 
Since installOnDevice is public I would suggest you to change the deployOnDevice to be just startOnDevice (or some better name)  and call

KristinaKoeva:
I stated my opinion in private but for the sake of visibility I will state it here again. I disagree with this code change. I believe that with this code change we are breaking the single responsibility principle. Currently we have a live sync service that is also responsible for calling debug. For me the better approach would be to call subsequently live sync and then debug in the debug command. As we previously discussed I made a small refactoring to your previous PR https://github.com/NativeScript/nativescript-cli/pull/1929 that does this. Take a look here https://github.com/NativeScript/nativescript-cli/tree/KristinaKoeva/DebugLiveSync

lib/services/platform-service.ts
tzraikov: Thanks for being honest. I excluded all the logic related to the debug service and moved it in the debug command. Now the single responsibility principle is not hurt. The debug command relies on messages to communicate with the livesync service. I checked your approach, it will not work when we use the --watch option.

tzraikov: Created a new method startOnDevice in platform-service

lib/services/livesync/ios-livesync-service.ts
tzraikov: The first restart was necessary because the stopApplication command doesn't work correctly. It is addressed now.

tzraikov:
Thanks for being honest. I excluded all the logic related to the debug service and moved it in the debug command. Now the single responsibility principle is not hurt. The debug command relies on messages to communicate with the livesync service. I checked your approach, it will not work when we use the --watch option.

lib/commands/debug.ts
```diff
@@ -1,14 +1,47 @@
...
+			let liveSyncServiceBase: any = this.$injector.resolve("liveSyncServiceBase");
+			let liveSyncProvider: ILiveSyncProvider = this.$injector.resolve("liveSyncProvider");
+
+			liveSyncServiceBase.on("sync", (device: Mobile.IDevice, data: ILiveSyncData) => {
+				let platformLiveSyncService: IPlatformLiveSyncService = this.$injector.resolve(liveSyncProvider.platformSpecificLiveSyncServices[data.platform.toLowerCase()], { _device: device });
+				let projectData: IProjectData = this.$injector.resolve("projectData");
+				let appId = device.isEmulator ? projectData.projectName : data.appIdentifier;
+				if (data.platform === this.$devicePlatformsConstants.iOS) {
+					try {
+						this.$childProcess.exec("/usr/bin/killall 'NativeScript Inspector'").wait();
```
rosen-vladimirov: Why not just `killall`, `/usr/bin` must be in the PATH variable.

lib/commands/debug.ts
```diff
@@ -1,14 +1,47 @@
...
+					} catch (e) {
```
rosen-vladimirov: you can add:
```TypeScript
this.$logger.trace(`Error while killing NativeScript inspector: ${e.message}`);
```
So in case there's some strange error on client side, we'll be able to at least see it when `--log trace` is passed.

lib/commands/debug.ts
```diff
@@ -1,14 +1,47 @@
...
+					if (device.isEmulator && this.$options.watch) {
+						try {
+							this.$childProcess.exec("/usr/bin/killall 'lldb'").wait();
+						} catch (e) {
+						}
+					}
+				}
+				device.applicationManager.stopApplication(appId).wait();
+				platformLiveSyncService.debugService.debug().wait();
+				return true;
```
rosen-vladimirov: What's the idea of this return? When will it be used?

lib/services/livesync/livesync-service-base.ts
```diff
@@ -212,9 +217,12 @@ class LiveSyncServiceBase implements ILiveSyncServiceBase {
 					}
 
 					this.$logger.info("Applying changes...");
-					platformLiveSyncService.refreshApplication(deviceAppData, localToDevicePaths, data.forceExecuteFullSync || !wasInstalled).wait();
+					if (!this.emit("sync", device, data)) {
```
rosen-vladimirov: According to nodejs documentation:
```
Returns true if the event had listeners, false otherwise.
```
https://nodejs.org/api/events.html#events_emitter_emit_eventname_arg1_arg2

So the refreshApplication method will be called only if there are no listeners (currently every other case except debug command I think). Is this the expected behavior?

lib/commands/debug.ts
tzraikov: You are correct, the exact path is not necessary in this case. I removed it.

lib/commands/debug.ts
tzraikov: Nice suggestion, added trace where necessary.

lib/commands/debug.ts
tzraikov: The return true is used to indicate that the event is processed and the livesync service should not execute the same action. This is necessary because the livesync service calls the refreshApplication function which behaves in different way from what the debug workflow expects. This way we keep the livesync service separate from debug workflow.

lib/services/livesync/livesync-service-base.ts
tzraikov: Actually yes, this is the expected behavior. See the previous comment.

rosen-vladimirov:
:+1:  after green build

tzraikov:
run ci

Issue #1881, opened 2016-06-27
Plugin add command should uninstall the plugin first


blagoev:
Getting this error when trying to install a new version of tns core modules widgets. The problem is that instead of telling me I have this already and suggest to uninstall it the command tries to reinstall the plugin over the other. This I guess causes peerDependencies failure. In general this peerDependencies error is very undescriptive.

c:\Work\nativescript\sample-Groceries>c:\Work\nativescript\nativescript-cli\bin\tns plugin add  k:\DailyBuilds\NativeScript\android-widgets\Release\tns-core-modules-widgets.tgz
(node:1020) fs: re-evaluating native module sources is not supported. If you are using the graceful-fs module, please update it to a more recent version.
The package rxjs@5.0.0-beta.6 does not satisfy its siblings' peerDependencies requirements!

PR #1942, opened 2016-07-22
return false if plugin is already installed

enchev:
Related to https://github.com/NativeScript/nativescript-cli/issues/1881


PanayotCankov:
:+100:

PR #1953, opened 2016-07-27
Fixed: Livesync not syncing on 2 devices when platform is not specified

tzraikov:
1. connect two devices (iOS and Android).
2. create a new project: `tns create testProject`
3. run `tns platform add ios` and `tns platform add android`
4. run `tns livesync`
5. only the Android device will be synced, there will be a random exception when syncing the iOS device.


rosen-vladimirov:
:+1:

tzraikov:
run ci

PR #1961, opened 2016-07-28
Added support for Xcode 8 and iOS 10

tzraikov:


dtopuzov:
:100:

enchev:
:+1:

lib/services/ios-project-service.ts
```diff
@@ -293,8 +293,8 @@ export class IOSProjectService extends projectServiceBaseLib.PlatformProjectServ
 			} else {
 				args = basicArgs.concat([
 					"-sdk", "iphonesimulator",
-					"-arch", "i386",
-					'VALID_ARCHS="i386"',
+					"-arch", "x86_64",
```
rosen-vladimirov: will this work with XCode 7 ?

tzraikov: Tested on Xcode 7.3.1

PR #1966, opened 2016-07-29
Don't livesync tns_modules folder for iOS device

jasssonpet:
This is done for performance improvements to avoid unarchiving Angular and NativeScript modules. At runtime the `tns_modules` from the deployed `ipa` bundle  will be used instead.

BTW, Livesync with `--watch` is not livesyncing `tns_modules` too, so at least this should be consistent now.

Related: https://github.com/NativeScript/ios-runtime/pull/611/files#diff-26b03154ad8867bf950c44b3bb6c9c86


enchev:
:+1:

PR #1969, opened 2016-08-01
Remove the outer future

KristinaKoeva:
The afterSyncActions are already futuristic, so we don't need the outer future


tzraikov:
👍

rosen-vladimirov:
I would vote against this change and add `.wait()` on the required methods. Here are my reasons:
* Somtimes future's callstack comes in incorrect order when there's no future wrapper.
* It's good practice to have the construction with 
```TypeScript
return (() => { }).future<T>()()
```
as it returns a new Fiber future. As you are executing an action, it may be executed outside of Fiber, but the construction above will create a FiberFuture and will pass. Without it the code will fail with "Can't wait without fiber".

rosen-vladimirov:
👍

PR #1975, opened 2016-08-03
Do not force LaunchScreen when adding AndroidManifest.xml in old projects

PanayotCankov:
Some arcane projects have no AndroidManifest.xml in App_Resources. When we add the new AndroidManifets it requires LaunchScreenTheme resources. Since these project had no splashscreen when they were created we had decided that we do not want to force feed them splash screens and sanitize the theme attribute and default theme meta in the AndroidManifest when we are coping it from the default template.


vchimev:
👍

PR #1977, opened 2016-08-03
Send special message if we live sync javascript file and the liveEdit option is on

KristinaKoeva:
Commin PR https://github.com/telerik/mobile-cli-lib/pull/773


enchev:
:+1:

lib/services/livesync/ios-device-livesync-service.ts
```diff
@@ -31,78 +32,140 @@ class IOSLiveSyncService extends DeviceLiveSyncServiceBase<Mobile.IiOSDevice> im
...
 
-	private attachEventHandlersIfNecessary(): void {
-		if(this.$options.watch) {
-			this.attachProcessExitHandlers();
-			this.attachSocketCloseEvent();
-		}
+	private liveEdit(localToDevicePaths: Mobile.ILocalToDevicePathData[]) {
+		return (() => {
+			_.each(localToDevicePaths, localToDevicePath => {
+				let content = this.$fs.readFile(localToDevicePath.getLocalPath()).wait().toString();
```
rosen-vladimirov: You can use `this.$fs.readText` instead of `readFile` and skip `.toString()`.

lib/services/livesync/ios-device-livesync-service.ts
```diff
@@ -31,78 +32,140 @@ class IOSLiveSyncService extends DeviceLiveSyncServiceBase<Mobile.IiOSDevice> im
...
+				});
+
+				this.sendMessage(message);
+			});
+		}).future<void>()();
 	}
 
-	private attachSocketCloseEvent(): void {
+	private attachEventHandlers(): void {
+		this.attachProcessExitHandlers();
```
rosen-vladimirov: you can replace this with
```TypeScript
this.$processService.attachToProcessExitSignals(this, this.destroySocket);
```
This will prevent multiple handlers for `exit`, `SIGINT` and `SIGTERM` events.

rosen-vladimirov:
I would advise you to change the sendMessage of IOSLiveSyncService to return future, which is resolved when all the data is written:
```TypeScript
private sendMessage(message: string): IFuture<void> {
	return (() => {
		let socketWriteFuture = new Future<void>();
		try {
			let length = Buffer.byteLength(message, "utf16le");
			let payload = new Buffer(length + 4);
			payload.writeInt32BE(length, 0);
			payload.write(message, 4, length, "utf16le");

			// Have to check how to add the handler only for current sending. In case error is not raised, the socket will remain alive and next message will add a new error handler. This will be a problem in a long living process
			this.socket.once("error", err => {
				this.$logger.trace("Error while sending message: ", err);
				
				this.destroySocket();
				
				socketWriteFuture.throw(err);
			});
			
			this.socket.write(payload, "utf16le", () => {
				socketWriteFuture.return();
			});
			
		} catch(err) {
			this.$logger.trace("Error while sending page reload:", err);
			this.destroySocket();
		}
		
		socketWriteFuture.wait();
	}).future<void>()();
}
```

My idea is to make sure the message is sent. According to [Node.js documentation](https://nodejs.org/api/net.html#net_socket_write_data_encoding_callback) `The optional callback parameter will be executed when the data is finally written out - this may not be immediately.`

tzraikov:
👍

lib/services/livesync/ios-device-livesync-service.ts
KristinaKoeva: Totally agree with you and if you don't mind I will do this in another PR because it may require additional changes

rosen-vladimirov:
:+1: after green build

PR #1981, opened 2016-08-05
Use the new processService for destroying the live sync socket

KristinaKoeva:


rosen-vladimirov:
:sunny:  :+1:  after green build

enchev:
:+1:

PR #1985, opened 2016-08-08
Fixed: Livesync+Debug on Android not working when there are changes in tns-core-modules

tzraikov:
## The issue:

1. Create a new {N} project.
2. Open it in Visual Studio Code. Run debug with "Sync on Android". This will start livesync and then debug.
3. Make a change in tns-core-modules. 
4. Run "Sync on Android" again. It will fail. This happens because the prepareProject called by livesync service will clean the android build folder with all apks. It is required by the debug service that is called after livesync.

## Proposal:

In that situation we change only the js file and therefore it is not necessary to clean the whole build folder.  We could check whether the change is in platforms folder of particular module.  It so, we should clean the project. If not, it is safe to skip the clean process.

### DIFF
--- a/lib/services/android-project-service.ts
+++ b/lib/services/android-project-service.ts
@@ -41,6 +41,7 @@ export class AndroidProjectService extends projectServiceBaseLib.PlatformProject
 		private $devicePlatformsConstants: Mobile.IDevicePlatformsConstants,
 		private $projectTemplatesService: IProjectTemplatesService,
 		private $xmlValidator: IXmlValidator,
+		private $config: IConfiguration,
 		private $npm: INodePackageManager) {
 		super($fs, $projectData, $projectDataService);
 		this._androidProjectPropertiesManagers = Object.create(null);
@@ -405,16 +406,18 @@ export class AndroidProjectService extends projectServiceBaseLib.PlatformProject
 	}
 
 	public beforePrepareAllPlugins(): IFuture<void> {
-		let buildOptions = this.getBuildOptions();
-		buildOptions.unshift("clean");
+		if (!this.$config.debugLivesync) {
+			let buildOptions = this.getBuildOptions();
 
-		let projectRoot = this.platformData.projectRoot;
-		let gradleBin = this.useGradleWrapper(projectRoot) ? path.join(projectRoot, "gradlew") : "gradle";
-		if (this.$hostInfo.isWindows) {
-			gradleBin += ".bat";
+			buildOptions.unshift("clean");
+
+			let projectRoot = this.platformData.projectRoot;
+			let gradleBin = this.useGradleWrapper(projectRoot) ? path.join(projectRoot, "gradlew") : "gradle";
+			if (this.$hostInfo.isWindows) {
+				gradleBin += ".bat";
+			}
+			this.spawn(gradleBin, buildOptions, { stdio: "inherit", cwd: this.platformData.projectRoot }).wait();
 		}
-		this.spawn(gradleBin, buildOptions, { stdio: "inherit", cwd: this.platformData.projectRoot }).wait();
-
 		return Future.fromResult();
 	}
 
--- a/test/npm-support.ts
+++ b/test/npm-support.ts
@@ -78,6 +78,7 @@ function createTestInjector(): IInjector {
 	testInjector.register("mobilePlatformsCapabilities", MobilePlatformsCapabilities);
 	testInjector.register("devicePlatformsConstants", DevicePlatformsConstants);
 	testInjector.register("xmlValidator", XmlValidator);
+	testInjector.register("config", StaticConfigLib.Configuration);
 
 	return testInjector;
 }
--- a/test/plugins-service.ts
+++ b/test/plugins-service.ts
@@ -33,6 +33,7 @@ import {DeviceAppDataProvider} from "../lib/providers/device-app-data-provider";
 import {MobilePlatformsCapabilities} from "../lib/mobile-platforms-capabilities";
 import {DevicePlatformsConstants} from "../lib/common/mobile/device-platforms-constants";
 import { XmlValidator } from "../lib/xml-validator";
+import StaticConfigLib = require("../lib/config");
 import * as path from "path";
 import * as temp from "temp";
 temp.track();
@@ -99,6 +100,7 @@ function createTestInjector() {
 		defaultTemplate: future.fromResult("")
 	});
 	testInjector.register("xmlValidator", XmlValidator);
+	testInjector.register("config", StaticConfigLib.Configuration);
 
 	return testInjector;
 }


tzraikov:
run ci

lib/commands/debug.ts
```diff
@@ -15,12 +15,12 @@
 			let applicationReloadAction = (deviceAppData: Mobile.IDeviceAppData, localToDevicePaths: Mobile.ILocalToDevicePathData[]): IFuture<void> => {
 				return (() => {
 					let projectData: IProjectData = this.$injector.resolve("projectData");
-
 					this.debugService.debugStop().wait();
-
-					let applicationId = deviceAppData.device.isEmulator ? projectData.projectName : deviceAppData.appIdentifier;
+					let applicationId = deviceAppData.appIdentifier;
+					if (deviceAppData.device.isEmulator && deviceAppData.platform.toLowerCase() === this.$devicePlatformsConstants.iOS.toLocaleLowerCase()) {
```
rosen-vladimirov: `toLocaleLowerCase` -> `toLowerCase` maybe

lib/services/livesync/platform-livesync-service-base.ts
```diff
@@ -78,7 +79,7 @@ export abstract class PlatformLiveSyncServiceBase implements IPlatformLiveSyncSe
 			device.applicationManager.checkForApplicationUpdates().wait();
 
 			let appIdentifier = this.liveSyncData.appIdentifier;
-			if (!device.applicationManager.isApplicationInstalled(appIdentifier).wait()) {
+			if (!device.applicationManager.isApplicationInstalled(appIdentifier).wait() || !this.$platformService.isPlatformBuilt(deviceAppData.platform.toLowerCase()).wait()) {
 				this.$logger.warn(`The application with id "${appIdentifier}" is not installed on device with identifier ${device.deviceInfo.identifier}.`);
```
rosen-vladimirov: The message indicates that the app is not installed, but in fact we can get here when the application is installed. Consider changing the message.

lib/services/android-project-service.ts
```diff
@@ -298,6 +298,10 @@ export class AndroidProjectService extends projectServiceBaseLib.PlatformProject
 		return this.$fs.exists(path.join(this.platformData.appDestinationDirectoryPath, constants.APP_FOLDER_NAME));
 	}
 
+	public isPlatformBuilt(projectRoot: string): IFuture<boolean> {
+		return this.$fs.exists(path.join(projectRoot, "build", "outputs", "apk"));
```
rosen-vladimirov: Is this enough? Shouldn't you check for existing file with extensions .apk inside this directory? What will happen in case it is empty?

lib/services/ios-project-service.ts
```diff
@@ -357,6 +357,10 @@ export class IOSProjectService extends projectServiceBaseLib.PlatformProjectServ
 		return this.$fs.exists(path.join(projectRoot, this.$projectData.projectName, constants.APP_FOLDER_NAME));
 	}
 
+	public isPlatformBuilt(projectRoot: string): IFuture<boolean> {
+		return this.$fs.exists(path.join(projectRoot, "build"));
```
rosen-vladimirov: Is this enough? Shouldn't you check for existing file with extensions `.ipa` or directory with name `*.app` inside this directory? What will happen in case it is empty?

lib/services/android-project-service.ts
Plamen5kov: I don't believe livesync should depend on the `platforms/android/build` folder at all. Folders inside build are temporary, they are being constantly changed and we can't be sure if this won't have unexpected behavior depending on the user actions.

lib/services/livesync/platform-livesync-service-base.ts
```diff
@@ -78,7 +79,7 @@ export abstract class PlatformLiveSyncServiceBase implements IPlatformLiveSyncSe
 			device.applicationManager.checkForApplicationUpdates().wait();
 
 			let appIdentifier = this.liveSyncData.appIdentifier;
-			if (!device.applicationManager.isApplicationInstalled(appIdentifier).wait()) {
+			if (!device.applicationManager.isApplicationInstalled(appIdentifier).wait() || !this.$platformService.isPlatformBuilt(deviceAppData.platform.toLowerCase()).wait()) {
```
rosen-vladimirov: Why do we need to be sure that the platform is built? In case the application is installed on device, shouldn't we just livesync the files (persuming there are no changes requiring build).

lib/tools/broccoli/node-modules-dest-copy.ts
```diff
@@ -69,7 +70,7 @@ export class DestCopy implements IBroccoliPlugin {
 				});
 			}
 		});
-		if (!_.isEmpty(this.dependencies)) {
+		if (!_.isEmpty(this.dependencies) && !this.$config.debugLivesync) {
```
Plamen5kov: I believe we can find a better way of controlling the behavior than making specific exceptions for every case that comes up. This will be a heavy burden on us in the future, makes the code hard to read and understand and put's a strain on future development.

lib/services/livesync/platform-livesync-service-base.ts
tzraikov: We need this because in some cases the build folder is removed (for example the prepareProject cleans the build folder on android when there is a change in tns-core-modules, or the user simply removes the folder). In this situation there is no apk file and we cannot guarantee that the application installed on device is the same application that we want. I think the best option in this scenario is to build the application again. This way we will ensure that the correct application is installed.

lib/tools/broccoli/node-modules-dest-copy.ts
tzraikov: I don't agree with this comment. Relations between services in CLI are fact and I don't see how we can avoid them. Conditions are normal part of the development process, we should accept this. If we want to extend and improve the existing livesync feature, we need to add some additional logic. Moving this condition to another place will not make our code better, it will just move the problem to another place. Even worse, it could require many lines of code just to make things working.  If you have better idea that can be applied, I would love to hear it.

lib/services/android-project-service.ts
tzraikov: Actually this is enough to work and it solves the issue. Of course it won't work if someone removes everything from the build folder explicitly (The CLI always removes the whole folder). We could add a check whether there are apk files, but who guarantees that those are the apk files generated by us? Maybe the user placed his own files there? Do you think that we should check this too?

lib/services/ios-project-service.ts
tzraikov: Check my comment about android-project-service. I can't imagine a situation caused by the CLI in which this folder would be empty. However, if you think that such situation may occur, I could extend the check.

tzraikov:
run ci

Plamen5kov:
This is a functionality that needs to be tested.

tzraikov:
@Plamen5kov 
Yes, as a general suggestion.. tests should be added. However, I don't see a single test related with livesync or debug functionality in our code. Does anyone have an idea how to test those features?

lib/tools/broccoli/node-modules-dest-copy.ts
Plamen5kov: All I can give is an opinion, as a person, who is not familiar with CLI. I am relying on you to give the best ideas and solutions based on the discussions we have and comments we provide.

lib/services/android-project-service.ts
Plamen5kov: I believe we should guard ourselves from undefined behavior.

rosen-vladimirov:
With current code `gradle clean` will be skipped in Debug + LiveSync scenario, no matter what's really changed. As I believe the code will evolve in the future and the current solution will be only part of a bigger effort (as far as I understand livesync will successfully detect what has changed in a user project and will rebuild it in case it's required) I give my :+1: . Yes, the current code resolves an issue and introduces new one (always skipping `gradle clean` in a specific scenario), but I believe that this has been considered and it's the easiest solution for a bug that is a show stopper for one of our internal teams.
:+1:  and I like the updated description of the issue.

lib/tools/broccoli/node-modules-dest-copy.ts
tzraikov: Please note that I saw the CLI code a month ago for the first time. So, I am just scratching the surface.  I think we both have relatively the same knowledge about it. I am just trying to fix some of its issues that annoy our customers and our internal teams. This has the greatest priority for me. So I count on your assistance, good will and support.

lib/services/android-project-service.ts
tzraikov: I believe we should not over-engineer things.

lib/tools/broccoli/node-modules-dest-copy.ts
tzraikov: I believe that one should give suggestions only when he understands the specific issue in detail and when he has something to offer.

PR #1988, opened 2016-08-09
Fixed: debug > stopApplication not working correctly on Android emulator

tzraikov:
The current logic in debug commnad changes the argument passed to stopApplication method from applicationId to projectName when using emulator. This behavior is correct in iOS, but it fails on Android. This fix corrects the issue.

### DIFF
--- a/lib/commands/debug.ts
+++ b/lib/commands/debug.ts
@@ -18,7 +18,10 @@
 
 					this.debugService.debugStop().wait();
 
-					let applicationId = deviceAppData.device.isEmulator ? projectData.projectName : deviceAppData.appIdentifier;
+					let applicationId = deviceAppData.appIdentifier;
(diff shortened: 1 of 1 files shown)


lib/commands/debug.ts
```diff
@@ -18,7 +18,10 @@
 
 					this.debugService.debugStop().wait();
 
-					let applicationId = deviceAppData.device.isEmulator ? projectData.projectName : deviceAppData.appIdentifier;
+					let applicationId = deviceAppData.appIdentifier;
+					if (deviceAppData.device.isEmulator && deviceAppData.platform.toLowerCase() === this.$devicePlatformsConstants.iOS.toLowerCase()) {
```
KristinaKoeva: I've seen this code in many many places. Do you think that it's reasonable to extract it in a helper method somewhere ?

PR #1993, opened 2016-08-10
Fixed: VisualStudioCode hangs when executing Sync on iOS action

tzraikov:


tzraikov:
run ci

enchev:
:+1:

lib/commands/debug.ts
```diff
@@ -21,6 +21,8 @@
 				return (() => {
 					let projectData: IProjectData = this.$injector.resolve("projectData");
 
+					this.$logger.info(`Successfully synced application ${deviceAppData.appIdentifier} on device ${deviceAppData.device.deviceInfo.identifier}.`);
```
KristinaKoeva: This message is repeated in the platform-livesync-service-base, so I vote for moving both messages into the livesync-service. This way the chance to change/delete it would be lower. 
Also it would be nice to add a comment why this message is so important

tzraikov: Good suggestion. I had some issues doing this in the current implementation of the livesync service. That is why I will do this in another commit.

PR #1994, opened 2016-08-10
Fixed: livesync + debug syncs wrong files on Android

tzraikov:
1. Create a new {N} project.
2. Run `tns livesync android --watch`.
3. Do some changes in XML and JS files.
4. Stop tns process.
5. Open Visual Studio Code and start `Sync on Android` action.
6. Change JS file - you will see that the JS file is correct, but the XML one is not.
7. Change XML file - you will see that the XML file is correct, but the JS file is not.

Should be merged on top of #1985

### DIFF
--- a/lib/services/android-debug-service.ts
+++ b/lib/services/android-debug-service.ts
@@ -171,7 +171,7 @@ class AndroidDebugService implements IDebugService {
 
 	private startAppWithDebugger(packageFile: string, packageName: string): IFuture<void> {
 		return (() => {
-			if (!this.$options.emulator) {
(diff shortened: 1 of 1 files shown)


kekelzpekel:
Hi, do you think this has some relation to my issue on stack overflow ? Basically tns livesync builds old code (regardless of IDE), but it builds even wrong tj (js).
http://stackoverflow.com/questions/38853359/nativescript-builds-old-code

tzraikov:
@kekelzpekel: Yes, this could be the same issue. It could not make it to our latest release - 2.2.0, however you can get it by installing our @next release.

PR #1997, opened 2016-08-11
Improved: The live sync successfull message appears twice in the code

tzraikov:
This is a follow-up PR to #1993. Currently the "Successfully synced application" message can be found in two files. This PR creates a single method called `finishLivesync` in `platform-livesync-service-base` which improves the code. Note that `livesync-service` is not the best place for such message because it is platform agnostic, the message however is not.


enchev:
:+1:

PR #2002, opened 2016-08-15
Added tests for debug + livesync

tzraikov:
1. Test that debugLivesync flag is true when executing debug --watch command
2. Test that beforePrepareAllPlugins will not call gradle when livesyncing
3. Test that beforePrepareAllPlugins will call gradle with clean option when *NOT* livesyncing


enchev:
:+1:

rosen-vladimirov:
:+1:

PR #2037, opened 2016-09-07
Use monkey to start Android applications

rosen-vladimirov:
Use the monkey tool to start Android applications on device. This way we do not have to care about the startPackageActivity name and we can start the application in case a custom activity is used.
Remove unused code and variables.

Fixes https://github.com/NativeScript/nativescript-cli/issues/2028


Mitko-Kerezov:
LGTM :+1:

dtopuzov:
:100:

PR #2038, opened 2016-09-07
Get two fixes for Android

rosen-vladimirov:
### Add checks for emulate on android
We need to check if there are any emulators installed for android to emulate on android.

### 	Use monkey to start Android applications
Use the monkey tool to start Android applications on device. This way we do not have to care about the startPackageActivity name and we can start the application in case a custom activity is used.
Remove unused code and variables.


TsvetanMilanov:
:+1:

PR #2041, opened 2016-09-08
Fix using of custom AndroidManifest.xml

rosen-vladimirov:
In case user had edited his AndroidManifest.xml, he may have removed the com.tns.ErrorReportActivity. In this case, NativeScript CLI thinks that the AndroidManifest.xml is not correct, creates a backup for it and extracts the default one from the template.
Remove the check for `com.tns.ErrorReportActivity`.


dtopuzov:
:+1:

PR #2044, opened 2016-09-08
Fix live edit on device

KristinaKoeva:
In case of live edit on device, the getDevicePath does not return the absolute device path. So pass a relative path instead and the runtime would handle it.


dtopuzov:
:100:

PR #2048, opened 2016-09-10
Introduce a --disable-npm-install option

hdeshev:
Allow for advanced uses that run `npm install` only once and want to avoid rerunning it on every command. We want that since always running `npm install` can be slow (e.g. always hitting the network for packages configured as remote git repos) and unpredictable (always running pre/post-install scripts).

This one speeds up `tns livesync --watch` too since it seems to run `npm install` on every device sync.


enchev:
run ci

PR #2051, opened 2016-09-12
Kill the process for iOS Simulator logs on SIGINT

rosen-vladimirov:
On some Mac OS machines, the process for reading iOS Simulator logs does not die when Ctrl + C is used. So CLI's process cannot die as well and Ctrl + Z should be used. However the processes for reading logs remain alive and zombies.
Make sure to kill the process on SIGINT, SIGTERM, etc. This way CLI's process should die as well.

Use new version of ios-sim-portable, where the child process is returned to the CLI, so we can kill it later.

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
     "glob": "^7.0.3",
     "iconv-lite": "0.4.11",
     "inquirer": "0.9.0",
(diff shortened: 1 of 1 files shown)


TsvetanMilanov:
:+1:

dtopuzov:
👍

PR #2060, opened 2016-09-14
Workaround Ctrl+C not exiting debug process

rosen-vladimirov:
During `tns debug ios` command, Ctrl + C does not kill the proecss. It looks like there are still resources that do not allow the process to die.
Implement a workaround - call process.exit when SIGINT (Ctrl + C) is received. There's no guarantee that this will release the other resources. However CLI's process will definitely die and console will be released.
Code is based on this PR: https://github.com/telerik/mobile-cli-lib/pull/791


rosen-vladimirov:
:+1:  from @dtopuzov

PR #2064, opened 2016-09-17
A proper fix for all Ctrl+C issues

tzraikov:
`tns run ios` - on device
`tns livesync ios --watch`
`tns debug --watch`
.. etc

Depends on:
https://github.com/telerik/mobile-cli-lib/pull/807

### DIFF
--- a/lib/services/ios-debug-service.ts
+++ b/lib/services/ios-debug-service.ts
@@ -31,9 +31,13 @@ class IOSDebugService implements IDebugService {
 		private $utils: IUtils,
 		private $iOSNotification: IiOSNotification,
 		private $iOSSocketRequestExecutor: IiOSSocketRequestExecutor,
-		private $socketProxyFactory: ISocketProxyFactory) { }
+		private $processService: IProcessService,
+		private $socketProxyFactory: ISocketProxyFactory) {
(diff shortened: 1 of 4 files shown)


lib/services/ios-debug-service.ts
```diff
@@ -167,13 +176,17 @@ class IOSDebugService implements IDebugService {
 		return (() => {
 			let timeout = this.getReadyForAttachTimeout();
 			this.$iOSSocketRequestExecutor.executeAttachRequest(device, timeout).wait();
-			this.wireDebuggerClient(() => device.connectToPort(inspectorBackendPort)).wait();
+			this.wireDebuggerClient(device).wait();
 		}).future<void>()();
 	}
 
-	private wireDebuggerClient(factory: () => net.Socket): IFuture<void> {
+	private wireDebuggerClient(device: Mobile.IiOSDevice): IFuture<void> {
```
rosen-vladimirov: You can make the device argument not mandatory
```TypeScript
private wireDebuggerClient(device?: Mobile.IiOSDevice): IFuture<void> {
```

This way you will not have to pass `null` when calling the method without device.

tzraikov: 👍

PR #2099, opened 2016-10-06
removing unnecessary functionality

Plamen5kov:
* `lib/services/android-project-service.ts`: unnecessary functionality deleting `AndroidManifest.xml`
* `lib/services/platform-service.ts`: unnecessary functionality checking for `tns_modules` in `app/` directory.

ping: @hshristov, @tzraikov @PanayotCankov, @hdeshev 

_Edit_: Expecting failing obsolete tests

### DIFF
--- a/lib/services/android-project-service.ts
+++ b/lib/services/android-project-service.ts
@@ -312,15 +312,7 @@ export class AndroidProjectService extends projectServiceBaseLib.PlatformProject
 
 	public ensureConfigurationFileInAppResources(): IFuture<void> {
 		return (() => {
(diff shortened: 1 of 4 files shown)


rosen-vladimirov:
What about removing `ensureConfigurationFileInAppResources` - this method was created only to add AndroidManifest.xml (Info.plist) in `app/App_Resources/<platform>/` dir. However this was possible in versions like 1.3, 1.5, 1.6... Maybe we should remove these methods as well

Plamen5kov:
Thank you, I'll check that out too.

PR #2100, opened 2016-10-06
Kill all references to broccoli.

hdeshev:
We were not using it for quite some time anyway.


lib/tools/node-modules/node-modules-builder.ts
```diff
@@ -9,10 +9,9 @@ import {sleep} from "../../../lib/common/helpers";
 
 let glob = require("glob");
 
-export class Builder implements IBroccoliBuilder {
+export class NodeModulesBuilder {
```
rosen-vladimirov: shouldn't this implement `INodeModulesBuilder `

hdeshev: Fixed!

PR #2108, opened 2016-10-07
refactoring tns install to do only npm install

Plamen5kov:
`tns install` command no longer tries to add platform and only does npm install. Deprecate soon so only `npm install` is used. Before this command was doing `npm install` + `platform add`. This PR removes the `platform add` functionality and soon the whole command will be deprecated.

ping: @hdeshev

### DIFF
--- a/lib/commands/install.ts
+++ b/lib/commands/install.ts
@@ -1,6 +1,5 @@
 export class InstallCommand implements ICommand {
 	constructor(private $projectData: IProjectData,
-		private $projectDataService: IProjectDataService,
 		private $pluginsService: IPluginsService,
 		private $fs: IFileSystem,
 		private $stringParameter: ICommandParameter,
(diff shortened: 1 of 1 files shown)


lib/commands/install.ts
```diff
@@ -1,12 +1,7 @@
-import {EOL} from "os";
-
 export class InstallCommand implements ICommand {
-	constructor(private $platformsData: IPlatformsData,
-		private $platformService: IPlatformService,
-		private $projectData: IProjectData,
+	constructor(private $projectData: IProjectData,
 		private $projectDataService: IProjectDataService,
```
rosen-vladimirov: this dependency is not used anymore

lib/commands/install.ts
```diff
@@ -21,26 +16,7 @@ export class InstallCommand implements ICommand {
 
 	private installProjectDependencies(): IFuture<void> {
 		return (() => {
```
rosen-vladimirov: you can modify `installProjectDependencies` in the following way, in order to skip creation of additional FiberFuture:
```TypeScript
private installProjectDependencies(): IFuture<void> {
    this.$pluginsService.ensureAllDependenciesAreInstalled();
}
```

Plamen5kov:
Closed in favor of killing the whole functionality in a future release, and introducing the runtime as a normal npm dependency.

PR #2118, opened 2016-10-11
Prepare plugins independently of node_modules

### LINKED ISSUE #2110: Prepare plugins without node_modules
Preparing platforms with the `--bundle` option does not prepare plugins since plugin discovery and preparation is coupled to the node_modules -> tns_modules preparation.

We need to decouple it an always run plugin native code preparation when preparing with `--bundle`

hdeshev:
- Extract dependency resolution from the node_modules copy code.
- Decouple plugin preparation from node_modules copying.
- Reuse resolved dependencies for plugin preparation.

Fixes #2110

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,9 @@ before_script:
 - npm install grunt
 - node_modules/.bin/grunt enableScripts:false
 - npm install
+- grunt rebuild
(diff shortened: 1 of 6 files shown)


Plamen5kov:
:+1:

PR #2123, opened 2016-10-12
Fixed: Livesync not working when started for the first time on device

### LINKED ISSUE #2072: First-time livesync fails on device
1. Create a new project with `tns create <project_name>`
2. Close all simulators and connect an iOS device.
3. Execute `tns platform add ios`
3. Execute `tns livesync ios --watch`
4. The project will start
5. Make a change in some js file. The livesync will fail to sync the file.

tzraikov:
Fixes #2072

### DIFF
--- a/lib/services/livesync/ios-platform-livesync-service.ts
+++ b/lib/services/livesync/ios-platform-livesync-service.ts
@@ -29,10 +29,11 @@ class IOSPlatformLiveSyncService extends PlatformLiveSyncServiceBase {
 					let localToDevicePaths = this.$projectFilesManager.createLocalToDevicePaths(deviceAppData, projectFilesPath, null, this.liveSyncData.excludedProjectDirsAndFiles);
 					let afterSyncAction: () => IFuture<void>;
 
+					this.transferFiles(deviceAppData, localToDevicePaths, this.liveSyncData.projectFilesPath, true).wait();
+
 					if(installed) {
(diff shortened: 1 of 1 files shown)


lib/services/livesync/ios-platform-livesync-service.ts
```diff
@@ -29,10 +29,11 @@ class IOSPlatformLiveSyncService extends PlatformLiveSyncServiceBase {
 					let localToDevicePaths = this.$projectFilesManager.createLocalToDevicePaths(deviceAppData, projectFilesPath, null, this.liveSyncData.excludedProjectDirsAndFiles);
 					let afterSyncAction: () => IFuture<void>;
 
+					this.transferFiles(deviceAppData, localToDevicePaths, this.liveSyncData.projectFilesPath, true).wait();
```
rosen-vladimirov: how will this code work in case the application is not installed on device? Where will it transfer the files as there's no sandbox on device (`/Library/Application Support` directory for current app)?

tzraikov: The idea is to synchronize all files the first time when livesync is called, because all consequent updates will be partial and it will bring inconsistence. The app will not be able to find all js files and it will crash.

rosen-vladimirov: Ok, the name `installed` confused me.

Issue #2128, opened 2016-10-14
Add option to subscribe to NativeScript email lists when installing CLI


tzraikov:

SvetoslavTsenov:
An environment variable CLI_NOPROMT=1  is not working. Still asking for email address

PR #2134, opened 2016-10-17
Added email registration for NativeScript mailing lists as post install step

tzraikov:
#2128

### DIFF
--- a/lib/commands/post-install.ts
+++ b/lib/commands/post-install.ts
@@ -1,0 +1,86 @@
+import {PostInstallCommand} from "../common/commands/post-install";
+import * as emailValidator  from "email-validator";
+import * as queryString from "querystring";
+import * as helpers from "../common/helpers";
+
+export class PostInstallCliCommand extends PostInstallCommand {
+
+	private logger: ILogger;
+
+	constructor($fs: IFileSystem,
+		private $httpClient: Server.IHttpClient,
+		private $prompter: IPrompter,
+		private $userSettingsService: IUserSettingsService,
+		$staticConfig: Config.IStaticConfig,
+		$commandsService: ICommandsService,
+		$htmlHelpService: IHtmlHelpService,
+		$options: ICommonOptions,
+		$doctorService: IDoctorService,
+		$analyticsService: IAnalyticsService,
+		$logger: ILogger) {
+            super($fs, $staticConfig, $commandsService, $htmlHelpService, $options, $doctorService, $analyticsService, $logger);
+            this.logger = $logger;
+	}
+
+	public execute(args: string[]): IFuture<void> {
+		return (() => {
+			super.execute(args).wait();
+
+			if (this.shouldAskForEmail()) {
+				this.logger.out("Leave your e-mail address here to subscribe for NativeScript newsletter and product updates, tips and tricks:");
+ 				let email = this.getEmail("(press Enter for blank)").wait();
+ 				this.$userSettingsService.saveSetting("EMAIL_REGISTERED", true).wait();
+ 				this.sendEmail(email);
+ 			}
+
+		}).future<void>()();
+	}
+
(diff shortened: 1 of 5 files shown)


lib/commands/post-install.ts
```diff
@@ -0,0 +1,91 @@
+import {PostInstallCommand} from "../common/commands/post-install";
+let emailValidatorModule = require('email-validator');
+let queryString = require('querystring');
+import * as helpers from "../common/helpers";
+
+export class PostInstallCliCommand extends PostInstallCommand {
+
+    private fileSystem: IFileSystem;
```
rosen-vladimirov: This property is just assigned, it's not used anywhere, consider removing it.

lib/commands/post-install.ts
```diff
@@ -0,0 +1,91 @@
...
+    private logger: ILogger;
```
rosen-vladimirov: same here - the property is just assigned, but not used anywhere

lib/commands/post-install.ts
```diff
@@ -0,0 +1,91 @@
...
+		$analyticsService: IAnalyticsService,
+		$logger: ILogger) {
+            super($fs, $staticConfig, $commandsService, $htmlHelpService, $options, $doctorService, $analyticsService, $logger);
+            this.fileSystem = $fs;
+            this.logger = $logger;
+	}
+
+	public execute(args: string[]): IFuture<void> {
+		return (() => {
+            super.execute(args).wait();
```
rosen-vladimirov: mixed spaces and tabs

lib/commands/post-install.ts
```diff
@@ -0,0 +1,91 @@
...
+
+           	if (this.shouldAskForEmail()) {
+                this.logger.out("Leave your e-mail address here to subscribe for NativeScript newsletter and product updates, tips and tricks:");
+ 				let email = this.getEmail("(press Enter for blank)").wait();
+ 				this.sendEmail(email);
+ 				this.$userSettingsService.saveSetting("EMAIL_REGISTERED", true).wait();
```
rosen-vladimirov: if the sendEmail command fails, next time when user installs NS CLI, he'll be prompted again (as the EMAIL_REGISTERED value will never be written in the settings file), is this expected? In case not, you can swap the two lines above.

lib/commands/post-install.ts
```diff
@@ -0,0 +1,91 @@
...
+ 			}
+
+		}).future<void>()();
+	}
+
+    private shouldAskForEmail(): boolean {
+        if (!helpers.isInteractive() || process.env.CLI_NOPROMPT === "1" || this.$userSettingsService.getSettingValue("EMAIL_REGISTERED").wait()) {
```
rosen-vladimirov: this could be simplified to:
```TypeScript
return helpers.isInteractive() && process.env.CLI_NOPROMPT !== "1" && !this.$userSettingsService.getSettingValue("EMAIL_REGISTERED").wait();
```

But I do not insist changing it.

lib/commands/post-install.ts
```diff
@@ -0,0 +1,91 @@
+import {PostInstallCommand} from "../common/commands/post-install";
+let emailValidatorModule = require('email-validator');
```
rosen-vladimirov: you can include the email-validator.d.ts file from here:
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/email-validator/email-validator.d.ts
Which will allow you to use 
```TypeScript
import * as emailValidator  from "email-validator";
```

But this is not a merge stopper :)

lib/commands/post-install.ts
```diff
@@ -0,0 +1,91 @@
...
+let queryString = require('querystring');
```
rosen-vladimirov: you can use 
```TypeScript
import * as querystring from "querystring";
```

PR #2136, opened 2016-10-17
don't use flatten logic

Plamen5kov:
Copies `/app/node_modules` as is in `/platforms/android/src/main/assets/app/tns_modules`

If this PR is merged, the runtimes require logic must implement [node require](https://nodejs.org/api/modules.html#modules_all_together) specification.

Currently works only for npm v.2 node_modules, so if the user does `npm install` using npm v.3.x as opposed to `tns install`(using npm v.2.x), `tns prepare android` command will execute incorrectly.

Related to: https://github.com/NativeScript/android-runtime/pull/587


Plamen5kov:
This PR is obsolete, instead follow [this one](https://github.com/NativeScript/nativescript-cli/pull/2152)

PR #2151, opened 2016-10-21
Add snapshot plugin to project package.json

jasssonpet:
This will enable the [snapshot plugin](https://www.npmjs.com/package/nativescript-dev-android-snapshot) for newly created apps.

### DIFF
--- a/lib/services/project-data-service.ts
+++ b/lib/services/project-data-service.ts
@@ -63,7 +63,10 @@ export class ProjectDataService implements IProjectDataService {
 					"description": "NativeScript Application",
 					"license": "SEE LICENSE IN <your-license-filename>",
 					"readme": "NativeScript Application",
-					"repository": "<fill-your-repository-here>"
+					"repository": "<fill-your-repository-here>",
+					"devDependencies": {
+						"nativescript-dev-android-snapshot": "^0.*.*"
+					}
 				}).wait();
 			}
 


dtopuzov:
It looks ok, but please DO NOT merge, it will hit the current CI infrastructure.

lib/services/project-data-service.ts
```diff
@@ -63,7 +63,10 @@ export class ProjectDataService implements IProjectDataService {
 					"description": "NativeScript Application",
 					"license": "SEE LICENSE IN <your-license-filename>",
 					"readme": "NativeScript Application",
-					"repository": "<fill-your-repository-here>"
+					"repository": "<fill-your-repository-here>",
+					"devDependencies": {
```
rosen-vladimirov: why the code is here? It can be added to the templates as all devDependencies from the template will be added to the currently created project:
https://github.com/NativeScript/nativescript-cli/blob/master/lib/services/project-service.ts#L76

https://github.com/NativeScript/nativescript-cli/blob/master/lib/services/project-service.ts#L100-L102

enchev: In this case I prefer to add this to the templates. @jasssonpet what do you think?

jasssonpet: I can move it to the templates, but are we sure there aren't any templates outside the NativeScript organization? Or should I add it to the hello-world template and it will be copied like the App_Resources are?

enchev: I've pull requests for all three templates. No other templates outside of the organization
NativeScript/template-hello-world-ng#24
NativeScript/template-hello-world#53
NativeScript/template-hello-world-ts#14

Issue #2149, opened 2016-10-20
Implement incremental prepare


tzraikov:
Currently all prepare steps are executed on every `run` and `debug` command. This process can be improved:

When executing `tns run`, `tns debug` or `tns livesync`:

- [x] Changes in app folder (ts,js,xml,css files) should not trigger prepare task
- [x] Changes in app resources should trigger prepare task
- [x] Changes in modules should trigger prepare task
- [x] Build task will be triggered only when a prepare task was executed (or when the target device has changed for iOS)

Other commands:

- [x] `tns prepare` should always execute the full prepare task
- [x] `tns build` should always execute the full prepare and build tasks

Issue #1989, opened 2016-08-10
TNS and Dependances vs DevDepandancies


NathanaelA:
I was helping @NathanWalker tonight with an issue; and it "appears" tns makes the assumption that if the module is NOT listed in the main package.json file devDependancy; that it is a valid dependency.     

However, when dealing with some larger projects a devDependancy will install its own devDependancies; and TNS will see these as "Dependancies" and copy them into `/platforms/android/src/main/assets/app/tns_modules/` -- then the gradle builder system will attempt to use them as valid dependancies.   In Nathan's case; the tar and tar-pack that are pulled in by the gyp tooling were being copied and attempted to be processed.   Which would cause the android tooling to fail with an error that I would bet you that Nathan would tell you is very unhelpful.

I believe this is a flawed assumption to assume anything not listed as a direct "devDependancy" is a valid dependancy.   I assume the best way to fix this would be to actually iterate through all the package.json files and make a real dependency tree -- probably cache it in the node_modules (use foldernames and main package.json file as as hash name.   I would also probably recommend that you make tns require npm 3 or later so that everything is installed in the node_modules folder rather than in sub-directories of the plugins, as this should make this a lot simpler to manage.

Plamen5kov:
fixed for 2.4.0 release.

Issue #1845, opened 2016-06-14
Remove NPM as a dependency, use the NPM installed on users' machine


ErjanGavalji:
I am copying here the discussion from https://github.com/NativeScript/nativescript-cli/issues/1740, for it deserves a separate thread...

ErjanGavalji:
By @enchev:
Hey @rosen-vladimirov,

Can you post here more info why we have our own nmp and why we do not use the instilled one. What problems we solved?

Thanks

ErjanGavalji:
By @rosen-vladimirov:
Hi @enchev ,
As you are asking, using `npm` as dependency resolves exactly this issue.  In case we were relying on user's npm, you wouldn't have any workaround for this particular problem and in fact nativescript-cli would have been unusable with npm 3. And btw this is not "our" npm, it's the real npm, we are just using it as dependency.

### How does npm work
NativeScript projects rely on npm and npm's structure. This means that each project has package.json with dependencies and devDependencies. When you call `npm install` all of your dependencies and devDependencies will be installed and placed in your `node_modules` directory. Different `npm` versions place the dependencies in a different way inside `node_modules` dir.
`npm` versions 2.x.x will install each dependency in it's own directory and each of it's dependencies will be installed inside it's node_modules directory. All dependencies of devDependencies will be installed in the respective `node_modules` dir.
`npm` versions 3.x.x tries to flatten the dependencies and install everything at your project's node_modules directory. So for example in case devDependency A has dependency B, which has dependency C, at the root node_modules you'll receive A, B and C directories.

### How does NativeScript CLI handle dependencies
Due to known behavior of NativeScript's require, all dependencies used in a project must be flattened when copied to `platforms/<platform>/.../tns_modules` directory. So when the project is prepared, CLI tries to flatten the dependencies from `node_modules` dir, filters the one which are `devDependencies` or dependencies of `devDependencies` or dependency of dependency of dependency ... of `devDependencies` and process them to `tns_modules` dir inside `platforms`.

### The problem
When npm 3.x.x is installed and user executes `npm install`, the flattening of modules makes it really hard for NativeScript CLI to understand from where is each package inside `node_modules` dir. It would require parsing all package.json files inside node_modules and it's subdirectories, checking each package and is it a dependency or devDependency, etc.

### The workaround
NativeScript CLI has `npm` 2.x.x as dependency and executes `npm install` with it. This allows us to have the "old" way of installing the dependencies, where it is much easier to filter the devDependencies.

[...]

Issue #1740, opened 2016-05-05
Dev Dependencies (like Gulp, etc) getting built and build is failing because of which Gulp integration is not working currently.


hdeshev:
_From @Headcult on April 26, 2016 11:30_

Thanks for the amazing work on integrating Angular 2 with NativeScript at such a good pace. Looking forward for future releases.

The main issue that I'm facing now is:
Dev Dependencies (like Gulp, etc) getting built and build is failing because of which Gulp integration is not working currently.

Can this be fixed at the earliest as we are unable to use Gulp within our project when using Angular 2 and NativeScript?

**My Dev Dependencies:**

`"devDependencies": {
        "nativescript-dev-typescript": "^0.3.1",
        "shelljs": "^0.5.3",
        "typescript": "^1.8.9",
        "gulp": "^3.9.1",
        "gulp-debug": "^2.1.2",
        "gulp-typescript": "^2.12.1",
        "del": "^2.2.0",
        "run-sequence": "^1.1.4"
    }`

**Build Failure:**

`:processDebugResources FAILED

FAILURE: Build failed with an exception.
- What went wrong:
  Execution failed for task ':processDebugResources'.
  > com.android.ide.common.process.ProcessException: org.gradle.process.internal.E
  xecException: Process 'command 'C:\Android\android-sdk\build-tools\23.0.3\aapt.e
  xe'' finished with non-zero exit value 1`

_Copied from original issue: NativeScript/nativescript-angular#195_

hdeshev:
I finally reproduced the problem, and I believe this to be a problem with how nativescript-cli handles npm dependencies. Steps to reproduce:

1. Start an empty project.
2. Run it using `$ tns run android`
3. Install gulp: ` $ npm install gulp --save-dev`
4. `$ tns run android`

Working around the problem is easy, but not obvious -- don't call `npm`, but let the tns tool do that for you by adding gulp to your `package.json` and running `tns install` or just `tns run android`

PeterStaev:
Was struggling with the same problem! I can confirm that using `tns install` instead of `npm install` does the trick. Thanks @hdeshev !

enchev:
Hey @rosen-vladimirov,

Can you post here more info why we have our own nmp and why we do not use the instilled one. What problems we solved?

Thanks

rosen-vladimirov:
Hi @enchev ,
[...]

PR #2152, opened 2016-10-21
Copy node_modules to platform on prepare (fix Node6/npm 3.x bug)

Pip3r4o:
Changes in this PR will allow copying of all `production` (non-dev) node_modules into the platform's `tns_modules` independent of whether they have been installed using `tns install` (npm 2.x) or npm install with `npm` version 3+.

The proposed implementation however will iterate through the production modules' package.jsons on every build. Traversal of all package's production dependencies is also a bit slower than before. I'll look into a way to address only changes on subsequent builds, and potentially optimize traversals.

Issue https://github.com/NativeScript/nativescript-cli/issues/2149 and the respective PR will most likely address preparing of modules on each rebuild. Has yet to be verified.

**There are currently 3 failing tests testing how and where npm_modules are copied into the platforms dir.**

The changes have been tested with the Groceries app (master branch) with both `tns install` (npm 2), and `npm install` (npm 3) 

Addresses issues: #1989, #1845, #1740

Also invalidates PR #2136 

Depends on changes in both runtimes: https://github.com/NativeScript/android-runtime/pull/587, https://github.com/NativeScript/ios-runtime/pull/673;

ping: @rosen-vladimirov @hdeshev @Plamen5kov @atanasovg

### DIFF
--- a/lib/tools/node-modules/node-modules-builder.ts
+++ b/lib/tools/node-modules/node-modules-builder.ts
@@ -3,9 +3,9 @@ import * as fs from "fs";
 import * as path from "path";
 import * as shelljs from "shelljs";
 import Future = require("fibers/future");
-import {NpmDependencyResolver, TnsModulesCopy, NpmPluginPrepare} from "./node-modules-dest-copy";
+import { TnsModulesCopy, NpmPluginPrepare } from "./node-modules-dest-copy";
 import * as fiberBootstrap from "../../common/fiber-bootstrap";
-import {sleep} from "../../../lib/common/helpers";
+import { sleep } from "../../../lib/common/helpers";
 
 let glob = require("glob");
 
@@ -37,7 +37,7 @@ export class NodeModulesBuilder implements INodeModulesBuilder {
 				}, (er: Error, files: string[]) => {
 					fiberBootstrap.run(() => {
 
-						while(this.$lockfile.check().wait()) {
+						while (this.$lockfile.check().wait()) {
 							sleep(10);
 						}
 
@@ -85,7 +85,7 @@ export class NodeModulesBuilder implements INodeModulesBuilder {
 						let intervalId = setInterval(() => {
 							fiberBootstrap.run(() => {
 								if (!this.$lockfile.check().wait() || future.isResolved()) {
-									if(!future.isResolved()) {
+									if (!future.isResolved()) {
 										future.return();
 									}
 									clearInterval(intervalId);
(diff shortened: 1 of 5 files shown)


lib/tools/node-modules/node-modules-builder.ts
```diff
@@ -133,27 +133,165 @@ export class NodeModulesBuilder implements INodeModulesBuilder {
...
-			let nodeModules = this.getChangedNodeModules(absoluteOutputPath, platform, lastModifiedTime).wait();
 
-			const resolver = new NpmDependencyResolver(this.$projectData.projectDir);
-			const resolvedDependencies = resolver.resolveDependencies(_.keys(nodeModules), platform);
+			let productionDependencies = this.getProductionDependencies(this.$projectData.projectDir);
+
+			// console.log(productionDependencies);
+
+			// TODO: Pip3r4o - result is not used currently
+			// let nodeModules = this.getChangedNodeModules(absoluteOutputPath, platform, lastModifiedTime).wait();
```
hdeshev: Can we get rid of commented out code?

lib/tools/node-modules/node-modules-builder.ts
```diff
@@ -133,27 +133,165 @@ export class NodeModulesBuilder implements INodeModulesBuilder {
...
-			npmPluginPrepare.preparePlugins(resolvedDependencies, platform);
+			npmPluginPrepare.preparePlugins(productionDependencies, platform);
 		}).future<void>()();
 	}
 
+	public getProductionDependencies(projectPath: string) {
+		var deps: any = [];
+		var seen: any = {};
+
+		var pJson = path.join(projectPath, "package.json");
```
hdeshev: Please, avoid weird abbreviations like `pJson`. How about `packageJsonPath`?

lib/tools/node-modules/node-modules-builder.ts
```diff
@@ -133,27 +133,165 @@ export class NodeModulesBuilder implements INodeModulesBuilder {
...
+					}
+
+					var dep = addDependency(deps, name, modulePath, 0);
+
+					traverseModule(modulePath, 0, dep);
+				}
+
+			}
+			// module is scoped 
+			else {
```
hdeshev: `} else {`

lib/tools/node-modules/node-modules-builder.ts
```diff
@@ -133,27 +133,165 @@ export class NodeModulesBuilder implements INodeModulesBuilder {
...
+				var moduleName = name.substring(scopeSeparatorIndex + 1, name.length);
+				var scopedModulePath = path.join(nodeModulesDir, scope, moduleName);
+
+				var exists = ensureModuleExists(scopedModulePath);
+
+				if (exists) {
+					var dep = addDependency(deps, name, scopedModulePath, 0);
+					traverseModule(scopedModulePath, depth, dep);
+				}
+				else {
```
hdeshev: `} else {`

lib/tools/node-modules/node-modules-builder.ts
```diff
@@ -133,27 +133,165 @@ export class NodeModulesBuilder implements INodeModulesBuilder {
...
+					traverseModule(modulePath, depth + 1, dep);
+				} else {
+					modulePath = path.join(nodeModulesDir, name); // /node_modules/<package>
+					exists = ensureModuleExists(modulePath);
+
+					if(!exists) {
+						return;
+					}
+
+					var dep = addDependency(deps, name, modulePath, 0);
```
hdeshev: Again, please avoid abbreviations: `dep`

lib/tools/node-modules/node-modules-builder.ts
```diff
@@ -133,27 +133,165 @@ export class NodeModulesBuilder implements INodeModulesBuilder {
...
+			var depth = 0;
+			var directory = path.join(nodeModulesDir, key);
+
+			// find and traverse child with name `key`, parent's directory -> dep.directory
+			traverseChild(key, directory, depth);
+		});
+
+		return filterUniqueDirectories(deps);
+
+		function traverseChild(name: string, currentModulePath: string, depth: number) {
```
hdeshev: Those inner functions are very confusing and unreadable. Consider moving them to methods of this class or an entirely new class.

lib/tools/node-modules/node-modules-builder.ts
```diff
@@ -133,27 +133,165 @@ export class NodeModulesBuilder implements INodeModulesBuilder {
...
+					if (!exists) {
+						return;
+					}
+
+					var dep = addDependency(deps, name, scopedModulePath, depth + 1);
+					traverseModule(scopedModulePath, depth + 1, dep);
+				}
+			}
+
+			function traverseModule(modulePath: string, depth: number, currentDependency: any) {
```
hdeshev: What is the difference when traversing a "child" and a "module". I think one of these functions needs a better name.

lib/tools/node-modules/node-modules-builder.ts
```diff
@@ -133,27 +133,165 @@ export class NodeModulesBuilder implements INodeModulesBuilder {
...
+				try {
+					var exists = fs.lstatSync(modulePath);
+					return exists.isDirectory();
+				} catch (e) {
+					return false;
+				}
+			}
+		}
+
+		function filterUniqueDirectories(dependencies: any) {
```
hdeshev: Can we use a map keyed by the dependency directory here (simulating a set collection) instead of doing linear traversals all the time?

Pip3r4o: Noted. I used a string map initially to check if the plugin's already present in the list, but that approach omitted corner cases where 2 plugins with different versions were present. I'll revisit it again

Pip3r4o:
@hdeshev about `2.` - currently we will pull both versions of the package, as extracted by `npm`, you can't tell by version alone if any of the packages' dependents are compatible with all, or just 1 of said package's version. 
Previously only the package with the greatest version will be pulled from node_modules, which is not always desired.

lib/tools/node-modules/node-modules-builder.ts
```diff
@@ -133,27 +133,156 @@ export class NodeModulesBuilder implements INodeModulesBuilder {
...
 	}
 
+	public getProductionDependencies(projectPath: string) {
+		let deps: any = [];
+		let seen: any = {};
+
+		let pJson = path.join(projectPath, "package.json");
+		let nodeModulesDir = path.join(projectPath, "node_modules");
+
+		let content = require(pJson);
```
rosen-vladimirov: I would advise using `$fs` as this will make the code more testable:
```TypeScript
this.$fs.readJson(pJson).wait();
```

lib/tools/node-modules/node-modules-builder.ts
```diff
@@ -133,27 +133,156 @@ export class NodeModulesBuilder implements INodeModulesBuilder {
...
+
+		Object.keys(content.dependencies).forEach((key) => {
```
rosen-vladimirov: if there's no dependencies section in the package.json, this code will fail. I would use lodash's `_.each`:
```TypeScript
_.keys(content.dependencies)
   .forEach(key => { 
    ....
    });
```

lib/tools/node-modules/node-modules-builder.ts
```diff
@@ -133,27 +133,156 @@ export class NodeModulesBuilder implements INodeModulesBuilder {
...
+				if (packageJsonExists) {
+					let packageJsonContents = require(packageJsonPath);
+
+					if (!!packageJsonContents.nativescript) {
+						// add `nativescript` property, necessary for resolving plugins
+						currentDependency.nativescript = packageJsonContents.nativescript;
+					}
+
+					if (packageJsonContents.dependencies) {
+						Object.keys(packageJsonContents.dependencies).forEach((key) => {
```
rosen-vladimirov: Same as above - it's safer to use:
```TypeScript
_.keys(packageJsonContent.dependencies).forEach(...
```

lib/tools/node-modules/node-modules-builder.ts
```diff
@@ -133,27 +133,156 @@ export class NodeModulesBuilder implements INodeModulesBuilder {
...
+							traverseChild(key, modulePath, depth);
+						});
+					}
+				}
+			}
+
+			function addDependency(deps: any[], name: string, directory: string, depth: number) {
```
rosen-vladimirov: this function needs return type

lib/tools/node-modules/node-modules-builder.ts
rosen-vladimirov: Also please add return type to the function

lib/tools/node-modules/node-modules-builder.ts
```diff
@@ -133,27 +133,156 @@ export class NodeModulesBuilder implements INodeModulesBuilder {
...
+				let dep: any = {};
```
rosen-vladimirov: you can use shorter syntax here:
```TypeScript
let dep = {
	name,
	directory,
	depth
};
```

lib/tools/node-modules/node-modules-builder.ts
```diff
@@ -133,27 +133,156 @@ export class NodeModulesBuilder implements INodeModulesBuilder {
...
+				} catch (e) {
+					return false;
+				}
+			}
+		}
+
+		function filterUniqueDirectories(dependencies: any) {
+			let unique: any = [];
+			let distinct: any = [];
+			for (let i in dependencies) {
```
rosen-vladimirov: You can get directly the dependency by using:
```TypeScript
for (let dep of dependencies) {
   if (distinct.indexOf(dep.directory) > -1) {
....
}
```

lib/tools/node-modules/node-modules-dest-copy.ts
```diff
@@ -10,96 +10,17 @@ export interface ILocalDependencyData extends IDependencyData {
...
 	constructor(
 		private outputRoot: string,
 		private $fs: IFileSystem
 	) {
 	}
 
-	public copyModules(dependencies: IDictionary<ILocalDependencyData>, platform: string): void {
-		_.each(dependencies, dependency => {
+	public copyModules(dependencies: any[], platform: string): void {
+		for (var entry in dependencies) {
```
rosen-vladimirov: var -> let

lib/tools/node-modules/node-modules-dest-copy.ts
```diff
@@ -10,96 +10,17 @@ export interface ILocalDependencyData extends IDependencyData {
...
+			var dependency = dependencies[entry];
```
rosen-vladimirov: var -> let

lib/tools/node-modules/node-modules-dest-copy.ts
```diff
@@ -110,22 +31,24 @@ export class TnsModulesCopy {
...
+			let isScoped = dependency.name.indexOf("@") === 0;
+			let targetDir = this.outputRoot;
+
+			if (isScoped) {
+				targetDir = path.join(this.outputRoot, dependency.name.substring(0, dependency.name.indexOf("/")));
+			}
+
+			shelljs.mkdir("-p", targetDir);
+			shelljs.cp("-Rf", dependency.directory, targetDir);
+			// console.log("Copied dependency: " + dependency.name);
```
rosen-vladimirov: can you remove the commented line or add it as log trace message:
```TypeScript
this.$logger.trace(`Copied dependency: ${dependency.name}`);
```

Pip3r4o:
@hdeshev @rosen-vladimirov addressed concerns with style and readability

lib/tools/node-modules/node-modules-dependencies-builder.ts
```diff
@@ -0,0 +1,143 @@
...
+            directory,
+            depth
+        };
+
+        this.resolvedDependencies.push(dependency);
+
+        return dependency;
+    }
+
+    private ensureModuleExists(modulePath: string): boolean {
```
hdeshev: I'd rename this to `moduleExists`. Nothing to ensure anyway.

lib/tools/node-modules/node-modules-dependencies-builder.ts
```diff
@@ -0,0 +1,143 @@
...
+        });
+
+        return this.resolvedDependencies;
+    }
+
+    private traverseDependency(name: string, currentModulePath: string, depth: number): void {
+        // check if key appears in a scoped module dependency
+        let isScoped = name.indexOf('@') === 0;
+
+        if (!isScoped) {
```
hdeshev: The scoped/unscoped branches seem to be mostly the same code copied/pasted over. Can we eliminate the duplication in a shared method and pass the different paths as parameters?

Pip3r4o: You're right! I seem to have missed it!

Plamen5kov:
:+1: looks good, the sooner this is in master, the better

PR #2156, opened 2016-10-26
Extract tns-core-modules on root level of tns_modules

Pip3r4o:
This PR aims to partially revert changes made in #2152 such that `tns-core-modules` submodules will once again be copied directly into `tns_modules`.

@hdeshev @dtopuzov


lib/tools/node-modules/node-modules-dest-copy.ts
```diff
@@ -30,6 +30,8 @@ export class TnsModulesCopy {
 				Future.wait(deleteFilesFutures);
 
 				shelljs.rm("-rf", path.join(tnsCoreModulesResourcePath, "node_modules"));
+				shelljs.cp("-Rf", path.join(tnsCoreModulesResourcePath, "*"), this.outputRoot);
```
hdeshev: Add a scare comment why we're doing all this crap, and we have a deal :grin:

Issue #2112, opened 2016-10-10
Update webpack plugin to run as a npm script


hdeshev:
Change the behavior for release webpacking so that it is run by a npm script. That script should be able to:
- webpack the application code, outputting bundles into the platform subfolder. Possibly introduce a `tns platform info` command to get the correct platform app dir.
- prepare/build/run the platform with the --bundle option

This change will remove much of the magic in the bundling process and make it easy to customize by just changing the respective `webpack.config.js` or `webpack.config.<platform>.js` files.

PR #2159, opened 2016-10-26
clean-app command installs the platform if not present.

hdeshev:
This allows for easier documentation workflows when using webpack.


hdeshev:
Related to #2112

Issue #2164, opened 2016-10-27
Deprecate support for Node 0.12


rosen-vladimirov:
Currently CLI can be used with Node.js 0.12.x. However this version will not be supported by Node.js Foundation after the end of 2016 - check the [table](https://github.com/nodejs/LTS). So we'll deprecate support for Node.js 0.12 (with 2.4 users will see warning message) and we'll remove it as supported version in 2.5.

PR #2165, opened 2016-10-27
Deprecate support for Node.js 0.12.x

rosen-vladimirov:
Deprecate support for Node.js 0.12.x as it will not be supported after 2016 by Node.js Foundation.

Implements: https://github.com/NativeScript/nativescript-cli/issues/2164


rosen-vladimirov:
run ci

PR #2172, opened 2016-10-28
Unzip NativeScript Inspector Application

KristinaKoeva:
Sadly $npmInstallationManager.install does not install the package, it only inserts it in the cache through the npm cache add command
Since npm cache add command does not execute scripts our posinstall script that extract the Inspector Application does not execute as well
So until this behavior is changed this ugly workaround should not be deleted


KristinaKoeva:
@Plamen5kov promised to fix this ugliness soon

PR #2184, opened 2016-11-01
Docs(README.md): Add docs for the --template option of tns create

sis0k0:


README.md
```diff
@@ -317,21 +317,41 @@ Run `tns help` to view all available commands in the browser. Run `tns help <Com
...
+```
+
+Or you can simply use the shorthand `tsc` and `ng` options.
+
+```Shell
+tns create MyApp --tsc
+tns create MyApp --ng
+```
+With the `template` option you can also specify a local or a remote path to the template that you want to use to create your project.
+For example, if you want to use the nightly build of the default JavaScript template, run the following command.
```
rosen-vladimirov: This is not correct - the code below will use latest master branch, not the nightly build of the template.
In order to use the nightly build (marked as tag `next` in npm) the command should be:
```Shell
tns create MyApp --template tns-template-hello-world@next
```
However I'm not sure we even have `next` for the template.

sis0k0: We don't have `@next` for the templates. Fetching the master branch from github is the only way to use the latest unofficial (or nightly in other words) template.

README.md
```diff
@@ -11,18 +11,18 @@ The NativeScript CLI lets you create, build, and deploy NativeScript-based proje
 * [Supported Platforms](#supported-platforms "The mobile platforms you can target with NativeScript")
 * [System Requirements](#system-requirements "The hardware and software requirements for setting up and working with the NativeScript CLI")
 * [Installation](#installation "How to configure and install the NativeScript CLI")
-	* [Install the NativeScript CLI](#install-the-nativescript-cli)
-	* [Configure Proxy Usage](#configure-proxy-usage)
+    * [Install the NativeScript CLI](#install-the-nativescript-cli)
```
rosen-vladimirov: well, CLI uses tabs 😺 
However this is not a big deal, it's just the README and I believe we'll have to use the same guidelines in all products (we'll have to discuss this further).

Issue #2027, opened 2016-09-01
Create tns update command


sitefinitysteve:
Right now to do an upgrade we need to visit the docs to remember all the commands :)

Would be fantastic to just roll everything up into 
"tns update" and it'll do the platform remove\add, and update core modules all with one command

firescript:
+1

dtopuzov:
This command should actually do those steps:
````
rm -rf hooks
rm -rf platforms
rm -rf node_modules
tns platform remove (for all installed platforms)
tns platform add (for all installed platforms)
tns plugin remove tns-core-modules
tns plugin add tns-core-modules
```
Not sure what to do with 3rd party plugins, should we also update to latest version?

SvetoslavTsenov:
Hey @tzraikov,
What about to add a check also for the package.json and if the version is lower than the released one to update project no matter if the platform is installed. So far as it is working fine but only if the platforms are already installed.

PR #2186, opened 2016-11-01
`tns update` should check for platforms installed in package.json but missing in platforms folder

tzraikov:
Additional fix for #2027


SvetoslavTsenov:
👍

PR #2194, opened 2016-11-04
Keep track of prepared plugins and avoid unnecessary cleans.

hdeshev:
The new dependency analysis algorithm always passes all plugins and we need to keep track of prepared ones, so that we don't clean and prepare everything unnecessary (most painful on Android).

Related to #2110, #2152

Ping @Plamen5kov

### DIFF
--- a/test/plugin-prepare.ts
+++ b/test/plugin-prepare.ts
@@ -1,0 +1,66 @@
+import {assert} from "chai";
+import {NpmPluginPrepare} from "../lib/tools/node-modules/node-modules-dest-copy";
+
+require("should");
+
+class TestNpmPluginPrepare extends NpmPluginPrepare {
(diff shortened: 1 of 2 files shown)


test/plugin-prepare.ts
```diff
@@ -0,0 +1,79 @@
...
+				name: "tns-core-modules-widgets",
+				version: "1.0.0",
+				nativescript: null,
+			}
+		};
+		pluginPrepare.preparePlugins(testDependencies, "android");
+		assert.deepEqual({}, pluginPrepare.preparedDependencies);
+	});
+
+	it("skips prepare if every plugin prepared", () => {
```
Plamen5kov: this test is same as the one above

Plamen5kov:
:+1:  after deleting unnecessary test

PR #2195, opened 2016-11-04
Remove use of lib/iOS folder for native plugins

jasssonpet:
They are added from their original `node_modules` folder instead.

Closes https://github.com/NativeScript/nativescript-cli/issues/2227.


NathanaelA:
Any chance this is going to show up in 2.4?

PR #2201, opened 2016-11-07
future proof {N} CLI for the next 10 years

blagoev:


blagoev:
This less risky than trying to refactor all the places this is used currently

blagoev:
In all its seriousness we intent to remove this version check after the 2.4 release

blagoev:
It turns out this is problematic so will close this in flavor of this one https://github.com/NativeScript/nativescript-cli/pull/2202

PR #2209, opened 2016-11-09
remove npm2 dependency

Plamen5kov:
## What does this PR do?
PR removes the local npm2 dependency from CLI. Fix for: https://github.com/NativeScript/nativescript-cli/issues/1845

_Description_
Cache functionality is removed from `lib/node-package-manager.ts` and `lib/npm-installation-manager`. Other changes are made along the way to accommodate the lack of the cache add cache unpack functionality.

### How did it work till now?
* On npm install template: template is saved to `tmp` folder in some cases and to cache in others, and then are extracted to app folder. If Template is not default, the default template was downloaded because it's the only one that has `App_Resources` folder. The `App_Resources` folder is copied to new template.
* On npm install core-framework (tns-ios, tns-android), tns-core-framework is saved to cache and extracted to `platforms/` folder
* On npm install `node-inspector`, package is saved and extracted to cache.
* On tns platform update we always updated iOS 

### How will it work after this PR
* On npm install, all packages are treated as normal node packages. That means all of them are installed in the node modules and in the case of the templates and the core-frameworks the packages are uninstalled after they are extracted to `app/` and to `platforms/` folders respectively. All templates will be expected to have an `App_Resources` folder with them.
* We check the current iOS proj file and the new one and if they are different, there is a failure on update.
* Currently the ios-node-inspector that is being installed through npm is in node_modules folder for each project. We might think of a way to make an exception for the inspector because of it's size.

> The implementation works with npm cache internally.

### Expected
Because not all templates have an `App_Resource` folder yet, there are tests that are expected to fail with: `Project's app/App_Resources must contain Android and iOS directories.`

### Related PR
These PR add `App_Resource` folder to templates that don't have it. From now on all templates will have to carry with them an `App_Resource` folder.
https://github.com/NativeScript/template-hello-world-ts/pull/21
https://github.com/NativeScript/template-hello-world-ng/pull/34

ping: @NativeScript/android-runtime @rosen-vladimirov @hshristov @tzraikov @hdeshev @KristinaKoeva

### DIFF
--- a/lib/definitions/project.d.ts
+++ b/lib/definitions/project.d.ts
@@ -27,11 +27,6 @@ interface IProjectDataService {
  */
 interface IProjectTemplatesService {
 	/**
-	 * Defines the path where unpacked default template can be found.
-	 */
-	defaultTemplatePath: IFuture<string>;
-
-	/**
 	 * Prepares template for project creation.
 	 * In case templateName is not provided, use defaultTemplatePath.
 	 * In case templateName is a special word, validated from us (for ex. typescript), resolve the real template name and add it to npm cache.
@@ -39,7 +34,7 @@ interface IProjectTemplatesService {
 	 * @param {string} templateName The name of the template.
 	 * @return {string} Path to the directory where extracted template can be found.
 	 */
-	prepareTemplate(templateName: string): IFuture<string>;
+	prepareTemplate(templateName: string, projectDir: string): IFuture<string>;
 }
 
 interface IPlatformProjectServiceBase {
@@ -77,12 +72,7 @@ interface IPlatformProjectService {
 	prepareProject(): IFuture<void>;
 	prepareAppResources(appResourcesDirectoryPath: string): IFuture<void>;
 	isPlatformPrepared(projectRoot: string): IFuture<boolean>;
-	canUpdatePlatform(currentVersion: string, newVersion: string): IFuture<boolean>;
-	/**
-	* Provides a platform specific update logic for the specified runtime versions.
-	* @return true in cases when the update procedure should continue.
-	*/
-	updatePlatform(currentVersion: string, newVersion: string, canUpdate: boolean, addPlatform?: Function, removePlatform?: (platforms: string[]) => IFuture<void>): IFuture<boolean>;
+	canUpdatePlatform(newInstalledModuleDir: string): IFuture<boolean>;
 	preparePluginNativeCode(pluginData: IPluginData, options?: any): IFuture<void>;
 	removePluginNativeCode(pluginData: IPluginData): IFuture<void>;
 	afterPrepareAllPlugins(): IFuture<void>;
--- a/lib/node-package-manager.ts
+++ b/lib/node-package-manager.ts
@@ -1,5 +1,4 @@
-import Future = require("fibers/future");
-import * as npm from "npm";
+import * as path from "path";
 
 interface INpmOpts {
 	config?: any;
@@ -8,34 +7,12 @@ interface INpmOpts {
 }
 
 export class NodePackageManager implements INodePackageManager {
-	constructor(private $childProcess: IChildProcess,
+	constructor(private $fs: IFileSystem,
+		private $hostInfo: IHostInfo,
+		private $errors: IErrors,
+		private $childProcess: IChildProcess,
 		private $logger: ILogger,
 		private $options: IOptions) { }
-
-	public getCache(): string {
-		return npm.cache;
-	}
-
-	public load(config?: any): IFuture<void> {
-		if (npm.config.loaded) {
-			let data = npm.config.sources.cli.data;
-			Object.keys(data).forEach(k => delete data[k]);
-			if (config) {
-				_.assign(data, config);
-			}
-			return Future.fromResult();
-		} else {
-			let future = new Future<void>();
-			npm.load(config, (err: Error) => {
-				if (err) {
-					future.throw(err);
-				} else {
-					future.return();
-				}
-			});
-			return future;
-		}
-	}
 
 	public install(packageName: string, pathToSave: string, config?: any): IFuture<any> {
 		return (() => {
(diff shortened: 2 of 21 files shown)


lib/node-package-manager.ts
```diff
@@ -8,139 +7,99 @@ interface INpmOpts {
...
-					//      'ng2-translate@2.5.0': '~2.0.0' } }
-					this.$logger.warn(err.message);
-					this.$logger.trace("Required peerDependencies are: ", err.peersDepending);
-				} else {
-					// All other errors should be handled by the caller code.
-					throw err;
-				}
+			return diff;
+		} catch (err) {
+			if (err.code === "EPEERINVALID") {
```
hdeshev: Since we're not using npm as a library anymore, will we actually get an exception with an EPEERINVALID `code` from the child process? Seems like we should either delete this code or handle invalid peer dependencies in another way.

lib/node-package-manager.ts
```diff
@@ -8,139 +7,99 @@ interface INpmOpts {
...
-				return;
-			}
-			if (this.$options.ignoreScripts) {
-				config = config || {};
-				config["ignore-scripts"] = true;
+		try {
+			let subDirsBefore:Array<string> = this.$fs.exists(path.join(pathToSave, "node_modules")).wait() ? this.$fs.readDirectory(path.join(pathToSave, "node_modules")).wait() : [];
+
+			let flags = this.getFlagsString(config);
+			this.$childProcess.exec(`npm install ${packageName} ${flags}`, { cwd: pathToSave }).wait();
```
hdeshev: This `exec` invocation looks like it's eating the child process stdout. I think we should inherit stdio here so that npm-related errors are visible to the user. Hmm, yes, we need stdin too -- I've heard rumors on the internet there are packages that block the console asking for email addresses on installation.

lib/services/ios-project-service.ts
```diff
@@ -436,48 +436,53 @@ export class IOSProjectService extends projectServiceBaseLib.PlatformProjectServ
...
+			if(!contentIsTheSame) {
+				this.$logger.warn(`The content of the current project file: ${currentXcodeProjectFile} and the new project file: ${newXcodeProjectFile} is different.`);
 			}
+			return contentIsTheSame;
 
-			return true;
 		}).future<boolean>()();
 	}
 
+	// Is this used?
```
hdeshev: I suggest we kill this commented out code block..

test/project-templates-service.ts
```diff
@@ -64,65 +66,76 @@ describe("project-templates-service", () => {
...
-			});
-
-			it("when after npm install the temp directory has only one subdir and it is node_modules", () =>{
-				testInjector = createTestInjector({shouldNpmInstallThrow: false, npmInstallationDirContents: ["node_modules"], npmInstallationDirNodeModulesContents: [expectedTemplatePath]});
-				projectTemplatesService = testInjector.resolve("projectTemplatesService");
-				let actualPathToTemplate = projectTemplatesService.prepareTemplate("validName").wait();
-				assert.strictEqual(path.basename(actualPathToTemplate), expectedTemplatePath);
-				assert.strictEqual(isDeleteDirectoryCalledForNodeModulesDir, true, "When correct path is returned, template's node_modules directory should be deleted.");
-			});
+			//there is no tmp directory anymore
```
hdeshev: Delete commented-out code.

hdeshev:
Squash those commits, please.

lib/node-package-manager.ts
Plamen5kov: You're right I hadn't thought about that, will make adjustments

lib/node-package-manager.ts
Plamen5kov: inside childProcess.exec we do a throw which is caught by the catch, so we don't consume the error: https://github.com/telerik/mobile-cli-lib/blob/16a4c64e6f1ac2c58d21d92a5da11bc73854451c/child-process.ts#L14

Plamen5kov: In the new version i use $childProcess.spawnFromEvent(...) so we can handle the stdout and stderr to keep previous behavior.

lib/node-package-manager.ts
Plamen5kov: With new implementation we handle the EPEERINVALID error same as before.

lib/services/plugins-service.ts
```diff
@@ -36,9 +36,24 @@ export class PluginsService implements IPluginsService {
...
 			this.ensure().wait();
-			let dependencyData = this.$npm.cache(plugin, undefined, PluginsService.NPM_CONFIG).wait();
-			if (dependencyData.nativescript) {
-				let pluginData = this.convertToPluginData(dependencyData);
+			let installedModuleNames = this.$npm.install(plugin, this.$projectData.projectDir, PluginsService.NPM_CONFIG);
+
+			//npmInstallResultStream is not very usefull (doesn't contain all the package information), so we need to find the real one inside node_modules
+			let foundValidPlugin = false;
+			let realNpmPackageJson:any;
+			for(let index in installedModuleNames) {
```
rosen-vladimirov: you can use `of` instead of `in` which will give you the name directly:
```TypeScript
for(let name of installedModuleNames) {
...
```
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of

lib/services/plugins-service.ts
```diff
@@ -36,9 +36,24 @@ export class PluginsService implements IPluginsService {
...
+				let name = installedModuleNames[index];
+				let pathToRealNpmPackageJson = `${this.$projectData.projectDir}/node_modules/${name}/package.json`;
+				realNpmPackageJson = require(pathToRealNpmPackageJson);
```
rosen-vladimirov: it's safer to use `this.$fs.readJson(pathToRealNpmPackageJson).wait()` as require caches results and also it does not work in case the file has BOM header

lib/definitions/project.d.ts
```diff
@@ -77,12 +72,12 @@ interface IPlatformProjectService {
...
 	prepareAppResources(appResourcesDirectoryPath: string): IFuture<void>;
 	isPlatformPrepared(projectRoot: string): IFuture<boolean>;
-	canUpdatePlatform(currentVersion: string, newVersion: string): IFuture<boolean>;
+	canUpdatePlatform(newInstalledModuleDir: string): IFuture<boolean>;
 	/**
 	* Provides a platform specific update logic for the specified runtime versions.
 	* @return true in cases when the update procedure should continue.
 	*/
-	updatePlatform(currentVersion: string, newVersion: string, canUpdate: boolean, addPlatform?: Function, removePlatform?: (platforms: string[]) => IFuture<void>): IFuture<boolean>;
+	// updatePlatform(currentVersion: string, newVersion: string, canUpdate: boolean, addPlatform?: Function, removePlatform?: (platforms: string[]) => IFuture<void>): IFuture<boolean>;
```
rosen-vladimirov: In case the method is commented, can you please delete it. Also the implementation should be removed.

lib/node-package-manager.ts
```diff
@@ -8,139 +7,109 @@ interface INpmOpts {
...
+		try {
+			let subDirsBefore:Array<string> = this.$fs.exists(path.join(pathToSave, "node_modules")).wait() ? this.$fs.readDirectory(path.join(pathToSave, "node_modules")).wait() : [];
 
-	public install(packageName: string, pathToSave: string, config?: any): IFuture<any> {
-		return (() => {
-			if (this.$options.disableNpmInstall) {
-				return;
+			let flags = this.getFlagsString(config, true);
+			let params = ["install", packageName];
+			for(let index in flags) {
```
Plamen5kov: tried, didn't work :D

lib/node-package-manager.ts
```diff
@@ -8,139 +7,109 @@ interface INpmOpts {
...
-				}
+			let subDirsAfter = this.$fs.readDirectory(path.join(pathToSave, "node_modules")).wait();
+
+			/** This diff is done in case the installed pakcage is a URL address, a path to local directory or a .tgz file
+			 *  in these cases we don't have the package name and we can't rely on "npm install --json"" option
+			 *  to get the project name because we have to parse the output from the stdout and we have no controll over it (so other messages may be mangled in stdout)
+			 * 	The solution is to traverse node_modules before and after install and get the name of the installed package,
+			 * 	even if it's installed through local path or URL. If command installes more than one package, all package names are returned.
+			 */
+			let diff = subDirsAfter.filter(x => subDirsBefore.indexOf(x) === -1);
```
Plamen5kov: @rosen-vladimirov, yes this behavior is  expected

lib/npm-installation-manager.ts
```diff
@@ -23,103 +13,51 @@ export class NpmInstallationManager implements INpmInstallationManager {
...
 
 			this.$lockfile.lock().wait();
 
 			try {
-				let packageToInstall = packageName;
-				let pathToSave = (opts && opts.pathToSave) || npm.cache;
+				let packageToInstall = this.$options.frameworkPath || packageName;
+				let pathToSave = projectDir;
 				let version = (opts && opts.version) || null;
+				let dependencyType = (opts && opts.dependencyType) || null;
```
Plamen5kov: dependency type is the npm flag that you want to use on install like --save --save-dev ...

lib/node-package-manager.ts
```diff
@@ -48,9 +24,41 @@ export class NodePackageManager implements INodePackageManager {
...
+					}
+				}
+
+				/** This diff is done in case the installed pakcage is a URL address, a path to local directory or a .tgz file
+				 *  in these cases we don't have the package name and we can't rely on "npm install --json"" option
+				 *  to get the project name because we have to parse the output from the stdout and we have no controll over it (so other messages may be mangled in stdout)
+				 * 	The solution is to traverse node_modules before and after install and get the name of the installed package,
+				 * 	even if it's installed through local path or URL. If command installes more than one package, all package names are returned.
+				 */
+				let diff = dependenciesAfter.filter(x => dependenciesBefore.indexOf(x) === -1);
```
rosen-vladimirov: I've found a case where this approach will not work. Consider the case where a plugin is already installed in package.json, but user wants to update it's version (or install it from .tgz this time). In this case the difference will be empty array and the plugin will not be installed. I would suggest to compare both the name of the plugin and the version. For example this code could do the job:
```TypeScript
let diff = _(jsonContentAfter.dependencies)
    .omitBy((val: string, key: string) => jsonContentBefore && jsonContentBefore.dependencies[key] && jsonContentBefore.dependencies[key] === val)
    .keys()
    .value();
```

lib/npm-installation-manager.ts
```diff
@@ -130,99 +68,59 @@ export class NpmInstallationManager implements INpmInstallationManager {
...
-		return ((): boolean => {
-			let shasumProperty = "dist.shasum";
-			let cachedPackagePath = this.getCachedPackagePath(packageName, version);
-			let packageInfo = this.$npm.view(`${packageName}@${version}`, shasumProperty).wait();
 
-			if (_.isEmpty(packageInfo)) {
-				// this package has not been published to npmjs.org yet - perhaps manually added via --framework-path
-				this.$logger.trace(`Checking shasum of package ${packageName}@${version}: skipped because the package was not found in npmjs.org`);
-				return true;
+			if(version === constants.PackageVersion.NEXT) {
```
rosen-vladimirov: why do you need special code for `next` version. In case we decide to use separate tag (for example `nextNext`) we shouldn't add it here. Npm supports installing of package directly by tag, for example `npm install nativescript@next` will work as expected. Is there a particular reason to have special methods for `next` and `latest` tags?

lib/npm-installation-manager.ts
```diff
@@ -130,99 +68,59 @@ export class NpmInstallationManager implements INpmInstallationManager {
...
+				version = this.getNextVersion(packageName).wait();
 			}
+			let installedModuleNames = this.npmInstall(packageName, pathToSave, version, dependencyType).wait();
+			let installedPackageName =  installedModuleNames[0];
```
rosen-vladimirov: Why `[0]` - is it possible to return more than one packages? In case yes - why do we get only the first one. In case not - why does the method return array?

lib/services/platform-service.ts
```diff
@@ -692,102 +689,43 @@ export class PlatformService implements IPlatformService {
...
+					this.updatePlatformCore(platformData, currentVersion, newVersion, canUpdate).wait();
 				} else if (semver.eq(currentVersion, newVersion)) {
 					this.$errors.fail("Current and new version are the same.");
 				} else {
-					this.updatePlatformCore(platformData, currentVersion, newVersion, canUpdate).wait();
+					this.$errors.fail(`Your current version: ${currentVersion} is higher than the one you're trying to install ${newVersion}.`);
 				}
 			} else {
-				this.updatePlatformCore(platformData, currentVersion, newVersion, canUpdate).wait();
+				this.$errors.failWithoutHelp("Native Platfrom cannot be updated.");
```
rosen-vladimirov: Platfrom -> Platform

lib/node-package-manager.ts
```diff
@@ -48,30 +24,54 @@ export class NodePackageManager implements INodePackageManager {
...
+					}
+				}
+
+				let flags = this.getFlagsString(config, true);
+				let params = ["install"];
+				if(packageName !== pathToSave) {
+					params.push(packageName); //because npm install ${pwd} on mac tries to install itself as a dependency (windows and linux have no such issues)
+				}
+				params = params.concat(flags);
+				this.$childProcess.spawnFromEvent("npm", params, "close", { cwd: pathToSave }).wait();
```
rosen-vladimirov: How does this code work on Windows? On Mac and Linux the executable name is `npm`, but on Windows it's `npm.cmd`

Plamen5kov: As I'm not calling a direct path to executable, I expect npm to be found in path.

rosen-vladimirov: There's no `npm` executable in PATH on Windows. There's `npm.cmd`

rosen-vladimirov: To be more precise in the explanation:
 - on Windows, the default command prompt is able to resolve npm to `npm.cmd` when you execute `npm install`
 - childProcess.spawn method does not launch new shell, it just starts executable in PATH (or passed with full path)

In the current case, childProcess is trying to start a new process called `npm` but it cannot be resolved as the real name of the executable on Windows is `npm.cmd`.
We have such code, you can take a look at it [here](https://github.com/telerik/mobile-cli-lib/blob/79633174cb36302a0ec4c1eecc382ff45f7898ed/appbuilder/services/npm-service.ts#L51-L61)

lib/node-package-manager.ts
```diff
@@ -48,30 +24,54 @@ export class NodePackageManager implements INodePackageManager {
...
+					this.$errors.failWithoutHelp(`The plugin ${packageName} is already installed`);
+				}
+				if(diff.length <= 0 && dependenciesBefore.length !== dependenciesAfter.length) {
+					this.$errors.failWithoutHelp(`Couldn't install package correctly`);
+				}
+
+				return diff;
 			} catch (err) {
-				if (err.code === "EPEERINVALID") {
+				if (err.message && err.message.indexOf("EPEERINVALID") !== -1) {
```
Plamen5kov: It's not, I'll fix it

lib/npm-installation-manager.ts
Plamen5kov: I've missed removing, this if, i was testing something out, and I've seem to forgot to remove it. The idea behind the next and last tags is readability only.

lib/npm-installation-manager.ts
Plamen5kov: On all the places we use node-package-manager, we pass a single package to be installed, and expect a single package to be installed. My thinking was for a later refactoring, where we could be able to use the node-package-manager everywhere. That means it could install one or multiple packages, hence saving the array result.

lib/node-package-manager.ts
Plamen5kov: I understand, I'll fix it, thank you.

Plamen5kov:
run CI

Plamen5kov:
@hdeshev @rosen-vladimirov, 5 tests concerning `App_Resources` folder are commented out so CI runs ok, but if you have any last notes for the PR, I would be happy to resolve them.

PR #2215, opened 2016-11-11
Some refactorings. socket-proxy-factory is a Factory so move out the stop logic out of it

KristinaKoeva:


KristinaKoeva:
run ci

PR #2229, opened 2016-11-15
Support Cocoapods 1.0.0 in NativeScript 1.7.1

rosen-vladimirov:
NativeScript 1.7.1 can work only with Cocoapods 0.39.0 or 0.38.2. However these versions are not longer usable and trying to execute `pod install` leads to error. This is described [here](http://blog.cocoapods.org/Sharding/).

Since NativeScript 2.0.0, CLI supports Cocoapods 1.0.0 or later. The changes here are a mix of around six older commits in NS CLI which were adding support for Cocoapods 1.0.0. I've extracted the minimum required changes, so we can use allow building of NativeScript 1.7.1 projects in AppBuilder.

The summary of the changes is:
Trying to run `pod install` with CocoaPods 1.0.0 issues an error in case the Podfile doesn't reference a target. Add the project target to the Podfile.
Upon specifying target to the Podfile, however, the path to the Pods' .xcconfig file changes - reflect that change in code.
In addition, truncate and write the Podfile upon each prepare in order to avoid duplication and errors.


rosen-vladimirov:
run ci

Issue #282, opened 2015-03-06
"tns platform add android --path TNSApp --symlink" does not work on Windows


dtopuzov:
Run:
node_modules.bin\tns platform add android --path TNS_App --symlink

Result:
EPERM, symlink 'C:\Users\topuzov\AppData\Roaming\npm-cache\tns-android\0.9.0\package\framework\assets\bindings'

Expected:
If tns supports symlink on Windows OS it should work (it works on OSX)
If tns does not support symlink on Windows OS then appropriate error message should be displayed

dtopuzov:
--symlink with --frameworkPath also fails:
```Shell
C:\Git\nativescript-cli-tests>node_modules\.bin\tns create TNS_App
Project TNS_App was successfully created

C:\Git\nativescript-cli-tests>node_modules\.bin\tns platform add android --frameworkPath c:\Git\nativescript-cli-tests\tns-android\package\package --path TNS_App --symlink
Copying template files...
cp: no such file or directory: c:\Git\nativescript-cli-tests\tns-android\package\package\framework\res

cp: no such file or directory: c:\Git\nativescript-cli-tests\tns-android\package\package\framework\.project

cp: no such file or directory: c:\Git\nativescript-cli-tests\tns-android\package\package\framework\.project
cp: no such file or directory: c:\Git\nativescript-cli-tests\tns-android\package\package\framework\AndroidManifest.xml

cp: no such file or directory: c:\Git\nativescript-cli-tests\tns-android\package\package\framework\.project
cp: no such file or directory: c:\Git\nativescript-cli-tests\tns-android\package\package\framework\AndroidManifest.xml
cp: no such file or directory: c:\Git\nativescript-cli-tests\tns-android\package\package\framework\project.properties

ENOENT, readdir 'c:\Git\nativescript-cli-tests\tns-android\package\package\framework\assets'

dtopuzov:
Still fails
c:\Git\temp>node_modules\.bin\tns platform add android --path TNS_App --symlink
Copying template files...
EPERM, symlink 'c:\Git\temp\TNS_App\platforms\android\build-tools\android-static-binding-generator'
# platform add

agupta1989:
+1 Same here with tns --version 1.0.1

tsonevn:
tns plaform add will not work with symlink as of 2.5

PR #2234, opened 2016-11-17
added logging on install

Plamen5kov:
fixing python tests

### DIFF
--- a/lib/node-package-manager.ts
+++ b/lib/node-package-manager.ts
@@ -25,7 +25,6 @@ export class NodePackageManager implements INodePackageManager {
 			}
 
 			let jsonContentBefore = this.$fs.readJson(path.join(pathToSave, "package.json")).wait();
-			// let dependenciesBefore: Array<string> = [];
 			let dependenciesBefore = _.keys(jsonContentBefore.dependencies).concat(_.keys(jsonContentBefore.devDependencies));
 
 			let flags = this.getFlagsString(config, true);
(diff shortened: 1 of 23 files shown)


Plamen5kov:
run CI

Plamen5kov:
run CI

lib/node-package-manager.ts
```diff
@@ -92,7 +92,13 @@ export class NodePackageManager implements INodePackageManager {
 	public view(packageName: string, config: any): IFuture<any> {
 		return (() => {
 			let flags = this.getFlagsString(config, false);
-			let viewResult = this.$childProcess.exec(`npm view ${packageName} ${flags}`).wait();
+			let viewResult: any;
+			try {
+				viewResult = this.$childProcess.exec(`npm view ${packageName} ${flags}`).wait();
+			} catch(e) {
+				this.$logger.out(e);
```
rosen-vladimirov: What's the reason to catch the error, print it and throw it again? In case `npm view` is used in a command, CLI's commandsService will print the error when it's thrown.

lib/npm-installation-manager.ts
```diff
@@ -70,6 +70,14 @@ export class NpmInstallationManager implements INpmInstallationManager {
 
 	private installCore(packageName: string, pathToSave: string, version: string, dependencyType: string): IFuture<string> {
 		return (() => {
+			if(packageName.indexOf(".") === 0 || packageName.indexOf(".tgz") >= 0) {
```
rosen-vladimirov: this will fail in case user passes path without `./` in front, for example in case I want to install local directory I can pass `./my-dir` or `my-dir`. Both will work

lib/npm-installation-manager.ts
```diff
@@ -70,6 +70,14 @@ export class NpmInstallationManager implements INpmInstallationManager {
...
+				let cwd = process.cwd();
+				let possiblePackageName = path.join(`${cwd}`, packageName);
```
rosen-vladimirov: instead of this you can use `path.resolve(packageName)` - this will work with both full and relative paths. In this case you can skip the above if as well and use:
```TypeScript
const possiblePackageName= path.resolve(packageName);
if (this.$fs.exists(possiblePackageName).wait()) {
...
}

lib/node-package-manager.ts
Plamen5kov: this is temporary, it will failWithoutHelp

Plamen5kov:
run CI

pkoleva:
Related issue https://github.com/NativeScript/nativescript-cli/issues/282

pkoleva:
run CI

Plamen5kov:
run CI

Plamen5kov:
run CI

PR #2232, opened 2016-11-16
Updating macOS setup script with latest macOS/package requirements

jasssonpet:
Fix several issues reported in https://github.com/NativeScript/nativescript-cli/issues/2139:
* Execute brew commands without sudo - https://github.com/NativeScript/nativescript-cli/issues/2228.
* Remove Java home folder shell evaluation introduced in https://github.com/NativeScript/nativescript-cli/commit/f80953dfcac0093bc70186bb6809a01d6bfdba98#diff-31337fd644449cf05dae54e2c2a41294R106.
* Workaround xcodeproj requiring Ruby 2.2 (macOS default is 2.0) - https://github.com/CocoaPods/Xcodeproj/pull/393.
* Write environment variables to `.zprofile` in addition to `.profile`.

Tested on macOS Sierra and El Capitan VMs.

PR #2236, opened 2016-11-17
Homebrew complains loudly when run as root. 

jeckhart:
This updates the Mac setup script to run homebrew as the native user instead of as root.


jeckhart:
Note: When I looked I didn't see another PR, but it looks like https://github.com/NativeScript/nativescript-cli/pull/2232 now covers this one as well.

Closing this one in favor of #2232 .

Issue #2204, opened 2016-11-08
Remove fibers from CLI and use async/await


rosen-vladimirov:
CLI uses [fibers](https://www.npmjs.com/package/fibers) in order to use sync code instead of callbacks. Fibers are great, but most of the people are not familiar with them. Also Promises are more popular now and they are integrated in Node.js. 
TypeScript allows using async/await for getting Promises results instead of chaining `.then`.
Here are the steps to be taken when fibers are removed and async/await is used instead:
- [x] Drop support for Node.js 0.12 (Deprecated in 2.4) PRs: https://github.com/telerik/mobile-cli-lib/pull/851 , https://github.com/NativeScript/nativescript-cli/pull/2320
- [x] Transpile to ES6 instead of ES5. TypeScript's async/await support is available when transpiling to ES6. PRs: https://github.com/telerik/mobile-cli-lib/pull/843 , https://github.com/NativeScript/nativescript-cli/pull/2323
- [x] Use sync API for Node core modules, like `fs` and `child_process`. Currently CLI uses async versions of the methods and wraps them with futures. In case they are replaced with Promises, each method that uses such method will have to be marked as async. However in case we use the sync versions of the methods (for example use `fs.existsSync` instead of `fs.exists` in CLI's [file-system.ts](https://github.com/telerik/mobile-cli-lib/blob/master/file-system.ts)). This way some of methods, which should be marked as `async`, will become absolutely synchronous and we'll not have to `await` them. PRs: https://github.com/telerik/mobile-cli-lib/pull/853, https://github.com/NativeScript/nativescript-cli/pull/2347 
- [ ] Fix hooks which use $fs, $childProcess, etc. as the code of the methods is changed.
- [x] Replace fibers with async/await in [mobile-cli-lib](https://github.com/telerik/mobile-cli-lib): PR: https://github.com/telerik/mobile-cli-lib/pull/865 , 
- [x] Replace fibers with async/await in nativescript-cli: https://github.com/NativeScript/nativescript-cli/pull/2411
- [x] Replace fibers with async/await in ios-sim-portable: https://github.com/telerik/ios-sim-portable/pull/83
- [ ] Test with Node.js 4
- [ ] Test with Node.js 6
- [ ] Test with Node.js 7

PR #2290, opened 2016-12-01
Transpile to ES6

rosen-vladimirov:
> NOTE: Merge only after https://github.com/telerik/mobile-cli-lib/pull/843 is merged

Add support for ES6 transpilation. Change dependency injection in order to match the `class` and `constructor` keywords.
Update istanbul to latest version as its previous versions do not support ES6.

Remove bluebird.d.ts as it conflicts with Promises from TypeScript's ES6 d.ts.
Remove bluebird as dependency - we will use native Promises from now on.

Add support for Node.js 4 and 5 - both versions have limited support for ES6 features, so in order to support them we have to:
 - pass --harmony flag to Node.js
 - do not use some features like default values of method arguments.
 - do not use spread operator - it's implementation in Node.js 4 is limited only to arrays. Even this implementation has issues. So use `.apply` instead of using spread operator. Remove all other usages of spread operator (for objects).

Add tests scripts for istanbul and mocha, as they both will need the `--harmony` flag as well, so we cannot execute them directly with Node.js 4 and 5. Instead we'll execute a custom script, that will start a new node process.

Introduce `test-scripts` directory with JavaScript executables to start the tests. Set them in package.json. They are required as when Node 4 or Node 5 is used, the tests will need `--harmony` flag as well.
Exclude `test-scripts` directory from npm package (add it to `.npmignore`) - users do not need these files.

Exclude `.vscode/launch.json` from `.gitignore` and add predefined launch configurations. They can be used for debugging CLI process and unit tests.

nativescript_launch_configs_2

Part of https://github.com/NativeScript/nativescript-cli/issues/2204

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,9 +29,12 @@ results
 scratch/
 .idea/
 .settings/
-.vscode/
+.vscode/**
+!.vscode/launch.json
 test-reports.xml
 
 npm-debug.log
 node_modules
 docs/html
+
+!test-scripts/*.js
--- a/.npmignore
+++ b/.npmignore
@@ -21,6 +21,7 @@ lib/**/*.ts
 lib/**/*.js.map
 
 test/
+test-scripts/
 .vscode
 lib/common/test/
 coverage/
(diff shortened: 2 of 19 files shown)


dtopuzov:
run ci

PR #2295, opened 2016-12-01
Fixed: CLI crashes/stops responding after ~80 console.logs

### LINKED ISSUE #2266: [ios][2.4.0 & @next] CLI crashes/stops responding after ~80 console.logs
### iOS Only
Since upgrading the cli to `2.4.0` I've noticed that after ~80 `console.log`'s the cli will hang and become unresponsive.

### Steps to reproduce:
```
let i: number, len: number = 1000
for (i = 0; i < len; i++) {
	console.log('i hope u dont crash :X', i)
}
```
It usually gets up to ~80 but never more than 100. The cli freezes, livesync no longer works, and strangely the app I'll be working on freezes for about 5 seconds.

I tried installing `nativescript@next` to see if this was resolved but it was not.

### Temporary Fix
Unable to reproduce when reverting back to `2.3.0`.

tzraikov:
Fixes #2266
> *Merge after* https://github.com/telerik/mobile-cli-lib/pull/847

### DIFF
--- a/lib/services/ios-log-filter.ts
+++ b/lib/services/ios-log-filter.ts
@@ -1,13 +1,14 @@
 let sourcemap = require("source-map");
 import * as path from "path";
+import fs = require("fs");
 
 export class IOSLogFilter implements Mobile.IPlatformLogFilter {
 
 	private $projectData: IProjectData;
+	private partialLine: string = null;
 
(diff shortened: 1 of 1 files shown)


tzraikov:
run ci

PR #2305, opened 2016-12-02
Fixed: TNS run command broken with using --watch

### LINKED ISSUE #2150: TNS run command broken with using --watch
According to the docs (and my experience) with the `tns run ios` command -- it rebuilds the app, deploys it and then runs it.   

However, if you add the cool magical `--watch` command the initial build/deploy is skipped and it acts exactly as if you did a `tns livesync ios --watch`.    This is completely different behavior that I don't expect.     When I use the `tns run ios --emulator --watch` I expect it do a full build the app, deploy it, run it and then livesync any changed files after that point.  

Right now because it is broken; you have to a `tns run ios --emulator` ctrl-c after it deploys it, and then start the `--watch` command.

TNS v2.30 - issue is on ios for sure.  No idea on Android as I still use my LiveEdit and never use the --watch command on android.

tzraikov:
Fixes #2150

### DIFF
--- a/bin/nativescript.js
+++ b/bin/nativescript.js
@@ -1,0 +1,4 @@
+#!/usr/bin/env node
+
+"use strict";
+require("../lib/nativescript-cli.js");
--- a/lib/commands/appstore-upload.ts
+++ b/lib/commands/appstore-upload.ts
@@ -65,7 +65,7 @@ export class PublishIOS implements ICommand {
 					};
 					this.$logger.info("Building .ipa with the selected mobile provision and/or certificate.");
 					// This is not very correct as if we build multiple targets we will try to sign all of them using the signing identity here.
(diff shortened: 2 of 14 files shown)


KristinaKoeva:
It seems to me that passing boolean parameter to the runPlatform, deployOnDevice and deployOnEmulator overcomplicates the interface. I would suggest separating the methods, so that you can call something like the following in your command

```
public executeCore(args: string[], buildConfig?: IBuildConfig): IFuture<void> {
                this.$platformService.deploy(args[0], buildConfig).wait();
   		if (this.$options.watch) {
  			return this.$usbLiveSyncService.liveSync(args[0]);
  		}

                return this.$platformService.startApplication(args[0], buildConfig).wait();
 	}
```
Where the deploy method would be responsible only for deploying an application on device/simulator and the startApplication would be responsible only for running an application on device/simulator

Issue #1122, opened 2015-10-30
semver.browser.js.gz file already in archive in sample-Groceries app


kyrisu:
My problem is similar to issue #601 but based on the notes there it should be fixed in version 1.1.2

I'm on version 1.4.3
```
$ tns --version
1.4.3
```

Unfortunately when I try to run sample-Groceries app from the tutorial I have the following error:

```
$ tns run android
Project successfully prepared
Picked up _JAVA_OPTIONS: -Dawt.useSystemAAFontSettings=on -Dswing.aatext=true -Dswing.defaultlaf=com.sun.java.swing.plaf.gtk.GTKLookAndFeel
execute: copyAarDependencies, addAarDependencies before configuration
:preBuild UP-TO-DATE
:preDebugBuild UP-TO-DATE
:checkDebugManifest
:preReleaseBuild UP-TO-DATE
:prepareChart201531016DevReleaseLibrary UP-TO-DATE
:prepareComAndroidSupportAppcompatV72221Library UP-TO-DATE
:prepareComAndroidSupportSupportV42221Library UP-TO-DATE
:prepareCommon201531016DevReleaseLibrary UP-TO-DATE
:prepareData201531016DevReleaseLibrary UP-TO-DATE
:prepareList201531016DevReleaseLibrary UP-TO-DATE
:preparePrimitives201531016DevReleaseLibrary UP-TO-DATE
:prepareDebugDependencies
:compileDebugAidl UP-TO-DATE
:compileDebugRenderscript UP-TO-DATE
:generateDebugBuildConfig UP-TO-DATE
:deleteJavaDir UP-TO-DATE
:cleanLocalAarFiles
:collectAllJars UP-TO-DATE
:ensureMetadataOutDir UP-TO-DATE
:buildMetadata UP-TO-DATE
:copyMetadata UP-TO-DATE
:generateDebugAssets UP-TO-DATE
:mergeDebugAssets UP-TO-DATE
:generateDebugResValues UP-TO-DATE
:generateDebugResources UP-TO-DATE
:mergeDebugResources UP-TO-DATE
:processDebugManifest UP-TO-DATE
:processDebugResourcesUnknown source file :       Unable to add '/home/kyrisu/dev_temp/sample-Groceries/platforms/android/build/intermediates/assets/debug/app/tns_modules/semver/semver.browser.js.gz': file already in archive (try '-u'?)
[...]

PR #2318, opened 2016-12-05
Update log4js version

losttime:
Updates transitive semver dependency which is blocking Android builds in some situations (#1122).


enchev:
run ci

PR #2324, opened 2016-12-07
Use https for submodule initialization

rosen-vladimirov:
At the moment the submodule is initialized via ssh. This way the command `git submodule update --init` will work only for users who have write access to mobile-cli-lib.
This also requires custom logic when travis is used. In current implementation, travis automatically calls `npm install` before executing the scripts, but at this point the submodule is not initialized yet (as we have declared submodules variable to false).
So the `npm install` executed by travis, will fail in case the postinstall script uses some code from the submodule.
In order to fix this two changes are applied:
 - change the sumodule to be used over https instead of ssh.
 - modify .travis.yml to use submodule correctly and remove two manual steps which are executed by travis by default, there's no need the execute them again.

NOTE: In case you have two-factor authentication, you'll not be able to commit to mobile-cli-lib over https. In this case you have to:
 - remove content of lib/common dir
 - clone over ssh: `$ git clone git@github.com:telerik/mobile-cli-lib.git .` (you must be inside lib/common dir).

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "lib/common"]
 	path = lib/common
-	url = git@github.com:telerik/mobile-cli-lib.git
+	url = https://github.com/telerik/mobile-cli-lib.git
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,13 +7,10 @@ language: node_js
 node_js:
 - '4'
 git:
-  submodules: false
+  submodules: true
 before_script:
-- sed -i 's/git@github.com:/https:\/\/github.com\//' .gitmodules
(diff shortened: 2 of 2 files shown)


dtopuzov:
run ci

rosen-vladimirov:
Ping @ErjanGavalji

rosen-vladimirov:
run ci

ErjanGavalji:
Agreed, the [dumb protocol of git](https://git-scm.com/book/en/v2/Git-Internals-Transfer-Protocols#The-Dumb-Protocol) has proven as the better option for submodules. That will be true until people learn caring about their security and using SSH :)

PR #2334, opened 2016-12-08
Fix CLI usage with Node.js 4

rosen-vladimirov:
When a package has to be used globally, npm automatically creates required executable based on the `bin` key in `package.json`.
In our case the executable, that's created, is a shell script (`.cmd` on Windows) that starts a node process and launches `<cli dir>/bin/nativescript.js`.
This makes it impossible for us to modify the arguments passed to Node.js. But for Node.js 4.x.x, we want to pass `--harmony` flag.
In order to fix this, modify nativescript.js to checks current node version and if we want to pass additional arguments to Node.js process, start a new Node process inside it.
In case the version of Node is 6.x.x or later, we'll use the old code, where the operations are in the same process.

For easier development, we've created several shell scripts (`.cmd` on Windows) inside CLI's bin directory. Developers of CLI add the bin dir to their `$PATH` variable instead of installing the CLI globally. In order to allow usage of `tns` directly from the terminal for them, we've created shell scripts that start the `bin/nativescript.js` executable.
However the `bin/nativescript.js` has shebang header and it can be executed directly. So in order to simplify the bin directory, rename `bin/nativescript.js` to `tns` and remove all other files.
This way when CLI developers add bin to their $PATH environment variable, they'll be able to call `tns` directly. Set the executable name in `package.json`, so when the CLI is installed globally, everything will work as before.


Mitko-Kerezov:
Seems legit 👍

PR #2335, opened 2016-12-08
Fix scripts

rosen-vladimirov:
Fix postinstall and preuninstall scripts which incorrectly try to start `bin/nativescript.js`, which does not exist anymore.


Mitko-Kerezov:
LGTM :+1:

PR #2337, opened 2016-12-08
Fix $injector to resolve lambdas

rosen-vladimirov:
As the transpilation had been changed to ES6 instead of ES5, there is a problem, when $injector should resolve lambdas.
When they are transpiled to ES5, they are converted to functions. However in ES6 they are still lambdas and current `annotate` method does not work correctly.
Fix the regex that's used for the parsing and add unit tests for the failing cases.


Mitko-Kerezov:
LGTM :+1:

PR #2340, opened 2016-12-12
Implement Automatic and Manual Xcode Signing Styles

PanayotCankov:
Upon iOS build for device the user will be prompted to choose between Automatic (the old approach with team id) and Manual (new approach using specific provisioning profile). Once the team id or provisioning profiles are selected, the Xcode project will be updated to match the Xcode 8 behavior where the "Automatically manage signing" checkbox is checked or not.

There are some issues to be resolved in: https://github.com/NativeScript/ios-mobileprovision-finder/issues


lib/services/ios-project-service.ts
```diff
@@ -300,9 +304,62 @@ export class IOSProjectService extends projectServiceBaseLib.PlatformProjectServ
...
+				if (buildConfig.codeSignIdentity || buildConfig.mobileProvisionIdentifier) {
+					signingStyle = "Manual";
+				} else if (buildConfig.teamIdentifier) {
+					signingStyle = "Automatic";
+				} else {
+					let signingStyles = [
+						"Manual - Select existing provisioning profile for use",
+						"Automatic - Select Team ID for signing and let Xcode select managed provisioning profile"
+					];
+					let signingStyleIndex = signingStyles.indexOf(this.$prompter.promptForChoice("Select codesiging style", signingStyles).wait());
```
rosen-vladimirov: What will happen in non-interactive terminal? The prompter will wait for input, but there will be no input, so maybe the execution will hang here.

lib/services/ios-project-service.ts
```diff
@@ -300,9 +304,62 @@ export class IOSProjectService extends projectServiceBaseLib.PlatformProjectServ
...
+							break;
+						case "Automatic":
+							buildConfig.teamIdentifier = this.getDevelopmentTeam();
+							this.persistDevelopmentTeam(buildConfig.teamIdentifier);
+							break;
+					}
+				}
+
+				switch(signingStyle) {
+					case "Manual": {
```
rosen-vladimirov: do you really need the curly brackets `{` - they seems inconsistent with the other code.

lib/services/ios-project-service.ts
```diff
@@ -1045,73 +1106,148 @@ We will now place an empty obsolete compatability white screen LauncScreen.xib f
...
-				return teamId;
-			}
 		}
-		let fileName = path.join(this.platformData.projectRoot, "teamid");
-		if (this.$fs.exists(fileName).wait()) {
-			return this.$fs.readText(fileName).wait();
+		return result;
+	}
+
+	private getProvisioningProfile(): provision.MobileProvision {
```
rosen-vladimirov: this method has async operations (`.wait()` inside it's body), so it must return `IFuture<provision.MobileProvision>`

lib/services/ios-project-service.ts
```diff
@@ -1045,73 +1106,148 @@ We will now place an empty obsolete compatability white screen LauncScreen.xib f
...
+		this.$logger.warn("You can specify the team id by setting the DEVELOPMENT_TEAM setting in build.xcconfig file located in App_Resources folder of your app, or by using the --teamId option when calling run, debug or livesync commnads.");
+		if (teams.length === 1) {
+			teamId = teams[0].id;
+			this.$logger.warn("Found and using the following development team installed on your system: " + teams[0].name + " (" + teams[0].id + ")");
+		} else if (teams.length > 0) {
+			let choices: string[] = [];
+			for (let team of teams) {
+				choices.push(team.name + " (" + team.id + ")");
+			}
+			let choice = this.$prompter.promptForChoice('Found multiple development teams, select one:', choices).wait();
```
rosen-vladimirov: As this is async operation, the method `getDevelopmentTeam` must return IFuture

lib/services/ios-project-service.ts
```diff
@@ -1045,73 +1106,148 @@ We will now place an empty obsolete compatability white screen LauncScreen.xib f
...
-				teamId = teams[0].id;
-				this.$logger.warn("Found and using the following development team installed on your system: " + teams[0].name + " (" + teams[0].id + ")");
-			} else if (teams.length > 0) {
-				let choices: string[] = [];
-				for (let team of teams) {
-					choices.push(team.name + " (" + team.id + ")");
+		return teamId;
+	}
+
+	private persistProvisioningProfiles(uuid: string) {
```
rosen-vladimirov: this method needs return type.

lib/services/ios-project-service.ts
```diff
@@ -1045,73 +1106,148 @@ We will now place an empty obsolete compatability white screen LauncScreen.xib f
...
-						break;
-					default:
-						break;
+				break;
+			default:
+				break;
+		}
+	}
+
+	private persistDevelopmentTeam(teamId: string) {
```
rosen-vladimirov: this method needs return type

lib/services/ios-project-service.ts
PanayotCankov: This was interactive before my intervention, read getDevelopmentTeam

lib/services/ios-project-service.ts
PanayotCankov: No I don't.

lib/services/ios-project-service.ts
PanayotCankov: readTeamId was not returning future before my intervention

PanayotCankov:
QSF build, as provisioning profiles from xcconfig were propagated as command line args to xcodebuild and as result tried to sign Cocoa pods. This PR: #2356 will resolve this.

PanayotCankov:
When provisioning profile is selected manually it should either be stored in xcconfig or in the pbxproj in the platforms/ios instead of passed as command line arg to xcodebuild, and thus used only for the App build. When it is passed as command line arg xcodebuild will try to sign and add the provisioning profile to the dynamic framework build for Cocoapods, instead they should not be signed.

PanayotCankov:
On clear project without `platforms` folder `tns run ios` (and the like) when trying to build for device will ask for provisioning profiles used for signings as follow:

 - Ask for 'Manual' or 'Automatic' code signing
    - When 'Manual'
        - Ask for a valid provisioning profile supporting the selected device, fail the build with a proper message if there are no matching provisioning profiles.
            - Ask for persistence choice "xccofig in app/App_Resources", "platforms/ios" or "do not persist"
               - When persisted in "xccofig in app/App_Resources" the profile is saved in the build.xcconfig and used to sign the app during the build.  
                  - On a subsequent `tns run ios` no dialogs are presented and the provisioning profile from the build.xcconfig is used to sign the app.
                  - If the `plaforms/ios` is deleted and subsequent `tns run ios` is issued the build.xcconfig will preserve the provisioning profile and it will be used to sign the app for device.  
               - When persisted in "platforms/ios" the selected provisioning profile is set in the Xcode project and if the platforms/ios/<app>.xcodeproj is opened in Xcode the App target will have its signing set to "Manual" and the selected provisioning profile will be selected in the comboboxes, just as if this is done in the Xcode itself.
                 - On subsequent `tns run ios` the provisioning profile will be used.
                 - It the `platforms/ios` is deleted and subsequent `tns run ios` is issued the provisioning profile selection will be lost and the client will be asked to select signing from ground up.
               - When not persisted, the selected provisioning profile will be used.  
                  - On subsequent `tns run ios` the provisioning profile selection will be lost and the client will be asked to select provisioning profile from ground up.
    - When 'Automatic'
       - Ask for a Team, fail with proper error message when there are no teams to select from.
            - Ask for persistence choice "xccofig in app/App_Resources", "platforms/ios" or "do not persist"
               - When persisted in "xccofig in app/App_Resources" the team id is saved in the build.xcconfig  
                  - On a subsequent `tns run ios` no dialogs are presented and the team id from the build.xcconfig is used to sign the app.
                  - If the `plaforms/ios` is deleted and subsequent `tns run ios` is issued the build.xcconfig will preserve the team id and it will be used to sign the app for device.  
               - When persisted in "platforms/ios" the selected team id is saved in the Xcode project and if the platforms/ios/<app>.xcodeproj is opened in Xcode the App target will have its signing set to "Automatic" and the selected team will be selected in the team combobox, just as if this is done in the Xcode itself.
                 - On subsequent `tns run ios` the team id will be used.
                 - It the `platforms/ios` is deleted and subsequent `tns run ios` is issued the team id will be lost and the client will be asked to select signing from ground up.
               - When not persisted, the selected team id will be used with automatic signing to sign the app.  
                  - On subsequent `tns run ios` the team id selection will be lost and the client will be asked to select signing from ground up.

Must be able to sign cocoa pods. This is important as at some points signing was passed as command line args and forced all build targets to use the provisioning specified. If the provisioning profile targets `org.nativescript.*` identifiers it may match the app target but pods are prefixed with different identifier (~`org.cocoapods.<podname>`) and the build fails to sign the cocoa pod targets.

For the future:
 - Currently switching the iOS device will not trigger validation nor new provisioning profile selection. It may be desirable to do such check.
 - Same provisioning profile will be set for both Debug and Release, it would be preferable to allow the provisioning profile selection to be forced from the command line.
 - We have publish command for iOS but there may be demand for distribution certificates to be selected during iOS build.

PanayotCankov:
Due to high complexity we will try to relax the supported scenarios and narrow the supported paths to:

 - Setting team id in the build.xcconfig should set Automatic signing style in the Xcode project and use the xcconfig team id for builds.
 - Setting provision in the build.xcconfig should set Manual signing style in the Xcode project and use the xcconfig defined provision.
 - Running `tns * ios --team-id` or `tns * ios --provision` will set Manual or Automatic signing style and set the team or provision in the native Xcode project. If the --team-id or --provision is not provided option, the CLI will just list the teams and profiles, (or interactive dialog?) Both display names and UUIDs should be allowed.

Issue #2301, opened 2016-12-02
There are currently 2 package.json files after tns create


valentinstoychev:
Currently, after executing `tns create`, there are two `package.json` files - one in the root folder and one in the `app` folder. These two files are with almost identical content and it is confusing for the user which one to use.

What is the reason for this?

![cli](https://cloud.githubusercontent.com/assets/4980822/20834036/904b4ebc-b89b-11e6-9d9f-5a7b3cace767.png)

rosen-vladimirov:
Hey @valentinstoychev , 
The `package.json` at the root is used to define the dependencies of the template, to persist some project specific information, etc.
The `package.json` in the `app` folder is coming from the template. It has several purposes:
- it allows us to have the templates published in npm.
- it defines the dependencies and devDependencies required to use the template - nativescript-cli copies them to root's package.json during creation.
- most important - this package.json is copied to native project and it defines the entry point of the application (`"main": "main.js"). At runtime this value is read and application is started from this point.

That's the current workflow and the reason why we have two package.json's. Of course this can be changed, but we should take care to handle all of the above requirements, so everything will continue working in the same manner :)

Plamen5kov:
I wouldn't say having two `package.json` files is confusing, since this has been the default workflow for more than an year and no one seems to have problems with multiple `package.json` files.

valentinstoychev:
@rosen-vladimirov there are two problems:
1. The `package,json` in the app folder contains mixed data - data for the template package and data for the application. The developer is not interested in the data for the template package. This is for our internal use only. We discussed with Plamen that this data can be deleted after `tns create` is executed. 

2. Having the dependencies of the application listed in two different places is misleading and confusing to the developer. I personaly lost time figuring out why a dependency listed in the `package.json` in the `app` folder is not working. Which `package.json` he need to edit in order to set the proper dependencies? there is no single source of truth (when looking from customer perspective).

Ultimately we will have only one package.json which will be application specific.

For the time being though in order not to make drastic changes to the way the CLI works we decided with Plamen to keep the two `package.json` files, but filter and remove the information that is misleading and not needed at runtime.

valentinstoychev:
`{

/* npm package properties

    "private": false,
    "nativescript": {
        "id": "org.nativescript.helloworldng"
    },
    "name": "tns-template-hello-world-ng",
    
    
    /*"version": "2.4.2",
    "author": "Telerik <support@telerik.com>",
[...]

PR #2346, opened 2016-12-13
delete all unnecessary keys from app/package.json

Plamen5kov:
See [issue](https://github.com/NativeScript/nativescript-cli/issues/2301) for more info.

_Problem_
People get confused because there are two `package.json` files, one in the root folder and one in `app/` folder. The two `package.json` files have the same dependencies and devDependencies, which is the main reason for the confusion.

_Solution_ 
The solution discussed with @valentinstoychev was to keep the `app/package.json` as simple as possible, and if possible to add some comments to each key in the `app/package.json` to explain what is the purpose of it. Because there are no comments in the `.json` format, IMO it's best to describe the functionality in the documentation.

This PR is not final and is only one possible solution. We can decide we want to apply different solution to the problem later on after a discussion. This is simply an easy fix.

Currently we keep only the `name`, `main`, `version` and `android` keys in the `app/package.json` if anyone thinks we should exclude or include more keys, please comment this PR or [this](https://github.com/NativeScript/nativescript-cli/issues/2301) issue.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -7,17 +7,23 @@
 		{
 			"type": "node",
 			"request": "launch",
-			"name": "Launch Program (Node 6+)",
-			"program": "${workspaceRoot}/lib/nativescript-cli.js",
-			"cwd": "${workspaceRoot}",
+			"cwd": "${workspaceRoot}/scratch",
 			"sourceMaps": true,
-			// define the arguments that you would like to pass to CLI, for example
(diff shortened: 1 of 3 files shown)


Plamen5kov:
run CI

dtopuzov:
run ci

pkoleva:
run CI

pkoleva:
run CI

Plamen5kov:
run CI

Plamen5kov:
run CI

PR #2351, opened 2016-12-14
Treat tns-core-modules as a normal npm module

Pip3r4o:
**Related to** https://github.com/NativeScript/nativescript-cli/pull/2156
The *temporary* logic for extracting every single piece of `tns-core-modules` at root level of `tns_modules` is removed as discussed back during the `2.4` release.

iOS livesync logic should in turn be revised, so we can ensure it doesn't break following these changes.


Plamen5kov:
The failing test is expected, I recommend merging this in master, because there are known issues that need to be resolved. 
ping: @enchev @dtopuzov

PR #2364, opened 2016-12-19
Plamen5kov/respect shelljserrors

Plamen5kov:
This PR should allow shelljs errors to be fatal and interrupt the execution of any command if necessary. The current implementation has accounted for several places where, the errors are expected and are handled accordingly.
One more thing is the minimal project template support version is upped from `1.3.0` to `1.5.0`.

This PR should be merged after: https://github.com/telerik/mobile-cli-lib/pull/861

### DIFF
--- a/lib/services/project-service.ts
+++ b/lib/services/project-service.ts
@@ -139,9 +139,6 @@ export class ProjectService implements IProjectService {
 			this.$fs.createDirectory(appDestinationPath);
 
 			shelljs.cp('-R', path.join(appSourcePath, "*"), appDestinationPath);
-			// Copy hidden files.
-			//TODO: plamen5kov: don't copy hidden files, until we find a reason to do it
-			// shelljs.cp('-R', path.join(appSourcePath, ".*"), appDestinationPath);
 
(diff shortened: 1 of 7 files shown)


Plamen5kov:
run CI

lib/services/project-service.ts
```diff
@@ -140,7 +140,8 @@ export class ProjectService implements IProjectService {
 
 			shelljs.cp('-R', path.join(appSourcePath, "*"), appDestinationPath);
 			// Copy hidden files.
-			shelljs.cp('-R', path.join(appSourcePath, ".*"), appDestinationPath);
+			//TODO: plamen5kov: don't copy hidden files, until we find a reason to do it
+			// shelljs.cp('-R', path.join(appSourcePath, ".*"), appDestinationPath);
```
rosen-vladimirov: Strongly disagree here. Why you need a reason for already existing code? Either remove it or keep it in the way it works today. Also I can reverse the question - why wouldn't you copy all files from template? Hidden files could be part of the template, so... why not copying them? And last - on Windows they are not hidden at all.

pkoleva:
run CI

Plamen5kov:
run CI

Issue #2270, opened 2016-11-28
App is not rebuilt after removing plugin with native dependencies


Pip3r4o:
Steps to reproduce:

```shell
tns create app
cd app
tns plugin add ../path-to-plugin
tns run android
tns plugin remove plugin-name
tns run android
```
The last command executed `tns run/debug android` will not trigger a rebuild because it implicitly applies `rebuild=false` flag although a rebuild is very often obligatory when dealing with native dependencies

PR #2369, opened 2016-12-21
Invoke gradle fullClean on android after removing a native plugin

Pip3r4o:
**Depends on**: https://github.com/NativeScript/android-runtime/pull/664

Moved native plugin clean logic to android build.gradle. Now when removing a plugin with android libraries in the form of jars and aars, the android build directory is invalidated, such that a full build will be necessary the next time the project is ran.

Addresses https://github.com/NativeScript/nativescript-cli/issues/2270

### DIFF
--- a/lib/services/android-project-service.ts
+++ b/lib/services/android-project-service.ts
@@ -359,8 +359,11 @@ export class AndroidProjectService extends projectServiceBaseLib.PlatformProject
 
 	public removePluginNativeCode(pluginData: IPluginData): void {
 		try {
-			this.$fs.deleteDirectory(path.join(this.platformData.projectRoot, "configurations", pluginData.name));
-			this.$fs.deleteDirectory(path.join(this.platformData.projectRoot, "src", pluginData.name));
+			// check whether the dependency that's being removed has native code
+			let pluginConfigDir = path.join(this.platformData.projectRoot, "configurations", pluginData.name);
+			if (this.$fs.exists(pluginConfigDir)) {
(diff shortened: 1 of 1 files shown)


lib/services/android-project-service.ts
```diff
@@ -396,18 +399,24 @@ export class AndroidProjectService extends projectServiceBaseLib.PlatformProject
...
-				gradleBin += ".bat";
-			}
-			this.spawn(gradleBin, buildOptions, { stdio: "inherit", cwd: this.platformData.projectRoot }).wait();
+
+			this.cleanProject(projectRoot, buildOptions);
 		}
 		return Future.fromResult();
 	}
 
+	private cleanProject(projectRoot: string, options: string[]): void {
```
rosen-vladimirov: the method has async operation (`.wait()` ), so it must return `IFuture<void>` instead of `void`.

Pip3r4o: If I `wait()` the asynchronous `spawn`, doesn't that mean that the method will execute synchronously, as the async operation is resolved within the method?

rosen-vladimirov: Yes, but think of the async method in a C# manner:
You can use `await` only in methods marked as `async`.
`Futures` allow such inconsistency - to have `.wait` in a pure sync method, but it's a common practice to wrap your code in a Future, when anywhere in the body you are using `.wait`. Also Promises and `async/await` in TypeScript does not allow having `await` without the method being marked as `async`, so it will be easier to replace the `IFuture` with `Promise` in the method definition.

Issue #1366, opened 2015-12-29
Make `tns run` use livesync by default


tjvantoll:
I’ve spent some time over the holidays playing with a variety of frameworks, and one thing I’ve noticed is that livesync functionality is increasingly the default behavior that developers expect. React Native watches by default, and Angular 2’s setup does as well.

My suggestion is to have `tns run` also setup/run `tns livesync` by default (and perhaps provide a `--no-livesync` flag to turn it off or something like that). IMO this would significantly help the getting started experience. One complication is that we probably want https://github.com/NativeScript/nativescript-cli/issues/763 implemented first, so that `tns run` continues to show console logs and stack traces.

> UPDATE (May 20th, 2016): #763 was implemented and shipped in 2.0.

One alternative is that we could include a handful of npm scripts in the default project’s `package.json`. Something like `npm run ios|android` that runs the app and sets up a watcher by default. For example in Angular 2 `npm start`, starts a browser tab and an HTTP server, watches for changes to code, and deploys those changes automatically.

stephenfeather:
Agree with @tjvantoll, can say I use `watchy` a good bit around my dev environments, having 2 or 3 terminals open, touching each other to trigger various build stages. 

Yes, that watched build, that LiveSync is high on the expectation list for new users.  Provided the user's environment can handle it, put up some info in the docs and turn it on.

(Side rant, xCode is the worst with its partial saves of your work as you are typing, triggering off broken process' down stream.)

sitefinitysteve:
+1 makes sense, one less command to confuse people...

teobugslayer:
As a work-around, livesync builds and deploys the project if needed.

I.e. instead of `run` finishing with `livesync`, `livesync` starts with `run`.

Perhaps we can make this more visible in the documentation?

tjvantoll:
@teobugslayer Ah ok, I didn’t know that. [This section of the getting started guide](http://docs.nativescript.org/start/getting-started#running-your-app) would probably be the place to do that, but I’ve been holding off promoting livesync there because it still hides errors and stack traces.

burkeholland:
I think that LiveSync should be the default experience for `tns run`

tjvantoll:
@valentinstoychev With the livesync improvements coming in 2.1 and 2.2, can we consider making it the default experience as this issue suggests?

sipacate:
Yup. I fully agree that would be the nicest thing to do from the developer experience.

[...]

Issue #2341, opened 2016-12-12
LiveSync after {N} update uses old app on the device


valentinstoychev:
Scenario: You have an app deployed on a device using livesync. You update to a new version of NativeScrpt (modules, or platforms) and then run livesync again for the same app.

The result is that the Livesync routine is not deploying the new app with the new version of the NativeScript framework, but is deploying only the changed files without the new app. What needs to be done is livesync to upload the new app to the device in order to make sure that the local version of the app is the same as the already deployed version. I also synced this with @tzraikov. 

This was the problem for this issue - https://github.com/NativeScript/NativeScript/issues/3281

Since using livesync is our recommended way of development process, we need to make sure that livesync is taking this scenario into consideration because this the primary user experience when updating to a new version of the framework.

This is related - https://github.com/NativeScript/nativescript-cli/issues/2330

Issue #2330, opened 2016-12-07
`tns run` command to watch for changes by default


tzraikov:
Currently when executing `tns run` command you should pass the `--watch` option in order to start watching for changes. We will change this behavior so that when you execute `tns run` command the watch to be on by default.
If you don't want to watch, we will provide a `--no-watch` option.
If you want to rebuild the application explicitly, we will provide the `--clean` option
> The deploy, prepare and build commands will not be changed. The `livesync` command will be deprecated in 2.5 and removed in 2.6.
We should update the docs accordingly.

vchimev:
Additional:
In order to run the app and release the console/terminal, `--just-launch` option should be used.

Issue #2315, opened 2016-12-05
tns debug ios --watch fails on newly created app if there is already deployed app with the same app id


ivanbuhov:
`tns debug ios --watch` fails on newly created app if there is an app with the same app id already deployed on the device/emulator. The CLI fails with the following error message:

```
ENOENT: no such file or directory, scandir '/Users/buhov/Desktop/MyApp/platforms/ios/build/emulator'
```

PR #2374, opened 2016-12-27
Chnged: `tns run` command to watch for changes by default

tzraikov:
Changes the run command behavior to watch for changes by executing livesync.
Changes the livesync behavior to watch for changes in all files. The project-changes-service detects whether the change requires a full build or files can be livesynced.
The livesync state is preserved between calls, so no extra files will be transferred on second run.

Fixes: #1366, #2276, #2277, #2303, #2315, #2330, #2341, and most parts of #2153

The new prepare, build, deploy, run, debug and livesync commands behavior is described here: https://github.com/NativeScript/nativescript-cli/wiki/Workflow-commands-spec

It is based on the one described in this issue:
https://github.com/NativeScript/nativescript-cli/issues/475

Currently there are two issues not related with livesync:
- the snapshot plugin executes in debug and modifies file times. This breaks the livesync logic. The change should be in snapshot module.
- changing the app id in package.json doesn't work on android because the gradle script uses the app.gradle file directly from App_Resources folder.

* Currently livesync doesn't work when adding/removing project dependencies. We should add this in our documentation.

**DO NOT MERGE BEFORE https://github.com/telerik/mobile-cli-lib/pull/863 is merged in mobile-cli-lib**


lib/services/livesync/livesync-service.ts
```diff
@@ -3,11 +3,9 @@ import * as helpers from "../../common/helpers";
 import * as path from "path";
 import * as semver from "semver";
 import * as fiberBootstrap from "../../common/fiber-bootstrap";
-
-let gaze = require("gaze");
+let choki = require("chokidar");
```
rosen-vladimirov: Also it's not added to package.json, so it will not be installed on any machine.

lib/services/project-changes-service.ts
```diff
@@ -0,0 +1,220 @@
...
+			let changed = fileStats.mtime.getTime() > this._outputProjectMtime;
+			if (!changed) {
+				let lFileStats = this.$fs.getLsStats(filePath);
+				changed = lFileStats.mtime.getTime() > this._outputProjectMtime;
+			}
+			if (changed) {
+				if (processFunc) {
+					this._newFiles ++;
+					let filePathRelative = path.relative(this.$projectData.projectDir, filePath);
+					let that = this;
```
rosen-vladimirov: I believe you do not need `that` variable here, you can pass `this` directly.

rosen-vladimirov:
I've tested the current implementation for Android and there's some strange behavior. It looks like no matter of the change I made, a package.json is always transferred to the device. This causes double restart of the application during livesync. You can check [this video](http://screencast.com/t/bt1fZ63a8) where the issue is visible.

rosen-vladimirov:
Based on the code and related issues in the descriptions, I assume the current PR should allow adding new dependencies in package.json during `tns run android` (while file system watcher is working). I've tried the following:
- In terminal 1: `tns run android`
- After it finishes, I've opened terminal 2 and I've tried `npm install --save lodash`. Before it finished, I received an error in terminal 1:
```
Executing before-prepare hook from D:\Work\nativescript-cli\scratch\app2303-1\hooks\before-prepare\nativescript-dev-android-snapshot.js
Preparing project...
Project successfully prepared (android)
Executing after-prepare hook from D:\Work\nativescript-cli\scratch\app2303-1\hooks\after-prepare\nativescript-dev-android-snapshot.js
Building project...

:config phase:  createDefaultIncludeFiles
        +found plugins: nativescript-barcodescanner
        +found plugins: tns-core-modules-widgets

:config phase:  createPluginsConfigFile
         Creating product flavors include.gradle file in D:\Work\nativescript-cli\scratch\app2303-1\platforms\android/configurations folder...

:config phase:  pluginExtend
        +applying configuration from: D:\Work\nativescript-cli\scratch\app2303-1\platforms\android\configurations\include.gradle
        +applying configuration from: D:\Work\nativescript-cli\scratch\app2303-1\platforms\android\configurations\nativescript-barcodescanner\include.gradle
        +applying configuration from: D:\Work\nativescript-cli\scratch\app2303-1\platforms\android\configurations\tns-core-modules-widgets\include.gradle

:config phase:  copyAarDependencies
> Configuring > 0/3 projects > root projectError: EPERM: operation not permitted, lstat 'D:\Work\nativescript-cli\scratch\app2303-1\node_modules\lodash'
1/3 projects
vladimirov@VLADIMIROV  /d/Work/nativescript-cli/scratch/app2303-1 (raikov/fix-2303c)
Observed package id 'build-tools;21.1.1' in inconsistent location 'D:\Program Files\android_sdk\build-tools\build-tools-21.1.1' (Expected 'D:\Program Files\android_sdk\build-tools\21.1.1')

* Where:
Build file 'D:\Work\nativescript-cli\scratch\app2303-1\platforms\android\build.gradle' line: 491

* What went wrong:
A problem occurred evaluating root project 'app23031'.
> Could not read path 'D:\Work\nativescript-cli\scratch\app2303-1\node_modules\lodash'.

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output.

BUILD FAILED

Total time: 6.524 secs
```

Should this work?

lib/services/livesync/livesync-service.ts
tzraikov: Yes, there is a reason. I observed that gaze is quite slow and unstable when watching for a large number of files, while chokidar handles this scenario fast and reliable. Chokidar is the preferred watching tool used in webpack.

I added the corresponding entry in package.json

lib/services/livesync/livesync-service.ts
```diff
@@ -122,25 +114,27 @@ class LiveSyncService implements ILiveSyncService {
...
+						that.$logger.info("Try saving it again or restart the livesync operation.");
+					}
+				}).future<void>()());
 			});
 		});
 
-		this.$processService.attachToProcessExitSignals(this, () => gazeWatcher.close());
+		this.$processService.attachToProcessExitSignals(this, () => {
+			watcher.unwatch(pattern);
+			process.exit();
```
tzraikov: I was not able to stop the process without calling process.exit. Maybe there is something specific in chokidar preventing the process from exiting. I didn't observed any side effects while testing this code. Watching for changes in all files is important in order to support predictable livesync. So I am open for suggestions.

lib/options.ts
```diff
@@ -38,10 +38,11 @@ export class Options extends commonOptionsLibPath.OptionsBase {
 			bundle: { type: OptionType.Boolean },
 			all: { type: OptionType.Boolean },
 			teamId: { type: OptionType.String },
-			rebuild: { type: OptionType.Boolean, default: true },
 			syncAllFiles: { type: OptionType.Boolean },
 			liveEdit: { type: OptionType.Boolean },
-			chrome: { type: OptionType.Boolean }
+			chrome: { type: OptionType.Boolean },
+			clean: { type: OptionType.Boolean },
+			watch: { type: OptionType.Boolean, default: true },
```
tzraikov: Technically, yes. Someone could argue that this way it is easier to add new entries, however different languages, different styles.. removed the comma.

lib/services/platform-service.ts
```diff
@@ -215,46 +214,45 @@ export class PlatformService implements IPlatformService {
...
-
-			if (!changeInfo.modulesChanged) {
+			if (!changesInfo || !changesInfo.modulesChanged) {
 				excludedDirs.push(constants.TNS_MODULES_FOLDER_NAME);
 			}
 
 			this.$projectFilesManager.processPlatformSpecificFiles(directoryPath, platform, excludedDirs);
 
-			if (changeInfo.configChanged || changeInfo.modulesChanged) {
+			if (!changesInfo || (changesInfo.configChanged || changesInfo.modulesChanged)) {
```
tzraikov: correct! 👍

lib/services/platform-service.ts
```diff
@@ -313,55 +311,191 @@ export class PlatformService implements IPlatformService {
...
+				buildTime: new Date().toString()
+			};
+			this.$fs.writeJson(buildInfoFile, buildInfo);
+			this.$logger.out("Project successfully built.");
+		}).future<void>()();
+	}
+
+	public shouldInstall(device: Mobile.IDevice): boolean {
+		if(this.$options.release) {
+			return true;
```
tzraikov: Humm, just wanted to ensure that everything is ok in release. In case if there are some artefacts from debug. It shouldn't be necessary.

lib/services/platform-service.ts
```diff
@@ -313,55 +311,191 @@ export class PlatformService implements IPlatformService {
...
+	private getDeviceBuildInfoFilePath(device: Mobile.IDevice): string {
+		let deviceAppData = this.$deviceAppDataFactory.create(this.$projectData.projectId, device.deviceInfo.platform, device);
+		let deviceRootPath = deviceAppData.deviceProjectRootPath;
+		if (device.deviceInfo.platform.toLowerCase() === this.$devicePlatformsConstants.Android.toLowerCase()) {
+			deviceRootPath = path.dirname(deviceRootPath);
+		}
+		return path.join(deviceRootPath, buildInfoFileName);
+	}
+
+	private getDeviceBuildInfo(device: Mobile.IDevice): IBuildInfo {
```
tzraikov: Changed the method to use IFuture. 👍

lib/services/project-changes-service.ts
tzraikov: Removed the variable. 👍

lib/services/platform-service.ts
tzraikov: Removed the code.

lib/services/livesync/platform-livesync-service.ts
```diff
@@ -184,30 +167,64 @@ export abstract class PlatformLiveSyncServiceBase implements IPlatformLiveSyncSe
...
 			return (() => {
-				let deviceAppData = this.$deviceAppDataFactory.create(this.liveSyncData.appIdentifier, this.$mobileHelper.normalizePlatformName(this.liveSyncData.platform), device);
-				let localToDevicePaths = this.$projectFilesManager.createLocalToDevicePaths(deviceAppData, this.liveSyncData.projectFilesPath, filesToSync, this.liveSyncData.excludedProjectDirsAndFiles);
-
-				fileSyncAction(deviceAppData, localToDevicePaths).wait();
+				let deviceAppData:Mobile.IDeviceAppData = null;
+				let localToDevicePaths:Mobile.ILocalToDevicePathData[] = null;
+				let isFullSync = false;
+				if (this.$options.clean || this.$projectChangesService.currentChanges.changesRequireBuild) {
+					this.$platformService.buildPlatform(this.liveSyncData.platform, { buildForDevice: !device.isEmulator }).wait();
```
tzraikov: Correct! Changed the logic to build only when necessary. 👍

tzraikov:
Regarding the scond rebuild on android, yes I confirm the issue. Like I pointed, it is caused by the android snapshot plugin. It modifies last modified times for different files explicitly every time it is executed (regardless if this is debug or release). I am not sure why it does this. It breaks the livesync  watching logic. My personal opinion is that it should be corrected in android snapshot plugin. It should be able to work without changing file times explicitly.

You shouldn't see this glitch if you remove the hooks related with android snapshot plugin from your project.

tzraikov:
No, currently adding dependencies while watching for changes will not work. This is caused by the fact that livesync syncs/rebuilds on every change. When adding a new dependency you are adding a lot of files at the same time. This will lead to unpredictable behaviour. In some cases it will cause endless sync/rebuild on every file change, in other cases it will break.  I was not able to find an easy solution for this scenario. We should state this explicitly in our documentation.

lib/services/livesync/livesync-service.ts
tzraikov: Just found a way to stop the watcher without process.exit. Changed the code.

vchimev:
Please make sure this PR addresses also the issue below:
- [ ] https://github.com/NativeScript/nativescript-cli/issues/2267#issuecomment-270123791

PR #2381, opened 2017-01-03
fix behavior with old templates

Plamen5kov:
This PR is to help the users understand what needs to be done, when the template they're using is old.

When there's an old template with missing `app.gradle` there's an error on copy added a verbose log that helps the user take care of updating the template.

This present's an opportunity for a known issues page in the documentation, to which we can refer when we feel the need to.

@valentinstoychev @tsonevn

### DIFF
--- a/lib/services/android-project-service.ts
+++ b/lib/services/android-project-service.ts
@@ -199,7 +199,7 @@ export class AndroidProjectService extends projectServiceBaseLib.PlatformProject
 			try {
 				shell.sed('-i', /__PACKAGE__/, this.$projectData.projectId, userAppGradleFilePath);
 			} catch(e) {
-				this.$errors.failWithoutHelp(` Error: ${e}.\nIt's possible you're using an old template. You can try updating it and try again.`);
+				this.$logger.warn(`\n${e}.\nIt's possible you're using an old template.`);
 			}
 		}).future<void>()();
 	}
--- a/lib/services/android-project-service.ts
+++ b/lib/services/android-project-service.ts
@@ -116,7 +116,13 @@ export class AndroidProjectService extends projectServiceBaseLib.PlatformProject
 			} else {
 				this.copy(this.platformData.projectRoot, frameworkDir, "src", "-R");
 			}
-			this.copy(this.platformData.projectRoot, frameworkDir, "build.gradle settings.gradle gradle.properties build-tools", "-Rf");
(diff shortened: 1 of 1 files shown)


lib/services/android-project-service.ts
```diff
@@ -190,7 +190,11 @@ export class AndroidProjectService extends projectServiceBaseLib.PlatformProject
 
 			// will replace applicationId in app/App_Resources/Android/app.gradle if it has not been edited by the user
 			let userAppGradleFilePath = path.join(this.$projectData.appResourcesDirectoryPath, this.$devicePlatformsConstants.Android, "app.gradle");
-			shell.sed('-i', /__PACKAGE__/, this.$projectData.projectId, userAppGradleFilePath);
+			try {
+				shell.sed('-i', /__PACKAGE__/, this.$projectData.projectId, userAppGradleFilePath);
+			} catch(e) {
+				this.$errors.failWithoutHelp(` Error: ${e}.\nIt's possible you're using an old template. You can try updating it and try again.`);
```
rosen-vladimirov: is there something really important in the current `app.gradle` that requires having it in the template. Can't we just warn the users and continue the execution? Will the project work without app.gradle in App_Resource?

Plamen5kov: The problem is it depends on what is in `app.gradle`, and because we can't guarantee if it's important or not, IMO it's best to stop execution right there and then. We had a similar problem, and it was very unpleasant to figure it out.

rosen-vladimirov: My question is - is there something really important in the default app.gradle that requires having it in all projects? Will the project work without the default app.gradle. In case "yes", this should be warning, not error.
Also what will happen in case I have app.gradle, but I do not have `__PACKAGE__` in it - the `sed` will probably fail - should this stop the execution?

Plamen5kov: IMO all projects we support should have an app_id in app.gradle, in case they don't the package_name of the AndroidManifest.xml is used implicitly instead, which for me is erroneous behavior when publishing an app. That's why a default android studio project sets explicitly the app_id and the package name.
On the second question:
>Also what will happen in case I have app.gradle, but I do not have \_\_PACKAGE__ in it - the sed will probably fail

The sed command doesn't fail, and in case there is a \_\_PACKAGE__ placeholder it replaces it as expected.

Having said that, it's true we can choose warning over failure, if the we agree this is the best behavior userwise.

PR #2391, opened 2017-01-09
Open specific version of Chrome Dev Tools

KristinaKoeva:
Otherwise we would have to update our logic on every Chrome update


KristinaKoeva:
Retargeted to release https://github.com/NativeScript/nativescript-cli/pull/2392

PR #2392, opened 2017-01-09
Open specific version of Chrome Dev Tools

KristinaKoeva:
Otherwise we would have to update our logic on every Chrome update

### DIFF
--- a/lib/services/ios-debug-service.ts
+++ b/lib/services/ios-debug-service.ts
@@ -2,6 +2,7 @@ import * as iOSDevice from "../common/mobile/ios/device/ios-device";
 import * as net from "net";
 import * as path from "path";
 import * as log4js from "log4js";
+import * as os from "os";
 import {ChildProcess} from "child_process";
 import byline = require("byline");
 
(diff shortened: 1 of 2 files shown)


lib/services/ios-debug-service.ts
```diff
@@ -203,7 +203,8 @@ class IOSDebugService implements IDebugService {
 			if (this.$options.chrome) {
 				this._socketProxy = this.$socketProxyFactory.createWebSocketProxy(factory);
 
-				this.$logger.info(`To start debugging, open the following URL in Chrome:\nchrome-devtools://devtools/bundled/inspector.html?experiments=true&v8only=true&ws=localhost:${this._socketProxy.options.port}\n`);
+				let commitSHA = "02e6bde1bbe34e43b309d4ef774b1168d25fd024"; // corresponds to 55.0.2883 Chrome version
```
rosen-vladimirov: you can extract the variable to `private static field` or `const` at the top of the file.

lib/services/ios-debug-service.ts
```diff
@@ -203,7 +203,8 @@ class IOSDebugService implements IDebugService {
...
+				this.$logger.info(`To start debugging, open the following URL in Chrome:\nchrome-devtools://devtools/remote/serve_file/@${commitSHA}/inspector.html?experiments=true&ws=localhost:${this._socketProxy.options.port}\n`.cyan);
```
rosen-vladimirov: `os.EOL` instead of `\n` maybe.

lib/services/ios-debug-service.ts
KristinaKoeva: I kind of disagree to move it at the top because it is only used in this method. Are you OK with the current version ? It is still a local variable but it's now a const.

Issue #2006, opened 2016-08-18
Clean up terminal output with verbose mode 


burkeholland:
Currently, the terminal outputs virtually everything from underlying build processes. This makes the terminal extremely cluttered, and it's hard to find log statements because there is so much noise.

For instance, currently when we run `tns livesync ios --emualtor --watch`, the terminal shells out the following text.

```
tns livesync ios --emulator --watch
Executing before-prepare hook from /Users/burkeholland/dev/burkeholland/nativescript-instagram/hooks/before-prepare/nativescript-dev-typescript.js
Found peer TypeScript 1.8.10
Project successfully prepared (ios)
Executing before-livesync hook from /Users/burkeholland/dev/burkeholland/nativescript-instagram/hooks/before-livesync/nativescript-angular-sync .js
Transferring project files...
Successfully transferred all files.
Applying changes...
Successfully synced application org.nativescript.nativescriptinstagram on device 70547B05-8B62-487D-A39C-5D7522A41B3A.
Executing before-watch hook from /Users/burkeholland/dev/burkeholland/nativescript-instagram/hooks/before-watch/nativescript-dev-typescript.js
Found peer TypeScript 1.8.10
8:23:55 AM - Compilation complete. Watching for file changes.

Aug 18 08:24:00 hollandmac backboardd[74984]: SecTaskCopyDebugDescription: nativescriptinst[93322]
Aug 18 08:24:00 hollandmac backboardd[74984]: SecTaskCopyDebugDescription: nativescriptinst[93322]
Aug 18 08:24:00 hollandmac backboardd[74984]: SecTaskCopyDebugDescription: nativescriptinst[93322]
Aug 18 08:24:00 hollandmac backboardd[74984]: SecTaskCopyDebugDescription: nativescriptinst[93322]
Aug 18 08:24:00 hollandmac backboardd[74984]: SecTaskLoadEntitlements failed error=22 cs_flags=200, task->pid_self=-1
Aug 18 08:24:00 hollandmac backboardd[74984]: SecTaskCopyDebugDescription: nativescriptinst[93322]
Aug 18 08:24:00 hollandmac nativescriptinstagram[93322]: assertion failed: 15G31 13E230: libxpc.dylib + 71506 [6F98A9DA-B46E-3A2B-88D4-92F1EC77477F]: 0x7d
[...]

PR #2393, opened 2017-01-09
Revisit the manual signing, revert the dialogs and add --provision switch

PanayotCankov:


lib/services/ios-project-service.ts
```diff
@@ -311,12 +328,85 @@ export class IOSProjectService extends projectServiceBaseLib.PlatformProjectServ
...
+			if (this.$options.provision) {
+				const pbxprojPath = path.join(projectRoot, this.$projectData.projectName + ".xcodeproj", "project.pbxproj");
+				const xcode = Xcode.open(pbxprojPath);
+				const signing = xcode.getSigning(this.$projectData.projectName);
+
+				let shouldUpdateXcode = false;
+				if (signing && signing.style === "Manual") {
+					for(let config in signing.configurations) {
+						let options = signing.configurations[config];
+						if (options.name !== this.$options.provision && options.name !== this.$options.provision) {
```
rosen-vladimirov: both statements check the same conditions, maybe we really want to be sure that `options.name !== this.$options.provision` 😄

lib/services/ios-project-service.ts
```diff
@@ -311,12 +328,85 @@ export class IOSProjectService extends projectServiceBaseLib.PlatformProjectServ
...
 
 		}).future<void>()();
 	}
 
+	private setupManualSigning(projectRoot: string, buildConfig?: IiOSBuildConfig): IFuture<void> {
+		return (() => {
+			if (this.$options.provision) {
+				const pbxprojPath = path.join(projectRoot, this.$projectData.projectName + ".xcodeproj", "project.pbxproj");
+				const xcode = Xcode.open(pbxprojPath);
+				const signing = xcode.getSigning(this.$projectData.projectName);
```
rosen-vladimirov: you can extract the whole code from here to line 358 in a separate method

package.json
```diff
@@ -58,7 +58,7 @@
     "mute-stream": "0.0.5",
     "open": "0.0.5",
     "osenv": "0.1.3",
-    "pbxproj-dom": "1.0.3",
+    "pbxproj-dom": "^1.0.7",
```
rosen-vladimirov: This is not very safe - in case `pbxproj-dom` is released with breaking change or some unexpected behavior for us in a minor version, we'll be broken on live. That's why we prefer strict versions in package.json

lib/services/ios-project-service.ts
```diff
@@ -248,6 +258,13 @@ export class IOSProjectService extends projectServiceBaseLib.PlatformProjectServ
 				basicArgs.push("-xcconfig", path.join(projectRoot, this.$projectData.projectName, "build.xcconfig"));
 			}
 
+			// if (this.$logger.getLevel() === "INFO") {
```
PanayotCankov: It partially addresses the https://github.com/NativeScript/nativescript-cli/issues/2006
I've tested it locally but QAs won't have time to push it for this release and I didn't bother to go the right way and make a separate branch with this change only to merge in separate PR later. I hope we will uncomment this shortly after the release.

lib/services/ios-project-service.ts
PanayotCankov: http://programmer.97things.oreilly.com/wiki/index.php/First_Write,_Second_Copy,_Third_Refactor

package.json
PanayotCankov: :+1:

PanayotCankov:
@rosen-vladimirov I will adress all the comments in following PRs, the version should probably be fixed ASAP but the rest will probably happen shortly after the release.

lib/services/ios-project-service.ts
PanayotCankov: It should have been .uuid

PR #2399, opened 2017-01-10
Fixed: tns emulate --device id not working

tzraikov:
Fixed: Devices services not initialized! error when executing `tns emulate ios`
Fixed: cannot specify which emulator to start


tzraikov:
run ci

PR #2404, opened 2017-01-11
Changed the syncAllFiles option to be true by default

tzraikov:
When set to false this command causes confusion because changes in node_modules are not synced!


vchimev:
run ci

vchimev:
run ci

PR #2413, opened 2017-01-12
Fixed: "tns emulate android" fails with a space in the name of the home directory

### LINKED ISSUE #2380: "tns emulate android" fails with a space in the name of the home directory
The command `tns emulate android` fails when there is a space in the name of the home directory. My version of nativescript-cli is 2.4.2 and I'm on Windows 10.
Output from the command:
```
Command failed: C:\Users\Jonathan Henriksen\AppData\Local\Android\android-sdk\tools\android list avd
'C:\Users\Jonathan' is not recognized as an internal or external command,
operable program or batch file.
```
It then goes on to print the command usage.

tzraikov:
Fixes #2380

### DIFF
--- a/lib/services/emulator-platform-service.ts
+++ b/lib/services/emulator-platform-service.ts
@@ -165,7 +165,7 @@ export class EmulatorPlatformService implements IEmulatorPlatformService {
     public getAndroidEmulators(): IFuture<IEmulatorInfo[]> {
         return (() => {
             let androidPath = path.join(process.env.ANDROID_HOME, "tools", "android");
-            let text:string = this.$childProcess.exec(`${androidPath} list avd`).wait();
+            let text:string = this.$childProcess.exec(`"${androidPath}" list avd`).wait();
             let notLoadedIndex = text.indexOf("The following");
             if (notLoadedIndex > 0) {
                 text = text.substring(0, notLoadedIndex);


lib/services/platform-service.ts
```diff
@@ -458,16 +458,25 @@ export class PlatformService implements IPlatformService {
...
 			let info = this.$emulatorPlatformService.getEmulatorInfo(platform, this.$options.device).wait();
 			if (info) {
 				if (!info.isRunning) {
 					this.$emulatorPlatformService.startEmulator(info).wait();
 				}
 				this.$options.device = null;
+			} else {
+				this.$devicesService.initialize({ platform: platform, deviceId: this.$options.device }).wait();
+				let found: Mobile.IDeviceInfo[] = [];
+				if (this.$devicesService.hasDevices) {
```
rosen-vladimirov: Why is this code required? `devicesService` will fail in case you pass `--device invalidIdentifier` and it does not find such identifier:
https://github.com/telerik/mobile-cli-lib/blob/master/mobile/mobile-core/devices-service.ts#L337
https://github.com/telerik/mobile-cli-lib/blob/master/mobile/mobile-core/devices-service.ts#L345
https://github.com/telerik/mobile-cli-lib/blob/master/mobile/mobile-core/devices-service.ts#L256

tzraikov: This code is necessary because deployPlatform method calls $devicesService.execute method in order to deploy on all specified devices:
https://github.com/NativeScript/nativescript-cli/blob/master/lib/services/platform-service.ts#L435

The execute method however starts the default emulator when there are no connected devices:
https://github.com/telerik/mobile-cli-lib/blob/9b2d88f4b23247e601fccb35be45c411ea3ddf87/mobile/mobile-core/devices-service.ts#L318 

All this is not expected when executing `tns emulate <platform> --device <device_id>`. The expected result is to fail with the following message:
"Cannot find device with name: <device_id>" as specified by our QAs. Note that this message is different from ERROR_NO_DEVICES message.
This code comes from PR #2412

PR #2416, opened 2017-01-12
Fixed: run command doesn't exit when using --justlaunch option

tzraikov:


lib/services/livesync/livesync-service.ts
```diff
@@ -109,6 +109,9 @@ class LiveSyncService implements ILiveSyncService {
 				this.$hooksService.executeBeforeHooks('watch').wait();
 				this.partialSync(liveSyncData[0].syncWorkingDirectory, watchForChangeActions);
 			}
+			if (this.$options.justlaunch) {
+				process.kill(process.pid);
```
rosen-vladimirov: This is really strange. What's the idea. Isn't this process.exit? What keeps the process alive?

tzraikov: No, this isn't process.exit. Process.exit exits the app unconditionally. This one simulates Ctrl+C and all ongoing tasks registered on SIGKILL will execute.

tzraikov: I suggest to polish this code after release.

slavchev:
just for reference https://github.com/NativeScript/nativescript-cli/pull/2396

PR #2423, opened 2017-01-13
Fixed: the app doesn't exit when using --release and --justlaunch

tzraikov:
The following command doesn't exit: `tns run ios --release --justlaunch` because of the watch option that is enabled by default.

### DIFF
--- a/lib/services/platform-service.ts
+++ b/lib/services/platform-service.ts
@@ -438,6 +438,9 @@ export class PlatformService implements IPlatformService {
 
 	public runPlatform(platform: string): IFuture<void> {
 		return (() => {
+			if (this.$options.justlaunch) {
+				this.$options.watch = false;
+			}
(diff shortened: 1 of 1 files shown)


lib/services/platform-service.ts
```diff
@@ -438,6 +438,9 @@ export class PlatformService implements IPlatformService {
 
 	public runPlatform(platform: string): IFuture<void> {
 		return (() => {
+			if (this.$options.release) {
```
rosen-vladimirov: shouldn't this be `this.$options.justlaunch`

tzraikov: That's correct! Thank you!

tzraikov: Pull request updated

tzraikov:
Yep, there are several ifs :)

PR #2429, opened 2017-01-16
Update SHA to latest common

Mitko-Kerezov:
`isEmulator` method was changed to a `get` property.
It was initially a `get` property but was subsequently changed to a method because there were some asynchronous operations in the `android` implementation.
These are no longer necessary so revert the change and make `isEmulator` a `get` property again.

Ping @rosen-vladimirov


rosen-vladimirov:
Please describe what's changed in mobile-cli-lib and why it should be updated.

PR #2430, opened 2017-01-16
Fixed: .nsbuildinfo/.nslivesyncinfo files should be app specific

tzraikov:
Currently .nsbuildinfo and .nslivesyncinfo files were written in global device folder instead of app specific folders. This will cause issues when deploying different {N} apps on the same device. Currently we can't write those files in app sandbox folder. This commit adds the app id as a  suffix to the corresponding file. We will continue searching for a better solution.


rosen-vladimirov:
Closing this one as https://github.com/NativeScript/nativescript-cli/pull/2434 provides better fix.

Issue #2260, opened 2016-11-25
Deprecate support for Node.js 4.x


vchimev:
We consider deprecating support for Node.js 4.x. For this goal, it is needed to update:

- [ ] Doctor Command - UPDATE: Instead of doctor, every single command warns about this.
- [x] Documentation
- [x] Windows Installer

rosen-vladimirov:
Can you explain in more details why do you want to deprecate LTS version of Node.js. It's recommended for users to use Node.js 4 or 6. Also this version will be supported until 2018-04-01 according to Node.js LTS organization: https://github.com/nodejs/LTS

stephenfeather:
Consider that you have competing frameworks whose tooling is stuck at NodeJS 4.x.
If you move up unnecessarily, those users would be required to carry 2 node installs during any transition phase.

valentinstoychev:
@rosen-vladimirov @stephenfeather - we are forced by the Angular requirements. See - https://github.com/angular/angular/blob/master/package.json#L10-L12

PR #2444, opened 2017-01-19
Deprecate support for Node.js 4

rosen-vladimirov:
Merge after https://github.com/telerik/mobile-cli-lib/pull/879
Deprecate support for Node.js `4.x.x` by adding message that users will see each time when they execute any command and their Node.js version is `4.x.x`.

Implements https://github.com/NativeScript/nativescript-cli/issues/2260


vchimev:
Does the doctor warn about this?

rosen-vladimirov:
@vchimev , nope. Instead of doctor, every single command warns about this ;)

Issue #2443, opened 2017-01-19
[2.5 preview] Initiating debug after a release build crashes the application


Pip3r4o:
The new fast run/livesync strategy does not build a debuggable APK after building once for release

```
tns create ts-my-app
cd ts-myapp
tns build/run android --release --key-store- .... ...
tns debug android # <-- app will not be rebuilt, and instead the release apk will be used
```

Executing the above will crash the application when we open the chrome-devtools frontend or debug through the official VSCode extension. 

**Edit**: app rebuilds, but release-specific artifacts remain in the build and pollute the debug output.

vchimev:
This issue seems unpleasant.

A workaround is to delete the `<app/platform/android>` folder and execute `tns debug android` again.

PR #2452, opened 2017-01-24
clean-up after release build in android to prevent runtime errors

Pip3r4o:
If app's built with the `--release` option, the settings are recorded in `.nsprepareinfo` in the `platforms/android` directory. If the file is present, and follow up builds are not built in release, then the project directory is cleaned manually to ensure that no build artifacts specific to just release (like the android snapshot binaries) are packaged into the debug application.

File changes:
- exposed an android-specific private method as public to the `IPlatformProjectService` and created a stub on the ios side.
- added conditions to clean the build directories on prepare when targeting `android`

**Addresses** https://github.com/NativeScript/nativescript-cli/issues/2443

### DIFF
--- a/lib/services/ios-project-service.ts
+++ b/lib/services/ios-project-service.ts
@@ -653,8 +653,13 @@ We will now place an empty obsolete compatability white screen LauncScreen.xib f
 				this.platformData.configurationFileName
 			);
 	}
+
 	public ensureConfigurationFileInAppResources(): void {
 		return null;
+	}
+
+	public cleanProject(projectRoot: string, options: string[]): IFuture<void> {
+		return Future.fromResult();
 	}
 
 	private mergeInfoPlists(): IFuture<void> {
--- a/lib/services/platform-service.ts
+++ b/lib/services/platform-service.ts
@@ -217,6 +217,16 @@ export class PlatformService implements IPlatformService {
 			this.ensurePlatformInstalled(platform).wait();
 			let changesInfo = this.$projectChangesService.checkForChanges(platform);
 			if (changesInfo.hasChanges) {
+				// android build artifacts need to be cleaned up when switching from release to debug builds
+				if (platform.toLowerCase() === "android") {
+					let previousPrepareInfo = this.$projectChangesService.getPrepareInfo(platform);
+					// clean up prepared plugins when not building for release
+					if (previousPrepareInfo && previousPrepareInfo.release !== this.$options.release) {
+						let platformData = this.$platformsData.getPlatformData(platform);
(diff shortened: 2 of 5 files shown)


lib/services/ios-project-service.ts
```diff
@@ -657,6 +657,10 @@ We will now place an empty obsolete compatability white screen LauncScreen.xib f
 		return null;
 	}
 
+	public cleanProject(projectRoot: string, options: string[]): IFuture<void> {
+		return null;
```
rosen-vladimirov: It's better to use:
```TypeScript
return Future.fromResult();
```

This way the code will not fail in case the method is called (currently in case you call `cleanProject(..).wait()` it will fail that it cannot read property wait fo null.

lib/services/platform-service.ts
```diff
@@ -217,6 +217,16 @@ export class PlatformService implements IPlatformService {
 			this.ensurePlatformInstalled(platform).wait();
 			let changesInfo = this.$projectChangesService.checkForChanges(platform);
 			if (changesInfo.hasChanges) {
+				// android build artifacts need to be cleaned up when switching from release to debug builds
+				if (platform.toLowerCase() === "android") {
```
rosen-vladimirov: As the whole logic is Android specific, instead of using this if, you can move the whole code inside AndroidProjectService's cleanProject method. Here you can just call:
```TypeScript
platformData.platformProjectService.cleanProject(platformData.projectRoot, []).wait();
```

lib/services/platform-service.ts
```diff
@@ -217,6 +217,16 @@ export class PlatformService implements IPlatformService {
...
+					let previousPrepareInfo = this.$projectChangesService.getPrepareInfo(platform);
+					// clean up prepared plugins when not building for release
+					if (previousPrepareInfo && previousPrepareInfo.release && !this.$options.release) {
```
rosen-vladimirov: Isn't it good idea to call the clean method in case the previous build was for debug config and the current one is for release? For example:
```TypeScript
if (previousPrepareInfo && previousPrepareInfo.release !== this.$options.release) {
```

Pip3r4o: Good point! Build artifacts should be cleaned when changing the build type. Don't know why I didn't leave it that way.

lib/services/platform-service.ts
Pip3r4o: Thought about that earlier, but I don't think `cleanProject` should depend on conditions specific to the prepare phase, and the *android* test prevents unnecessary prepareInfo checks when preparing an ios project.

rosen-vladimirov:
👍  After green build, of course ;)

Issue #2460, opened 2017-01-30
CLI does not process all native frameworks


dtopuzov:
Steps to reproduce:

```
npm i -g nativescript@next
git clone https://github.com/EddyVerbruggen/nativescript-barcodescanner
cd nativescript-barcodescanner
tns plugin remove nativescript-barcodescanner --path demo
tns plugin add nativescript-barcodescanner --path demo
tns build ios --path demo
```

Result:
At least one native framework (usually Widgets) is missing in final package:
`platforms/ios/build/emulator/demo.app/Frameworks/`

Notes: CLI 2.5.0 is ok, the problem is valid only for nativescript@next and it is introduced with https://github.com/NativeScript/nativescript-cli/pull/2411

PR #2465, opened 2017-01-31
fix for  #2460

Plamen5kov:
Problem:
See more in this issue.
Because of the way promises are executed, a new project was created twice and was overridden twice with only one framework file being added each time. The proper implementation reads the previously changed .proj file and augments it with the new .framework file.

Fix for:  https://github.com/NativeScript/nativescript-cli/issues/2460

### DIFF
--- a/lib/services/plugins-service.ts
+++ b/lib/services/plugins-service.ts
@@ -110,7 +110,7 @@ export class PluginsService implements IPluginsService {
 		let appFolderExists = this.$fs.exists(path.join(platformData.appDestinationDirectoryPath, constants.APP_FOLDER_NAME));
 		if (appFolderExists) {
 			this.preparePluginScripts(pluginData, platform);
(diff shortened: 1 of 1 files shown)


lib/services/plugins-service.ts
```diff
@@ -109,8 +109,8 @@ export class PluginsService implements IPluginsService {
 
 		let appFolderExists = this.$fs.exists(path.join(platformData.appDestinationDirectoryPath, constants.APP_FOLDER_NAME));
 		if (appFolderExists) {
-			this.preparePluginScripts(pluginData, platform);
-			this.preparePluginNativeCode(pluginData, platform);
+			await this.preparePluginScripts(pluginData, platform);
```
rosen-vladimirov: `preparePluginScripts` returns void, not Promise, so no need to use await here.

PR #2483, opened 2017-02-03
Expose createProject to public API

rosen-vladimirov:
In order to allow creating a new NativeScript project when {N} CLI is required as a library, we should add createProject method to public API. This requires some modifications in method's implementation - we cannot rely on `$options` as it is not populated correctly when the CLI is used as a library. Pass object describing the project that we have to create. Fix unit tests accordingly and uncomment (and fix) all tests in project-service.

Add unit tests that will check if a method is really exposed in public API.

Add PublicAPI.md file, which describes all methods that can be used when using CLI as a library.

### DIFF
--- a/lib/services/project-service.ts
+++ b/lib/services/project-service.ts
@@ -1,6 +1,7 @@
 import * as constants from "../constants";
 import * as path from "path";
 import * as shelljs from "shelljs";
+import { exportedPromise } from "../common/decorators";
 
 export class ProjectService implements IProjectService {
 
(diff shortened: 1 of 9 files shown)


lib/services/project-service.ts
```diff
@@ -11,22 +12,29 @@ export class ProjectService implements IProjectService {
 		private $projectDataService: IProjectDataService,
 		private $projectHelper: IProjectHelper,
 		private $projectNameService: IProjectNameService,
-		private $projectTemplatesService: IProjectTemplatesService,
-		private $options: IOptions) { }
+		private $projectTemplatesService: IProjectTemplatesService) { }
+
+	@exportedPromise("projectService")
+	public async createProject(projectOptions: { projectName: string, template?: string, appId?: string, pathToProject?: string, force?: boolean, ignoreScripts?: boolean }): Promise<void> {
```
Mitko-Kerezov: `public async createProject(projectOptions: IProjectSettings): Promise<void> {`

test/nativescript-cli-lib.ts
```diff
@@ -12,13 +12,34 @@ describe("nativescript-cli-lib", () => {
...
-	it("resolves publicly available module - deviceEmitter, when it is required", () => {
-		const pathToEntryPoint = path.join(__dirname, "..", "lib", "nativescript-cli-lib.js").replace(/\\/g, "\\\\");
-		// HACK: If we try to require the entry point directly, the below code will fail as mocha requires all test files before starting the tests.
-		// When the files are required, $injector.register adds each dependency to $injector's cache.
-		// For example $injector.register("errors", Errors) will add the errors module with its resolver (Errors) to $injector's cache.
-		// Calling $injector.require("errors", <path to errors file>), that's executed in our bootstrap, will fail, as the module errors is already in the cache.
-		// In order to workaround this problem, start new process and assert there. This way all files will not be required in it and $injector.require(...) will work correctly.
-		childProcess.execSync(`"${process.execPath}" ${nodeArgs.join(" ")} -e "var assert = require('chai').assert; var result = require('${pathToEntryPoint}'); assert.ok(result.deviceEmitter);"`);
+	const publicApi: any = {
+		"deviceEmitter": null,
```
Mitko-Kerezov: You don't need the quotation marks around the keys of this object

test/project-commands.ts
```diff
@@ -10,8 +10,8 @@ let isProjectCreated: boolean;
 let dummyArgs = ["dummyArgsString"];
 
 class ProjectServiceMock implements IProjectService {
-	async createProject(projectName: string, selectedTemplate?: string): Promise<void> {
-		selectedTemplateName = selectedTemplate;
+	async createProject(projectOptions: { projectName: string, template?: string, appId?: string, pathToProject?: string }): Promise<void> {
```
Mitko-Kerezov: `projectOptions: IProjectSettings`

test/project-service.ts
```diff
@@ -70,11 +69,14 @@ class ProjectIntegrationTest {
 		let tnsCoreModulesRecord = packageJsonContent["dependencies"][constants.TNS_CORE_MODULES_NAME];
 		assert.isTrue(tnsCoreModulesRecord !== null);
 
-		let sourceDir = projectSourceDirectory || options.template;
+		let sourceDir = projectSourceDirectory;
 
-		let expectedFiles = fs.enumerateFilesInDirectorySync(sourceDir);
+		// Hidden files (starting with dots ".") are not copied.
+		let expectedFiles = fs.enumerateFilesInDirectorySync(sourceDir, (file, stat) => { return stat.isDirectory() || !_.startsWith(path.basename(file), "."); });
```
Mitko-Kerezov: You can optionally use the short fat arrow syntax here

PR #2484, opened 2017-02-03
node_modules is not watched by default

Plamen5kov:
If you want to watch for changes in node_modules or package.json a --sync-all-files or --syncAllFiles flag needs to be passed. By default only the app/ folder is being watched during any livesync.

Furthermore even with the --syncAllFiles flag enabled only production dependencies will be watched and livesynced.


Plamen5kov:
run CI

Issue #2478, opened 2017-02-02
Add tracking for Angular projects


teobugslayer:
It would be great to track the adoption level of NG2 and pure JS in NativeScript.

We are interested in two metrics:

 - [x] How much projects are created - i.e. tracking `tns create --ng`
 - [x] How much NG is used actively. In this case, the chosen performance metrics is this: when the user builds a project, check whether the project uses NG2.

These metrics should be tracked separately.

PR #2492, opened 2017-02-06
Add tracking for the project types that users are creating and working with

rosen-vladimirov:
## Track from which template a project is created
Add tracking from which template a project is created. This will give us better picture of the usage of CLI and the types of projects that the users are creating.

## Track project type when deploy/run/livesync is executed  …
Track the project type (Angular, TypeScript, JavaScript) when any of the commands is executed:
* prepare
* deploy
* run
* livesync

This will allow us to better understand the type of projects that the users are building.

Implements https://github.com/NativeScript/nativescript-cli/issues/2478

### DIFF
--- a/lib/services/platform-service.ts
+++ b/lib/services/platform-service.ts
@@ -12,6 +12,7 @@ let clui = require("clui");
 const buildInfoFileName = ".nsbuildinfo";
 
 export class PlatformService implements IPlatformService {
+	private _trackedProjectFilePath: string = null;
 
 	constructor(private $devicesService: Mobile.IDevicesService,
 		private $errors: IErrors,
@@ -38,7 +39,8 @@ export class PlatformService implements IPlatformService {
 		private $deviceAppDataFactory: Mobile.IDeviceAppDataFactory,
 		private $projectChangesService: IProjectChangesService,
 		private $emulatorPlatformService: IEmulatorPlatformService,
-		private $childProcess: IChildProcess) { }
+		private $childProcess: IChildProcess,
+		private $analyticsService: IAnalyticsService) { }
 
 	public addPlatforms(platforms: string[]): IFuture<void> {
 		return (() => {
@@ -196,6 +198,8 @@ export class PlatformService implements IPlatformService {
 		return (() => {
 			this.validatePlatform(platform);
 
+			this.trackProjectType().wait();
+
 			//We need dev-dependencies here, so before-prepare hooks will be executed correctly.
 			try {
 				this.$pluginsService.ensureAllDependenciesAreInstalled().wait();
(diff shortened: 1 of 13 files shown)


lib/services/project-templates-service.ts
```diff
@@ -6,31 +6,32 @@ temp.track();
...
-					// Use the original template name, specified by user as it may be case-sensitive.
-					realTemplatePath = this.prepareNativeScriptTemplate(name, version, projectDir).wait();
-				}
-			} else {
-				realTemplatePath = this.prepareNativeScriptTemplate(constants.RESERVED_TEMPLATE_NAMES["default"], null/*version*/, projectDir).wait();
+					name = data[0];
+
+				version = data[1];
+
+				templateName = constants.RESERVED_TEMPLATE_NAMES[name.toLowerCase()] || name;
```
Plamen5kov: Missing case when template is case sensitive folder.
If the developer uses `App` as `originalTemplateName` and then uses `app` these can be two different apps. I'm OK with us not supporting this scenario, and either way there aren't such tests, but thought you should know.

test/project-service.ts
```diff
@@ -153,7 +154,7 @@ describe("Project Service Tests", () => {
 				"readme": "dummy",
 				"repository": "dummy"
 			});
-			npmInstallationManager.install("tns-template-hello-world", defaultTemplateDir, {dependencyType: "save"}).wait();
+			npmInstallationManager.install("tns-template-hello-world", defaultTemplateDir, { dependencyType: "save" }).wait();
```
Plamen5kov: can use `constants.RESERVED_TEMPLATE_NAMES["default"]` instead of `"tns-template-hello-world"`

lib/services/project-templates-service.ts
rosen-vladimirov: In fact the case is handled - when `originalTemplateName` is `app`, the `name` variable will be set to `app`.
At this point `templateName` will be set to `app`.

At another point, when user passes `--template App`, `originalTemplateName` will be set to `App`, the `name` variable will be set to `App` and the `templateName` will be set to `App` again.

lib/services/platform-service.ts
```diff
@@ -372,9 +376,24 @@ export class PlatformService implements IPlatformService {
...
 	}
 
+	public trackProjectType(): IFuture<void> {
+		return (() => {
+			// Track each project once per process.
+			// In long living process, where we may work with multiple projects, we would like to track the information for each of them.
+			if (this.$projectData && (this.$projectData.projectFilePath !== this._trackedProjectFilePath)) {
+				this._trackedProjectFilePath = this.$projectData.projectFilePath;
+
+				this.$analyticsService.track("Working with project type", this.$projectData.projectType);
```
rosen-vladimirov: missing `.wait()` here

PR #2495, opened 2017-02-06
Remove ios native dependencies

TsvetanMilanov:
We need to remove our native dependencies - ref and ffi because they are casuing problems with each version of NodeJS which has new version of V8 in it. We are using ref and ffi to communicate with iTunes and execute fs and application management operations on iOS devices. This operations are moved in new dependency - device-lib. The device lib has JavaScript wrapper which we use in the CLI. Everything in the ios device file system and ios device application manager is replaced with the exposed methods from the device lib.

Common lib reference: https://github.com/telerik/mobile-cli-lib/pull/889


lib/device-sockets/ios/socket-request-executor.ts
```diff
@@ -31,37 +27,38 @@ export class IOSSocketRequestExecutor implements IiOSSocketRequestExecutor {
...
+			// before we send the PostNotification.
+			const promisesToWait = [this.$iOSNotificationService.awaitNotification(deviceIdentifier, this.$iOSNotification.appLaunching, timeout),
+			this.$iOSNotificationService.awaitNotification(deviceIdentifier, this.$iOSNotification.readyForAttach, readyForAttachTimeout)];
+			if (shouldBreak) {
+				await this.$iOSNotificationService.postNotificationAndAttachForData(deviceIdentifier, this.$iOSNotification.waitForDebug);
+			}
 
-			await this.$iOSNotificationService.awaitNotification(npc, this.$iOSNotification.readyForAttach, readyForAttachTimeout);
+			await this.$iOSNotificationService.postNotificationAndAttachForData(deviceIdentifier, this.$iOSNotification.attachRequest);
+			await Promise.all(promisesToWait);
```
rosen-vladimirov: hm, is `Promise.all` the correct method here. The order before was:
- appLaunching
- readyForAttach
- waitForDebug (optional)
- attachRequest

Now, with current code, it is:
- appLaunching, readyForAttach (simultaneously)
- waitForDebug
- attachRequest

Or even in some cases:
- waitForDebug
- attachRequest
- appLaunching, readyForAttach (simultaneously)

Can you elaborate why is the change of the order?

Issue #2496, opened 2017-02-06
TNS platform clean android 


NathanaelA:
I ran into a corner case; but one that is annoying if you aren't paying attention.

If you are building a product using a prior version of the runtime; lets say 2.4.1 and you do a 
`tns platform clean android` you will end up with 2.5.0 of the runtime.    When I do a "clean" I expect to maintain the SAME version of the runtimes; I don't expect to have part of my project upgraded...

Plamen5kov:
@NathanaelA you're absolutely right, I'll fix it as soon as possible.

PR #2502, opened 2017-02-07
don't add platform after tns platform clean is ran

Plamen5kov:
see more info [here](https://github.com/NativeScript/nativescript-cli/issues/2496).
In short when `tns platform clean <platform>` is ran, the command will now only clean the platform, instead of removing it and then adding the latest platform package.

### DIFF
--- a/lib/commands/platform-clean.ts
+++ b/lib/commands/platform-clean.ts
@@ -6,7 +6,6 @@ export class CleanCommand implements ICommand {
 
 	public async execute(args: string[]): Promise<void> {
 		this.$platformService.removePlatforms(args);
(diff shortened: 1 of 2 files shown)


Plamen5kov:
@rosen-vladimirov  I'll update the documentation, but we don't need to add a platform. The platform will remain the same in the package.json so any command will install the correct platform. No need for clean to install the platform.

rosen-vladimirov:
I believe we no longer need these changes.

Plamen5kov:
Closing because of: https://github.com/NativeScript/nativescript-cli/pull/2615

Issue #2476, opened 2017-02-02
tns run ios issue with 2.5.0 - unable to sync files on iOS Simulator


vchimev:
Output:
```
Unable to sync files. Error is: cp: copyFileSync: could not write to dest file (code=ENOENT):/Users/toddanglin/Library/Developer/CoreSimulator/Devices/50DFBA8C-A077-4EB4-B49F-93D2FAE13810/data/Containers/Bundle/Application/6878EDAF-504B-4366-9CE4-B5A97870EA3C/ns-demo.app/Resources/en.lproj/Localizable.strings
```

Seems to fail with the same error on different files in “/Resources/en.lproj”. Here it’s “Localizable.strings.” Sometimes it’s “InfoPlist.strings”.

Plamen5kov:
How to reproduce: 
Add folder to `<project_name>/app` folder once an application is created with any template.

JuanDelgadillo:
I got the same problem over here. I'm using the plugin `nativescript-i18n` and when I make a first change after run `tns run ios` it refresh the app but with next changes it throws that error and not reload the app anymore. I'm using tns 2.5.0

toddanglin:
+1 I've go the same situation as @JuanDelgadillo. Building on Mac for iOS, targeting the simulator.

`tns run ios --emualtor` will work the first time and do one LiveSync. After that, it errors-out with the `copyFileSync` errors. Current workaround is to simply kill the running command and restart `tns run`.

ludwiktrammer:
I'm experiencing the problem while using `nativescript-dev-sass`. I'm able to run the app using "tns run android", but as soon as I touch a single .scss file NativeScript tries to sync and everything blow ups. Here is the relevant issue in `nativescript-dev-sass` bugtracker: https://github.com/toddanglin/nativescript-dev-sass/issues/18

vance:
I still get this in `tns --version 2.5.4`. The first time is fine, but on 2nd run, my assets directory fails every time since I added images to it. 

```
[1] Transferring project files...
[1] Unable to sync files. Error is: EEXIST: file already exists, mkdir '... /assets '
```

lukeramsden:
I still get this aswell.

pkoleva:
Hey, @vance @lukeramsden,
We've released 2.5.5 and 3.0.1 with another fix that references this problem. Would you like to give it a try and write back with the results. Meanwhile I'll reopen the issue.

dtopuzov:
[...]

PR #2508, opened 2017-02-08
Fix livesync/run issue

rosen-vladimirov:
### Fix livesync when directories are modified
In case you try adding/removing directories in your `app` dir, `chokidar` (the file system watcher that we are using now instead of `gaze`) raises `addDir`/`unlinkDir` events.
However we do not have handles for these events, so we do not execute anything. After that, when we add a file in the newly created dir, `chokidar` sends `add` event, we handle it and try to process the file.
This works fine for iOS and Android devices, but it does not work at all for iOS Simulator, as we have not transferred the new directory to `platforms` dir.
Add required handles for `addDir` and `unlinkDir` methods.

Also currently there's a problem when already existing directory is renamed. In this case its modified time (`mtime`) is not changed, so the projectChangesService disregards the change and doesn't transfer it to `platforms` dir. After that, in case you modify any file inside the renamed dir, you'll see ENOENT error. In order to fix this, check the time of last status change (`ctime`) of the directory/file.

### LiveSync only existing files during `--watch`
During `--watch`, when a change is detected, the project is prepared and after that we are trying to move the prepared file (from `platforms/<platform>/...` dir) to the device.
However some plugins modify the content of the `platforms` directory on afterPrepare. For example `nativescript-dev-sass` allows you to use `.scss` files, on `beforePrepare` they are "transpiled" to `.css` files.
After that, on `afterPrepare` the plugin itself removes the file from `platforms/<platform>/...` dir.
CLI detects the change in `.scss` file, prepares the project (which moves the `.scss` file to `platforms` dir) and after that tries to move it from `platforms` to the device. During the last step, the `.scss` file is already removed from `platforms` directory, so our code fails.
Meanwhile, the `beforePrepare` hook of the plugin has created/modified the `.css` file inside `app` dir. This will trigger new iteration, where the file will be sent to device.

In order to fix the error when the `.scss` file is modified, we'll execute livesync only if the modified files exist in `platforms` dir.

Fixes https://github.com/NativeScript/nativescript-cli/issues/2476

### DIFF
--- a/lib/services/livesync/livesync-service.ts
+++ b/lib/services/livesync/livesync-service.ts
@@ -138,6 +138,7 @@ class LiveSyncService implements ILiveSyncService {
 			fiberBootstrap.run(() => {
 				that.$dispatcher.dispatch(() => (() => {
 					try {
+						that.$logger.trace(`Event '${event}' triggered for path: '${filePath}'`);
 						filePath = path.join(syncWorkingDirectory, filePath);
 						for (let i = 0; i < onChangedActions.length; i++) {
 							onChangedActions[i](event, filePath, that.$dispatcher);
--- a/lib/services/livesync/platform-livesync-service.ts
+++ b/lib/services/livesync/platform-livesync-service.ts
@@ -47,8 +47,8 @@ export abstract class PlatformLiveSyncServiceBase implements IPlatformLiveSyncSe
 					}
 
 					if (postAction) {
-					 	this.finishLivesync(deviceAppData).wait();
-					 	return postAction(deviceAppData).wait();
+						this.finishLivesync(deviceAppData).wait();
+						return postAction(deviceAppData).wait();
 					}
 
 					this.refreshApplication(deviceAppData, localToDevicePaths, true).wait();
@@ -65,7 +65,7 @@ export abstract class PlatformLiveSyncServiceBase implements IPlatformLiveSyncSe
 			return;
 		}
 
-		if (event === "add" || event === "addDir"|| event === "change") {
+		if (event === "add" || event === "addDir" || event === "change") {
 			this.batchSync(filePath, dispatcher, afterFileSyncAction);
 		} else if (event === "unlink" || event === "unlinkDir") {
 			this.syncRemovedFile(filePath, afterFileSyncAction).wait();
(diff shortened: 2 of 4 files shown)


rosen-vladimirov:
run ci

dtopuzov:
It looks with those changes we will not handle cases when user add new files, which is important case for getting started. All tutorials start with: `tns run android`, modify this file, then add new file, write code, require new file in the old file and so on.

rosen-vladimirov:
@dtopuzov when user adds new file, it will be successfully synced to device. No changes in this behavior.
This PR addresses two problems:
1) We skip prepare when a new directory is added to app dir, which leads to ENOENT error (easily reproducible with iOS Simulator).
2) Some plugins modify the content of the files in the `platforms` dir and if I have to be precise, nativescript-dev-sass plugin removes some files from `platforms` directory. Meanwhile CLI tries to transfer the removed files from `platforms` dir to device. But they do not exist in `platforms` dir anymore (they still exist in `app` dir). Due to incorrect behavior in previous versions of NS CLI (2.4.x and earlier), the error was not visible to the end users. However we've fixed the incorrect behavior and now the error is shown. This PR addresses this by checking the files in `platforms` dir before trying to send them to device.
What happens before this PR when `nativescript-dev-sass` plugin is used:
1) tns run android - project is prepared and started on device. File system watcher is started.
2) modify .scss file in app dir
3) CLI's watcher detects this change and starts livesync logic. The logic itself executes the following steps:
3.1) execute beforePrepare hook - at this point `nativescript-dev-sass` plugin will convert .scss file to .css in app dir
3.2) execute prepare logic from CLI - this will move all changed files (.scss in this case) to platforms dir
3.3) execute afterPrepare hook - at this point  `nativescript-dev-sass` plugin deletes the .scss file from platforms dir.
3.4) Transfer modified file (.scss) from platforms dir to device. Before this PR, CLI fails at this point and shows warning to the user
So I've modified step 3.4 to check if the file (.scss in the example described here), really exists inside `platforms` dir.

Please note that step 3.1 will trigger additional iteration of the whole cycle, as the .css file is modified.

irman:
How can I get this fix? I see it's on release branch but how do I update it with npm? There's no `2.5.1` listed on the npm repo too.

image

Plamen5kov:
@irman, we haven't published it yet, but it should be published by the end of the week.

irman:
@Plamen5kov great to hear that. Can't wait. 😄 

Thanks!

vincentpalita:
@Plamen5kov 

is the package already available on npm? It seems to but I still having the same issue with nativescript-dev-sass.

Thanks for your help

Plamen5kov:
@vincentpalita yes, 2.5.1 cli is published, and if you do a `tns --version` you should see printed out version 2.5.1. Please check you have the latest CLI installed, and if the problem still persists, we'll continue investigating the problem.

vincentpalita:
@Plamen5kov,

Ok I had to remove manually the version then install it back and it is now ... working :)

That's great!
Thank you so much

PR #2527, opened 2017-02-12
Pass current stdio to npm install process

### LINKED ISSUE #2521: How to request user input during plugin installation?
Hi,

Thanks!

rosen-vladimirov:
Whenever CLI spawns `npm install <smth>`, we do not set stdin and stdout
of the spawned process. This way in case any of the npm scripts executed
during install requires input from user, it cannot be handled in any
way.

So pass the current stdin and stdout to the spawned npm process. This
way in case a postinstall script is executed and it requires input from
user, the users will be able to enter required information in the same
place where `tns` command has been executed (same terminal).

This fixes postinstall execution of `nativescript-plugin-firebase`. Fixes #2521

### DIFF
--- a/lib/node-package-manager.ts
+++ b/lib/node-package-manager.ts
@@ -45,7 +45,7 @@ export class NodePackageManager implements INodePackageManager {
 				}
 			}
 			try {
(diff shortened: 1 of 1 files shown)


EddyVerbruggen:
THANKS!

Issue #2487, opened 2017-02-03
Error: Exception: The plugin tns-android@2.5.0 is already installed


grasshoppermouse:
Just updated to 2.5. I can add the ios platform and run on the emulator, but I can't add the android platform (OS X 10.12.2).

UPDATE: when I create a new test project, `tns run android` works. But I still can't add the android platform to my existing project. Pointers appreciated.

```
van-aimc-l2dnmp:aapa-conference-app hagen$ node --version
v7.5.0
van-aimc-l2dnmp:aapa-conference-app hagen$ npm --version
4.1.2
van-aimc-l2dnmp:aapa-conference-app hagen$ tns --version
2.5.0
van-aimc-l2dnmp:aapa-conference-app hagen$ tns doctor
Verifying CocoaPods. This may take more than a minute, please be patient.
  ◝ Installing iOS runtime.nativescript-check-cocoapods@0.0.1 /private/var/folders/6p/vb6bmpxd65v_hpd91kzj3jhh0000gq/T/nativescript-check-cocoapods11713-24129-lbxqsa.djp4vrhpvi
└── tns-ios@2.5.0 


  ◜ Verifying CocoaPods. This may take some time, please be patient..
Your components are up-to-date.

No issues were detected.
van-aimc-l2dnmp:aapa-conference-app hagen$ npm uninstall tns-android
- tns-android@2.5.0 node_modules/tns-android
van-aimc-l2dnmp:aapa-conference-app hagen$ npm cache clean
van-aimc-l2dnmp:aapa-conference-app hagen$ tns platform add android --log trace
execFile: /usr/local/opt/android-sdk/platform-tools/adb "help"
execFile: /usr/local/opt/android-sdk/platform-tools/adb "help"
spawn: /usr/local/opt/android-sdk/platform-tools/adb "--help"
Socket {
  connecting: false,
  _hadError: false,
  _handle: 
   Pipe {
     bytesRead: 0,
     _externalStream: {},
     fd: 15,
     writeQueueSize: 0,
     owner: [Circular],
     onread: [Function: onread],
     reading: true },
  _parent: null,
  _host: null,
  _readableState: 
   ReadableState {
     objectMode: false,
     highWaterMark: 16384,
     buffer: BufferList { head: null, tail: null, length: 0 },
     length: 0,
     pipes: null,
     pipesCount: 0,
     flowing: null,
     ended: false,
     endEmitted: false,
     reading: true,
     sync: false,
     needReadable: true,
[...]

PR #2531, opened 2017-02-13
Rollback package.json if plugin installation fails

Pip3r4o:
Implement rollback of package.json contents of project if npm installation fails for whatever reason during installation of a NS plugin or platform.

Addresses #2487

### DIFF
--- a/lib/node-package-manager.ts
+++ b/lib/node-package-manager.ts
@@ -17,7 +17,8 @@ export class NodePackageManager implements INodePackageManager {
 			config["ignore-scripts"] = true;
 		}
 
-		let jsonContentBefore = this.$fs.readJson(path.join(pathToSave, "package.json"));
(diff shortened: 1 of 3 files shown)


Pip3r4o:
@Plamen5kov I'd be unable to do that in the node-package-manager service as I don't want to pollute it with additional parameters (package json paths). So I might have to do it manually for every place where npm.install is called.

Plamen5kov:
@Pip3r4o https://github.com/NativeScript/nativescript-cli/blob/97cf288e34db4e4adc862112d9744d7412146302/lib/node-package-manager.ts#L20 the path and the content already exist in node-package-manager, so there won't be any polution. We also need a mechanism to handle ctrl + C signal in various places in cli, to revert the project state. @rosen-vladimirov what's your oppinion?

PR #2532, opened 2017-02-13
tns-android/ios are installed as dev dependencies

Plamen5kov:
Previously `tns-android` and `tns-ios` have been installed in nativescript key in the root `package.json`, which is illogical since these platforms are just regular npm packages.
With this PR, both platforms are installed as `devDependencies` and the `package.json` looks like:
```
{
  "nativescript": {
    "id": "org.nativescript.<app_name>"
  },
  "dependencies": {
    ....
  },
  "devDependencies": {
    "tns-android": "^2.5.0"
  }
}
```
This change does not introduce any breaking changes, but it's a first step towards future improvements.

### DIFF
--- a/lib/services/platform-service.ts
+++ b/lib/services/platform-service.ts
@@ -82,7 +82,7 @@ export class PlatformService implements IPlatformService {
 		let packageToInstall = "";
 		let npmOptions: IStringDictionary = {
 			pathToSave: path.join(this.$projectData.platformsDir, platform),
-			dependencyType: "save"
+			dependencyType: "save-dev"
 		};
 
 		if (!this.$options.frameworkPath) {
@@ -97,8 +97,7 @@ export class PlatformService implements IPlatformService {
 			let frameworkDir = path.join(downloadedPackagePath, constants.PROJECT_FRAMEWORK_FOLDER_NAME);
 			frameworkDir = path.resolve(frameworkDir);
 
-			let coreModuleName = await this.addPlatformCore(platformData, frameworkDir);
-			await this.$npm.uninstall(coreModuleName, { save: true }, this.$projectData.projectDir);
+			await this.addPlatformCore(platformData, frameworkDir);
 		} catch (err) {
 			this.$fs.deleteDirectory(platformPath);
 			throw err;
@@ -129,17 +128,15 @@ export class PlatformService implements IPlatformService {
 		if (customTemplateOptions) {
 			frameworkPackageNameData.template = customTemplateOptions.selectedTemplate;
 		}
-		this.$projectDataService.setValue(platformData.frameworkPackageName, frameworkPackageNameData);
 
 		return coreModuleName;
-
 	}
 
 	private async getPathToPlatformTemplate(selectedTemplate: string, frameworkPackageName: string): Promise<any> {
 		if (!selectedTemplate) {
 			// read data from package.json's nativescript key
 			// check the nativescript.tns-<platform>.template value
-			let nativescriptPlatformData = this.$projectDataService.getValue(frameworkPackageName);
+			let nativescriptPlatformData = this.$projectDataService.getValue(frameworkPackageName, constants.DEV_DEPENDENCIES);
 			selectedTemplate = nativescriptPlatformData && nativescriptPlatformData.template;
 		}
 
@@ -577,6 +574,7 @@ export class PlatformService implements IPlatformService {
 			let platformDir = path.join(this.$projectData.platformsDir, platform);
 			this.$fs.deleteDirectory(platformDir);
 			this.$projectDataService.removeProperty(platformData.frameworkPackageName);
+			this.$npm.uninstall(platformData.frameworkPackageName, {"save-dev": true});
 
 			this.$logger.out(`Platform ${platform} successfully removed.`);
 		});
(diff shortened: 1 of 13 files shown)


lib/services/platform-service.ts
```diff
@@ -82,7 +82,7 @@ export class PlatformService implements IPlatformService {
 		let packageToInstall = "";
 		let npmOptions: IStringDictionary = {
```
rosen-vladimirov: can we add `--save-exact` to `npmOptions`, as currently we are adding the framework package with `^`, for example:
```
"tns-android": "^2.5.0"
```
This means that once we have `tns-android` 2.6.0, and I execute `rm -rf node_modules && npm i` I'll get the new framework.

lib/tools/node-modules/node-modules-builder.ts
```diff
@@ -126,6 +126,14 @@ export class NodeModulesBuilder implements INodeModulesBuilder {
 		let dependenciesBuilder = this.$injector.resolve(NodeModulesDependenciesBuilder, {});
 		let productionDependencies = dependenciesBuilder.getProductionDependencies(this.$projectData.projectDir);
 
+		let prodDependenciesArr = [];
```
rosen-vladimirov: The `for` below can be replaced with `map`:
```TypeScript
const prodDependenciesArr = _.map(productionDependencies, prodDep => prodDep.name);
```

lib/tools/node-modules/node-modules-builder.ts
```diff
@@ -126,6 +126,14 @@ export class NodeModulesBuilder implements INodeModulesBuilder {
...
+		for (let i in productionDependencies) {
+			prodDependenciesArr.push(productionDependencies[i].name);
+		}
+
+		let prodDependenciesFilePath = path.join(this.$projectData.projectDir, "platforms", "android", "productionDependencies.json");
+		this.$fs.writeJson(prodDependenciesFilePath, prodDependenciesArr);
```
rosen-vladimirov: what's the purpose of this file? Why should we write it to platforms dir?

rosen-vladimirov:
Another thing came to my mind, that could be a showstopper at the moment - `tns-ios` package can be installed only on macOS. So in case I work with my project on macOS and test the iOS version, I'll end up with `tns-ios` in the devDependencies. After that, anyone else will be unable to use the same project on other OS, as the `npm install` command will fail.
Another `problem` that I see (but maybe it's not such an issue), is that if you have both runtimes in devDependencies and you would like to build your project for Android for example, you'll have both runtimes in `node_modules` (as `npm install` will be called at the root fo the project). This will slow down the process a little bit, but I think we can live with it.

lib/tools/node-modules/node-modules-builder.ts
Plamen5kov: I'll add comments to the code, but the basic idea is, we need to tell gradle what are the production dependencies, that CLI found, so we search for dependencies only in production modules, it saves time, and it's less error prone.

rosen-vladimirov: Why gradle needs to know which are the production dependencies? As far as I remember, CLI copies only production dependencies to `platforms/android` dir.
Also... this code is executed for iOS builds as well.

Plamen5kov:
@rosen-vladimirov 
>After that, anyone else will be unable to use the same project on other OS, as the npm install command will fail.

Haven't thought about that. Maybe we can use `--force` as a workaround in this particular case, or add additional logic to be able to build for android at least if possible.

---
>After that I've tried tns prepare android, but CLI didn't respect my version and installed tns-android 2.5.0. It also modified my package.json back to: 2.5.0. Is this expected?

It's not expected, I'll look into it.

---
>As tns-android is written in package.json, we can remove magic dependencies installed by CLI when adding android platform: https://github.com/NativeScript/nativescript-cli/blob/master/lib/services/android-project-service.ts#L14-L19 - they can be added as dependencies of tns-android package. This way CLI will not take care of installing them.

I agree, it would be a great thing to add template dependencies to the actual template. Will do it, for sure.

---
>Related to the point above - how to determine the version of Android/iOS that I have in platforms dir? As now I have tns-android in my package.json, but I've modified it's version, how can I determine in the package installed in node_modules is the same as the one I have in my platforms dir.

I believe we should always check the version of the platform from the `package.json` in `node_modules/tns-android(ios)` so on that point, on a change in the `tns-android(ios)` packages we can do a platform remove -> add so we update/downgrade to what the user has specified in the `package.json`. This way the project will stay consistent: `package.json` version == `node_modules/tns-android(ios)` version == `platforms/android(ios)` version.

rosen-vladimirov:
@Plamen5kov 
> After that, anyone else will be unable to use the same project on other OS, as the npm install command will fail.
> Haven't thought about that. Maybe we can use --force as a workaround in this particular case, or add additional logic to be able to build for android at least if possible.

The problem is that `npm install` command will fail in this case. So l suggest enabling installation of `tns-ios` on Windows (modification of the `tns-ios` package).

lib/tools/node-modules/node-modules-builder.ts
Plamen5kov: Because gradle needs to go through the `node_modules` folder and get all `.aar` so it can include them into the build. Currently we traverse all of the `node_modules` which is error prone and it's a waste of time, when we know which dependencies we should use.

dtopuzov:
run ci

rosen-vladimirov:
@Plamen5kov if this PR won't make it for 3.0.0, maybe we should close it, what do you think?

Issue #2528, opened 2017-02-13
`platforms\android` locked after `tns run`


dtopuzov:
Since Android runtime now use gradle daemon {N} is almost unusable on Windows because gradle deamon locks platform/android folder and user can not cleanup the project.

```
npm install -g nativescript@next
tns create TestApp
tns platform add android@next
tns run android
```
Then try to delete `TestApp` folder.

CLI: @next
Android Runtime: @next

dtopuzov:
In order to unlock folders on Windows end users should perform `tns platform remove android`.

PR #2533, opened 2017-02-13
stop gradle daemons when removing platform

Pip3r4o:
The android runtime (3.0) has had its gradle wrapper version updated where daemon processes are used by default.

Daemons are used to prevent the overhead of having to initialize the JVM on every build command, thus lowering the build time dramatically. The fact that the process remains alive appears to be a problem in Windows environments as the daemon locks the `platforms/android` directory.

The changes in this PR ensure that the currently running daemons will be terminated, and the lock over the platforms directory released, so that it may be freely removed.

Addresses #2528

### DIFF
--- a/lib/services/platform-service.ts
+++ b/lib/services/platform-service.ts
@@ -570,19 +570,21 @@ export class PlatformService implements IPlatformService {
 		this.$logger.info(`Copied file '${packageFile}' to '${targetPath}'.`);
 	}
 
-	public removePlatforms(platforms: string[]): void {
+	public async removePlatforms(platforms: string[]): Promise<void> {
 		this.$projectDataService.initialize(this.$projectData.projectDir);
 
-		_.each(platforms, platform => {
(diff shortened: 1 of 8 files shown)


lib/services/platform-service.ts
```diff
@@ -570,19 +571,31 @@ export class PlatformService implements IPlatformService {
...
 			let platformDir = path.join(this.$projectData.platformsDir, platform);
 			this.$fs.deleteDirectory(platformDir);
 			this.$projectDataService.removeProperty(platformData.frameworkPackageName);
 
 			this.$logger.out(`Platform ${platform} successfully removed.`);
-		});
+		}
+	}
+
+	private async gradleStop(projectRoot: string): Promise<ISpawnResult> {
```
Plamen5kov: looks great, just move logic to `android-project-service.ts`, and call through `platformData.platformProjectService...`

PR #2535, opened 2017-02-14
removed --copy-from from help

Plamen5kov:
--copy-from option is deleted in 2.5 release, so we should remove it from the docs


rosen-vladimirov:
`--copy-from` option has been removed in master branch after 2.5 release, so it's still available in 2.5.x
This PR should target master branch

PR #2545, opened 2017-02-15
Write variables to .bash_profile

dtopuzov:
Write ANDROID_HOME and JAVA_HOME variables in `.bash_profile` instead of `.profile`


Plamen5kov:
@dtopuzov is this related to some issue?

PR #2573, opened 2017-02-24
Fix creating multiple projects in a single process

rosen-vladimirov:
In case the CLI is `required` as a library (used in a long living process, not as a command-line interface), trying to create a project multiple times leads to errors.
The problem is in the `projectDataService` which caches the path to first created project file (package.json) and uses it for all consecutive operations.
In other cases, when CLI had to use the projectDataService, it always had to call `projectDataService.initialize` method.

Instead of relying on a magically called initialize method, modify the service methods to work with any project dir and execute the required operation for it.
This will fix the creation of multiple projects as all operations will work with currently passed project directory.
Also remove the code that creates a `package.json` file during `loadProjectFile` in projectDataService - it's required only when a project is created, so move it there.

Rename the methods getValue, setValue and removeProperty to getNSValue, setNSValue, removeNSProperty as they are intended to modify the nativescript key in package.json.
We can later expose getValue, setValue and removeProperty (they are currently private).

### DIFF
--- a/lib/services/project-data-service.ts
+++ b/lib/services/project-data-service.ts
@@ -1,75 +1,109 @@
 import * as path from "path";
-import * as assert from "assert";
+
+interface IProjectFileData {
+	projectData: any;
+	projectFilePath: string;
+}
 
 export class ProjectDataService implements IProjectDataService {
 	private static DEPENDENCIES_KEY_NAME = "dependencies";
 
-	private projectFilePath: string;
-	private projectData: IDictionary<any>;
-	private projectFileIndent: string;
-
 	constructor(private $fs: IFileSystem,
-		private $staticConfig: IStaticConfig) {
+		private $staticConfig: IStaticConfig,
+		private $logger: ILogger) {
 	}
 
-	public initialize(projectDir: string): void {
-		if (!this.projectFilePath) {
-			this.projectFilePath = path.join(projectDir, this.$staticConfig.PROJECT_FILE_NAME);
-		}
+	public getNSValue(projectDir: string, propertyName: string): any {
+		return this.getValue(projectDir, this.getNativeScriptPropertyName(propertyName));
 	}
 
-	public getValue(propertyName: string): any {
-		this.loadProjectFile();
-		return this.projectData ? this.projectData[this.$staticConfig.CLIENT_NAME_KEY_IN_PROJECT_FILE][propertyName] : null;
+	public setNSValue(projectDir: string, key: string, value: any): void {
+		this.setValue(projectDir, this.getNativeScriptPropertyName(key), value);
 	}
 
-	public setValue(key: string, value: any): void {
-		this.loadProjectFile();
-		if (!this.projectData[this.$staticConfig.CLIENT_NAME_KEY_IN_PROJECT_FILE]) {
-			this.projectData[this.$staticConfig.CLIENT_NAME_KEY_IN_PROJECT_FILE] = Object.create(null);
-		}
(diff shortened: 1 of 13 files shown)


lib/services/project-data-service.ts
```diff
@@ -1,74 +1,124 @@
...
+		const projectFileContent = this.$fs.readText(projectFilePath);
+		const indent = projectFileContent ? this.detectIndent(projectFileContent) : "\t";
 
-		this.projectFileIndent = projectFileContent ? this.detectIndent(projectFileContent) : "\t";
+		const projectData = projectFileContent ? JSON.parse(projectFileContent) : Object.create(null);
 
-		this.projectData = projectFileContent ? JSON.parse(projectFileContent) : Object.create(null);
+		return {
+			projectData,
+			indent,
```
TsvetanMilanov: I think we don't need to detect the `indent`. We will detect it here https://github.com/telerik/mobile-cli-lib/blob/master/file-system.ts#L199

test/services/project-data-service.ts
```diff
@@ -0,0 +1,236 @@
...
+};
+
+describe("projectDataService", () => {
+	const generateJsonDataFromTestData = (currentTestData: any, skipNativeScriptKey?: boolean) => {
+		const props = currentTestData.propertyName.split(".");
+		let data: any = {};
+		let currentData: any = skipNativeScriptKey ? data : (data[CLIENT_NAME_KEY_IN_PROJECT_FILE] = {});
+
+		_.each(props, (prop, index: number) => {
+
```
TsvetanMilanov: Can you remove this blank line?

test/services/project-data-service.ts
```diff
@@ -0,0 +1,236 @@
...
+		const generateExpectedDataFromTestData = (currentTestData: any) => {
+			const props = currentTestData.propertyName.split(".");
+			props.splice(props.length - 1, 1);
+
+			let data: any = {};
+			let currentData: any = data[CLIENT_NAME_KEY_IN_PROJECT_FILE] = {};
+
+			_.each(props, (prop) => {
+
```
TsvetanMilanov: Can you remove this blank line?

test/services/project-data-service.ts
```diff
@@ -0,0 +1,236 @@
...
+				const projectDataService: IProjectDataService = testInjector.resolve("projectDataService");
+				projectDataService.removeNSProperty("projectDir", currentTestData.propertyName);
+
+				assert.deepEqual(dataPassedToWriteJson, generateExpectedDataFromTestData(currentTestData));
+				assert.isTrue(!!dataPassedToWriteJson[CLIENT_NAME_KEY_IN_PROJECT_FILE], "Data passed to write JSON must contain nativescript key.");
+			});
+
+		});
+
+		it("does not override already existing property of project data", () => {
```
TsvetanMilanov: Maybe the description should be `removes only the selected property`?

lib/services/project-data-service.ts
```diff
@@ -1,75 +1,109 @@
...
-		}
+	public getNSValue(projectDir: string, propertyName: string): any {
+		return this.getValue(projectDir, this.getNativeScriptPropertyName(propertyName));
 	}
 
-	public getValue(propertyName: string): any {
-		this.loadProjectFile();
-		return this.projectData ? this.projectData[this.$staticConfig.CLIENT_NAME_KEY_IN_PROJECT_FILE][propertyName] : null;
+	public setNSValue(projectDir: string, key: string, value: any): void {
+		return this.setValue(projectDir, this.getNativeScriptPropertyName(key), value);
```
Mitko-Kerezov: Gratuitous `return` statement here

PR #2574, opened 2017-02-24
Allow local builds when {N} CLI is required as a library

Mitko-Kerezov:
* Get rid of dependency in services
* Expose `build` method in `platformService` which enables building when {N} CLI is required

### DIFF
--- a/lib/commands/emulate.ts
+++ b/lib/commands/emulate.ts
@@ -1,17 +1,36 @@
 export class EmulateCommandBase {
-	constructor(private $platformService: IPlatformService) { }
+	constructor(private $options: IOptions,
+		private $projectData: IProjectData,
+		private $platformService: IPlatformService) {
+			this.$projectData.initializeProjectData();
+		}
 
 	public async executeCore(args: string[]): Promise<void> {
-		return this.$platformService.emulatePlatform(args[0]);
+		this.$options.emulator = true;
+		const appFilesUpdaterOptions: IAppFilesUpdaterOptions = { bundle: this.$options.bundle, release: this.$options.release };
+		const emulateOptions: IEmulatePlatformOptions = {
+			avd: this.$options.avd,
+			clean: this.$options.clean,
+			device: this.$options.device,
+			release: this.$options.release,
+			emulator: this.$options.emulator,
+			projectDir: this.$options.path,
+			justlaunch: this.$options.justlaunch,
+			availableDevices: this.$options.availableDevices,
+			platformTemplate: this.$options.platformTemplate
+		};
+		return this.$platformService.emulatePlatform(args[0], appFilesUpdaterOptions, emulateOptions, this.$projectData, this.$options.provision);
 	}
 }
 
 export class EmulateIosCommand extends EmulateCommandBase implements ICommand {
 	public allowedParameters: ICommandParameter[] = [];
 
-	constructor($platformService: IPlatformService,
+	constructor($options: IOptions,
+		$projectData: IProjectData,
+		$platformService: IPlatformService,
 		private $platformsData: IPlatformsData) {
-		super($platformService);
+		super($options, $projectData, $platformService);
 	}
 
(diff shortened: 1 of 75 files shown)


lib/services/platform-service.ts
```diff
@@ -351,25 +348,31 @@ export class PlatformService implements IPlatformService {
...
+			this._trackedProjectFilePath = projectData.projectFilePath;
 
-			await this.$analyticsService.track("Working with project type", this.$projectData.projectType);
+			await this.$analyticsService.track("Working with project type", projectData.projectType);
 		}
 	}
+	@exportedPromise("platformService")
+	public async build(platform: string, platformBuildOptions: IPlatformBuildData, platformTemplate?: string): Promise<void> {
+		this.$projectData.initializeProjectData(platformBuildOptions.projectDir);
+		await this.preparePlatform(platform, platformBuildOptions, platformTemplate, this.$projectData);
```
Mitko-Kerezov: Given that there are this many changes already I'd personally prefer a separate PR

lib/services/platform-service.ts
```diff
@@ -530,19 +532,19 @@ export class PlatformService implements IPlatformService {
...
+	public async cleanDestinationApp(platform: string, appFilesUpdaterOptions: IAppFilesUpdaterOptions, platformTemplate: string, projectData: IProjectData): Promise<void> {
+		await this.ensurePlatformInstalled(platform, platformTemplate, projectData);
 
-		const appSourceDirectoryPath = path.join(this.$projectData.projectDir, constants.APP_FOLDER_NAME);
-		let platformData = this.$platformsData.getPlatformData(platform);
+		const appSourceDirectoryPath = path.join(projectData.projectDir, constants.APP_FOLDER_NAME);
+		let platformData = this.$platformsData.getPlatformData(platform, projectData);
 		let appDestinationDirectoryPath = path.join(platformData.appDestinationDirectoryPath, constants.APP_FOLDER_NAME);
-		const appUpdater = new AppFilesUpdater(appSourceDirectoryPath, appDestinationDirectoryPath, this.$options, this.$fs);
+		const appUpdater = new AppFilesUpdater(appSourceDirectoryPath, appDestinationDirectoryPath, appFilesUpdaterOptions, this.$fs);
```
Mitko-Kerezov: I am a bit reluctant to change this in this PR as `AppFilesUpdater` class has a `options` dependency which isn't `$options` and more importantly should be resolved as `$options`.
If it is okay with you I'm willing to make a subsequent PR @rosen-vladimirov

lib/commands/emulate.ts
```diff
@@ -1,17 +1,36 @@
...
-	constructor(private $platformService: IPlatformService) { }
+	constructor(private $options: IOptions,
+		private $projectData: IProjectData,
+		private $platformService: IPlatformService) {
+			this.$projectData.initializeProjectData();
+		}
 
 	public async executeCore(args: string[]): Promise<void> {
-		return this.$platformService.emulatePlatform(args[0]);
+		this.$options.emulator = true;
```
Mitko-Kerezov: As I have not removed the `$options` dependency from **all** services I cannot guarantee that the code will run the same if I remove this.

lib/commands/emulate.ts
```diff
@@ -1,17 +1,36 @@
...
+		const appFilesUpdaterOptions: IAppFilesUpdaterOptions = { bundle: this.$options.bundle, release: this.$options.release };
+		const emulateOptions: IEmulatePlatformOptions = {
+			avd: this.$options.avd,
+			clean: this.$options.clean,
+			device: this.$options.device,
+			release: this.$options.release,
+			emulator: this.$options.emulator,
```
Mitko-Kerezov: I decided to reuse the variable 😇

Mitko-Kerezov:
Ping @rosen-vladimirov fixed comments and rebased

rosen-vladimirov:
run ci

PR #2579, opened 2017-02-28
Upgrade ws module

Mitko-Kerezov:
The newer version doesn't require `bufferutil` or `utf-8-validate` both of which have native V8 code which we are trying to get rid of.
Debugging still works with this new version - tested on three different OS X machines with `tns debug ios --chrome` command.

Ping @rosen-vladimirov @Plamen5kov


rosen-vladimirov:
run ci

PR #2594, opened 2017-03-09
Respect runtime version written in package json

Mitko-Kerezov:
Instead of always installing the latest runtime version, respect the one written in `package.json` `nativescript` key.

Ping @TsvetanMilanov @rosen-vladimirov @Plamen5kov

### DIFF
--- a/lib/services/platform-service.ts
+++ b/lib/services/platform-service.ts
@@ -67,6 +67,10 @@ export class PlatformService extends EventEmitter implements IPlatformService {
 		}
 
 		let platformData = this.$platformsData.getPlatformData(platform, projectData);
+		let currentPlatformData: any = this.$projectDataService.getNSValue(projectData.projectDir, platformData.frameworkPackageName);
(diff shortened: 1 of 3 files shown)


lib/services/platform-service.ts
```diff
@@ -67,6 +67,10 @@ export class PlatformService extends EventEmitter implements IPlatformService {
 		}
 
 		let platformData = this.$platformsData.getPlatformData(platform, projectData);
+		let currentPlatformData: any = this.$projectDataService.getNSValue(projectData.projectDir, platformData.frameworkPackageName);
```
TsvetanMilanov: `this.$projectDataService.getNSValue` can return `null` can you add check if it is null before you try to access `currentPlatformData[constants.VERSION_STRING]`

Mitko-Kerezov: Added a check on the line below, nice catch 🎣

Mitko-Kerezov:
Ping @TsvetanMilanov with fixed comments and @Plamen5kov with added tests

Plamen5kov:
:+1: after green build

PR #2595, opened 2017-03-09
take last parameter of id for packageName

Plamen5kov:
_problem_:
We needed id set in package.json to be exactly 3 components long.

_solution_
Take the last component of the id instead of the third one.

### DIFF
--- a/lib/services/android-project-service.ts
+++ b/lib/services/android-project-service.ts
@@ -214,7 +214,8 @@ export class AndroidProjectService extends projectServiceBaseLib.PlatformProject
 	private getProjectNameFromId(projectData: IProjectData): string {
 		let id: string;
 		if (projectData && projectData.projectId) {
-			id = projectData.projectId.split(".")[2];
+			let idParts = projectData.projectId.split(".");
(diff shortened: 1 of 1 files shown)


vchimev:
run ci

rosen-vladimirov:
I do not understand what's the problem here and what do we fix. Can you provide more details? Ping @Plamen5kov , @pkoleva

Plamen5kov:
@rosen-vladimirov to answer your question:
_For android:_
https://developer.android.com/studio/build/application-id.html
>It must have at least two segments (one or more dots).
Each segment must start with a letter.
All characters must be alphanumeric or an underscore [a-zA-Z0-9_].

_For iOS_
https://developer.apple.com/library/content/documentation/General/Conceptual/DevPedia-CocoaCore/AppID.html

>For an explicit App ID to match an app, the Team ID in the App ID must equal the Team ID associated with the app, and the bundle ID search string must equal the bundle ID for the app. The bundle ID is a unique identifier that identifies a single app and cannot be used by other teams.

There's no mention of length restrictions.

Issue #2566, opened 2017-02-23
`tns plugin search` does not work


dtopuzov:
c:\Git\nativescript-cli-tests>node_modules\.bin\tns plugin search nativescript camera --all
```
Searching npm please be patient........
Available NativeScript plugins:
┌────────┬───────────┬─────────────┐
│ Plugin │ Version   │ Description │
│ 0      │           │             │
│ 1      │           │             │
│ 10     │           │             │
│ 11     │           │             │
│ 12     │           │             │
│ 13     │           │             │
│ 14     │           │             │
│ 15     │           │             │
│ 16     │           │             │
│ 17     │           │             │
│ 18     │           │             │
│ 19     │           │             │
│ 2      │           │             │
│ 20     │           │             │
│ 21     │           │             │
│ 22     │           │             │
│ 23     │           │             │
│ 24     │           │             │
│ 25     │           │             │
│ 26     │           │             │
│ 27     │           │             │
│ 28     │           │             │
│ 29     │           │             │
│ 3      │           │             │
│ 30     │           │             │
│ 31     │           │             │
│ 32     │           │             │
│ 33     │           │             │
│ 34     │           │             │
│ 35     │           │             │
│ 36     │           │             │
│ 37     │           │             │
│ 38     │           │             │
│ 39     │           │             │
│ 4      │           │             │
│ 40     │           │             │
│ 41     │           │             │
│ 42     │           │             │
│ 43     │           │             │
│ 44     │           │             │
│ 45     │           │             │
│ 46     │           │             │
│ 47     │           │             │
│ 5      │           │             │
│ 6      │           │             │
│ 7      │           │             │
│ 8      │           │             │
│ 9      │           │             │
└────────┴───────────┴─────────────┘
```
Notes: I think we use `npm search` which has known issues with Node6

vchimev:
Command removed in the associated pull-request.

PR #2600, opened 2017-03-10
Remove the find plugin command. We don't see a good use case of its u…

etabakov:
…sage.


vchimev:
Related to https://github.com/NativeScript/nativescript-cli/issues/2566.

PR #2602, opened 2017-03-10
Add isvalid {N} project method

TsvetanMilanov:
We need a method in the public API to check if a directory contains valid NativeScript project.

### DIFF
--- a/lib/services/project-service.ts
+++ b/lib/services/project-service.ts
@@ -1,7 +1,7 @@
 import * as constants from "../constants";
 import * as path from "path";
 import * as shelljs from "shelljs";
-import { exportedPromise } from "../common/decorators";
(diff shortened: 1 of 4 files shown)


lib/services/project-service.ts
```diff
@@ -66,6 +67,17 @@ export class ProjectService implements IProjectService {
 		this.$logger.printMarkdown("Project `%s` was successfully created.", projectName);
 	}
 
+	@exported("projectService")
+	public isValidNativeScriptProject(pathToProject?: string): boolean {
```
rosen-vladimirov: Why the parameter is not mandatory?

TsvetanMilanov: Because if it is `null` or `undefined` the method will still work. It will check the current directory.

PR #2605, opened 2017-03-13
provided more information to plugin authors

Plamen5kov:
_Problem_:
Plugin authors do unnecessary changes, because they lack the information about files about to be synced. This workflow causes too much trouble, which can be easily avoided if cli provides more information to the plugin authors, who will then be able to handle the erroneous behavior.

_Solution_:
Introduced `hookArgs.filesToSync` parameter which contains all changed files as a string array. If no files are marked as changed `filesToSync` is `undefined`.

Plugin authors needed more information on what files have changed during livesync and we now provide this information through `hookArgs.filesToSync`, which is a string array full with all detected changed files. `hookArgs` is a parameter of the `before-(after)-prepare-hook.js` files, that gets resolved by an injector.

### DIFF
--- a/lib/services/platform-service.ts
+++ b/lib/services/platform-service.ts
@@ -190,7 +190,7 @@ export class PlatformService extends EventEmitter implements IPlatformService {
 		return _.filter(this.$platformsData.platformsNames, p => { return this.isPlatformPrepared(p, projectData); });
 	}
 
-	public async preparePlatform(platform: string, appFilesUpdaterOptions: IAppFilesUpdaterOptions, platformTemplate: string, projectData: IProjectData, platformSpecificData: IPlatformSpecificData): Promise<boolean> {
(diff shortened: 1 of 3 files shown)


lib/services/platform-service.ts
```diff
@@ -257,7 +257,7 @@ export class PlatformService extends EventEmitter implements IPlatformService {
 	}
 
 	@helpers.hook('prepare')
-	private async preparePlatformCore(platform: string, appFilesUpdaterOptions: IAppFilesUpdaterOptions, projectData: IProjectData, platformSpecificData: IPlatformSpecificData, changesInfo?: IProjectChangesInfo): Promise<void> {
+	private async preparePlatformCore(platform: string, appFilesUpdaterOptions: IAppFilesUpdaterOptions, projectData: IProjectData, platformSpecificData: IPlatformSpecificData, changesInfo?: IProjectChangesInfo, filesToSync?: Array<String>): Promise<void> {
```
Plamen5kov: Sure

PR #2607, opened 2017-03-13
Fix Hardcoded IPA export method.

vbresults:
https://github.com/vbresults-llc/nativescript-cli/commit/1f4e041cdf56a0946a70276cf323e6f571d27ab0

### DIFF
--- a/lib/commands/appstore-upload.ts
+++ b/lib/commands/appstore-upload.ts
@@ -34,6 +34,8 @@ export class PublishIOS implements ICommand {
 			mobileProvisionIdentifier = args[2],
 			codeSignIdentity = args[3],
 			teamID = this.$options.teamId,
+			exportMethod = this.$options.exportMethod ? this.$options.exportMethod : null,
+			xcargs = this.$options.xcargs ? this.$options.xcargs : null,
 			ipaFilePath = this.$options.ipa ? path.resolve(this.$options.ipa) : null;
 
 		if (!username) {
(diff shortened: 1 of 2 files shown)


Plamen5kov:
@vbresults this PR seems ok, but what causes the need for it. it will be really useful for me if you could spare some time to explain the reasoning behind it. What problem does this PR solve?

vbresults:
@Plamen5kov The problem is that ns-build is hardcoded to only export one type of IPA (app-store) when there are four (development, ad-hoc, enterprise, and finally app-store).

In a CI envrionment, the current hardcoded implementation breaks automatic signing for non-app-store builds when we just want to build an IPA (rather than run it).

lib/services/ios-project-service.ts
```diff
@@ -186,7 +186,7 @@ export class IOSProjectService extends projectServiceBaseLib.PlatformProjectServ
 	/**
 	 * Exports .xcarchive for AppStore distribution.
 	 */
-	public async exportArchive(projectData: IProjectData, options: { archivePath: string, exportDir?: string, teamID?: string }): Promise<string> {
+	public async exportArchive(projectData: IProjectData, options: { archivePath: string, exportDir?: string, teamID?: string, exportMethod?: string, xcargs?: string }): Promise<string> {
```
Plamen5kov: You've changed the signature of the method, but you don't call it from anywhere with the additional `xcargs` parameter. How do you plan on using the changes?

rosen-vladimirov: Same is valid for the new `exportMethod` option.

vbresults: @Plamen5kov @rosen-vladimirov Would I change it here?  https://github.com/NativeScript/nativescript-cli/blob/00b80f71f6e45955c09b091adc884e74e41aca47/lib/commands/appstore-upload.ts

The filename is appstore-upload so I'm not sure if this would be where to change it, since I'm not uploading to the app store

vbresults: @Plamen5kov @rosen-vladimirov Updated commit.

lib/commands/appstore-upload.ts
```diff
@@ -34,6 +34,8 @@ export class PublishIOS implements ICommand {
 			mobileProvisionIdentifier = args[2],
 			codeSignIdentity = args[3],
 			teamID = this.$options.teamId,
+			exportMethod = this.$options.exportMethod ? this.$options.exportMethod : null,
```
Plamen5kov: lib/commands/appstore-upload.ts(37,33): error TS2339: Property 'exportMethod' does not exist on type 'IOptions'.
lib/commands/appstore-upload.ts(37,62): error TS2339: Property 'exportMethod' does not exist on type 'IOptions'.
lib/commands/appstore-upload.ts(38,27): error TS2339: Property 'xcargs' does not exist on type 'IOptions'.
lib/commands/appstore-upload.ts(38,50): error TS2339: Property 'xcargs' does not exist on type 'IOptions'.

from travis: please fix

Plamen5kov: @vbresults you can run tests like so: `grunt test` in the root folder of nativesript-cli.

vbresults:
@Plamen5kov I have run into many problems and overall the current architecture for automating builds (or lack of it) is not suitable.  Fastlane should be integrated so that there is full flexibility in this regard.

I will submit a PR at a later time with a permanent and less maintenance-intensive fastlane integration rather than this patch, which will solve symptoms like this (hardcoded IPA method) implicitly.

PR #2608, opened 2017-03-14
release option triggers build every time

Plamen5kov:
_Problem_:
When running release builds twice with changes to a file between them, these changes are not respected.

_Solution_
Build every time in release. With gradle 3.0 that'll happen fast so it won't be slowing down development.

### DIFF
--- a/lib/services/project-changes-service.ts
+++ b/lib/services/project-changes-service.ts
@@ -86,7 +86,7 @@ export class ProjectChangesService implements IProjectChangesService {
 				this._prepareInfo.iOSProvisioningProfileUUID = nextCommandProvisionUUID;
 			}
 		}
-		if (this.$options.bundle !== this._prepareInfo.bundle || this.$options.release !== this._prepareInfo.release) {
+		if (this.$options.bundle !== this._prepareInfo.bundle || this.$options.release) {
 			this._changesInfo.appFilesChanged = true;
 			this._changesInfo.appResourcesChanged = true;
 			this._changesInfo.modulesChanged = true;


rosen-vladimirov:
What about iOS?
`Build every time in release. With gradle 3.0 that'll happen fast so it won't be slowing down development.`
Is Xcode 8.2.1 fast?

Plamen5kov:
When I tested it, it wasn't slower than gradle build. And the expectation is release will be done once the app is ready, and the need to run with `--release` option will not be prevailing.

rosen-vladimirov:
I'm not sure this is the best solution. Ping @hshristov , @dtopuzov , @vchimev , @vakrilov , @hdeshev , @etabakov , @Mitko-Kerezov , @yyosifov 
What do you think - is it okay when user writes `tns run/build <platform> --release` to skip all checks that make the build faster and execute full prepare + build?

vchimev:
```
Problem:
When running release builds twice with changes to a file between them, these changes are not respected.
```

What are these changes that are not respected? Any related issue logged?

I'm not sure if it is the best option always to execute full prepare and build in release, but I'm definitely concerned to identify the problems and resolve them in **debug**, so users have fast and correct development experience. Then if we do it for **debug**, may be there won't be problem to do the same and for **release**.

Plamen5kov:
@vchimev I've linked this PR with the related issue. It's shown at the bottom of this PR.

Plamen5kov:
run CI

yyosifov:
@rosen-vladimirov do we have some analytics on the Options passed to the command? Can we see how often do users execute "run/build" with --release flag? I guess if it's less than 5-10% this quick fix will do the job and we can improve it later?

Plamen5kov:
closing this, because it's fixed in master.

Issue #2513, opened 2017-02-08
Allow templates to omit App_Resources


tjvantoll:
As the NativeScript 2.5 release, NativeScript templates now _must_ provide a complete `App_Resources` folder in order to work. I believe we should revert that change, and make `App_Resources` optional for templates. Here’s why:

- The easier we make templates to create the more people will create them. We don’t have many templates right now and it’d be great to have more community-written ones.
- Very few templates will actually need to change files in `App_Resources`. Few templates will need to add images, and none will want to alter the default splash screens or icons. That means most template authors will do what I did, and copy/paste the `App_Resources` folder from the default hello world template.
- We update the `App_Resources` folder in the default template fairly often, and I want those updates to make it into my templates. To accomplish this currently I have to manually check for `App_Resources` updates with each release, copy and paste the folder again, and publish a brand new version of each of my templates.

I realize that this change was made for performance reasons—that we don’t want users to download two templates when using `App_Resources`-free templates. But I believe the developer convenience outweighs the moderate performance gains. And most users will continue to use one of our three default templates, all of which include an `App_Resources` folder, and will therefore not be subject to this small performance hit.

nraboy:
Thanks!

Plamen5kov:
@nraboy @tjvantoll The performance hit wasn't the big problem which led to the decision. The actual problem was, the behavior we had was error prone, because CLI augmented the App_Resources folder when it was present, but it didn't contain all the necessary files. The second reason is the templates can differ and while the copy/paste solution might work in some cases, an angular template can differ a lot from a vanilla js template, and we might tie our hands on using more than we need. In third place is the performance issue (additional traffic). 

As @tjvantoll said:
>That means most template authors will do what I did, and copy/paste the App_Resources folder from the default hello world template.

One possible solution for developers making templates to add the `App_Resources` folder in their templates and publish them.

---
>We update the App_Resources folder in the default template fairly often, and I want those updates to make it into my templates. 

That's exactly right, the templates are updated often, but what happens when the angular templates `App_Resources` is different from the vanilla one. What should the "default" one be. We'll be making decisions instead of the template author, which has proven error prone.

tjvantoll:
> The actual problem was, the behavior we had was error prone, because CLI augmented the App_Resources folder when it was present, but it didn't contain all the necessary files.

[...]

PR #2609, opened 2017-03-14
add App_Resources from default template if missing #2513

yyosifov:
If the template specified by the --template parameter doesn't contain the App_Resources folder, we take the App_Resources folder from the default template and use that instead.

### DIFF
--- a/lib/services/project-service.ts
+++ b/lib/services/project-service.ts
@@ -48,6 +48,8 @@ export class ProjectService implements IProjectService {
 			let templatePath = await this.$projectTemplatesService.prepareTemplate(selectedTemplate, projectDir);
 			await this.extractTemplate(projectDir, templatePath);
 
+			await this.ensureAppResourcesExist(projectDir);
+
 			let packageName = constants.TNS_CORE_MODULES_NAME;
 			await this.$npm.install(packageName, projectDir, { save: true, "save-exact": true });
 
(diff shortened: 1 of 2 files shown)


lib/services/project-service.ts
```diff
@@ -102,6 +104,27 @@ export class ProjectService implements IProjectService {
 		this.$fs.createDirectory(path.join(projectDir, "platforms"));
 	}
 
+	private async ensureAppResourcesExist(projectDir: string): Promise<void> {
+		let appPath = path.join(projectDir, constants.APP_FOLDER_NAME),
```
rosen-vladimirov: these could be `const` instead of `let`

lib/services/project-service.ts
```diff
@@ -102,6 +104,27 @@ export class ProjectService implements IProjectService {
...
+			appResourcesDestinationPath = path.join(appPath, constants.APP_RESOURCES_FOLDER_NAME);
+
+		if (!this.$fs.exists(appResourcesDestinationPath)) {
+			this.$fs.createDirectory(appResourcesDestinationPath);
+
+			// the template installed doesn't have App_Resources -> get from a default template
+			let defaultTemplateName = constants.RESERVED_TEMPLATE_NAMES["default"];
+			await this.$npm.install(defaultTemplateName, projectDir, { save: true, });
+			let defaultTemplateAppResourcesPath = path.join(projectDir, constants.NODE_MODULES_FOLDER_NAME,
```
rosen-vladimirov: same here - these could be `const`

lib/services/project-service.ts
yyosifov: I agree `const` is the better approach. I'll leave it as `let` now, following the whole style of the file - where all other same-constructed paths are with `let`. I'll take care to use const in the future :)

PR #2610, opened 2017-03-14
Fix getting info for Android tools

rosen-vladimirov:
Due to changes in Android SDK, we have to update the checks in CLI. While gathering system information, we check the android executable, which is no longer returning correct results.

We use the android executable to find information about installed Android SDKs and to construct correct paths based on ANDROID_HOME. Fix this by listing directories inside ANDROID_HOME and find the information about installed SDKs from there.
Fix messages pointing to `android` executable to point to `sdkmanager` (in case it exists).

In order to fix this, we rely on the emulator executable, which is the real thing we need as it is the one that allows us to work with Android Emulators (this is changed in mobile-cli-lib).
Fix sys-info checks and get correct path to emulator according to latest changes.

PR in mobile-cli-lib https://github.com/telerik/mobile-cli-lib/pull/906

### DIFF
--- a/lib/android-tools-info.ts
+++ b/lib/android-tools-info.ts
@@ -14,7 +14,6 @@ export class AndroidToolsInfo implements IAndroidToolsInfo {
 	private showWarningsAsErrors: boolean;
 	private toolsInfo: IAndroidToolsInfoData;
 	private selectedCompileSdk: number;
-	private installedTargetsCache: string[] = null;
 	private androidHome = process.env["ANDROID_HOME"];
 
 	constructor(private $childProcess: IChildProcess,
@@ -313,22 +312,16 @@ export class AndroidToolsInfo implements IAndroidToolsInfo {
 		return parseInt(androidSdkString.replace(`${AndroidToolsInfo.ANDROID_TARGET_PREFIX}-`, ""));
 	}
 
+	@cache()
 	private getInstalledTargets(): string[] {
-		if (!this.installedTargetsCache) {
-			try {
-				const pathToInstalledTargets = path.join(this.androidHome, "platforms");
-				if (!this.$fs.exists(pathToInstalledTargets)) {
-					throw new Error("No Android Targets installed.");
-				}
-
-				this.installedTargetsCache = this.$fs.readDirectory(pathToInstalledTargets);
-				this.$logger.trace("Installed Android Targets are: ", this.installedTargetsCache);
-			} catch (err) {
-				this.$logger.trace("Unable to get Android targets. Error is: " + err);
-			}
+		let installedTargets: string[] = [];
+		const pathToInstalledTargets = path.join(this.androidHome, "platforms");
+		if (!this.$fs.exists(pathToInstalledTargets)) {
+			installedTargets = this.$fs.readDirectory(pathToInstalledTargets);
+			this.$logger.trace("Installed Android Targets are: ", installedTargets);
 		}
 
-		return this.installedTargetsCache;
+		return installedTargets;
 	}
 
 	private getMaxSupportedVersion(): number {
--- a/lib/android-tools-info.ts
+++ b/lib/android-tools-info.ts
@@ -318,10 +318,11 @@ export class AndroidToolsInfo implements IAndroidToolsInfo {
 	private getInstalledTargets(): string[] {
 		let installedTargets: string[] = [];
 		const pathToInstalledTargets = path.join(this.androidHome, "platforms");
-		if (!this.$fs.exists(pathToInstalledTargets)) {
+		if (this.$fs.exists(pathToInstalledTargets)) {
 			installedTargets = this.$fs.readDirectory(pathToInstalledTargets);
-			this.$logger.trace("Installed Android Targets are: ", installedTargets);
 		}
+
+		this.$logger.trace("Installed Android Targets are: ", installedTargets);
 
 		return installedTargets;
 	}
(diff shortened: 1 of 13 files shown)


lib/android-tools-info.ts
```diff
@@ -186,6 +149,28 @@ export class AndroidToolsInfo implements IAndroidToolsInfo {
 		return null;
 	}
 
+	private _cachedAndroidHomeValidationResult: boolean = null;
```
rosen-vladimirov: use `@cache` decorator here instead of variable

lib/android-tools-info.ts
```diff
@@ -318,44 +303,23 @@ export class AndroidToolsInfo implements IAndroidToolsInfo {
 	private async getInstalledTargets(): Promise<string[]> {
```
rosen-vladimirov: this method no longer needs to return Promise - its operations are sync now

lib/services/android-project-service.ts
```diff
@@ -204,10 +203,11 @@ export class AndroidProjectService extends projectServiceBaseLib.PlatformProject
 		}
 	}
 
-	public async interpolateConfigurationFile(projectData: IProjectData, platformSpecificData: IPlatformSpecificData): Promise<void> {
+// TODO: Check if we can make this method sync.
```
Pip3r4o: The method is already sync. Delete the comment?

lib/android-tools-info.ts
```diff
@@ -186,6 +139,51 @@ export class AndroidToolsInfo implements IAndroidToolsInfo {
 		return null;
 	}
 
+	@cache()
+	public validateAndroidHomeEnvVariable(options?: { showWarningsAsErrors: boolean }): boolean {
```
yyosifov: Do we need tests for this functionality? Will the cache work as expected here, don't we need to somehow include the "androidHome" to the cacheKey?

lib/android-tools-info.ts
```diff
@@ -186,6 +139,51 @@ export class AndroidToolsInfo implements IAndroidToolsInfo {
...
+				"where you will find `tools` and `platform-tools` directories.");
+			androidHomeValidationResult = false;
+		}
+
+		return androidHomeValidationResult;
+	}
+
+	@cache()
+	private getPathToSdkManagementTool(): string {
+		const sdkmanagerName = "sdkmanager";
```
yyosifov: Although I understand the "sdkmanager" is lowercase, the variables containing "Sdkmanager" look little less readable to me. (sdkmanagerName, pahtToSdkmanager)

lib/android-tools-info.ts
```diff
@@ -186,6 +139,51 @@ export class AndroidToolsInfo implements IAndroidToolsInfo {
...
+		let sdkManagementToolPath = sdkmanagerName;
+
+		const isAndroidHomeValid = this.validateAndroidHomeEnvVariable();
+
+		if (isAndroidHomeValid) {
```
yyosifov: if it's not valid, we'll return "sdkmanager" as result. Shouldn't we throw an error in case the AndroidHome is invalid?

lib/android-tools-info.ts
```diff
@@ -284,8 +282,8 @@ export class AndroidToolsInfo implements IAndroidToolsInfo {
 		return buildToolsVersion;
 	}
 
-	private async getAppCompatRange(): Promise<string> {
-		let compileSdkVersion = await this.getCompileSdk();
+	private getAppCompatRange(): string {
+		let compileSdkVersion = this.getCompileSdk();
```
yyosifov: why not name the "getCompileSdk()" method as "getCompileSdkVersion"? It returns a number and it's assigned to a variable sdkVersion, therefore I think it would be more correctly named this way

lib/android-tools-info.ts
```diff
@@ -306,56 +304,35 @@ export class AndroidToolsInfo implements IAndroidToolsInfo {
...
 	}
 
 	private parseAndroidSdkString(androidSdkString: string): number {
 		return parseInt(androidSdkString.replace(`${AndroidToolsInfo.ANDROID_TARGET_PREFIX}-`, ""));
 	}
 
-	private async getInstalledTargets(): Promise<string[]> {
+	private getInstalledTargets(): string[] {
 		if (!this.installedTargetsCache) {
 			try {
```
TsvetanMilanov: I think we don't need this `try/catch`. You can move the logger.trace in the `if` instead of `throw new Error`and do the rest in the `else`.

lib/android-tools-info.ts
rosen-vladimirov: The cache decorator has tests [here](https://github.com/telerik/mobile-cli-lib/blob/master/test/unit-tests/decorators.ts#L421-L494)
The `androidHome` will not be changed between executions of the method (I've also change androidHome to be only getter). 

Regarding tests for the whole androidToolsInfo - we'll move it to `nativescript-doctor` package very soon and we'll add tests there.

lib/android-tools-info.ts
rosen-vladimirov: This method is used just to get the path that we'll show in the `additionalMessage`. The idea is that if something is not configured properly, `tns doctor` command to show some suggestions how to fix the issue. 
In case you execute any other command, CLI will fail immediately in case the ANDROID_HOME is not set. The default is `sdkmanager` as when the ANDROID_HOME is not set, most probably the user does not have any Android SDK. After installing it, he'll have the new SDK, which requires usage of `sdkmanager` not `android`.

PR #2611, opened 2017-03-14
Fix getting info for Android tools

rosen-vladimirov:
Due to changes in Android SDK, we have to update the checks in CLI. While gathering system information, we check the android executable, which is no longer returning correct results.

We use the android executable to find information about installed Android SDKs and to construct correct paths based on ANDROID_HOME. Fix this by listing directories inside ANDROID_HOME and find the information about installed SDKs from there.
Fix messages pointing to `android` executable to point to `sdkmanager` (in case it exists).

In order to fix this, we rely on the emulator executable, which is the real thing we need as it is the one that allows us to work with Android Emulators (this is changed in mobile-cli-lib).
Fix sys-info checks and get correct path to emulator according to latest changes. 
PR In mobile-cli-lib: https://github.com/telerik/mobile-cli-lib/pull/907

### DIFF
--- a/lib/android-tools-info.ts
+++ b/lib/android-tools-info.ts
@@ -1,6 +1,6 @@
 import * as path from "path";
 import * as semver from "semver";
-import {EOL} from "os";
+import { EOL } from "os";
 
 export class AndroidToolsInfo implements IAndroidToolsInfo {
 	private static ANDROID_TARGET_PREFIX = "android";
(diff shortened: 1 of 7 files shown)


lib/android-tools-info.ts
```diff
@@ -343,45 +337,24 @@ export class AndroidToolsInfo implements IAndroidToolsInfo {
 		return (() => {
 			if (!this.installedTargetsCache) {
 				try {
```
TsvetanMilanov: I think we don't need this `try/catch`. You can move the logger.trace in the `if` instead of `throw new Error`and do the rest in the `else`.

lib/sys-info.ts
```diff
@@ -1,21 +1,21 @@
...
+		protected $iTunesValidator: Mobile.IiTunesValidator,
+		protected $logger: ILogger,
+		protected $winreg: IWinReg,
+		protected $androidEmulatorServices: Mobile.IAndroidEmulatorServices,
+		private $androidToolsInfo: IAndroidToolsInfo) {
+		super($childProcess, $hostInfo, $iTunesValidator, $logger, $winreg, $androidEmulatorServices);
 	}
 
-	public getSysInfo(pathToPackageJson: string, androidToolsInfo?: {pathToAdb: string, pathToAndroid: string}): IFuture<ISysInfoData> {
+	public getSysInfo(pathToPackageJson: string, androidToolsInfo?: { pathToAdb: string, pathToAndroid: string }): IFuture<ISysInfoData> {
```
TsvetanMilanov: We don't need `pathToAndroid: string` in the `androidToolsInfo`.

lib/android-tools-info.ts
rosen-vladimirov: I've added the mentioned else statement, but I prefer having the try-catch as well.

PR #2612, opened 2017-03-15
add check to the doctor that the 'six' python package is installed on…

yyosifov:
When Python is installed on Mac via Brew it's missing the "six" python package, which is installed by default when using the default Python installer for Mac. The "six" python package is required from the CLI in order to work properly.

So, we added the check for "six" python package presence in the TNS Doctor so that it detects the missing piece and gives advice how to fix your environment.

### DIFF
--- a/lib/services/doctor-service.ts
+++ b/lib/services/doctor-service.ts
@@ -105,7 +105,8 @@ class DoctorService implements IDoctorService {
 
 		let androidToolsIssues = this.$androidToolsInfo.validateInfo();
 		let javaVersionIssue = await this.$androidToolsInfo.validateJavacVersion(sysInfo.javacVersion);
-		let doctorResult = result || androidToolsIssues || javaVersionIssue;
+		let pythonIssues = await this.validatePythonPackages();
+		let doctorResult = result || androidToolsIssues || javaVersionIssue || pythonIssues;
 
 		if (!configOptions || configOptions.trackResult) {
 			await this.$analyticsService.track("DoctorEnvironmentSetup", doctorResult ? "incorrect" : "correct");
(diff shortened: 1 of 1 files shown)


lib/services/doctor-service.ts
```diff
@@ -213,5 +214,22 @@ class DoctorService implements IDoctorService {
...
+
+	private async validatePythonPackages(): Promise<boolean> {
+		let hasInvalidPackages = false;
+		if (this.$hostInfo.isDarwin) {
+			try {
+				let queryForSixPackage = await this.$childProcess.exec(`pip freeze | grep '^six=' | wc -l`);
+				let sixPackagesFoundCount = parseInt(queryForSixPackage);
+				if (sixPackagesFoundCount === 0) {
+					hasInvalidPackages = true;
+					this.$logger.error("Python 'six' package not found. Please install it by running 'pip install six' from the terminal.");
```
rosen-vladimirov: Can you follow the same structure of messages as the others warnings:
```TypeScript
this.$logger.warn("Python 'six' package not found.").
this.$logger.out("<information why do I need this package and what issues will occur in case I do not install it>. <information how to install this package>");
```

The concept is - `warning` message (in yellow) that something is missing and additional information how will this affect me and how to fix the issue (text will be in the default terminal color).

For example I do not know what's the idea of this package and I will not install it on my machine without evidence that I need it.

lib/services/doctor-service.ts
```diff
@@ -213,5 +214,22 @@ class DoctorService implements IDoctorService {
...
+				}
+			} catch (error) {
+				this.$logger.error("Cannot retrieve python installed packages. Please, make sure you have both 'python' and 'pip' installed.");
```
rosen-vladimirov: Same as above - `logger.warn` + `logger.out`.
Additionally it's a good practice to add the real error to the trace logs:
```TypeScript
this.$logger.trace(`Error while validating Python packages. Error is: ${error.message}`);
```

yyosifov:
run ci

PR #2614, opened 2017-03-16
refactoring in order to start moving functionality from emulate to run command

Plamen5kov:
* fixed up a couple of interfaces
* renamed methods
* removed some obsolete code
* moved `--available-devices` flag to run command

### DIFF
--- a/test/npm-support.ts
+++ b/test/npm-support.ts
@@ -159,7 +159,7 @@ async function setupProject(dependencies?: any): Promise<any> {
 				ensureConfigurationFileInAppResources: (): any => null,
 				interpolateConfigurationFile: (): void => undefined,
 				isPlatformPrepared: (projectRoot: string) => false
-				validatePlugins: async (projectData: IProjectData) => Promise.resolve()
+				validatePlugins: (projectData: IProjectData) => Promise.resolve()
 			}
 		};
 	};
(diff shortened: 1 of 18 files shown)


test/npm-support.ts
```diff
@@ -158,7 +158,8 @@ async function setupProject(dependencies?: any): Promise<any> {
 				processConfigurationFilesFromAppResources: () => Promise.resolve(),
 				ensureConfigurationFileInAppResources: (): any => null,
 				interpolateConfigurationFile: () => Promise.resolve(),
-				isPlatformPrepared: (projectRoot: string) => false
+				isPlatformPrepared: (projectRoot: string) => false,
+				validatePlugins: async (projectData: IProjectData) => Promise.resolve()
```
rosen-vladimirov: no need of async in this case - the method returns Promise.

test/platform-service.ts
```diff
@@ -57,6 +57,9 @@ function createTestInjector() {
 		},
 		ensureAllDependenciesAreInstalled: () => {
 			return Promise.resolve();
+		},
+		validate: async (platformData: IPlatformData, projectData: IProjectData) => {
```
rosen-vladimirov: no need of async here.

PR #2615, opened 2017-03-16
platform clean should preserve the nativescript version used #2496

etabakov:
Consider the following situation: There is a new version of NativeScript available (3.0) and I have existing projects using older version (2.5). I don't want to upgrade immediately for various reasons (too many effort required on my side, I'm just before releasing the app, etc). The current behaviour of the `tns platform clean` command is that it will first remove the specified platform and then the latest available platform (3.0) will be added. This is inconvenience in the described scenario.

As a developer, I would expect when I run `tns platform clean` the version of the platform added to be the same as before running the command (2.5).

Related to issue: [TNS platform clean android #2496](https://github.com/NativeScript/nativescript-cli/issues/2496)

### DIFF
--- a/lib/commands/platform-clean.ts
+++ b/lib/commands/platform-clean.ts
@@ -9,8 +9,7 @@ export class CleanCommand implements ICommand {
 		}
 
 	public async execute(args: string[]): Promise<void> {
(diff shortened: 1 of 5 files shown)


etabakov:
Thanks for the feedback! More detailed description is provided.

PR #2618, opened 2017-03-16
Fix hooks execution when using CLI as lib and fix package.json

rosen-vladimirov:
When NS CLI is used as library, project hooks are not executed as hooksService gets the project dir from `$options` or current working dir.
Fix this by getting the projectDir from the hookArgs object. The object is constructed from method parameters. We check if there's `projectDir` or `projectData.projectDir` arg.
Also fix the caching of hooks - each time when we try to execute them, we'll search the directories for hooks. This is required for long living process.

Additionally use semver's package definitions from `@types` repo.
Rename two `.d.ts` files which were incorrectly set as `.ts`.

Add zipstream package - it's used in `$fs.zipFiles` method, but somehow it's missing from this package.json, so trying to use the mentioned method fails


rosen-vladimirov:
run ci

PR #2619, opened 2017-03-17
Allow cloud builds from the CLI

rosen-vladimirov:
Add new package that allows cloud builds from CLI.

Also in case CLI's executed in non-interactive terminal and there's no DEVELOPMENT_TEAM set, `tns build ios --for-device` fails with message "Console is not interactive and no default action specified" which does not give any info to the users.
So in this case add more descriptive error message.


rosen-vladimirov:
Closing this due to latest requirements to install `nativescript-cloud` outside of `nativescript-cli`.

PR #2630, opened 2017-03-21
Fix isValidNativeScript project

rosen-vladimirov:
The isValidNativeScript project method is used in Fusion. However the implementation is not correct. Fix it to have correct behavior.
In order to fix it, modify projectData.initialize method - there's obsolete code for migration from .tnsproject to package.json - we do not need it anymore, so remove it.
Also fix a case where failing builds do not fail the build process (incorrect argument passed to childProcess.spawnFromEvent).

Update documentation for PublicAPI.

Also in case CLI's executed in non-interactive terminal and there's no DEVELOPMENT_TEAM set, tns build ios --for-device fails with message "Console is not interactive and no default action specified" which does not give any info to the users.
So in this case add more descriptive error message.

### DIFF
--- a/lib/project-data.ts
+++ b/lib/project-data.ts
@@ -42,7 +42,8 @@ export class ProjectData implements IProjectData {
 		private $errors: IErrors,
 		private $projectHelper: IProjectHelper,
 		private $staticConfig: IStaticConfig,
-		private $options: IOptions) { }
+		private $options: IOptions,
+		private $logger: ILogger) { }
 
 	public initializeProjectData(projectDir?: string): void {
 		projectDir = projectDir || this.$projectHelper.projectDir;
@@ -51,15 +52,14 @@ export class ProjectData implements IProjectData {
 			const projectFilePath = path.join(projectDir, this.$staticConfig.PROJECT_FILE_NAME);
 			let data: any = null;
 
-			if (projectFilePath) {
+			if (this.$fs.exists(projectFilePath)) {
 				let fileContent: any = null;
 				try {
 					fileContent = this.$fs.readJson(projectFilePath);
 					data = fileContent[this.$staticConfig.CLIENT_NAME_KEY_IN_PROJECT_FILE];
 				} catch (err) {
-					this.$errors.failWithoutHelp(
-						`The project file ${this.projectFilePath} is corrupted.` + EOL +
-						"Consider restoring an earlier version from your source control or backup." + EOL +
+					this.$errors.failWithoutHelp(`The project file ${this.projectFilePath} is corrupted. ${EOL}` +
+						`Consider restoring an earlier version from your source control or backup.${EOL}` +
 						`Additional technical info: ${err.toString()}`);
 				}
 
@@ -80,8 +80,11 @@ export class ProjectData implements IProjectData {
 			}
 		}
 
+		const currentDir = path.resolve(".");
+		this.$logger.trace(`Unable to find project. projectDir: ${projectDir}, options.path: ${this.$options.path}, ${currentDir}`);
+
 		// This is the case when no project file found
-		this.$errors.fail("No project found at or above '%s' and neither was a --path specified.", projectDir || this.$options.path || path.resolve("."));
+		this.$errors.fail("No project found at or above '%s' and neither was a --path specified.", projectDir || this.$options.path || currentDir);
 	}
 
 	private getProjectType(): string {
--- a/lib/project-data.ts
+++ b/lib/project-data.ts
@@ -9,8 +9,6 @@ interface IProjectType {
 }
 
 export class ProjectData implements IProjectData {
-	private static OLD_PROJECT_FILE_NAME = ".tnsproject";
-
 	/**
 	 * NOTE: Order of the elements is important as the TypeScript dependencies are commonly included in Angular project as well.
 	 */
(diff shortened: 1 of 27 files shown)


lib/project-data.ts
```diff
@@ -42,45 +40,48 @@ export class ProjectData implements IProjectData {
...
+	public initializeProjectData(projectDir?: string): void {
 		projectDir = projectDir || this.$projectHelper.projectDir;
 		// If no project found, projectDir should be null
 		if (projectDir) {
-			this.initializeProjectDataCore(projectDir);
+			const projectFilePath = path.join(projectDir, this.$staticConfig.PROJECT_FILE_NAME);
 			let data: any = null;
 
-			if (this.$fs.exists(this.projectFilePath)) {
+			if (projectFilePath) {
```
yyosifov: how can this ever be undefined? Why did we remove the exists? It might just not be a valid project?

lib/project-data.ts
```diff
@@ -42,45 +40,48 @@ export class ProjectData implements IProjectData {
...
-						suppressCommandHelp: true
-					},
-						this.projectFilePath, err.toString());
+						`Additional technical info: ${err.toString()}`);
 				}
 
 				if (data) {
+					this.projectDir = projectDir;
+					this.projectName = this.$projectHelper.sanitizeName(path.basename(projectDir));
+					this.platformsDir = path.join(projectDir, constants.PLATFORMS_DIR_NAME);
```
yyosifov: platformsDir/appDirectoryPath/appResourcesDirectoryPath seem like "Computed properties" for me - they all are coming from the projectDir. Why not initialize only the project dir, and create getters for these properties so that we don't hit a bug if sometime we miss updating one of the computed ones? Do we have such a practice?

lib/project-data.ts
```diff
@@ -42,45 +40,48 @@ export class ProjectData implements IProjectData {
...
+
+					return;
 				}
 			}
-		} else { // This is the case when no project file found
-			this.tryToUpgradeProject();
 		}
+
+		// This is the case when no project file found
+		this.$errors.fail("No project found at or above '%s' and neither was a --path specified.", projectDir || this.$options.path || path.resolve("."));
```
yyosifov: shall we add trace here, too?

lib/providers/livesync-provider.ts
```diff
@@ -45,10 +45,10 @@ export class LiveSyncProvider implements ILiveSyncProvider {
 		await this.$platformService.buildPlatform(device.deviceInfo.platform, buildConfig, projectData);
 		let platformData = this.$platformsData.getPlatformData(device.deviceInfo.platform, projectData);
 		if (device.isEmulator) {
-			return this.$platformService.getLatestApplicationPackageForEmulator(platformData).packageName;
+			return this.$platformService.getLatestApplicationPackageForEmulator(platformData, { isReleaseBuild: buildConfig.release }).packageName;
 		}
 
-		return this.$platformService.getLatestApplicationPackageForDevice(platformData).packageName;
+		return this.$platformService.getLatestApplicationPackageForDevice(platformData, { isReleaseBuild: buildConfig.release }).packageName;
```
yyosifov: Do you think we can rename from `isReleaseBuild` to `release` so that it's more consistent with the IBuildConfig?

yyosifov: Do you think we can directly pass on the `buildConfig` property, so that we have it in handy in case we need other property in the future?

lib/services/android-project-service.ts
```diff
@@ -57,12 +57,19 @@ export class AndroidProjectService extends projectServiceBaseLib.PlatformProject
...
 				projectRoot: projectRoot,
 				deviceBuildOutputPath: path.join(projectRoot, "build", "outputs", "apk"),
-				validPackageNamesForDevice: [
-					`${packageName}-debug.apk`,
-					`${packageName}-release.apk`,
-					`${projectData.projectName}-debug.apk`,
-					`${projectData.projectName}-release.apk`
-				],
+				getValidPackageNames: (buildOptions: { isReleaseBuild?: boolean, isForDevice?: boolean }): string[] => {
+					if (buildOptions.isReleaseBuild) {
```
yyosifov: Can be rewritten as:
`
const buildMode = buildOptions.isReleaseBuild ? "release" : "debug"; // or constants
return [
  `${packageName}-${buildMode}.apk`,
  `${projectData.projectName}-${buildMode}.apk` 
];
`

lib/services/android-project-service.ts
```diff
@@ -262,7 +269,7 @@ export class AndroidProjectService extends projectServiceBaseLib.PlatformProject
 			await this.spawn(gradleBin,
 				buildOptions,
 				{ stdio: buildConfig.buildOutputStdio || "inherit", cwd: this.getPlatformData(projectData).projectRoot },
-				{ emitOptions: { eventName: constants.BUILD_OUTPUT_EVENT_NAME }, throwError: false });
+				{ emitOptions: { eventName: constants.BUILD_OUTPUT_EVENT_NAME }, throwError: true });
```
yyosifov: 👍

lib/services/android-project-service.ts
```diff
@@ -426,6 +433,10 @@ export class AndroidProjectService extends projectServiceBaseLib.PlatformProject
 	}
 
 	public async cleanProject(projectRoot: string, options: string[], projectData: IProjectData): Promise<void> {
+		// In case options are not passed, we'll use the default ones for cleaning the project.
+		// In case we do not do this, we'll required android-23 (default in build.gradle) for cleaning the project.
+		options = options.length === 0 ? this.getBuildOptions({ release: false }, projectData) : options;
```
yyosifov: Do you think it would be a good idea to "extend"/"merge" the options passed with the "default options" which the getBuildOptions returns? So that we don't end up with fewer options than we need?

lib/services/android-project-service.ts
```diff
@@ -426,6 +433,10 @@ export class AndroidProjectService extends projectServiceBaseLib.PlatformProject
 	}
 
 	public async cleanProject(projectRoot: string, options: string[], projectData: IProjectData): Promise<void> {
+		// In case options are not passed, we'll use the default ones for cleaning the project.
+		// In case we do not do this, we'll required android-23 (default in build.gradle) for cleaning the project.
```
yyosifov: small typo: "we'll required" to "we'll require" I guess?

lib/services/project-changes-service.ts
```diff
@@ -38,18 +38,17 @@ export class ProjectChangesService implements IProjectChangesService {
...
-		private $options: IOptions,
 		private $fs: IFileSystem) {
 	}
 
 	public get currentChanges(): IProjectChangesInfo {
 		return this._changesInfo;
 	}
 
-	public checkForChanges(platform: string, projectData: IProjectData): IProjectChangesInfo {
+	public checkForChanges(platform: string, projectData: IProjectData, buildOptions: IProjectChangesOptions): IProjectChangesInfo {
```
yyosifov: Weird naming for me: "buildOptions for ProjectChangesOptions"

lib/services/project-service.ts
```diff
@@ -73,11 +73,10 @@ export class ProjectService implements IProjectService {
 	public isValidNativeScriptProject(pathToProject?: string): boolean {
```
yyosifov: I was wondering, do we want to return some errors if we can identify. I mean - OK, it's not a valid NativeScript project but what's the actual issue? Can the user repair his project to be valid - in case he can, I'd suggest we somehow give him a hint what needs to be fixed? What do you think?

lib/project-data.ts
rosen-vladimirov: I would like to fail in case projectFilePath does not exist or in case there's no `data` in the file. There are two cases where we'll fail with the same error (that a project is not found), that's why I've left the throw statement at the end of the method.

rosen-vladimirov: Thanks for pointing this, I've deleted it by mistake.

lib/project-data.ts
rosen-vladimirov: I like the idea, but there's a reason to set the properties here - we are accessing them in many, many places, so computing each of them every single time will require reading the directory, the package.json, etc. This will make a lot of I/O operations. I can cache the properties after first execution, but this will not work for Fusion, which is a long living process.
So I'll keep this code until I figure out a better way for handling this.

lib/project-data.ts
```diff
@@ -42,45 +40,48 @@ export class ProjectData implements IProjectData {
...
+			if (projectFilePath) {
 				let fileContent: any = null;
 				try {
-					fileContent = this.$fs.readJson(this.projectFilePath);
+					fileContent = this.$fs.readJson(projectFilePath);
 					data = fileContent[this.$staticConfig.CLIENT_NAME_KEY_IN_PROJECT_FILE];
 				} catch (err) {
-					this.$errors.fail({
-						formatStr: "The project file %s is corrupted." + EOL +
+					this.$errors.failWithoutHelp(
```
rosen-vladimirov: I'll use three template strings in order to keep the readability of the code and keep the formatting of the message.

lib/services/project-service.ts
rosen-vladimirov: This method is created for Fusion, where when the user selects a folder, we try to check if this is a valid project.
All NativeScript CLI users will see the message from `initializeProjectData` method.

rosen-vladimirov:
Ping @yyosifov , @Mitko-Kerezov  - comments are fixed

lib/services/ios-debug-service.ts
```diff
@@ -93,10 +93,10 @@ class IOSDebugService implements IDebugService {
 		}
 	}
 
-	private async emulatorDebugBrk(projectData: IProjectData, shouldBreak?: boolean): Promise<void> {
+	private async emulatorDebugBrk(projectData: IProjectData, buildConfig: IBuildConfig, shouldBreak?: boolean, ): Promise<void> {
```
Mitko-Kerezov: Gratuitous comma at the end of args list

lib/project-data.ts
yyosifov: I think I didn't explain it correctly. What was my idea is that if you set the `projectDir`, all other properties - `projectName, platformsDir, appDirectoryPath` can be computed from the projectDir variable - without reading anything from the package.json

i.e. something like:

`
    private _platformsDir: string;
    get platformsDir():boolean {
        if(!this._platformsDir) {
             this._platformsDir = path.join(this.projectDir, constants.PLATFORMS_DIR_NAME);
        }
        return this._platformsDir;
    }
`

so that we don't forget to initialize the `platformsDir` somewhere.

PR #2640, opened 2017-03-23
Fix removing of file on android during livesync

rosen-vladimirov:
When migrating from fibers to promises, we've placed incorrect await, which prevents us when trying to remove a file from project during livesync operation.
Fix the code, so removed files will be synced correctly.


yyosifov:
@rosen-vladimirov Great! Do we have an issue logged about this so that we Connect it here?

rosen-vladimirov:
@yyosifov no, I've found it by mistake, but I'll add an issue and will link it. Thanks ;)

Issue #2633, opened 2017-03-22
Deprecate `tns livesync` command


dtopuzov:
Since {N} 2.5.0 we say: 
`This command is deprecated. It will be removed in the next version of NativeScript CLI. Use the "run" command instead.`

It is time to remove `tns livesync` command.

Note: Please also remove the command from both command-line and online help.

PR #2643, opened 2017-03-23
Remove livesync command. 

etabakov:
Livesync was deprecated and run should be used instead.

Connected issue: [#2633 Deprecate Livesync command](https://github.com/NativeScript/nativescript-cli/issues/2633)


etabakov:
Sure, I was just looking into how to do this. On its way.

rosen-vladimirov:
Unfortunately you'll have to edit the files manually... we were thinking for some way to generate the "related articles" section automatically, but we didn't have time to implement it.

PR #2646, opened 2017-03-23
Stop stopping device detection interval

Mitko-Kerezov:
Delete @exportedPromise decorator as we no longer need it
Stop stopping device detection interval when deploying on iOS devices. These operations no longer conflict as the code is located in an external dependency.

Merge after https://github.com/telerik/mobile-cli-lib/pull/917


lib/services/local-build-service.ts
```diff
@@ -1,4 +1,4 @@
-// import { exportedPromise } from "../common/decorators";
+// import { exported } from "../common/decorators";
```
rosen-vladimirov: why are we updating the comment  instead of removing it?

Issue #2349, opened 2016-12-13
Command "emulate" and options "--emulator/--device/--avd" are a bit confusing to use


pkoleva:
**emulate** command is used to run an app on an emulator but we already have options to specify on what device to run the app.
The different options that one can use to define deployment target are behaving differently
**--device** - can be used to specify which device to be targeted and accepts device id and device name for some commands -> related issue [2345](https://github.com/NativeScript/nativescript-cli/issues/2345)

**run --emulator = emulate**, but the option does not accept any parameters - related [2329](https://github.com/NativeScript/nativescript-cli/issues/2329)

If emulate command is removed, to specify which device you'll use, you'll have to type the below command
`tns run android/ios --emulator --device {DeviceName/Id}`

I think we should agree on some kind of flow that covers the most common use cases.

For me the way I use these options is something like that ->
**Option 1 - no emulate command**
`tns run/debug/deploy/livesync --emulator` - starts emulator and ignores all connected real devices
`tns run/debug/deploy/livesync --emulator {EmulatorName/Id}` - starts/connects to specified emulator or if it doesn't exist give a meaningful error message
'tns run/debug/deploy/livesync --device {EmulatorName/Id}' - starts/connects to specified emulator or if it doesn't exist give a meaningful error message
**Option 2 - no --emulator option**
`tns emulate` - starts emulator and runs app on it, ignores all connected real devices
`tns emulate --device {DeviceName/Id}` - starts emulator with specified name and runs the app on it
'tns run/debug/deploy/livesync --device {EmulatorName/Id}' - starts/connects to specified emulator or if it doesn't exist give a meaningful error message

enchev:
I like Option 2.

tzraikov:
The *--avd* option is depricated and it is not supported for some time now.
We plan to remove the emulate command (as discussed with @valentinstoychev):
https://github.com/NativeScript/nativescript-cli/issues/2329

pkoleva:
the option may be deprecated but it is available, as well as --geny option, as @rosen-vladimirov reminded me recently. Also I am not sure removing the command helps a lot - it will still be confusing to use `tns run android/ios --emulator --device {DeviceName/Id}` for me.

tzraikov:
It is not necessary to use the `--emulator option` when using `--device`. 
Currently we fail with an error message when the `--avd` option is used. I would opt for removing this option.
I was not aware that we have a `--geny` option. Maybe we should remove it?

[...]

PR #2647, opened 2017-03-24
Plamen5kov/move emulate to run

Plamen5kov:
do not merge before: https://github.com/telerik/mobile-cli-lib/pull/918

What does this PR do?
* deprecating emulate command (warning to use run command instead)
* updated docs/help
* used introduced function to start emulators/simulators if necessary and whole logic for starting emulators is there

related issue: https://github.com/NativeScript/nativescript-cli/issues/2349


docs/man_pages/project/testing/run-android.md
```diff
@@ -10,19 +10,20 @@ Start an emulator and run the app inside it | `$ tns run android --emulator [<Em
 Runs your project on a connected Android device or in a native Android emulator, if configured. This is shorthand for prepare, build and deploy. While your app is running, prints the output from the application in the console and watches for changes in your code. Once a change is detected, it synchronizes the change with all selected devices and restarts/refreshes the application.
 
 ### Options
+* `--available-devices` - Shows a list of available emulators to be started and a list of already connected devices.
```
yyosifov: this is very very weird to me? So you're telling me I have to run `tns run android --available-devices` to get the available devices and not run the android app?

docs/man_pages/project/testing/run-android.md
```diff
@@ -10,19 +10,20 @@ Start an emulator and run the app inside it | `$ tns run android --emulator [<Em
...
+* `--device` - Specifies a connected device/emulator to start and run the app.
+* `--emulator` - If set, runs the app in all available and configured android emulators. It will start an emulator if none are already running.
```
yyosifov: what does "runs the app in all available and configured android emulators"?

docs/man_pages/project/testing/run-android.md
```diff
@@ -10,19 +10,20 @@ Start an emulator and run the app inside it | `$ tns run android --emulator [<Em
...
+* `--justlaunch` - If set, does not print the application output in the console.
```
yyosifov: besides that, this works for both emulator on device right?

I'd suggest: "--justlaunch - when specified starts the application on the emulator/device and omits all the application output in the console"

docs/man_pages/project/testing/run-android.md
Plamen5kov: The emulate command had this behavior, and we decided to move the functionality to the run command. If it were up to me, I'd advise the user to use `tns device` command.

docs/man_pages/project/testing/run-android.md
Plamen5kov: runs app on all running emulators

docs/man_pages/project/testing/run-android.md
yyosifov: who are we then?

docs/man_pages/project/testing/run-android.md
Plamen5kov: >besides that, this works for both emulator on device right?

this? 
* the `--device` flag works with emulators/simulators/real devices
* the `--emulator` flag works only with emulators/simulators
`--justlaunch` doesn't start the application, it only cuts off the console, once the application is running.

docs/man_pages/project/testing/run-android.md
Pip3r4o: I agree with @yyosifov, `--available-devices` should be removed, and users should use `tns devices` instead. IF you think it would affect users, then put a `to-be-deprecated message` when `--available-devices` is used so we can remove it in a following release.

Plamen5kov: ok, no problem. Will move the functionality to the `tns device` command.

docs/man_pages/project/testing/run-android.md
yyosifov: That sounds strange to me. My doubt would be what exactly am I launching? Anyway, you can leave it as is for now.

PR #2648, opened 2017-03-24
Fix Android builds when gradle args and project dir have spaces

rosen-vladimirov:
In case you try building application, it will fail in case the following conditions are applied:
* build on Windows
* build for Android
* build in release
* path to project dir has spaces
* path to certificate (keystore) has spaces OR the certificate(keystore) name itself has spaces OR the password has spaces

The problem is in Node.js's `child_process` when `spawn` is used - on Windows it has issues when both the command (in our case this is the FULL PATH to gradlew.bat executable in platforms dir of the project) and ANY of the arguments passed to `child_process's spawn` have spaces.
As we are also setting the current working directory of the spawned process to be the path to `<project dir>/platforms/android`, change the `command` passed to `spawn` to be just `gradlew`. This way we'll never have spaces in the command and we'll not hit this issue.

More information for the Node.js issue: https://github.com/nodejs/node-v0.x-archive/issues/25895

There's a similar issue (linked in the comments of the above) for newer node.js versions. The suggested solution there is to use `{ shell: true}` parameter to `spawn` method, but this will require handling of all spaces in the args on our own.

### DIFF
--- a/lib/services/android-project-service.ts
+++ b/lib/services/android-project-service.ts
@@ -6,6 +6,7 @@ import * as projectServiceBaseLib from "./platform-project-service-base";
 import { DeviceAndroidDebugBridge } from "../common/mobile/android/device-android-debug-bridge";
 import { EOL } from "os";
 import { Configurations } from "../common/constants";
+import { SpawnOptions } from "child_process";
 
 export class AndroidProjectService extends projectServiceBaseLib.PlatformProjectServiceBase implements IPlatformProjectService {
 	private static VALUES_DIRNAME = "values";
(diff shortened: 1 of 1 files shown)


lib/services/android-project-service.ts
```diff
@@ -417,24 +415,14 @@ export class AndroidProjectService extends projectServiceBaseLib.PlatformProject
 	}
 
 	public stopServices(projectRoot: string): Promise<ISpawnResult> {
-		let gradleBin = path.join(projectRoot, "gradlew");
-		if (this.$hostInfo.isWindows) {
-			gradleBin += ".bat";
-		}
-
-		return this.$childProcess.spawnFromEvent(gradleBin, ["--stop", "--quiet"], "close", { stdio: "inherit", cwd: projectRoot });
+		// TODO: Check if we have to pass compileSdk and other options here
```
yyosifov: can we help with this TODO?

lib/services/android-project-service.ts
```diff
@@ -504,6 +492,19 @@ export class AndroidProjectService extends projectServiceBaseLib.PlatformProject
 
 		return versionInManifest;
 	}
+
+	private async executeGradleCommand(projectRoot: string, gradleArgs: string[], childProcessOpts?: SpawnOptions, spawnFromEventOptions?: ISpawnFromEventOptions): Promise<ISpawnResult> {
+		const gradleBin = this.$hostInfo.isWindows ? "gradlew.bat" : "./gradlew";
```
yyosifov: I would rename this const to "gradlew" as it points to the wrapper, not gradle itself.

lib/services/android-project-service.ts
rosen-vladimirov: ping @Plamen5kov , @Pip3r4o  for this

rosen-vladimirov: I'm not sure why we need to call stop and should we pass other options...

rosen-vladimirov: I've just tested it without args and it's working fine, so I'll remove the TODOD

PR #2656, opened 2017-03-24
fix the podfile spec

vance:


CocoaPods.md
```diff
@@ -62,7 +62,8 @@ Create a `package.json` file with the following content:
 }
 ```
 
-Create a [Podfile](https://guides.cocoapods.org/syntax/podfile.html) which describes the dependency to the library that you want to use. Move it to the `platforms/ios` folder.
+Create a [
```
Pip3r4o: What exactly are the changes here? I see that you removed the hyperlink

vance:
my mistake, I had tried to manually install the pod prior to `tns add` --- a possible mistake others might make since most pods have this as part of the "setup" process in their readme. The problem is it creates extra lock files etc. that if you're a beginner, you may not know you have to nuke the whole platform folder before you can proceed.

PR #2660, opened 2017-03-27
Add command for setting proxy

rosen-vladimirov:
At the moment users can set proxy by manually editing the file `<cli installation dir>/config/config.json`.
However this has two issues:
 - the proxy does not support authentication
 - if you install newer version of CLI, the file is overwritten, so you have to set it again.

Fix both issues by:
 - creating command that sets the proxy.
 - (currently) for Windows introduce support for authenticated proxy - username and password are stored via Windows Credential Manager API functions.

Expected workflow:
`$ tns <command that makes http request>`
Your proxy requires a username and password. You can run
        tns proxy set <hostname> <port> <username> <password>
In order to supply NativeScript with the credentials needed.

`$ tns proxy set 127.0.0.1 8888 user pass`
Sucessfully setup proxy.

`$ tns proxy get`
Hostname: 127.0.0.1
Port: 8888
Proxy is Enabled

`$ tns proxy clear`
Sucessfully cleared proxy.

`$ tns proxy get`
No proxy set

### DIFF
--- a/docs/man_pages/general/autocomplete-disable.md
+++ b/docs/man_pages/general/autocomplete-disable.md
@@ -8,7 +8,7 @@ General | `$ tns autocomplete disable`
 Disables command-line completion for bash and zsh shells. You need to restart the shell to complete the operation.
 
 <% if(isHtml) { %>> <% } %>NOTE: This operation might modify the `.bash_profile`, `.bashrc` and `.zshrc` files.
-<% if(isHtml) { %> 
+<% if(isHtml) { %>
 ### Related Commands
 
 Command | Description
@@ -20,4 +20,7 @@ Command | Description
 [error-reporting](error-reporting.html) | Configures anonymous error reporting for the NativeScript CLI.
 [help](help.html) | Lists the available commands or shows information about the selected command.
 [doctor](doctor.html) | Checks your system for configuration problems which might prevent the NativeScript CLI from working properly.
+[proxy](proxy.html) | Displays proxy settings.
+[proxy clear](proxy-clear.html) | Clears proxy settings.
+[proxy set](proxy-set.html) | Sets proxy settings.
 <% } %>
--- a/lib/config.ts
+++ b/lib/config.ts
@@ -6,7 +6,6 @@ export class Configuration extends ConfigBase implements IConfiguration { // Use
 	CI_LOGGER = false;
 	DEBUG = false;
 	TYPESCRIPT_COMPILER_OPTIONS = {};
-	USE_PROXY = false;
 	ANDROID_DEBUG_UI: string = null;
 	USE_POD_SANDBOX: boolean = false;
 	debugLivesync: boolean = false;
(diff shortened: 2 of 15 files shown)


docs/man_pages/general/autocomplete-disable.md
```diff
@@ -20,4 +20,7 @@ Command | Description
 [error-reporting](error-reporting.html) | Configures anonymous error reporting for the NativeScript CLI.
 [help](help.html) | Lists the available commands or shows information about the selected command.
 [doctor](doctor.html) | Checks your system for configuration problems which might prevent the NativeScript CLI from working properly.
+[proxy](proxy.html) | Displays proxy settings.
+[proxy clear](proxy-clear.html) | Clears proxy settings.
+[proxy set](proxy-set.html) | Sets proxy settings.
```
yyosifov: the `proxy get` or `proxy list` command is not documented anywhere?

yyosifov: How do I update a proxy to be enabled/disabled? Can I do it via command or by manually editing the config.json?

lib/config.ts
```diff
@@ -6,7 +6,6 @@ export class Configuration extends ConfigBase implements IConfiguration { // Use
 	CI_LOGGER = false;
 	DEBUG = false;
 	TYPESCRIPT_COMPILER_OPTIONS = {};
-	USE_PROXY = false;
```
Plamen5kov: do we consider having a way to keep this functionality of enable/disable proxy?

Plamen5kov: I believe a short article describing what is using a proxy good for and how to set it up, would be helpful for the users.

rosen-vladimirov: The functionality is introduced as command, so there's no need to manually edit any file. When you use the command, it will create required file (not the config.json ), so there's no need for the user to do anything else - just use the commands.
Where do you think is the appropriate place for such an article?

docs/man_pages/general/autocomplete-disable.md
rosen-vladimirov: You either set it or clear it.

rosen-vladimirov: there's no proxy list command - the default command is `proxy get` , so in case you write just `tns proxy` it will execute `proxy get` command.

lib/config.ts
Plamen5kov: Somewhere in the docs, better ask @PetyaSotirova

Plamen5kov: > The functionality is introduced as command, so there's no need to manually edit any file. 

I ment will we still have a way to enable/disable the proxy in any way through the newly created command.

rosen-vladimirov: There's no way to disable it - set it or clear it. The problem with disabling is that the credentials will be kept when they are not used, which is somehow a security risk. That's why we've decided to skip the disabling - set or clear only.

Issue #2654, opened 2017-03-24
`tns run ios --justlaunch` does not release the console


dtopuzov:
Steps to reproduce:
```
tns create TestApp
tns run ios --path TestApp --emulator --justlaunch
```
Expected:
Console is released when app is started

Actual:
Console is not released.

It looks valid only for `tns run ios --justlaunch`, `tns run android --justlaunch` is fine.

Found in `nativescript@3.0.0-2017.3.23.4`

PR #2662, opened 2017-03-27
Fix ios-device-operations dispose

TsvetanMilanov:
We need to check if the justlaunch flag is passed before we set the should dispose property to false. If we set it to false unconditionally when we pass justlaunch flag the process will never exit.

Fixes https://github.com/NativeScript/nativescript-cli/issues/2654

### DIFF
--- a/lib/commands/debug.ts
+++ b/lib/commands/debug.ts
@@ -83,9 +83,10 @@ export class DebugIOSCommand extends DebugPlatformCommand {
 		$platformService: IPlatformService,
 		$options: IOptions,
 		$projectData: IProjectData,
-		$platformsData: IPlatformsData) {
-
+		$platformsData: IPlatformsData,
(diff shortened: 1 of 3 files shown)


lib/commands/debug.ts
```diff
@@ -83,9 +83,10 @@ export class DebugIOSCommand extends DebugPlatformCommand {
 		$platformService: IPlatformService,
 		$options: IOptions,
 		$projectData: IProjectData,
-		$platformsData: IPlatformsData) {
-
+		$platformsData: IPlatformsData,
+		private $iosDeviceOperations: IIOSDeviceOperations) {
```
rosen-vladimirov: why is this private? It's used only in constructor

hypery2k:
is this fix going to be back ported to 2.5?

PR #2663, opened 2017-03-27
skip the hidden files when livesyncing

yyosifov:
Add filter for files which are hidden in linux/Mac to be skipped when livesync-ing the app. Some IDEs create them to backup the currently edited file and this causes a re-deploy/re-run of the Application although it's not necessary.


rosen-vladimirov:
@Plamen5kov platformLiveSyncService is the correct service used by NativeScript CLI:
https://github.com/NativeScript/nativescript-cli/blob/master/lib/services/livesync/livesync-service.ts#L82

yyosifov:
@Plamen5kov Sure, we can debug together then :)

Plamen5kov:
My bad, debugging was broken on my machine.

PR #2665, opened 2017-03-28
Fix watch incorrectly being overrided by justlaunch always

yyosifov:
The Watch Option was completely overriden by the Justlaunch option, every time. This is a newly introduced issue. Now, the justlaunch will take precedence if it's set to true and the watch will be set to false in this case, no matter what's passed for watch.

### DIFF
--- a/lib/options.ts
+++ b/lib/options.ts
@@ -62,7 +62,7 @@ export class Options extends commonOptionsLibPath.OptionsBase {
 
 		let that = (<any>this);
 		// if justlaunch is set, it takes precedence over the --watch flag and the default true value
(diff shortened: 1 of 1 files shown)


Plamen5kov:
@yyosifov what's the case `--justlaunch` isn't working as expected?
I tried running: `tns run android --justlaunch` and the console is interrupted as expected.

yyosifov:
@Plamen5kov  Justlaunch is working as expected. What happened is that someone set `watch = !justlaunch` which means that if I say `--no-watch` it still will be set to the value of `justlaunch`, which by default is `false`.

PR #2677, opened 2017-03-30
Add debug method to the public API

TsvetanMilanov:
We need to add debug service to the public API of the {N} CLI. This service should have method which starts debugging and returns the chrome devtools url. With this service it will be possible to use the CLI as a library and start debugging.
This PR includes:
- Return the results from the action in devicesService.execute(action)
- Remove the this.$options from both android and ios debug services
- Remove the projectData dependency from both android and ios debug services
- Add the new debug service which will be used in the public API

Common lib reference: https://github.com/telerik/mobile-cli-lib/pull/927

### DIFF
--- a/lib/services/android-debug-service.ts
+++ b/lib/services/android-debug-service.ts
@@ -1,4 +1,4 @@
-import { sleep, getAvailablePort } from "../common/helpers";
+import { sleep } from "../common/helpers";
 import { ChildProcess } from "child_process";
 import { DebugServiceBase } from "./debug-service-base";
 
@@ -23,7 +23,8 @@ class AndroidDebugService extends DebugServiceBase implements IPlatformDebugServ
 		private $logger: ILogger,
 		private $config: IConfiguration,
 		private $androidDeviceDiscovery: Mobile.IDeviceDiscovery,
-		private $androidProcessService: Mobile.IAndroidProcessService) {
+		private $androidProcessService: Mobile.IAndroidProcessService,
+		private $net: INet) {
 		super();
 	}
 
@@ -69,7 +70,7 @@ class AndroidDebugService extends DebugServiceBase implements IPlatformDebugServ
 		if (match) {
 			port = parseInt(match[1]);
 		} else {
-			port = await getAvailablePort(40000);
+			port = await this.$net.getAvailablePortInRange(40000);
 
 			await this.unixSocketForward(port, `${unixSocketName}`);
 		}
@@ -127,7 +128,7 @@ class AndroidDebugService extends DebugServiceBase implements IPlatformDebugServ
 		let startDebuggerCommand = ["am", "broadcast", "-a", `\"${packageName}-debug\"`, "--ez", "enable", "true"];
 		await this.device.adb.executeShellCommand(startDebuggerCommand);
 
-		if (debugOptions.client) {
+		if (debugOptions.chrome) {
 			let port = await this.getForwardedLocalDebugPortForPackageName(deviceId, packageName);
 			return `chrome-devtools://devtools/bundled/inspector.html?experiments=true&ws=localhost:${port}`;
 		}
--- a/lib/services/android-debug-service.ts
+++ b/lib/services/android-debug-service.ts
@@ -1,20 +1,10 @@
-import * as net from "net";
-import * as os from "os";
-import { sleep } from "../common/helpers";
+import { sleep, getAvailablePort } from "../common/helpers";
 import { ChildProcess } from "child_process";
+import { DebugServiceBase } from "./debug-service-base";
 
-class AndroidDebugService implements IDebugService {
+class AndroidDebugService extends DebugServiceBase implements IPlatformDebugService {
 	private _device: Mobile.IAndroidDevice = null;
 	private _debuggerClientProcess: ChildProcess;
-
-	constructor(private $devicesService: Mobile.IDevicesService,
-		private $platformService: IPlatformService,
-		private $platformsData: IPlatformsData,
-		private $logger: ILogger,
-		private $options: IOptions,
-		private $errors: IErrors,
-		private $config: IConfiguration,
-		private $androidDeviceDiscovery: Mobile.IDeviceDiscovery) { }
 
 	public get platform() {
 		return "android";
@@ -28,51 +18,42 @@ class AndroidDebugService implements IDebugService {
 		this._device = newDevice;
 	}
 
-	public async debug(projectData: IProjectData, buildConfig: IBuildConfig): Promise<void> {
-		return this.$options.emulator
-			? this.debugOnEmulator(projectData, buildConfig)
-			: this.debugOnDevice(projectData, buildConfig);
+	constructor(private $devicesService: Mobile.IDevicesService,
+		private $errors: IErrors,
+		private $logger: ILogger,
+		private $config: IConfiguration,
+		private $androidDeviceDiscovery: Mobile.IDeviceDiscovery,
+		private $androidProcessService: Mobile.IAndroidProcessService) {
+		super();
 	}
 
-	private async debugOnEmulator(projectData: IProjectData, buildConfig: IBuildConfig): Promise<void> {
+	public async debug(debugData: IDebugData, debugOptions: IDebugOptions): Promise<string[]> {
+		return debugOptions.emulator
+			? this.debugOnEmulator(debugData, debugOptions)
+			: this.debugOnDevice(debugData, debugOptions);
+	}
+
+	public async debugStart(debugData: IDebugData, debugOptions: IDebugOptions): Promise<void> {
+		await this.$devicesService.initialize({ platform: this.platform, deviceId: debugData.deviceIdentifier });
+		let action = (device: Mobile.IAndroidDevice): Promise<void> => {
+			this.device = device;
+			return this.debugStartCore(debugData.applicationIdentifier, debugOptions);
+		};
+
+		await this.$devicesService.execute(action, this.getCanExecuteAction(debugData.deviceIdentifier));
+	}
+
+	public async debugStop(): Promise<void> {
+		this.stopDebuggerClient();
+		return;
+	}
+
+	private async debugOnEmulator(debugData: IDebugData, debugOptions: IDebugOptions): Promise<string[]> {
 		// Assure we've detected the emulator as device
 		// For example in case deployOnEmulator had stated new emulator instance
 		// we need some time to detect it. Let's force detection.
 		await this.$androidDeviceDiscovery.startLookingForDevices();
-		await this.debugOnDevice(projectData, buildConfig);
-	}
-
-	private isPortAvailable(candidatePort: number): Promise<boolean> {
-		return new Promise<boolean>((resolve, reject) => {
-			let isResolved = false;
-			let server = net.createServer();
-
-			server.on("error", (err: Error) => {
-				if (!isResolved) {
-					isResolved = true;
-					resolve(false);
-				}
-			});
-
-			server.once("close", () => {
-				if (!isResolved) { // "close" will be emitted right after "error"
-					isResolved = true;
-					resolve(true);
-				}
-			});
-
-			server.on("listening", (err: Error) => {
-				if (err && !isResolved) {
-					isResolved = true;
-					resolve(false);
-				}
-
-				server.close();
-			});
-
-			server.listen(candidatePort, "localhost");
-
-		});
+		return this.debugOnDevice(debugData, debugOptions);
 	}
 
 	private async getForwardedLocalDebugPortForPackageName(deviceId: string, packageName: string): Promise<number> {
@@ -88,15 +69,7 @@ class AndroidDebugService implements IDebugService {
 		if (match) {
 			port = parseInt(match[1]);
 		} else {
-			let candidatePort = 40000;
-
-			for (; ! await this.isPortAvailable(candidatePort); ++candidatePort) {
-				if (candidatePort > 65534) {
-					this.$errors.failWithoutHelp("Unable to find free local port.");
-				}
-			}
-
-			port = candidatePort;
+			port = await getAvailablePort(40000);
 
 			await this.unixSocketForward(port, `${unixSocketName}`);
 		}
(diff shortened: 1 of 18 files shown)


lib/services/debug-data-service.ts
```diff
@@ -0,0 +1,43 @@
...
+			applicationIdentifier: this.$projectData.projectId,
+			projectDir: this.$projectData.projectDir,
+			deviceIdentifier: options.device,
+			pathToAppPackage: this.getPathToAppPackage(debugService, options, buildConfig),
+			projectName: this.$projectData.projectName
+		};
+	}
+
+	private getPathToAppPackage(debugService: IPlatformDebugService, options: IOptions, buildConfig: IBuildConfig): string {
+		if (debugService.platform === this.$devicePlatformsConstants.Android) {
```
Mitko-Kerezov: I think it'd be better to use `$mobileHelper.isAndroidPlatform` here in case the casing differs

lib/services/debug-data-service.ts
```diff
@@ -0,0 +1,43 @@
...
+			if (!options.start && !options.emulator) {
+				const platformData = this.getPlatformData(debugService);
+
+				return this.$platformService.getLatestApplicationPackageForDevice(platformData, buildConfig).packageName;
+			} else {
+				return null;
+			}
+		} else if (debugService.platform === this.$devicePlatformsConstants.iOS) {
```
Mitko-Kerezov: Same here `$mobileHelper.isiOSPlatform`

lib/services/debug-service-base.ts
```diff
@@ -0,0 +1,25 @@
+import { EventEmitter } from "events";
+
+export abstract class DebugServiceBase extends EventEmitter implements IPlatformDebugService {
+	constructor() {
```
Mitko-Kerezov: Isn't such a constructor gratuitous?

lib/services/debug-service.ts
```diff
@@ -0,0 +1,65 @@
...
+			this.$errors.failWithoutHelp(`Can't find device with identifier ${debugData.deviceIdentifier}`);
+		}
+
+		const debugOptions: IDebugOptions = _.merge({}, options || {});
+		debugOptions.start = true;
+
+		// TODO: Check if app is running.
+		const isAppRunning = true;
+		let result: string[];
+		if (device.deviceInfo.platform === this.$devicePlatformsConstants.iOS) {
```
Mitko-Kerezov: Same remark about `$mobileHelper.isiOSPlatform`

lib/services/debug-service.ts
```diff
@@ -0,0 +1,65 @@
...
+					this.$errors.failWithoutHelp(`Application ${debugData.applicationIdentifier} is not running. To be able to debug the application on Windows you must run it.`);
+				}
+
+				debugOptions.emulator = false;
+			} else if (!this.$hostInfo.isDarwin) {
+				this.$errors.failWithoutHelp(`Debugging on iOS devices is not supported for ${platform()} yet.`);
+			}
+
+			result = await debugService.debug(debugData, debugOptions);
+		} else if (device.deviceInfo.platform === this.$devicePlatformsConstants.Android) {
```
Mitko-Kerezov: `$mobileHelper`

lib/services/ios-debug-service.ts
```diff
@@ -21,55 +23,55 @@ class IOSDebugService implements IDebugService {
...
 			}
 		}
 	}
 
-	public async debugStart(projectData: IProjectData, buildConfig: IBuildConfig): Promise<void> {
-		await this.$devicesService.initialize({ platform: this.platform, deviceId: this.$options.device });
-		this.$devicesService.execute(async (device: Mobile.IiOSDevice) => await device.isEmulator ? this.emulatorDebugBrk(projectData, buildConfig) : this.debugBrkCore(device, projectData));
+	public async debugStart(debugData: IDebugData, debugOptions: IDebugOptions): Promise<void> {
+		await this.$devicesService.initialize({ platform: this.platform, deviceId: debugData.deviceIdentifier });
+		const action = async (device: Mobile.IiOSDevice) => await device.isEmulator ? this.emulatorDebugBrk(debugData, false, debugOptions) : this.debugBrkCore(device, debugData, debugOptions);
```
Mitko-Kerezov: The `await` operator here seems misplaced

lib/services/test-execution-service.ts
```diff
@@ -128,22 +130,33 @@ class TestExecutionService implements ITestExecutionService {
...
-					release: this.$options.release,
-					provision: this.$options.provision,
-					teamId: this.$options.teamId
-				};
-
-				const buildConfig: IBuildConfig = _.merge({ buildForDevice: this.$options.forDevice }, deployOptions);
-
 				if (this.$options.debugBrk) {
-					await this.getDebugService(platform).debug(projectData, buildConfig);
+					const debugService = this.getDebugService(platform); const deployOptions: IDeployPlatformOptions = {
```
Mitko-Kerezov: `deployOptions` can be declared above the `if` statement in order to avoid repetition

test/debug.ts
```diff
@@ -43,6 +43,9 @@ function createTestInjector(): IInjector {
 	testInjector.register("projectTemplatesService", {});
 	testInjector.register("xmlValidator", {});
 	testInjector.register("npm", {});
+	testInjector.register("debugDataService", {
+		createDebugData: () => { return {}; }
```
Mitko-Kerezov: You can use the short-hand syntax here: 
`createDebugData: () => ({})`

lib/services/debug-data-service.ts
```diff
@@ -0,0 +1,43 @@
...
+		};
+	}
+
+	private getPathToAppPackage(debugService: IPlatformDebugService, options: IOptions, buildConfig: IBuildConfig): string {
+		if (debugService.platform === this.$devicePlatformsConstants.Android) {
+			if (!options.start && !options.emulator) {
+				const platformData = this.getPlatformData(debugService);
+
+				return this.$platformService.getLatestApplicationPackageForDevice(platformData, buildConfig).packageName;
+			} else {
```
rosen-vladimirov: do we really need the else? Can't you place the `return null` at the end of the method?

lib/services/debug-data-service.ts
```diff
@@ -0,0 +1,43 @@
...
+				const platformData = this.getPlatformData(debugService);
+
+				return this.$platformService.getLatestApplicationPackageForEmulator(platformData, buildConfig).packageName;
+			} else {
+				return null;
+			}
+		}
+	}
+
+	protected getPlatformData(debugService: IPlatformDebugService): IPlatformData {
```
rosen-vladimirov: why is this method protected and not private?

lib/services/debug-service.ts
```diff
@@ -0,0 +1,65 @@
+import { platform } from "os";
+import { EventEmitter } from "events";
+import { CONNECTION_ERROR_EVENT_NAME } from "../constants";
+
+class DebugService extends EventEmitter {
```
rosen-vladimirov: shouldn't you implement IDebugService  as well ?

lib/services/debug-service.ts
```diff
@@ -0,0 +1,65 @@
...
+		private $hostInfo: IHostInfo,
+		private $devicePlatformsConstants: Mobile.IDevicePlatformsConstants) {
+		super();
+	}
+
+	public async debug(debugData: IDebugData, options: IDebugOptions): Promise<string> {
+		const device = this.$devicesService.getDeviceByIdentifier(debugData.deviceIdentifier);
+		const debugService = this.getDebugService(device);
+
+		debugService.on(CONNECTION_ERROR_EVENT_NAME, (e: Error) => this.emit(CONNECTION_ERROR_EVENT_NAME, e));
```
rosen-vladimirov: who will remove this handler? In case you call `debug` method three times, you'll have three handlers for the error event. So in case an error is raised, the one who listens for DebugService's `CONNECTION_ERROR_EVENT_NAME` event will receive three errors.

rosen-vladimirov: Also why the handler is attached before the `if (!device)` statement. It looks like the handler has meaning only when we have a device.

lib/services/debug-service.ts
```diff
@@ -0,0 +1,65 @@
...
+
+		if (!device) {
+			this.$errors.failWithoutHelp(`Can't find device with identifier ${debugData.deviceIdentifier}`);
+		}
+
+		const debugOptions: IDebugOptions = _.merge({}, options || {});
```
rosen-vladimirov: I believe you shouldn't worry if options is undefined, lodash works fine in these cases:
```TypeScript
const debugOptions: IDebugOptions = _.merge({}, options);
```

lib/services/debug-service.ts
```diff
@@ -0,0 +1,65 @@
...
+		debugOptions.start = true;
+
+		// TODO: Check if app is running.
```
rosen-vladimirov: Can you provide info that it's currently not possible for iOS (describe why it is TODO and why we have not done it).

lib/services/debug-service.ts
```diff
@@ -0,0 +1,65 @@
...
+		const isAppRunning = true;
+		let result: string[];
+		if (device.deviceInfo.platform === this.$devicePlatformsConstants.iOS) {
+			debugOptions.chrome = true;
```
rosen-vladimirov: why we always set it to true? What happened with the safari (ios) inspector?

lib/services/debug-service.ts
```diff
@@ -0,0 +1,65 @@
...
+				}
+
+				debugOptions.emulator = false;
+			} else if (!this.$hostInfo.isDarwin) {
+				this.$errors.failWithoutHelp(`Debugging on iOS devices is not supported for ${platform()} yet.`);
+			}
+
+			result = await debugService.debug(debugData, debugOptions);
+		} else if (device.deviceInfo.platform === this.$devicePlatformsConstants.Android) {
+			debugOptions.client = true;
```
rosen-vladimirov: shouldn't this be valid for iOS instead of Android?

lib/services/debug-service.ts
rosen-vladimirov: Maybe this should be `debugOptions.client`?

lib/services/ios-debug-service.ts
```diff
@@ -123,86 +123,99 @@ class IOSDebugService implements IDebugService {
...
 
 		let iOSEmulator = <Mobile.IiOSSimulatorService>this.$iOSEmulatorServices;
 		await iOSEmulator.postDarwinNotification(attachRequestMessage);
+		return result;
 	}
 
-	private async deviceDebugBrk(projectData: IProjectData, buildConfig: IBuildConfig, shouldBreak?: boolean): Promise<void> {
-		await this.$devicesService.initialize({ platform: this.platform, deviceId: this.$options.device });
-		this.$devicesService.execute(async (device: iOSDevice.IOSDevice) => {
+	private async deviceDebugBrk(debugData: IDebugData, shouldBreak: boolean, debugOptions: IDebugOptions): Promise<string[]> {
```
rosen-vladimirov: isn't `shouldBreak` a debug option? Can't we place it in debugOptions object?
P.S. I do not like passing boolean arguments.

lib/services/ios-debug-service.ts
```diff
@@ -123,86 +123,99 @@ class IOSDebugService implements IDebugService {
...
+			await startApplicationAction;
 
-	private async debugBrkCore(device: Mobile.IiOSDevice, projectData: IProjectData, shouldBreak?: boolean): Promise<void> {
-		let timeout = this.$utils.getMilliSecondsTimeout(TIMEOUT_SECONDS);
-		await this.$iOSSocketRequestExecutor.executeLaunchRequest(device.deviceInfo.identifier, timeout, timeout, projectData.projectId, shouldBreak);
-		await this.wireDebuggerClient(projectData, device);
+			return result;
+		};
+
+		const result = await this.$devicesService.execute(action, this.getCanExecuteAction(debugData.deviceIdentifier));
```
rosen-vladimirov: result`s` sounds better

lib/services/ios-debug-service.ts
```diff
@@ -123,86 +123,99 @@ class IOSDebugService implements IDebugService {
...
 	}
 
-	private async deviceStartCore(device: Mobile.IiOSDevice, projectData: IProjectData): Promise<void> {
-		let timeout = this.$utils.getMilliSecondsTimeout(TIMEOUT_SECONDS);
-		await this.$iOSSocketRequestExecutor.executeAttachRequest(device, timeout, projectData.projectId);
-		await this.wireDebuggerClient(projectData, device);
+	private async deviceStart(debugData: IDebugData, debugOptions: IDebugOptions): Promise<string[]> {
+		await this.$devicesService.initialize({ platform: this.platform, deviceId: debugData.deviceIdentifier });
+		const action = async (device: Mobile.IiOSDevice) => device.isEmulator ? await this.emulatorStart(debugData, debugOptions) : await this.deviceStartCore(device, debugData, debugOptions);
+		const result = await this.$devicesService.execute(action, this.getCanExecuteAction(debugData.deviceIdentifier));
```
rosen-vladimirov: results

lib/services/test-execution-service.ts
rosen-vladimirov: That's why they've been declared there so far 😃

lib/services/android-debug-service.ts
```diff
@@ -88,15 +69,7 @@ class AndroidDebugService implements IDebugService {
...
-			let candidatePort = 40000;
-
-			for (; ! await this.isPortAvailable(candidatePort); ++candidatePort) {
-				if (candidatePort > 65534) {
-					this.$errors.failWithoutHelp("Unable to find free local port.");
-				}
-			}
-
-			port = candidatePort;
+			port = await getAvailablePort(40000);
```
TsvetanMilanov: Because we want to start chrome devtools on port 40000 each time we start debugging. That's why we don't get random free port. If we get random free port, the users should copy the url each time `tns debug` is executed.

lib/services/debug-service.ts
TsvetanMilanov: This debug service has different debug method. In ios/android debug service the debug method returns array of the chrome devtools urls. In this service we want to start debugging only on one device and return the url only for it.

lib/services/debug-service.ts
TsvetanMilanov: `debugOptions.client` in the android debug service is the same as `debugOptions.chrome` in ios debug service. I changed the android debug service to use `debugOptions.chrome`.

Issue #2668, opened 2017-03-28
Add support for Xcode 8.3


jamessharp:
*iOS only* (tns 2.5.3)

Xcode 8.3 has fully removed the ability to do an `xcrun PackageApplication` as is performed when building/running a nativescript app on a development device (see [this line of code here](https://github.com/NativeScript/nativescript-cli/blob/v2.5.3/lib/services/ios-project-service.ts#L431)). This means that `tns run ios` and friends don't work under Xcode 8.3, failing with the following error:

```
xcrun: error: unable to find utility "PackageApplication", not a developer tool or in PATH
Command xcrun failed with exit code 72
```

Some googling around has suggested moving to using `xcodebuild`, and the cordova guys have done the update in this PR: https://github.com/apache/cordova-ios/pull/257

Let me know if you need any further info

NickIliev:
Also reported via t.1100681

svzi:
Any feedback on this? Specially on how to fix would be really much appreciated!

It happens on my machine as well and now I'm unable to run apps on my iOS device... This is really a deal-breaker!

dtopuzov:
We work on fixing the issue and we will provide fix for for {N} CLI 3.0.0 and hot fix for {N} 2.5.*

mn-martin:
> and we will also release hot fix for {N} 2.5.*

Really glad to hear that because we're close to release of an app and we wont have the time for an update to 3.0.0. I'm sure we're not alone with this scenario.

dxshindeo:
> Really glad to hear that because we're close to release of an app and we wont have the time for an update to 3.0.0. I'm sure we're not alone with this scenario.
Same...

teerasej:
> Really glad to hear that because we're close to release of an app and we wont have the time for an update to 3.0.0. I'm sure we're not alone with this scenario.

Same here...

dunqan:
As a temporary workaround, you can still build and run your app by opening workspace directly in XCode. Especially for the release purposes.

mn-martin:
@dunqan Thank you for that hint. By the way: The run command seems to work just fine with the iOS emulator. (Which helps shorten the dev-times compared to the Xcode workspace route if features can be tested within the emulator)

dxshindeo:
> As a temporary workaround, you can still build and run your app by opening workspace in XCode directly. Especially for the release purposes.

Doesn't work for me. 
[...]

PR #2678, opened 2017-03-31
Plamen5kov/use xcodebuild instead of xcrun

yyosifov:
xccrun PackageApplication is now deprecated in XCode 8.3, therefore we're switching to xcodebuild and execute two steps:

- archive the application
- export the `.xarchive` to `.ipa`

related issue: https://github.com/NativeScript/nativescript-cli/issues/2668

### DIFF
--- a/lib/services/ios-project-service.ts
+++ b/lib/services/ios-project-service.ts
@@ -248,8 +248,7 @@ export class IOSProjectService extends projectServiceBaseLib.PlatformProjectServ
 			"-exportPath", exportPath,
 			"-exportOptionsPlist", platformData.configurationFilePath
 		];
-
-		await this.$childProcess.spawnFromEvent("xcodebuild", args, "exit", 
+		await this.$childProcess.spawnFromEvent("xcodebuild", args, "exit",
 			{ stdio: buildConfig.buildOutputStdio || 'inherit', cwd: this.getPlatformData(projectData).projectRoot },
 			{ emitOptions: { eventName: constants.BUILD_OUTPUT_EVENT_NAME }, throwError: false });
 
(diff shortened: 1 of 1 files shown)


dtopuzov:
run ci

dtopuzov:
run ci

lib/services/ios-project-service.ts
```diff
@@ -230,6 +230,31 @@ export class IOSProjectService extends projectServiceBaseLib.PlatformProjectServ
...
+		let exportFile = path.join(exportPath, projectData.projectName + ".ipa");
+
+		let args = ["-exportArchive",
+			"-archivePath", archivePath,
+			"-exportPath", exportPath,
+			"-exportOptionsPlist", platformData.configurationFilePath
+		];
+		await this.$childProcess.spawnFromEvent("xcodebuild", args, "exit",
+			{ stdio: buildConfig.buildOutputStdio || 'inherit', cwd: this.getPlatformData(projectData).projectRoot },
+			{ emitOptions: { eventName: constants.BUILD_OUTPUT_EVENT_NAME }, throwError: false });
```
rosen-vladimirov: shouldn't throwError be true here?

lib/services/ios-project-service.ts
```diff
@@ -230,6 +230,31 @@ export class IOSProjectService extends projectServiceBaseLib.PlatformProjectServ
 		return exportFile;
 	}
 
+	/**
+	 * Exports .xcarchive for AppStore distribution.
+	 */
+	public async exportDevelopmentArchive(projectData: IProjectData, buildConfig: IBuildConfig, options: { archivePath: string, exportDir?: string, teamID?: string }): Promise<string> {
```
rosen-vladimirov: if it's public, it should be added to the interface

yyosifov: @rosen-vladimirov This doesn't have to be public

lib/services/ios-project-service.ts
yyosifov: @rosen-vladimirov Not sure, we've set this in the other spawn. I wanted to keep the flags the same?

rosen-vladimirov: @yyosifov this has been a mistake - it should be true whenever we want to break the operation (i.e. the next operation has no meaning in case this one fails)

yyosifov:
@rosen-vladimirov you can have another look, please :)

PR #2690, opened 2017-04-04
Track the amount and device type of builds and livesync operations

rosen-vladimirov:
In order to unsderstand the workflow of our users, we need to track how many builds they are executing, how many livesync operations, etc. We also need to know if the operation is for device or emulator (iOSSimulator).
This will allow us to focus the development and provide better features in the future.

### DIFF
--- a/lib/services/livesync/platform-livesync-service.ts
+++ b/lib/services/livesync/platform-livesync-service.ts
@@ -12,10 +12,6 @@ export abstract class PlatformLiveSyncServiceBase implements IPlatformLiveSyncSe
 
 	protected liveSyncData: ILiveSyncData;
 
-	private get $analyticsService(): IAnalyticsService {
-		return this.$injector.resolve("analyticsService");
-	}
-
 	constructor(_liveSyncData: ILiveSyncData,
 		private $devicesService: Mobile.IDevicesService,
 		private $mobileHelper: Mobile.IMobileHelper,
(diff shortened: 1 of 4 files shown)


lib/services/livesync/platform-livesync-service.ts
```diff
@@ -39,6 +43,10 @@ export abstract class PlatformLiveSyncServiceBase implements IPlatformLiveSyncSe
 			let canExecute = this.getCanExecuteAction(platform, appIdentifier);
 			let action = (device: Mobile.IDevice): IFuture<void> => {
 				return (() => {
+					const normalizePlatformName = this.$mobileHelper.normalizePlatformName(platform);
```
yyosifov: Do you think it would be possible to extract some function which can track both events? (build/livesync) Do we have other places where we can track such information?

lib/services/livesync/platform-livesync-service.ts
```diff
@@ -39,6 +43,10 @@ export abstract class PlatformLiveSyncServiceBase implements IPlatformLiveSyncSe
...
+					const deviceType = device.isEmulator ? "emulator" : "device";
```
yyosifov: from the Device do we get to know the particular OS of the device? Do we want to track this?

rosen-vladimirov: we can get such information, but I'm not sure we want it at the moment

lib/services/livesync/platform-livesync-service.ts
rosen-vladimirov: @yyosifov I can introduce new method in platform-service and call it from here. 
I've tried to find the best place for tracking LiveSync operation and it seems this is the one - only the "full sync" will be tracked (the first operation that's executed when you call `tns run <platform>`. In case you break it and call it again, we'll track it again. In case you change any file during `tns run <platform>` is working, we'll not track it, as we've already sent data for it.

rosen-vladimirov:
ping @yyosifov  - I've introduced new method in platformService and I've also added track for Deploy operation and for the device OS version.

dtopuzov:
run ci

PR #2693, opened 2017-04-06
Install karma peer dependencies on `test init`

Pip3r4o:
### Problem:
Often when running `tns test init` it is unclear that there are unmet peer dependencies that are critical to `karma`'s proper execution.

### Proposed Solution:
When executing `test init` in a NativeScript, project karma testing frameworks' (`karma-jasmine/qunit/mocha`) peer dependencies will be implicitly installed. These dependencies are `jasmine-core`, `qunit`, and `mocha` respectively, depending on the framework selected from the interactive console prompt.

### DIFF
--- a/lib/commands/test-init.ts
+++ b/lib/commands/test-init.ts
@@ -53,7 +53,7 @@ class TestInitCommand implements ICommand {
 						'save-dev': true
 					});
 				} catch (e) {
-					this.$logger.info(e.message);
+					this.$logger.error(e.message);
 				}
 			}
 		}
--- a/lib/commands/test-init.ts
+++ b/lib/commands/test-init.ts
@@ -37,6 +37,18 @@ class TestInitCommand implements ICommand {
 				'save-dev': true,
 				optional: false,
 			});
+
+			let modulePath = path.join(projectDir, "node_modules", mod);
+			let modulePackageJsonPath = path.join(modulePath, "package.json");
+			let modulePackageJsonContent = this.$fs.readJson(modulePackageJsonPath);
+			let modulePeerDependencies = modulePackageJsonContent.peerDependencies || {};
+
+			for (let peerDependency in modulePeerDependencies) {
+				let dependencyVersion = modulePeerDependencies[peerDependency] || "*";
+				await this.$npm.install(`${peerDependency}@${dependencyVersion}`, projectDir, {
+					'save-dev': true
+				});
+			}
 		}
 
 		await this.$pluginsService.add('nativescript-unit-test-runner', this.$projectData);
--- a/lib/commands/test-init.ts
+++ b/lib/commands/test-init.ts
@@ -38,16 +38,23 @@ class TestInitCommand implements ICommand {
 				optional: false,
 			});
 
-			let modulePath = path.join(projectDir, "node_modules", mod);
-			let modulePackageJsonPath = path.join(modulePath, "package.json");
-			let modulePackageJsonContent = this.$fs.readJson(modulePackageJsonPath);
-			let modulePeerDependencies = modulePackageJsonContent.peerDependencies || {};
+			const modulePath = path.join(projectDir, "node_modules", mod);
+			const modulePackageJsonPath = path.join(modulePath, "package.json");
+			const modulePackageJsonContent = this.$fs.readJson(modulePackageJsonPath);
+			const modulePeerDependencies = modulePackageJsonContent.peerDependencies || {};
 
 			for (let peerDependency in modulePeerDependencies) {
 				let dependencyVersion = modulePeerDependencies[peerDependency] || "*";
-				await this.$npm.install(`${peerDependency}@${dependencyVersion}`, projectDir, {
-					'save-dev': true
-				});
+
+				// catch errors when a peerDependency is already installed
+				// e.g karma is installed; karma-jasmine depends on karma and will try to install it again
+				try {
+					await this.$npm.install(`${peerDependency}@${dependencyVersion}`, projectDir, {
+						'save-dev': true
+					});
+				} catch (e) {
+					this.$logger.info(e.message);
+				}
 			}
 		}
 
--- a/lib/node-package-manager.ts
+++ b/lib/node-package-manager.ts
@@ -76,10 +76,10 @@ export class NodePackageManager implements INodePackageManager {
 		let diff = dependencyDiff.concat(devDependencyDiff);
 
 		if (diff.length <= 0 && dependenciesBefore.length === dependenciesAfter.length && packageName !== pathToSave) {
-			this.$errors.failWithoutHelp(`The plugin ${packageName} is already installed`);
+			this.$logger.warn(`The plugin ${packageName} is already installed`);
 		}
 		if (diff.length <= 0 && dependenciesBefore.length !== dependenciesAfter.length) {
-			this.$errors.failWithoutHelp(`Couldn't install package correctly`);
+			this.$logger.warn(`Couldn't install package ${packageName} correctly`);
 		}
 
 		return diff;


lib/commands/test-init.ts
```diff
@@ -37,6 +37,25 @@ class TestInitCommand implements ICommand {
 				'save-dev': true,
 				optional: false,
 			});
+
+			let modulePath = path.join(projectDir, "node_modules", mod);
```
yyosifov: These all can be "const"

lib/commands/test-init.ts
```diff
@@ -37,6 +37,25 @@ class TestInitCommand implements ICommand {
...
+			let modulePackageJsonPath = path.join(modulePath, "package.json");
+			let modulePackageJsonContent = this.$fs.readJson(modulePackageJsonPath);
+			let modulePeerDependencies = modulePackageJsonContent.peerDependencies || {};
```
yyosifov: shouldn't it be || [] to be slightly more-correct

lib/commands/test-init.ts
```diff
@@ -37,6 +37,25 @@ class TestInitCommand implements ICommand {
...
+
+			for (let peerDependency in modulePeerDependencies) {
```
yyosifov: I think a good practice when iterating an object's keys in JS is to check for own properties only. You can use lodash if you want to: http://lodash.com/docs/4.17.4#forOwn

lib/commands/test-init.ts
```diff
@@ -37,6 +37,25 @@ class TestInitCommand implements ICommand {
...
+				let dependencyVersion = modulePeerDependencies[peerDependency] || "*";
+
+				// catch errors when a peerDependency is already installed
+				// e.g karma is installed; karma-jasmine depends on karma and will try to install it again
+				try {
+					await this.$npm.install(`${peerDependency}@${dependencyVersion}`, projectDir, {
```
yyosifov: I think we should have the dependencyVersion in quotes, so that it works for dependencies defined like this:

    "karma": "1.x || ",
    "karma-chrome-launcher": "1.x || ~0.2.2",
    "karma-firefox-launcher": "1.x || ~0.1.7",

Pip3r4o: Explicitly adding quotes actually confused the `npm.install` call. The `"1.x || ~0.2.2"` style versions seem to work just alright

lib/commands/test-init.ts
Pip3r4o: Thanks for the advice!

Is it really necessary though to check for own property on a JS object when we know it's been parsed from a json file?

lib/commands/test-init.ts
yyosifov: You can use the `npm-installation-manager` and `npmInstall` in order to not have duplicated npm.install statements in our code and avoid errors as the one above

lib/commands/test-init.ts
```diff
@@ -37,6 +37,25 @@ class TestInitCommand implements ICommand {
...
+						'save-dev': true
+					});
+				} catch (e) {
+					this.$logger.info(e.message);
```
yyosifov: if we cannot install the karma dependencies, what happens? will the `test init` work properly? In case not, I'd suggest we use the $errors service here. You can test by installing from local package and adding an invalid package name in the peerDependencies?

lib/commands/test-init.ts
```diff
@@ -37,6 +37,25 @@ class TestInitCommand implements ICommand {
...
+			let modulePeerDependencies = modulePackageJsonContent.peerDependencies || {};
+
+			for (let peerDependency in modulePeerDependencies) {
+				let dependencyVersion = modulePeerDependencies[peerDependency] || "*";
+
+				// catch errors when a peerDependency is already installed
+				// e.g karma is installed; karma-jasmine depends on karma and will try to install it again
+				try {
+					await this.$npm.install(`${peerDependency}@${dependencyVersion}`, projectDir, {
+						'save-dev': true
```
yyosifov: why do we want to save it to `devDependencies`? If the entry point is always the `test init` there's no need to do this?

lib/commands/test-init.ts
```diff
@@ -37,6 +37,25 @@ class TestInitCommand implements ICommand {
...
+			let modulePath = path.join(projectDir, "node_modules", mod);
+			let modulePackageJsonPath = path.join(modulePath, "package.json");
+			let modulePackageJsonContent = this.$fs.readJson(modulePackageJsonPath);
+			let modulePeerDependencies = modulePackageJsonContent.peerDependencies || {};
+
+			for (let peerDependency in modulePeerDependencies) {
+				let dependencyVersion = modulePeerDependencies[peerDependency] || "*";
+
+				// catch errors when a peerDependency is already installed
+				// e.g karma is installed; karma-jasmine depends on karma and will try to install it again
```
yyosifov: If it's already installed, why would it fail?

lib/commands/test-init.ts
Pip3r4o: I don't understand your point. Testing frameworks should be installed as devDependencies so they are not included in built projects, but so that they may still be installed subsequently, for example after removing node_modules.

lib/commands/test-init.ts
yyosifov: @Pip3r4o What i meant is try to run `npm install karma-chrome-launcher@1.x || ~0.2.2` and it will fail. Therefore we need quotes -> `npm install karma-chrome-launcher@"1.x || ~0.2.2"`. The above versions are pasted from https://github.com/karma-runner/karma-jasmine/blob/master/package.json so they are valid.

lib/commands/test-init.ts
yyosifov: Sure, it's ok in this case. But it's good to be a habit when iterating over object's properties, that's why I mentioned it.

lib/commands/test-init.ts
Pip3r4o: If you were to manually run `npm install karma@...` without quotes in the package - it will certainly fail. The `$npm` service however manages the installation with versed appended in the end in the form that you see it.

lib/commands/test-init.ts
yyosifov: You can disregard this. I didn't see the reason to duplicate them from `peerDependencies` - they're already present and installed from their presence there. Why add more changes to the package.json

Pip3r4o: Because otherwise the next time you decide to reinstall node_modules you would have to run `test init` again to ensure the testing framework previously used is installed.

lib/commands/test-init.ts
Pip3r4o: The node-package-manager's install() that I used will fail the build - https://github.com/NativeScript/nativescript-cli/blob/master/lib/node-package-manager.ts#L78

I am unaware why that check is performed, even though the npm package is installed anyway at an earlier point - https://github.com/NativeScript/nativescript-cli/blob/master/lib/node-package-manager.ts#L42.

@rosen-vladimirov can you say what the reasons for that are?

rosen-vladimirov: I'm not really sure why there's an error there. Maybe it should be just warning. The code there is trying to find the name of the installed package based on the difference in package.json before installing the package and after that. In case the package is already installed, there will be no difference in package.json, so we are unable to determine the real name of the installed package (in case you are installing from github url or tarball, you do not know the real name of the package).

Pip3r4o:
@rosen-vladimirov I also made changes to the Node-package-manager so that it no longer throws errors if there is no difference in the package.json before and after installing a dependency.

lib/commands/test-init.ts
rosen-vladimirov: As the comment says, in case the dependency `karma` is already installed, the current implementation will reinstall it, but it may also change its version, which I do not think is expected. For example in case I have installed `karma`: `1.6.0`, but the karma-jasmine has peerDependency: `"karma":"~0.9"`, I'll end-up with the version `~0.9`. Can we skip the installation of dependencies in case we already have the same dependency installed.

Pip3r4o:
@rosen-vladimirov per your comment I get the idea that you are suggesting we have the logic to not reinstall dependencies at all in the node-package-manager service. With its current implementation even if there is no difference in the `before` and `after` of package.json, the package would still be reinstalled.

In the particular example you provided, if you had manually installed `karma` of a different version than the one provided by the `init` command, your version would be overwritten, but then the more recent version's peer dependency of `karma-jasmine` should itself have a valid peerDependency to karma. If we left the karma installation as is, per your example, the project will be left with unmet peer dependencies.

PR #2698, opened 2017-04-06
We passed incorrectly provision in-place of projectId

PanayotCankov:
We pass provision in-place of projectId downstream since any is assignable from everything and everything is assignable to any.

### DIFF
--- a/lib/services/platform-service.ts
+++ b/lib/services/platform-service.ts
@@ -256,18 +256,18 @@ export class PlatformService extends EventEmitter implements IPlatformService {
 		return true;
 	}
 
-	public async validateOptions(provision: any, projectData: IProjectData, platform?: string): Promise<boolean> {
+	public async validateOptions(provision: true | string, projectData: IProjectData, platform?: string): Promise<boolean> {
 		if (platform) {
 			platform = this.$mobileHelper.normalizePlatformName(platform);
 			this.$logger.trace("Validate options for platform: " + platform);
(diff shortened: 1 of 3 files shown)


lib/services/platform-service.ts
```diff
@@ -256,18 +256,18 @@ export class PlatformService extends EventEmitter implements IPlatformService {
 		return true;
 	}
 
-	public async validateOptions(provision: any, projectData: IProjectData, platform?: string): Promise<boolean> {
+	public async validateOptions(provision: true | string, projectData: IProjectData, platform?: string): Promise<boolean> {
```
rosen-vladimirov: I believe the type was any as there's a chance to get here with object, not boolean or string value (at least in the current code in master branch). For example [here](https://github.com/NativeScript/nativescript-cli/blob/ddd8328a30d33e997d975107b58c240d0b0c9a59/lib/commands/build.ts#L49) - in case you call:
```
$ tns build ios --provision.prop1
```
here you'll receive an object: ` { prop1: true }`
Maybe we should guard the `$options.provision`, but I agree to handle this later (separate PR) as it is not expected to pass such values to `--provision`.

PanayotCankov: :+1:

PanayotCankov: And if it were string it would still fail and allow us to switch them.
I can't wait for this: https://github.com/Microsoft/TypeScript/issues/202
It will let us better type such strings and limit where they can be used.

PR #2703, opened 2017-04-07
Detach event handlers after action is done

Mitko-Kerezov:
In order to not emit `buildOutput` event multiple times during build in a long running process (fusion) we have to detach event handlers after build is done.

Merge after https://github.com/telerik/mobile-cli-lib/pull/940

### DIFF
--- a/lib/services/android-project-service.ts
+++ b/lib/services/android-project-service.ts
@@ -4,6 +4,7 @@ import * as constants from "../constants";
 import * as semver from "semver";
 import * as projectServiceBaseLib from "./platform-project-service-base";
 import { DeviceAndroidDebugBridge } from "../common/mobile/android/device-android-debug-bridge";
+import { attachAwaitDetach } from "../common/helpers";
 import { EOL } from "os";
 import { Configurations } from "../common/constants";
 import { SpawnOptions } from "child_process";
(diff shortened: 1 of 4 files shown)


lib/services/android-project-service.ts
```diff
@@ -257,13 +258,17 @@ export class AndroidProjectService extends projectServiceBaseLib.PlatformProject
...
-			});
+			};
 
-			await this.executeGradleCommand(this.getPlatformData(projectData).projectRoot, buildOptions,
-				{ stdio: buildConfig.buildOutputStdio || "inherit" },
-				{ emitOptions: { eventName: constants.BUILD_OUTPUT_EVENT_NAME }, throwError: true });
+			await attachAwaitDetach(constants.BUILD_OUTPUT_EVENT_NAME,
+				this.$childProcess,
+				handler,
+				this.executeGradleCommand(this.getPlatformData(projectData).projectRoot,
```
rosen-vladimirov: In case this method fails before returning the Promise (for example in case the implementation is:
```TypeScript
myMethod(): Promise<any> {
    // some sync operation that throws error
    return new Promise<any>((resolve, reject) => {
        // execute some operations
        resolve();
    });
}
```
current code will fail during evaluation of method arguments and will never emit error event for the operation. Is this expected?

Note: I know we are using `async` keyword in such cases that will always generate Promise for us and we should never end up in a case where we fail before creating the Promise, but anyway, I want to be sure that's the behaviour we want.

lib/services/platform-service.ts
```diff
@@ -416,10 +417,13 @@ export class PlatformService extends EventEmitter implements IPlatformService {
 		await this.trackActionForPlatform({ action: "Build", platform, isForDevice });
 
 		let platformData = this.$platformsData.getPlatformData(platform, projectData);
-		platformData.platformProjectService.on(constants.BUILD_OUTPUT_EVENT_NAME, (data: any) => {
+		const handler = (data: any) => {
 			this.emit(constants.BUILD_OUTPUT_EVENT_NAME, data);
-		});
-		await platformData.platformProjectService.buildProject(platformData.projectRoot, projectData, buildConfig);
+			this.$logger.printInfoMessageOnSameLine(data.data.toString());
```
rosen-vladimirov: why do we have this line and we do no have it in the other cases when we use attachAwaitDetach?

Mitko-Kerezov: We need in order to keep the current behavior when using `tns` as a CLI.
Whenever we spawn a child process with `pipe` instead of `inherit` as its stdio, we will not see the output on the console.
In short you have 2 cases:
1. You spawn from Fusion(`tns` is `require`-d) - then you spawn with `pipe` and you need this line in order to see the output on the console and you need the event handler in order to capture output and forward it to fusion
2. You spawn from CLI (`tns` executed from command line) - you then spawn with `inherit`, this event handler will not execute because all of the output is printed on `process.stdout`. 

Bottom line - you need this in the unlikely case that you spawn with `pipe` from the terminal so that you see the build output on the console.

lib/services/android-project-service.ts
Mitko-Kerezov: IMO it should be the caller's responsibility to keep track of the arguments. If one of the arguments need to be prepared in some way (e.g. to be `try...catch`-ed) it should be done prior to calling the method.
As for the example you've provided I see no problem with it - failing during evaluation of the arguments will lead to the event handler not being attached in the first place and this method's main target is preventing leakage of event handlers.

PR #2704, opened 2017-04-07
Fix connect ECONNREFUSED error when trying to debug on iOS Sim

rosen-vladimirov:
When trying to debug on iOS Simulator (via NativeScript Inspector), we often receive `Error: connect ECONNREFUSED 127.0.0.1:18181` error.
There are two different problems. First one is incorrect identifying of the application's PID - we are using `_.trimStart` method totally incorrectly and in case your application identifier has numbers in it, they'll also be replaced in the search PID string. For example in case your app id is `org.nativescript.app10`, and the real PID of the running application is 18129, the current usage of `trimStart` method will give you the PID 8129.
Fix this by applying regular expression and find the correct PID.

The second problem is that there's some delay between opening the NativeScript Inspector and the debugged application on the device. In many cases the first time when we try to connect, we receive the error ECONNREFUSED. However in the previous implementation, we were trying to connect multiple times. Get back the old code and adjust it to work with Promises instead of the old sync execution.

### Second commit:

Fix debug command:  
- `tns debug ios` prints message that you have to open `null` url in Chrome - when you do not pass `--chrome` to this command, it will use NativeScript inspector, so hide the incorrect message.
- debugData is constructed too early in the command - in case the application has not been built before calling `tns debug ...`, construction of debugData will fail as CLI will not be able to find the latest built package. In order to fix this make the `pathToAppPackage` not mandatory (we do not need it for debug commands when `--start` is passed) and populate it after successful deploy of the application. This way it will have correct value. Delete most of the DebugDataService as most of the methods are not needed with these changes.
- remove the check if `--chrome` is passed in order to print the url when `tns debug android` is used. The check was incorrect (it should check the value of `options.client`), but in fact we do not need it - the idea of the check was to suppress starting of Node Inspector, however we do not start it anymore no matter of the passed flags. So remove the incorrect check.

### DIFF
--- a/lib/commands/debug.ts
+++ b/lib/commands/debug.ts
@@ -6,7 +6,6 @@ export abstract class DebugPlatformCommand implements ICommand {
 	constructor(private debugService: IPlatformDebugService,
 		private $devicesService: Mobile.IDevicesService,
 		private $injector: IInjector,
-		private $logger: ILogger,
 		private $devicePlatformsConstants: Mobile.IDevicePlatformsConstants,
 		private $config: IConfiguration,
 		private $usbLiveSyncService: ILiveSyncService,
@@ -14,7 +13,8 @@ export abstract class DebugPlatformCommand implements ICommand {
 		protected $platformService: IPlatformService,
 		protected $projectData: IProjectData,
 		protected $options: IOptions,
-		protected $platformsData: IPlatformsData) {
+		protected $platformsData: IPlatformsData,
+		protected $logger: ILogger) {
 		this.$projectData.initializeProjectData();
 	}
 
@@ -31,9 +31,7 @@ export abstract class DebugPlatformCommand implements ICommand {
 			teamId: this.$options.teamId
 		};
 
-		const buildConfig: IBuildConfig = _.merge({ buildForDevice: this.$options.forDevice }, deployOptions);
-
-		const debugData = this.$debugDataService.createDebugData(this.debugService, this.$options, buildConfig);
+		let debugData = this.$debugDataService.createDebugData(this.$projectData, this.$options);
 
 		await this.$platformService.trackProjectType(this.$projectData);
 
@@ -56,6 +54,9 @@ export abstract class DebugPlatformCommand implements ICommand {
 			}
 
 			await deviceAppData.device.applicationManager.stopApplication(applicationId);
+
+			const buildConfig: IBuildConfig = _.merge({ buildForDevice: this.$options.forDevice }, deployOptions);
+			debugData.pathToAppPackage = this.$platformService.lastOutputPath(this.debugService.platform, buildConfig, projectData);
 
 			this.printDebugInformation(await this.debugService.debug(debugData, debugOptions));
 		};
@@ -80,7 +81,7 @@ export abstract class DebugPlatformCommand implements ICommand {
 		return true;
 	}
 
-	private printDebugInformation(information: string[]): void {
+	protected printDebugInformation(information: string[]): void {
 		_.each(information, i => {
 			this.$logger.info(`To start debugging, open the following URL in Chrome:${EOL}${i}${EOL}`.cyan);
 		});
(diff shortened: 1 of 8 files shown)


lib/device-sockets/ios/socket-proxy-factory.ts
```diff
@@ -31,32 +34,33 @@ export class SocketProxyFactory extends EventEmitter implements ISocketProxyFact
...
-				}
+				backendSocket.on("end", () => {
+					this.$logger.info("Backend socket closed!");
+					if (!(this.$config.debugLivesync && this.$options.watch)) {
+						process.exit(0);
+					}
+				});
+
+				frontendSocket.on("close", () => {
+					console.log("frontend socket closed");
```
yyosifov: shouldn't this be $logger

lib/device-sockets/ios/socket-proxy-factory.ts
```diff
@@ -31,32 +34,33 @@ export class SocketProxyFactory extends EventEmitter implements ISocketProxyFact
...
+					if (!(<any>backendSocket).destroyed) {
+						backendSocket.destroy();
+					}
+				});
+				backendSocket.on("close", () => {
+					console.log("backend socket closed");
```
yyosifov: shouldn't this be $logger

lib/services/ios-debug-service.ts
```diff
@@ -103,11 +103,18 @@ class IOSDebugService extends DebugServiceBase implements IPlatformDebugService
 
 		let lineStream = byline(child_process.stdout);
 		this._childProcess = child_process;
+		const pidRegExp = new RegExp(`${debugData.applicationIdentifier}:\\s?(\\d+)`);
```
yyosifov: Would be nice to have Unit Tests for this Regular Expression. I'd suggest we extract this to some method/service so that we can test it and cover with more expressions?

lib/services/ios-debug-service.ts
```diff
@@ -103,11 +103,18 @@ class IOSDebugService extends DebugServiceBase implements IPlatformDebugService
...
 
 		lineStream.on('data', (line: NodeBuffer) => {
 			let lineText = line.toString();
 			if (lineText && _.startsWith(lineText, debugData.applicationIdentifier)) {
-				let pid = _.trimStart(lineText, debugData.applicationIdentifier + ": ");
+				const pidMatch = lineText.match(pidRegExp);
+
+				if (!pidMatch) {
+					this.$logger.trace(`Line ${lineText} does not contain the searched pattern: ${pidRegExp}.`);
```
yyosifov: if there's no pidMatch, what happens with `const pid = pidMatch[1]` below? Won't an `undefined exception` occur?

rosen-vladimirov: I've forgotten to commit a `return` here 😄  Thanks for pointing it.

lib/commands/debug.ts
```diff
@@ -101,22 +102,30 @@ export class DebugIOSCommand extends DebugPlatformCommand {
...
 		$iosDeviceOperations.setShouldDispose(this.$options.justlaunch);
 	}
 
 	public async canExecute(args: string[]): Promise<boolean> {
 		return await super.canExecute(args) && await this.$platformService.validateOptions(this.$options.provision, this.$projectData, this.$platformsData.availablePlatforms.iOS);
 	}
+
+	protected printDebugInformation(information: string[]): void {
+		if (this.$options.chrome) {
+			_.each(information, i => {
```
rosen-vladimirov: call `super.printDebugInformation` here instead of the each

lib/services/ios-debug-service.ts
rosen-vladimirov: Fixed

lib/services/ios-debug-service.ts
rosen-vladimirov: I'm not sure where's the appropriate place for this code. I agree that it needs tests and I'll add them in a separate PR.

rosen-vladimirov: Nevermind - I've created a new helper method and added unit tests for it: https://github.com/telerik/mobile-cli-lib/pull/942

rosen-vladimirov:
ping @yyosifov - all comments are fixed

Issue #2501, opened 2017-02-07
Manual signing with distribution provisioning profile fails with NS 2.5


dbenninger:
When specifying the provisioning profile with the new command line parameter "--provision", the "CODE_SIGN_IDENTITY = iPhone Distribution" in build.xcconfig is ignored and "iPhone Developer" is used. The xcode build fails with the error message 'Provisioning profile "<NAME>" doesn't include signing certificate "iPhone Developer: ..."' because it cannot find a developer certificate for the adhoc distribution provisioning profile.

To reproduce:
1. Use NS 2.5 to create a test app: tns create "SignTest" --ng
2. In the file app/App_Resources/iOS/build.xcconfig, enable CODE_SIGN_IDENTITY = iPhone Distribution;
3. Create an adhoc distribution provisioning profile for your SignTest app id and link it to your distribution certificate
4. Install the distribution provisioning profile and distribution certificate locally
5. Run the build with: tns build ios --release --for-device --provision (UUID)
6. The build fails with: Provisioning profile "..." doesn't include signing certificate "iPhone Developer: ..."

Manually editing the generated project.pbxproj and replacing "iPhone Developer" with "iPhone Distribution" solves the issue.

Opening the generated project settings in xcode shows you the same error message:
![xcode](https://cloud.githubusercontent.com/assets/13940547/22684231/c9c2a8d4-ed1b-11e6-9772-cf51b5d0eedd.png)

vbresults:
Confirmed.

patrickpereira:
Any fixes for this one?

spike1292:
having the same problem. 
is there a workaround?

vbresults:
@spike1292 Edit ios-project-service.js in nativescript-cli, look for the phrase "iPhone Developer"

pkoleva:
Hey guys, sorry for not updating you. We haven't got the time to fix this with all the improvements around 3.0.0 RC. We'll consider it for a future release. I also have the same issue and am constantly reminding about it :). For now I recommend you to not use this option and as a workaround set PROVISIONING_PROFILE = <name> in build.xcconfig.

spike1292:
@pkoleva thank you for the update, will try to use the `PROVISIONING_PROFILE` workaround.
We where using `PROVISIONING_PROFILE_SPECIFIER` but this is not working any more, any chance you will also support the specifier?

PanayotCankov:
[...]

PR #2705, opened 2017-04-11
Make --provision work with distribution certificates and able to update the project after initial prepare

PanayotCankov:
- Issuing `tns <prepare|build|deploy etc.> ios --provision <provision uuid or name/specifier>` will now update the native Xcode and invalidate the prepare step and native build if necessary (in case it was used with different provision than the last time).

 - In case in the xcconfig was used a PROVISIONING_PROFILE_SPECIFIER the build will set a "Manual" signing style in the Xcode project similar to the behaviour triggered with PROVISIONING_PROFILE. Also when one of these two is provided in the xcconfig, if the --provision was never used, any DEVELOPMENT_TEAM or CODE_SIGN_SPECIFIED will be cleared off the pbxproj so the values from the xcconfig can be respected. There is room for improvement to allow a behaviour that once --provsion was used, if it is not used in subsequent command and the xcconfig has singing flags it would clear the pbxproj and fallback to the xcconfig.

 - The path to the pbxproj is now changes and should respect scenarios when the id of the project does not match the folder name of the project.

### DIFF
--- a/lib/services/ios-project-service.ts
+++ b/lib/services/ios-project-service.ts
@@ -2,7 +2,6 @@ import * as path from "path";
 import * as shell from "shelljs";
 import * as os from "os";
 import * as semver from "semver";
-import * as xcode from "xcode";
 import * as constants from "../constants";
 import * as helpers from "../common/helpers";
 import { attachAwaitDetach } from "../common/helpers";
@@ -11,7 +10,6 @@ import { PlistSession } from "plist-merge-patch";
 import { EOL } from "os";
 import * as temp from "temp";
 import * as plist from "plist";
-import { Xcode } from "pbxproj-dom/xcode";
 import { IOSProvisionService } from "./ios-provision-service";
 
 export class IOSProjectService extends projectServiceBaseLib.PlatformProjectServiceBase implements IPlatformProjectService {
@@ -42,7 +40,9 @@ export class IOSProjectService extends projectServiceBaseLib.PlatformProjectServ
 		private $pluginVariablesService: IPluginVariablesService,
 		private $xcprojService: IXcprojService,
 		private $iOSProvisionService: IOSProvisionService,
-		private $sysInfo: ISysInfo) {
+		private $sysInfo: ISysInfo,
+		private $pbxprojDomXcode: IPbxprojDomXcode,
+		private $xcode: IXcode) {
 		super($fs, $projectDataService);
 	}
 
@@ -379,10 +379,9 @@ export class IOSProjectService extends projectServiceBaseLib.PlatformProjectServ
 		await this.createIpa(projectRoot, projectData, buildConfig);
 	}
 
-	private async setupSigningFromProvision(projectRoot: string, projectData: IProjectData, provision?: any): Promise<void> {
+	private async setupSigningFromProvision(projectRoot: string, projectData: IProjectData, provision?: string): Promise<void> {
 		if (provision) {
-			const pbxprojPath = path.join(projectRoot, projectData.projectName + ".xcodeproj", "project.pbxproj");
-			const xcode = Xcode.open(pbxprojPath);
+			const xcode = this.$pbxprojDomXcode.Xcode.open(this.getPbxProjPath(projectData));
 			const signing = xcode.getSigning(projectData.projectName);
 
 			let shouldUpdateXcode = false;
@@ -399,8 +398,6 @@ export class IOSProjectService extends projectServiceBaseLib.PlatformProjectServ
 			}
 
 			if (shouldUpdateXcode) {
-				// This is slow, it read through 260 mobileprovision files on my machine and does quite some checking whether provisioning profiles and devices will match.
-				// That's why we try to avoid id by checking in the Xcode first.
 				const pickStart = Date.now();
 				const mobileprovision = await this.$iOSProvisionService.pick(provision, projectData.projectId);
 				const pickEnd = Date.now();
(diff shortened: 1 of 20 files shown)


PanayotCankov:
Resolves #2501

lib/services/ios-project-service.ts
```diff
@@ -421,11 +421,16 @@ export class IOSProjectService extends projectServiceBaseLib.PlatformProjectServ
...
 
 	private async setupSigningForDevice(projectRoot: string, buildConfig: IiOSBuildConfig, projectData: IProjectData): Promise<void> {
-		const pbxprojPath = path.join(projectRoot, projectData.projectName + ".xcodeproj", "project.pbxproj");
-		const xcode = Xcode.open(pbxprojPath);
+		const xcode = Xcode.open(this.getPbxProjPath(projectData));
 		const signing = xcode.getSigning(projectData.projectName);
 
-		if ((this.readXCConfigProvisioningProfile(projectData) || this.readXCConfigProvisioningProfileForIPhoneOs(projectData)) && (!signing || signing.style !== "Manual")) {
+		const hasProvisioningProfileInXCConfig =
+			this.readXCConfigProvisioningProfile(projectData) ||
```
Mitko-Kerezov: I think the ordering here should be the other way around `[sdk=iphoneos*]` lines in .xcconfig are more specific than the ones without and it makes more sense that specific ones are taken into account before general ones.
In addition, `PROVISIONING_PROFILE_SPECIFIER` is a newer term than `PROVISIONING_PROFILE` and the latter is close to being deprecated I think.
So I think the order should be:
```TypeScript
this.readXCConfigProvisioningProfileSpecifierForIPhoneOs(projectData) ||
this.readXCConfigProvisioningProfileSpecifier(projectData) ||
this.readXCConfigProvisioningProfileForIPhoneOs(projectData) ||
this.readXCConfigProvisioningProfile(projectData)
```

lib/services/ios-project-service.ts
```diff
@@ -837,6 +842,41 @@ We will now place an empty obsolete compatability white screen LauncScreen.xib f
...
 
+	public checkForChanges(changesInfo: IProjectChangesInfo, options: IProjectChangesOptions, projectData: IProjectData): void {
+		const nextCommandProvisionUUID = options.provision;
+
+		if (typeof nextCommandProvisionUUID === "string") {
+			const pbxprojPath = this.getPbxProjPath(projectData);
+
+			// Check if the native project's signing is set to the provided provision...
+			let signingChanged = false;
+			if (!fs.existsSync(pbxprojPath)) {
```
PanayotCankov: yes

lib/services/ios-project-service.ts
PanayotCankov: ok

lib/services/ios-project-service.ts
```diff
@@ -844,6 +848,41 @@ We will now place an empty obsolete compatability white screen LauncScreen.xib f
...
+				} else {
+					signingChanged = true;
+				}
+			} else {
+				signingChanged = true;
+			}
+
+			if (signingChanged) {
+				changesInfo.nativeChanged = true;
+				changesInfo.configChanged = true;
```
yyosifov: Do you think we need to prepare the project if `configChanged=true`?

lib/services/ios-project-service.ts
```diff
@@ -844,6 +848,41 @@ We will now place an empty obsolete compatability white screen LauncScreen.xib f
 		return Promise.resolve();
 	}
 
+	public checkForChanges(changesInfo: IProjectChangesInfo, options: IProjectChangesOptions, projectData: IProjectData): void {
+		const nextCommandProvisionUUID = options.provision;
+
+		if (typeof nextCommandProvisionUUID === "string") {
```
yyosifov: what if it's not a string? Shall we throw an error or log something to the user - otherwise he won't know something's wrong and we don't use the provided provision?

lib/services/ios-project-service.ts
```diff
@@ -844,6 +848,41 @@ We will now place an empty obsolete compatability white screen LauncScreen.xib f
 		return Promise.resolve();
 	}
 
+	public checkForChanges(changesInfo: IProjectChangesInfo, options: IProjectChangesOptions, projectData: IProjectData): void {
```
yyosifov: It would be great if we add few tests for this method

lib/services/ios-project-service.ts
```diff
@@ -844,6 +848,41 @@ We will now place an empty obsolete compatability white screen LauncScreen.xib f
...
+
+		if (typeof nextCommandProvisionUUID === "string") {
+			const pbxprojPath = this.getPbxProjPath(projectData);
+
+			// Check if the native project's signing is set to the provided provision...
+			let signingChanged = false;
+			if (this.$fs.exists(pbxprojPath)) {
+				const xcode = Xcode.open(pbxprojPath);
+				const signing = xcode.getSigning(projectData.projectName);
+				if (signing && signing.style === "Manual") {
```
yyosifov: We can add the "Manual" as a constant

lib/services/ios-project-service.ts
```diff
@@ -844,6 +848,41 @@ We will now place an empty obsolete compatability white screen LauncScreen.xib f
...
+					for (let name in signing.configurations) {
+						let config = signing.configurations[name];
+						if (config.uuid !== nextCommandProvisionUUID && config.name !== nextCommandProvisionUUID) {
+							signingChanged = true;
+							break;
+						}
+					}
+				} else {
+					signingChanged = true;
```
yyosifov: Can we have a comment or something here? It's not clear to me, why if the signing is not Manual it means it's changed?

lib/services/ios-project-service.ts
PanayotCankov: No, the nativeChanged and configChanged are not necessary here. "appResourcesChanged" does the job but is not properly named for the case. Should I introduce a new flag?

lib/services/ios-project-service.ts
PanayotCankov: It is `"Manual" | "Automatic"`, the TypeScript compiler does enough of type checking for these types. E.g. it will error if you compare it with "Manual1".

lib/services/ios-project-service.ts
```diff
@@ -844,6 +846,35 @@ We will now place an empty obsolete compatability white screen LauncScreen.xib f
...
+		const provision = options.provision;
+		if (provision !== undefined) {
+			// Check if the native project's signing is set to the provided provision...
+			const pbxprojPath = this.getPbxProjPath(projectData);
+
+			if (this.$fs.exists(pbxprojPath)) {
+				const xcode = this.$pbxprojDomXcode.Xcode.open(pbxprojPath);
+				const signing = xcode.getSigning(projectData.projectName);
+				if (signing && signing.style === "Manual") {
+					for (let name in signing.configurations) {
```
rosen-vladimirov: Can't you use:
```TypeScript
for (let config of signing.configurations) {
 ...
}
```

lib/services/pbxproj-dom-xcode.ts
```diff
@@ -0,0 +1,7 @@
+import * as PbxprojDomXcodeModule from "pbxproj-dom/xcode";
+
+declare global {
+	type IPbxprojDomXcode = typeof PbxprojDomXcodeModule;
+}
+
+$injector.register("pbxprojDomXcode", () => require("pbxproj-dom/xcode"), true);
```
rosen-vladimirov: there's no need to pass `true` here - in case the value is undefined, `true` is set as default:
https://github.com/telerik/mobile-cli-lib/blob/master/yok.ts#L284

lib/services/ios-project-service.ts
PanayotCankov: Its an object, not an array, my node throws [Symbol.iterator] is not a function

rosen-vladimirov: My mistake, I didn't notice this is an object.

PR #2708, opened 2017-04-12
Change livesync dir permissions on device even if app is not restarted

Pip3r4o:
Original issue in android-runtime: https://github.com/NativeScript/android-runtime/issues/591

Changing .css and .xml files in a NativeScript project would not change permissions in the temporary sync dir on the device. This could result in showing outdated views on subsequent installs of the application on the same device without using the CLI's livesync service.

The proposed changes will always chmod 777 the sync directory, thus allowing the runtime to remove any previous livesync artifacts.

Relies on changes made in https://github.com/telerik/mobile-cli-lib/pull/943


dtopuzov:
+1

rosen-vladimirov:
Merging as the problem is in the tests. Approved by @dtopuzov

PR #2713, opened 2017-04-13
Fix debug android command

rosen-vladimirov:
### Fix Android debugging
When executing `tns debug android` for slow emulators/devices we have logic to wait several seconds until debugger is attached.
However we've missed to `await` the `sleep` method and in fact we are not waiting at all. Add the missing await, which fixes the `tns debug android` command.

### Remove unused code
A while ago, we had a logic to find out which node_modules should be processed. However it is not used anymore (removed in https://github.com/NativeScript/nativescript-cli/pull/2152/ ).
So remove the code - we do not need it anymore.

### DIFF
--- a/lib/services/android-debug-service.ts
+++ b/lib/services/android-debug-service.ts
@@ -165,20 +165,20 @@ class AndroidDebugService extends DebugServiceBase implements IPlatformDebugServ
 	private async waitForDebugger(packageName: String): Promise<void> {
 		let waitText: string = `0 /data/local/tmp/${packageName}-debugger-started`;
 		let maxWait = 12;
-		let debugerStarted: boolean = false;
-		while (maxWait > 0 && !debugerStarted) {
+		let debuggerStarted: boolean = false;
+		while (maxWait > 0 && !debuggerStarted) {
 			let forwardsResult = await this.device.adb.executeShellCommand(["ls", "-s", `/data/local/tmp/${packageName}-debugger-started`]);
 
 			maxWait--;
 
-			debugerStarted = forwardsResult.indexOf(waitText) === -1;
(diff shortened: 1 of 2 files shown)


lib/services/android-debug-service.ts
```diff
@@ -174,7 +174,7 @@ class AndroidDebugService extends DebugServiceBase implements IPlatformDebugServ
 			debugerStarted = forwardsResult.indexOf(waitText) === -1;
 
 			if (!debugerStarted) {
-				sleep(500);
+				await sleep(500);
```
yyosifov: Great! :)

Small typo: `debugerStarted` => `debuggerStarted`

What do you think of introducing a function wait.Until(function, condition, sleepIntervalMs) which implements a loop which sleeps for several iterations or until a condition is met? This way, we can introduce unit tests for it, reuse it and don't worry about such bugs anymore? I've seen this implemented and it was working good.

yyosifov: Here's the snippet I've been referring to (from another project and in C#):

```
public class PollingService : IPollingService
	{
		public bool WaitUntil(Func<bool> successCriteria, int millisecondsInterval, int iterations)
		{
			var succeeded = false;
			for (int i = 0; i < iterations; i++)
			{
				if (successCriteria())
				{
					succeeded = true;
					break;
				}
				else if (i < iterations - 1)
				{
					Thread.Sleep(millisecondsInterval);
				}
			}

			return succeeded;
		}
	}
```

rosen-vladimirov: I like the idea and I suggest to implement it in another PR (where we will add tests for it). We can use it on several places in CLI (at least two :) )
Thanks for the suggestion.

PR #2717, opened 2017-04-13
Fix debug for iOS devices

rosen-vladimirov:
Currently the `tns debug ios` command tries to find already built result for iOS Simulator. The problem is incorrect value passed to `buildForDevice` property. It should be set based on the current device.


rosen-vladimirov:
Merging as the tests are incorrect.

PR #2724, opened 2017-04-19
Implement extensibility model for CLI

rosen-vladimirov:
Implement extensibilty of CLI that allows anyone to add packages that add new functionality to NativeScript CLI. The packages are installed in a specific directory, so they are persisted through CLI's updates.
The directory where extensions are installed contains a package.json and each extension is npm package installed there.
The extensions can be maintained in two different ways:
- navigate to the directory where extensions are installed and use `npm` for install/uninstall/update of packages.
- use CLI's commands to update them: `tns extension install <name>`, `tns extension uninstall <name>`, `tns extension`

Implement extensibilityService that executes all operations and expose it to public API. In {N} CLI the extensions are loaded in the entry point, before parsing command line arguments. This way extensions can add new commands.
In Fusion, after CLI is required as a library, the `extensibilityService.loadExtensions` method should be called. It returns array of Promises - one for each installed extension.

Add help for the new commands, but do not link the new commands in other commands help for the moment.

Add unit tests for the new service.

Update PublicAPI.md with information about new methods that can be called. Also remove the TypeScript part of the `Usage` sections  - they are just confusing people. Keep only ES6 JavaScript examples.

> NOTE: A detailed guide how to create extension will be added to the repository later.

### DIFF
--- a/PublicAPI.md
+++ b/PublicAPI.md
@@ -3,35 +3,17 @@ Public API
 
 This document describes all methods that can be invoked when NativeScript CLI is required as library, i.e.
 
-<table>
-	<tr>
-        <td>
-        	JavaScript
-        </td>
-        <td>
-        	TypeScript
-        </td>
-    </tr>
-    <tr>
-    	<td>
-<pre lang="javascript">
+```JavaScript
 const tns = require("nativescript");
-</pre>
-        </td>
-    	<td>
-<pre lang="typescript">
-import * as tns from "nativescript";
-</pre>
-        </td>
-    </tr>
-
-</table>
+```
 
 ## Module projectService
 
 `projectService` modules allow you to create new NativeScript application.
 
-* `createProject(projectSettings: IProjectSettings): Promise<void>` - Creates new NativeScript application. By passing `projectSettings` argument you specify the name of the application, the template that will be used, etc.:
+### createProject
+* Description: `createProject(projectSettings: IProjectSettings): Promise<void>` - Creates new NativeScript application. By passing `projectSettings` argument you specify the name of the application, the template that will be used, etc.:
+
 ```TypeScript
 /**
  * Describes available settings when creating new NativeScript application.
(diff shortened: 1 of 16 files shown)


lib/services/extensibility-service.ts
```diff
@@ -0,0 +1,135 @@
...
+		private $requireService: IRequireService) {
+	}
+
+	public async installExtension(extensionName: string): Promise<IExtensionData> {
+		this.$logger.trace(`Start installation of extension '${extensionName}'.`);
+
+		await this.assertExtensionsSetup();
+
+		const npmOpts: any = { save: true };
+		npmOpts["save-exact"] = true;
```
Mitko-Kerezov: why initialize the object's properties separately:
```JavaScript
const npmOpts: any = { 
  save: true,
  ["save-exact"]: true
};
```

lib/services/extensibility-service.ts
```diff
@@ -0,0 +1,135 @@
...
+			this.$logger.trace(`Extension ${extensionName} is not installed, starting installation.`);
+			await this.installExtension(extensionName);
+		}
+
+		this.$logger.trace(`Extension ${extensionName} is installed.`);
+	}
+
+	@cache()
+	private async assertExtensionsSetup(): Promise<void> {
+		this.assertExtensionsDirExists();
```
Mitko-Kerezov: From a semantic point of view I think it'd be more natural if `this.assertPackageJsonExists` calls `this.assertExtensionsDirExists` and you just call `this.assertPackageJsonExists` here.
That's my 2¢ of course hence definitely not a merge-stopper

lib/services/extensibility-service.ts
```diff
@@ -0,0 +1,135 @@
...
+
+			// create default package.json
+			this.$fs.writeJson(this.pathToPackageJson, {
+				name: "nativescript-extensibility",
+				version: "1.0.0",
+				description: "The place where all packages that extend CLI will be installed.",
+				license: "Apache-2.0",
+				readme: "The place where all packages that extend CLI will be installed.",
+				repository: "none",
+				dependencies: {}
```
Mitko-Kerezov: You can skip writing a lot of the keys for example: the `license`, `repository` and `dependencies` are gratuitous IMO

lib/bootstrap.ts
```diff
@@ -125,3 +125,10 @@ $injector.require("projectChangesService", "./services/project-changes-service")
 $injector.require("emulatorPlatformService", "./services/emulator-platform-service");
 
 $injector.require("staticConfig", "./config");
+
+$injector.require("requireService", "./services/require-service");
+
+$injector.requireCommand("extension|*list", "./commands/extensibility/list-extensions");
+$injector.requireCommand("extension|install", "./commands/extensibility/install-extension");
+$injector.requireCommand("extension|uninstall", "./commands/extensibility/uninstall-extension");
+$injector.requirePublicClass("extensibilityService", "./services/extensibility-service");
```
Mitko-Kerezov: This needn't be `requirePublicClass` - it can be a regular `requirePublic` with decorators inside the class

lib/services/extensibility-service.ts
```diff
@@ -0,0 +1,135 @@
+import * as path from "path";
+import { cache } from "../common/decorators";
+import * as constants from "../constants";
+
+export class ExtensibilityService implements IExtensibilityService {
+	private get pathToExtensions(): string {
+		return path.join(this.$options.profileDir, "extensions");
```
Mitko-Kerezov: I think the time is ripe to move `this.$options.profileDir` into a separate method/service or whatnot and get it from here because we'd really like to keep our services clear of `$options`.

As this requires some additional effort in finding all of the occurrences of `this.$options.profileDir` where the new method may be used I'd like to see this in a separate PR I think, ping @rosen-vladimirov for an opinion

test/services/extensibility-service.ts
```diff
@@ -0,0 +1,636 @@
...
+		});
+
+		describe("passes correct arguments to npm install", () => {
+			const getArgsPassedToNpmInstallDuringInstallExtensionCall = async (userSpecifiedValue: string, testInjector?: IInjector): Promise<any> => {
+				testInjector = testInjector || getTestInjector();
+				const fs: IFileSystem = testInjector.resolve("fs");
+				fs.exists = (pathToCheck: string): boolean => true;
+
+				fs.readDirectory = (dir: string): string[] => {
+					return [userSpecifiedValue];
```
Mitko-Kerezov: You could use the short-hand syntax here:
`fs.readDirectory = (dir: string) => [userSpecifiedValue]`

test/services/extensibility-service.ts
```diff
@@ -0,0 +1,636 @@
...
+				};
+
+				const npm: INodePackageManager = testInjector.resolve("npm");
+				let argsPassedToNpmInstall: any = {};
+				npm.install = async (packageName: string, pathToSave: string, config?: any): Promise<any> => {
+					argsPassedToNpmInstall.packageName = packageName;
+					argsPassedToNpmInstall.patToSave = pathToSave;
```
Mitko-Kerezov: Minor typo patToSave -> pathToSave

test/services/extensibility-service.ts
```diff
@@ -0,0 +1,636 @@
...
+		});
+
+		it("throws error that has extensionName property when unable to load extension", async () => {
+			const expectedErrorMessage = "Require failed";
+
+			const extensionName = "extension1";
+			const testInjector = getTestInjector();
+			const fs: IFileSystem = testInjector.resolve("fs");
+			fs.exists = (pathToCheck: string): boolean => {
+				return path.basename(pathToCheck) !== extensionName;
```
Mitko-Kerezov: You can use short-hand syntax for the fat arrow function here

test/services/extensibility-service.ts
```diff
@@ -0,0 +1,636 @@
...
+			};
+
+			fs.readDirectory = (dir: string): string[] => {
+				return [extensionName];
```
Mitko-Kerezov: And here

test/services/extensibility-service.ts
```diff
@@ -0,0 +1,636 @@
...
+			};
+
+			const npm: INodePackageManager = testInjector.resolve("npm");
+			npm.install = async (packageName: string, pathToSave: string, config?: any): Promise<any> => {
+				return [extensionName];
```
Mitko-Kerezov: Here too

test/services/extensibility-service.ts
```diff
@@ -0,0 +1,636 @@
...
+				const actualResult = await Promise.all(extensibilityService.loadExtensions());
+				assert.deepEqual(actualResult, expectedResults);
+			});
+
+			it("installs extensions that are available in package.json, but are not available in node_modules", async () => {
+				const testInjector = getTestInjector();
+				const fs: IFileSystem = testInjector.resolve("fs");
+				const extensionNames = ["extension1", "extension2", "extension3"];
+				fs.exists = (pathToCheck: string): boolean => {
+					return path.basename(pathToCheck) !== extensionNames[0];
```
Mitko-Kerezov: And here

lib/services/extensibility-service.ts
rosen-vladimirov: OMG I've forgotten this once again... 
Thanks.

PublicAPI.md
```diff
@@ -73,71 +55,130 @@ interface IProjectSettings {
...
+```
+
+### installExtension
+Installs specified extensions and loads it in the current process, so the functionality that it adds can be used immediately.
+
+* Definition:
+```TypeScript
+/**
+ * Installs and loads specified extension.
+ * @param {string} extensionName Name of the extension to be installed. It may contain version as well, i.e. myPackage, myPackage@1.0.0, myPackage.tgz, https://github.com/myOrganization/myPackage, etc.
```
rosen-vladimirov: In fact it's correct, but it requires to have git in your PATH.

lib/services/extensibility-service.ts
rosen-vladimirov: Totally agree and I'll implement it in a separate PR!

lib/services/extensibility-service.ts
rosen-vladimirov: In case they are missing, on each installation of new extension, npm prints warnings. That's why I've added them.

PR #2736, opened 2017-04-24
Do not start emulator when `--available-devices` is passed

rosen-vladimirov:
In case there's no devices attached and no emulators running, trying `<cli name> devices <platform> --available-devices` will start emulator.
In order to fix this, modify the `startEmulatorIfNecessary` method to skip the starting in case `skipInferPlatform` option is passed. This option indicates that we are not concerned of specific platform, so the method does not know which is the target platform for which to start emulator.
Add unit test for this behavior.


rosen-vladimirov:
run ci

rosen-vladimirov:
run ci

rosen-vladimirov:
run ci

PR #2740, opened 2017-04-26
Improve debugService public API

rosen-vladimirov:
When using Chrome DevTools to debug iOS applications, CLI returns a url that points to a specific commit of the dev tools. This way, in case a new version of the dev tools introduces a breaking change, the debugging will still work.
However, in some cases (inside Electron app), we cannot use the remote url, so we must use the bundled one. So introduce a new option (only available when requiring CLI as a library), that defines that the returned url will use the bundled dev tools.
The default behavior (used in CLI), will still return the url that includes remote url.
Also change the definition of `debug` method in the interface, so now the DebugService can safely implement it.
Add a new check in the debug service - in case the device's status is not Connected, we are unable to start debug operation. So fail with correct error message in this case.
Add JSDocs for all debug related interfaces.
Add documentation in PublicAPI.md for the `debugService`.
Add debugService to tests of public API.
Add tests for debugService.
Add verification that in case device's OS is not supported (for example WP8), we'll throw correct error.

### DIFF
--- a/PublicAPI.md
+++ b/PublicAPI.md
@@ -180,10 +180,10 @@ for (let promise of loadExtensionsPromises) {
 }
 ```
 
-## settings
-`settings` module provides a way to configure various settings.
+## settingsService
+`settingsService` module provides a way to configure various settings.
 
-### set
+### setSettings
 Used to set various settings in order to modify the behavior of some methods.
 * Auxiliary interfaces:
 ```TypeScript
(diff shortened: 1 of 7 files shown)


test/services/debug-service.ts
```diff
@@ -0,0 +1,288 @@
...
+			};
+
+			_.each(["android", "iOS"], platform => {
+				describe(`${platform}DebugService's debug method`, () => {
+					describe("on macOS", () => {
+						it("passes chrome and start as true, when no debugOptions are passed to debugService", async () => {
+							await assertPassedDebugOptions(platform);
+						});
+
+						it("when calling debug service with chrome and start set to false, should disregarded them and set both to true", async () => {
```
Mitko-Kerezov: ...should **disregard** them...

test/services/debug-service.ts
```diff
@@ -0,0 +1,288 @@
...
+								assert.isFalse(passedDebugOptions.emulator);
+							}
+						};
+
+						it("passes chrome and start as true, when no debugOptions are passed to debugService", async () => {
+							const passedDebugOptions = await assertPassedDebugOptions(platform, null, { isWindows: true, isDarwin: false });
+							assertEmulatorOption(passedDebugOptions);
+						});
+
+						it("when calling debug service with chrome and start set to false, should disregarded them and set both to true", async () => {
```
Mitko-Kerezov: disregarded -> disregard

lib/services/ios-debug-service.ts
```diff
@@ -197,7 +197,13 @@ class IOSDebugService extends DebugServiceBase implements IPlatformDebugService
 			this._socketProxy = await this.$socketProxyFactory.createWebSocketProxy(this.getSocketFactory(device));
 
 			const commitSHA = "02e6bde1bbe34e43b309d4ef774b1168d25fd024"; // corresponds to 55.0.2883 Chrome version
-			return `chrome-devtools://devtools/remote/serve_file/@${commitSHA}/inspector.html?experiments=true&ws=localhost:${this._socketProxy.options.port}`;
+			let chromeDevToolsPrefix = `chrome-devtools://devtools/remote/serve_file/@${commitSHA}`;
```
Mitko-Kerezov: You can extract `chrome-devtools://devtools/` to a constant. Not a merge-stopper though

PublicAPI.md
```diff
@@ -219,6 +219,125 @@ interface ISettingsService {
...
+
+### debug
+The `debug` method allows starting a debug operation for specified application on a specific device. The method returns a Promise, which is resolved with a url. The url should be opened in Chrome DevTools in order to debug the application.
+
+The returned Promise will be rejected in case any error occurs. It will also be rejected in case:
+* Specified deviceIdentifier is not found in current list of attached devices.
+* The device, specified as deviceIdentifier is connected but not trusted.
+* The specified application is not installed on the device.
+* Trying to debug applications on connected iOS device on Linux.
+* In case the application is not running on Android device.
```
TsvetanMilanov: This is valid for iOS devices too. The only difference for now is that if the device is iOS the method will fail after 9 seconds not immediately.

PublicAPI.md
```diff
@@ -219,6 +219,125 @@ interface ISettingsService {
...
+* In case the application is not built in debug configuration on Android device.
```
TsvetanMilanov: This is valid for iOS devices too.

lib/services/debug-service.ts
```diff
@@ -35,9 +37,16 @@ class DebugService extends EventEmitter {
 		// After we find a way to check on iOS we should use it here.
 		const isAppRunning = true;
 		let result: string[];
-		debugOptions.chrome = !debugOptions.client;
+		debugOptions.chrome = true;
```
TsvetanMilanov: Why we want to set the `debugOptions.chrome` to true here? If we keep the old value we will default to chrome debugging but we will keep the app inspector debugging option.

rosen-vladimirov: There's no plans to support AppInspector at the moment, I want to be sure that passing different options will always use Chrome DevTools. Once we decide to change this, we'll modify the code and the tests.

lib/services/ios-debug-service.ts
```diff
@@ -196,8 +196,17 @@ class IOSDebugService extends DebugServiceBase implements IPlatformDebugService
 		if (debugOptions.chrome) {
 			this._socketProxy = await this.$socketProxyFactory.createWebSocketProxy(this.getSocketFactory(device));
 
-			const commitSHA = "02e6bde1bbe34e43b309d4ef774b1168d25fd024"; // corresponds to 55.0.2883 Chrome version
-			return `chrome-devtools://devtools/remote/serve_file/@${commitSHA}/inspector.html?experiments=true&ws=localhost:${this._socketProxy.options.port}`;
+			 // corresponds to 55.0.2883 Chrome version
```
Mitko-Kerezov: I do believe this comment belongs to line `205`

PR #2747, opened 2017-04-27
Change check for SIX to not use the PIP but python's help command.

yyosifov:
Update the Doctor to check for the Python's SIX package via executing `help("six")` in python, instead of using PIP as by default it appears Mac has the Python and Six package, but can be missing the PIP.


yyosifov:
This will enter from the release branch, where it's already merged

PR #2753, opened 2017-04-28
Add support for Custom entitlements files

yyosifov:
Adding Entitlements support:

- We now create by default an entitlements file for you
- If you add an `app.entitlements` file in the App_Resources folder it will be added and merged
- If you have plugins with `app.entitlements` in the `platforms/ios` folder they will be considered and merged
- If you haven't set the `CODE_SIGN_ENTITLEMENTS` property in xcconfig, we set it by default to the default entitlements file we produce.

Small refactoring of the XCConfig read property functionality - extracted, fixed several issues and added tests to cover all.


test/ios-entitlements-service.ts
```diff
@@ -0,0 +1,173 @@
...
+	};
+
+	let injector: IInjector,
+		platformsData: any,
+		projectData: IProjectData,
+		fs: IFileSystem,
+		iOSEntitlementsService: IOSEntitlementsService,
+		destinationFilePath: string;
+
+	beforeEach(function() {
```
rosen-vladimirov: `beforeEach( () => {`

test/ios-entitlements-service.ts
```diff
@@ -0,0 +1,173 @@
...
+
+		iOSEntitlementsService = injector.resolve("iOSEntitlementsService");
+		destinationFilePath = iOSEntitlementsService.getPlatformsEntitlementsPath(projectData);
+	});
+
+	describe("Ensure paths constructed are correct", () => {
+		it("Ensure destination entitlements relative path is calculated correctly.", () => {
+			const expected = "testApp/testApp.entitlements";
+			let actual = iOSEntitlementsService.getPlatformsEntitlementsRelativePath(projectData);
+			assert.equal(actual, expected);
```
rosen-vladimirov: `assert.deepEqual` maybe?

test/ios-entitlements-service.ts
```diff
@@ -0,0 +1,173 @@
...
+    <key>nameKey</key>
+    <string>appResources</string>
+  </dict>
+</plist>`;
+
+		function assertContent(actual: string, expected: string) {
+			let strip = (x: string) => {
+				return x.replace(EOL, '').trim();
+			};
+			assert.equal(strip(actual), strip(expected));
```
rosen-vladimirov: `deepEqual` maybe

test/ios-project-service.ts
```diff
@@ -720,8 +744,115 @@ describe("iOS Project Service Signing", () => {
...
+
+			assert.isTrue(fs.exists(destinationFilePath), 'Target build xcconfig is missing for release: ' + release);
+			let expected = {
+				'CODE_SIGN_ENTITLEMENTS': iOSEntitlementsService.getPlatformsEntitlementsRelativePath(projectData)
+			};
+			assertPropertyValues(expected, destinationFilePath, testInjector);
+		}
+	});
+
+	it("The user specified entitlements property takes percedence", async () => {
```
rosen-vladimirov: `percedence` -> `precedence` maybe

test/xcconfig-service.ts
```diff
@@ -0,0 +1,188 @@
...
+		testInjector.register('xCConfigService', XCConfigService);
+
+		return testInjector;
+	};
+
+	const assertPropertyValues = (expected: any, injector: IInjector) => {
+		let service = getXCConfigService(injector);
+		_.forOwn(expected, (value, key) => {
+			let actual = service.readPropertyValue("any", key);
+			assert.equal(actual, value);
```
rosen-vladimirov: deepEqual

lib/constants.ts
```diff
@@ -16,6 +16,9 @@ export const TEST_RUNNER_NAME = "nativescript-unit-test-runner";
 export const LIVESYNC_EXCLUDED_FILE_PATTERNS = ["**/*.js.map", "**/*.ts"];
 export const XML_FILE_EXTENSION = ".xml";
 export const PLATFORMS_DIR_NAME = "platforms";
+export const IOS_PLATFORM_NAME = "ios";
+export const IOS_PLATFORM_NORMALIZED_NAME = "iOS";
```
yyosifov: I'll do it, although there are already some constants in ios-project-service for "ios" and it turned out my code pretty ugly from:
`constants.IOS_PLATFORM_NORMALIZED_NAME`
to:
`this.$mobileHelper.normalizePlatformName(this.$devicePlatformsConstants.iOS)`

lib/services/ios-project-service.ts
```diff
@@ -40,9 +42,11 @@ export class IOSProjectService extends projectServiceBaseLib.PlatformProjectServ
 		private $pluginVariablesService: IPluginVariablesService,
 		private $xcprojService: IXcprojService,
 		private $iOSProvisionService: IOSProvisionService,
-		private $sysInfo: ISysInfo,
 		private $pbxprojDomXcode: IPbxprojDomXcode,
-		private $xcode: IXcode) {
+		private $xcode: IXcode,
+		private $iOSEntitlementsService: IOSEntitlementsService,
+		private $sysInfo: ISysInfo,
+		private xCConfigService: XCConfigService) {
```
yyosifov: Fixed. What does this `$` stand for anyway?

test/ios-entitlements-service.ts
yyosifov: That's a single string (see `expected` value). Do we need deepEqual in this case?

test/ios-entitlements-service.ts
yyosifov: again, these are string? Do we need it here?

test/xcconfig-service.ts
yyosifov: These are strings. Do we need it here?

rosen-vladimirov: nope, sorry, my mistake, equal is fine

rynop:
This is great guys, thanks. Is this [documented](https://docs.nativescript.org/) anywhere?  I think this function is common across many plugins, and some go as far as even [duplicating this logic](https://github.com/EddyVerbruggen/nativescript-plugin-firebase/blob/master/src/scripts/entitlements-before-prepare.js#L12).

If it is not documented, I'd be happy to write it up.  Just looking for guidance on where in the current docs to put it.

PR #2789, opened 2017-05-07
Fix getting production dependencies code

rosen-vladimirov:
CLI has logic to find which are the "production" dependencies, i.e. which should be copied from `node_modules` to `platforms` dir.
However, when the project has a lot of dependencies (more than 15), on some machines the code leads to error: "Maximum callstack size exceeded". On other machines the code tooks significant time to execute.
After investigation, it turned out the recursion inside `node-modules-dependencies-builder` is incorrect and it adds each package many times to the result array.
Fix the recursion and change the class NodeModulesDependenciesBuilder to be stateless - instead of using properties in `this` object when calculating the production dependencies, the methods will persist the results through the passed args.
This way the whole class can be safely added to `$injector` and used whenever we need the production dependencies. Each time the calculation is started from the beginning, which is the requirement for long living process, where the project may change.
Fix the type of the result, which leads to fix in several other services, where the result has been expected as `IDictionary<smth>`. However it's never been dictionary, it's always been an array. The code, that expected dictionary has been working because the `_.values` method of lodash (used all over the places where the incorrect type of data has been expected), returns the same array when the passed argument is array.
Fix the tests that incorrectly expected dictionary with keys "0", "1", "2", etc.
Remove the usage of Node.js's `fs` module from `NodeModulesDependenciesBuilder` - replace it with `$fs` which allows easir writing of tests.
Require the `nodeModulesDependenciesBuilder` in bootstrap, so it can be correctly resolved by `$injector`.

Add unit tests for `nodeModulesDependenciesBuilder`.

NOTE: On my machine the initial prepare of a project with a lot of dependencies took around 2 minutes. 
After fixing this issue, it need only 20 seconds.
The problem was that each dependency is added multiple times to the `productionDependencies` object and it turned out the result contains more than 1600 items. After fixing the issue, the object contains around 100 (which is the real number of the production dependencies I have in my project).

### DIFF
--- a/lib/tools/node-modules/node-modules-dependencies-builder.ts
+++ b/lib/tools/node-modules/node-modules-dependencies-builder.ts
@@ -1,77 +1,63 @@
 import * as path from "path";
-import * as fs from "fs";
+import { NODE_MODULES_FOLDER_NAME, PACKAGE_JSON_FILE_NAME } from "../../constants";
 
 export class NodeModulesDependenciesBuilder implements INodeModulesDependenciesBuilder {
-	private projectPath: string;
-	private rootNodeModulesPath: string;
-	private resolvedDependencies: any[];
-	private seen: any;
+	public constructor(private $fs: IFileSystem) { }
 
-	public constructor(private $fs: IFileSystem) {
-		this.seen = {};
-		this.resolvedDependencies = [];
+	public getProductionDependencies(projectPath: string): IDependencyData[] {
+		const rootNodeModulesPath = path.join(projectPath, NODE_MODULES_FOLDER_NAME);
+		const projectPackageJsonPath = path.join(projectPath, PACKAGE_JSON_FILE_NAME);
+		const packageJsonContent = this.$fs.readJson(projectPackageJsonPath);
+		const dependencies = packageJsonContent && packageJsonContent.dependencies;
+
+		let resolvedDependencies: IDependencyData[] = [];
+
+		_.keys(dependencies)
+			.forEach(dependencyName => {
+				const depth = 0;
+				const directory = path.join(rootNodeModulesPath, dependencyName);
+
+				// find and traverse child with name `key`, parent's directory -> dep.directory
+				this.traverseDependency(resolvedDependencies, rootNodeModulesPath, dependencyName, directory, depth);
+			});
+
+		return resolvedDependencies;
 	}
 
-	public getProductionDependencies(projectPath: string): any[] {
-		this.projectPath = projectPath;
-		this.rootNodeModulesPath = path.join(this.projectPath, "node_modules");
-
-		let projectPackageJsonpath = path.join(this.projectPath, "package.json");
-		let packageJsonContent = this.$fs.readJson(projectPackageJsonpath);
-
-		_.keys(packageJsonContent.dependencies).forEach(dependencyName => {
-			let depth = 0;
-			let directory = path.join(this.rootNodeModulesPath, dependencyName);
-
-			// find and traverse child with name `key`, parent's directory -> dep.directory
-			this.traverseDependency(dependencyName, directory, depth);
-		});
-
-		return this.resolvedDependencies;
-	}
-
-	private traverseDependency(name: string, currentModulePath: string, depth: number): void {
+	private traverseDependency(resolvedDependencies: IDependencyData[], rootNodeModulesPath: string, name: string, currentModulePath: string, depth: number): void {
 		// Check if child has been extracted in the parent's node modules, AND THEN in `node_modules`
 		// Slower, but prevents copying wrong versions if multiple of the same module are installed
 		// Will also prevent copying project's devDependency's version if current module depends on another version
-		let modulePath = path.join(currentModulePath, "node_modules", name); // node_modules/parent/node_modules/<package>
-		let alternativeModulePath = path.join(this.rootNodeModulesPath, name);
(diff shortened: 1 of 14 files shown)


yyosifov:
Shall we add this to release 3.0.1?

rosen-vladimirov:
@yyosifov I'm not sure about this - it has higher potential to break some stuff. But in case you, @pkoleva  and @dtopuzov  think it is okay to go in release branch, I'll retarget the PR

lib/tools/node-modules/node-modules-dependencies-builder.ts
```diff
@@ -1,77 +1,63 @@
...
-		this.seen = {};
-		this.resolvedDependencies = [];
-	}
+	public constructor(private $fs: IFileSystem) { }
 
-	public getProductionDependencies(projectPath: string): any[] {
-		this.projectPath = projectPath;
-		this.rootNodeModulesPath = path.join(this.projectPath, "node_modules");
+	public getProductionDependencies(projectPath: string): IDependencyData[] {
+		const rootNodeModulesPath = path.join(projectPath, NODE_MODULES_FOLDER_NAME);
```
yyosifov: Just wanted to share a personal opinion here. So, I think it's been a little bit more complicated than neccessary.

Why can't we combine this in one method and not have nested and hard to debug (for me) calls like:

---- many traverseDependency calls
-------- findModule
------------ readModuleDependencies
---------------- many traverse dependency calls

This nesting sounds too much to me. I'd suggest two approaches for simplification:

### If we want to use recursion, why not rewrite it as: ###

1. many traverseDependency(root_module, depth: 0)
   2.0 if (alreadyResolved(root_module)) return;
   2.1 let currentModule = findCurrentModule
   2.2 resolved[ currentModule ] = currentModule;
   2.3 for(dep in currentModule.dependencies) traverseDependency(dep, depth + 1)

This way we can remove the 3-4 nested methods.

### We can use iteration instead ###
Another suggestion would be to just simulate a queue and turn this into a breadth-first-search, which I would find easier to debug, too. This way we will never care about the `callstack` anymore and we can `debug it` easier

0. for(module in all_initial_modules) resolved[module]  = module
1. queue.push(all_initial_modules)
2. while(queue.length > 0) 
    2.1 let currentModule = queue.shift()
    2.2. for(dep in currentModule.dependencies) 
    2.3.      if (!alreadyResolved(dep)) {
    2.4.           resolved[dep] = dep;
    2.5.           queue.push(dep);
    
Thanks to your tests, we can refactor it at some point if we agree on it.

rosen-vladimirov: We were discussing the same, but I didn't have time to implement it last week. As it looks like we all find the current code difficult to read, I've invested some time to implement your second suggestion. 
Thanks for the idea.

lib/tools/node-modules/node-modules-dependencies-builder.ts
```diff
@@ -1,111 +1,98 @@
...
+			const resolvedDependency = this.findModule(rootNodeModulesPath, currentModule.parentDir, currentModule.name, currentModule.depth);
 
-	private traverseDependency(name: string, currentModulePath: string, depth: number): void {
-		// Check if child has been extracted in the parent's node modules, AND THEN in `node_modules`
-		// Slower, but prevents copying wrong versions if multiple of the same module are installed
-		// Will also prevent copying project's devDependency's version if current module depends on another version
-		let modulePath = path.join(currentModulePath, "node_modules", name); // node_modules/parent/node_modules/<package>
-		let alternativeModulePath = path.join(this.rootNodeModulesPath, name);
+			if (resolvedDependency && !_.some(resolvedDependencies, r => r.directory === resolvedDependency.directory)) {
+				const deps = _.map(resolvedDependency.dependencies, d => ({ name: d, parentDir: resolvedDependency.directory, depth: resolvedDependency.depth + 1 }));
```
yyosifov: I think we need to filter out here for:

1. Not add to the Queue dependencies we've already resolved (i.e. if 2 modules depend on lodash, we'll add both for traversing)
2. If we have Cyclic reference in the Dependencies, we can add them forever as we don't check when adding to the queue.

Pip3r4o: What if 2 modules depend and work with different versions of lodash? Shouldn't we traverse both? At least that's one of the problems the original algorithm was trying to solve.

yyosifov:
Great work! Thanks @rosen-vladimirov

lib/tools/node-modules/node-modules-dependencies-builder.ts
rosen-vladimirov: @yyosifov  the checks are added.
@Pip3r4o the logic is the same, I've just replaced the recursion as it was hard to read and also incorrect on some places, which caused the issue. For example my testing project has 171 production dependencies, the previous implementation returned 1671. The new implementation is more than 5times faster for large projects (230 ms with old code, 40ms with new one), but the real benefit is during project preparation, where we do not iterate over 1671 objects.

Pip3r4o: My response was to yyosifov's remark. I, by no means, believe that the original recursion was absolutely correct! Great job on the optimization!

yyosifov: @Pip3r4o sure, he checks for the full path of the dependency, so it won't happen.

PR #2790, opened 2017-05-09
Fix installation of tns-core-modules during project creation

rosen-vladimirov:
CLI tries to install `tns-core-modules` during project creation. In case the template does not have `tns-core-modules` in it, CLI will install latest version of `tns-core-modules` which may (and most probably is) be incompatible with the runtime and other dependencies of the project.
In such cases ensure the CLI installs version of tns-core-modules that matches its own version, i.e. when CLI is 2.5.2, it should install latest 2.5.x core modules.
Also ensure exact versions are installed and persisted by `npmInstallationManager`.

### DIFF
--- a/lib/services/project-service.ts
+++ b/lib/services/project-service.ts
@@ -14,7 +14,8 @@ export class ProjectService implements IProjectService {
 		private $projectHelper: IProjectHelper,
 		private $projectNameService: IProjectNameService,
 		private $projectTemplatesService: IProjectTemplatesService,
-		private $staticConfig: IStaticConfig) { }
+		private $staticConfig: IStaticConfig,
+		private $npmInstallationManager: INpmInstallationManager) { }
 
 	@exported("projectService")
 	public async createProject(projectOptions: IProjectSettings): Promise<void> {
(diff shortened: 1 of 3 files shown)


lib/services/project-service.ts
```diff
@@ -50,10 +51,12 @@ export class ProjectService implements IProjectService {
 
 			await this.ensureAppResourcesExist(projectDir);
 
-			let packageName = constants.TNS_CORE_MODULES_NAME;
-			await this.$npm.install(packageName, projectDir, { save: true, "save-exact": true });
-
 			let templatePackageJsonData = this.getDataFromJson(templatePath);
+
+			if (!(templatePackageJsonData && templatePackageJsonData.dependencies && templatePackageJsonData.dependencies[constants.TNS_CORE_MODULES_NAME])) {
```
yyosifov: @rosen-vladimirov  Isn't it more correct to install the correct version of the Template? I.e. if you install 3.0.0 template and you override the tns_core_modules, won't still there be a possibility that this 3.0.0 template doesn't work with them? For pure NS I don't know, but for NS+Angular I think it's likely?

yyosifov: What if the Template contains the `tns-core-modules-widgets` package? Shall we add it to the list of specific dependencies.

rosen-vladimirov: @yyosifov the template versioning may differ and we also do not have control over the templates - basically everything is template. 
Also the current code will install `tns-core-modules` only in case they are not defined in the template itself.

rosen-vladimirov: the `tns-core-modules-widgets` are dependency of the `tns-core-modules`. In case they exist in the template, it means someone has a very good reason to add them. I suggest to skip checks for them until we find a proper use-case.

PR #2793, opened 2017-05-09
Respect current version of CLI when searching for runtimes

yyosifov:
When searching for the ios/android runtimes respect the current version of the CLI and don't search only for matching versions - i.e. if I have CLI 2.5.4 don't look for runtimes >= 2.5.4 only, as they can be available only up to 2.5.2 which is still fine and working for us. So I've changed it to look for 2.5.x or major.minor.*any* compatible versions of the packages, when searching for them.

Replaces $npm.install with the $npmInstallationManager.install 

Updated the semver, to add the prerelease function missing in the last version


yyosifov:
run ci

yyosifov:
@dtopuzov can you confirm this regression test fails in 2.5.4 official, too? So that we can proceed with the merge with 1 regression test failing (as it's fixed in 3.x)

PR #2802, opened 2017-05-10
Remove .html from the files included in the fast-livesync to force fu…

yyosifov:
When syncing changes from .html files do a full app-restart. It was included for fast-sync, non-app-restart behavior which caused issues in Angular apps.


dtopuzov:
run ci

PR #2806, opened 2017-05-11
Fix UserId passing to Analytics Service

yyosifov:
Fix the UserId retrieval in the Analytics Service, when initializing as we have to await the Promise.


yyosifov:
@rosen-vladimirov I've updated the `mobile-cli-lib` to latest `release` branch
@dtopuzov I haven't created a issue for this. Can you please help me with defining one or just we can test it out?

PR #2817, opened 2017-05-16
Add awaitWriteFinish option to choki in livesync to wait file to be r…

yyosifov:
…eady to be read


yyosifov:
@dtopuzov Can you please confirm this doesn't break the Tests or something else? The affected users have confirmed this fixes the IDE's temporary files problem, so we just need a confirmation that it doesn't break everyone else :)

PR #2819, opened 2017-05-16
Exclude .css/.xml from fast sync files for angular

yyosifov:
Don't fast-reload on .css/.xml changes for Angular apps, as they still break the router and the App freezes.


yyosifov:
@rosen-vladimirov the experts on Angular have fixed the issue with the refresh for .css/html. I'll close and create PR to revert the .html change ;) So good

PR #2827, opened 2017-05-21
Fix unit tests on non-macOS

rosen-vladimirov:
Currently unit tests from master branch are failing on Windows and Linux. Fix them by applying the following:
- Fix expected path in entitlements test - use `path.join` instead of hardcoded `/` - on Windows the path separator is `\`
- Ignore tests for merging xcconfig files as they can work only on macOS, where ruby is installed. The tests will be ran only in case you are executing them on macOS.
- Require "colors" module in test-bootstrap. During test execution I've noticed some "undefined" statements printed on the terminal. It turned out these are `logger.warn` statements. However the logic in the logger is to print `<message>.yellow`.
When `colors` is not required, the strings do not have `yellow` property, so instead of seeing the real message, the loggger has been printing "undefined".

### DIFF
--- a/test/test-bootstrap.ts
+++ b/test/test-bootstrap.ts
@@ -9,13 +9,16 @@ const cliGlobal = <ICliGlobal>global;
 cliGlobal._ = require("lodash");
 cliGlobal.$injector = require("../lib/common/yok").injector;
 
+// Requiring colors will modify the prototype of String
+// We need it as in some places we use <string>.<color>, which is undefined when colors is not required
+// So we sometimes miss warnings in the tests as we receive "undefined".
+require("colors");
+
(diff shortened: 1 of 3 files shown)


test/test-bootstrap.ts
```diff
@@ -9,13 +9,16 @@ const cliGlobal = <ICliGlobal>global;
...
+
 use(require("chai-as-promised"));
 
 $injector.register("analyticsService", {
-	trackException: (): { wait(): void } => {
-		return {
-			wait: () => undefined
-		};
+	trackException: async (exception: any, message: string): Promise<void> => {
+		// Intentially left blank.
```
TsvetanMilanov: `Intentially` -> `Intentionally`?

PR #2833, opened 2017-05-22
Prevent starting a simulator when build --forDevice

yyosifov:
When we're executing

`tns build {platform} --forDevice`

we don't need to fire up the Simulator. We've changed the common lib to respect a flag named `isBuildForDevice` and in this case not to start the Simulator if no real devices are found attached.

Here we're just passing this `isBuildForDevicе` when running build forDevice


dtopuzov:
run ci

yyosifov:
run ci

PR #2835, opened 2017-05-22
Fix prepare scoped dependency with native code

Pip3r4o:
Preparing a scoped dependency containing `platforms` wouldn't remove them in the prepared `tns_modules` location because the path built was wrong.


Pip3r4o:
Closed in favor of https://github.com/NativeScript/nativescript-cli/issues/2837

PR #2837, opened 2017-05-22
Flatten scoped module when copying into android/src

Pip3r4o:
Flatten scoped npm plugins containing native code when copying into android/src. This is necessary to properly handle plugin flavors (F1, F2, etc.) when building for android.

```
@progress/nativescript-ui-pro -> nativescript-ui-pro
tns-core-modules -> tns-core-modules
```

branches off commit https://github.com/NativeScript/nativescript-cli/pull/2835
Edit: Includes fix:
Preparing a scoped dependency containing platforms wouldn't remove them in the prepared tns_modules location because the path built was wrong.

`assets/app/tns_modules/@scope/nativescript-with-native-dep` used to preserve its `platforms` dir

### DIFF
--- a/lib/tools/node-modules/node-modules-dest-copy.ts
+++ b/lib/tools/node-modules/node-modules-dest-copy.ts
@@ -36,18 +36,20 @@ export class TnsModulesCopy {
 
 	private copyDependencyDir(dependency: IDependencyData): void {
 		if (dependency.depth === 0) {
+			const targetPackageDir = path.join(this.outputRoot, dependency.name);
+
+			shelljs.mkdir("-p", targetPackageDir);
+
 			let isScoped = dependency.name.indexOf("@") === 0;
-			let targetDir = this.outputRoot;
 
 			if (isScoped) {
-				targetDir = path.join(this.outputRoot, dependency.name.substring(0, dependency.name.indexOf("/")));
+				// copy module into tns_modules/@scope/module instead of tns_modules/module
+				shelljs.cp("-Rf", dependency.directory, path.join(this.outputRoot, dependency.name.substring(0, dependency.name.indexOf("/"))));
+			} else {
+				shelljs.cp("-Rf", dependency.directory, this.outputRoot);
 			}
 
-			shelljs.mkdir("-p", targetDir);
-			shelljs.cp("-Rf", dependency.directory, targetDir);
-
 			// remove platform-specific files (processed separately by plugin services)
-			const targetPackageDir = path.join(isScoped ? this.outputRoot : targetDir, dependency.name);
 			shelljs.rm("-rf", path.join(targetPackageDir, "platforms"));
 		}
 	}
--- a/lib/tools/node-modules/node-modules-dest-copy.ts
+++ b/lib/tools/node-modules/node-modules-dest-copy.ts
@@ -46,8 +46,8 @@ export class TnsModulesCopy {
 			shelljs.mkdir("-p", targetDir);
 			shelljs.cp("-Rf", dependency.directory, targetDir);
 
-			//remove platform-specific files (processed separately by plugin services)
-			const targetPackageDir = path.join(targetDir, dependency.name);
+			// remove platform-specific files (processed separately by plugin services)
+			const targetPackageDir = path.join(isScoped ? this.outputRoot : targetDir, dependency.name);
 			shelljs.rm("-rf", path.join(targetPackageDir, "platforms"));
 		}
 	}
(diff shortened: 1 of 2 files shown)


lib/tools/node-modules/node-modules-dest-copy.ts
```diff
@@ -46,8 +46,8 @@ export class TnsModulesCopy {
 			shelljs.mkdir("-p", targetDir);
 			shelljs.cp("-Rf", dependency.directory, targetDir);
 
-			//remove platform-specific files (processed separately by plugin services)
-			const targetPackageDir = path.join(targetDir, dependency.name);
+			// remove platform-specific files (processed separately by plugin services)
+			const targetPackageDir = path.join(isScoped ? this.outputRoot : targetDir, dependency.name);
```
rosen-vladimirov: Is this really correct? If I understand correctly, the idea here is to copy `<project dir>/node_modules/@<scoped_name>/<package_name>` to `<projectDir>/platforms/<path to tns_modules>/@<scoped_name>/<package_name>` and remove the platforms dir from the destination. From this code it looks like we'll try to delete `<projectDir>/platforms/<path to tns_modules>/platforms` dir. Am I correct?

Pip3r4o: Assuming we are looking at a scoped dependency, `dependency.name` will yield `@scope/module`, and `this.outputRoot` is `platforms/.../app/tns_modules` meaning the path to be removed on the following line will look like this: `app/tns_modules/@scope/module/platforms`. Or am I missing something?

rosen-vladimirov: Ok, it looks like I'm still sleeping 😃 
Anyway, the whole method is confusing - we shouldn't have different logic for scoped dependencies, I suggest the following change:
```TypeScript
private copyDependencyDir(dependency: IDependencyData): void {
	if (dependency.depth === 0) {
		const targetPackageDir = path.join(this.outputRoot, dependency.name);

		shelljs.mkdir("-p", targetPackageDir);
		shelljs.cp("-Rf", dependency.directory, targetPackageDir);

		//remove platform-specific files (processed separately by plugin services)
		shelljs.rm("-rf", path.join(targetPackageDir, "platforms"));
	}
}
```

Also do you have an idea why are we checking only dependencies on level 0 ?

Pip3r4o: Now that I look at the method, I'm feeling stupid for not noticing the unnecessary complexity the scoped dep check adds to it. Either I was sleeping while writing that, or shelljs's mkdir didn't work recursively back in October :D, or we were trying to simulate proper scoped dependency dir structure in case the user had npm 2, although I am not sure I am recalling correctly.

The dependency.depth === 0 check was added as an optimization step with the assumption that root-level production node_modules will bring along all other production dependencies, possibly done with npm 2 support in mind.

rosen-vladimirov:
After green build

Pip3r4o:
@rosen-vladimirov I decided to keep the isScoped logic for the reasons written here -> https://github.com/NativeScript/nativescript-cli/pull/2837/files#diff-181af290f4b97ed6aa6fc82961eed07eR40. 

For example the code as per your suggestion would copy `lodash` into `tns_modules/lodash/lodash` and not `tns_modules/lodash`; While not making a @scope dir disrupted shelljs' copy.

Pip3r4o:
Reversed the depth === 0 checks as they are necessary to prevent the same module with different versions  clashing with itself.

zhangsuimi:
i have a similar problem now, it happens when gradle is building

```
Preparing project...
Successfully prepared plugin nativescript-theme-core for android.
Successfully prepared plugin nativescript-ui-listview for android.
Successfully prepared plugin tns-core-modules for android.
Successfully prepared plugin nativescript-ui-core for android.
Successfully prepared plugin tns-core-modules-widgets for android.
Project successfully prepared (Android)
Building project...
Gradle build...

> IDLE
Error: read ECONNRESET
    at exports._errnoException (util.js:1020:11)
    at TCP.onread (net.j<-------------> 0% INITIALIZING [1s]
         + applying user-defined configuration from C:\Users\lenovo\appname\app

\App_Resources\Androi
d\app.gradle
Configuration 'compile' in project ':app' is deprecated. Use 'implementation' instead.
Configuration 'debugCompile' in project ':app' is deprecated. Use 'debugImplementation' 

instead.
         + adding nativescript runtime package dependency: nativescript-optimized
         + adding aar plugin dependency: C:\Users\lenovo\appname\node_modules\nativescript-

ui-listvi
ew\platforms\android\TNSListView-release.aar
         + adding aar plugin dependency: C:\Users\lenovo\appname\node_modules\nativescript-

ui-core\p
latforms\android\TNSCore-release.aar
         + adding aar plugin dependency: C:\Users\lenovo\appname\node_modules\tns-core-

modules-widge
ts\platforms\android\widgets-release.aar

* What went wrong:
A problem occurred configuring project ':app'.
> Flavor 'nativescript-ui-listview' has unknown dimension 'nativescript-ui-listview'.

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to 

get more l
og output.

* Get more help at https://help.gradle.org

BUILD FAILED in 1m 16s
```

what should i do ?

PR #2848, opened 2017-05-26
Fix permissions on ~/.local/share/.nativescript-cli with Sudo runs

yyosifov:
Update lib/common after fix when persisting locally the user-settings to change the owner to be the `current user`, not the `root`.


vchimev:
run ci

pkoleva:
The team said you deserve this
sn

PR #2863, opened 2017-05-30
Add CFBundleURLSchemes when building for iOS in debug.

TomaNikolov:
We will use custom URL scheme to open iOS app from the browser. 
This way we will support app restart on Windows. 
Related to:  https://github.com/NativeScript/ios-runtime/issues/746

### DIFF
--- a/lib/services/ios-project-service.ts
+++ b/lib/services/ios-project-service.ts
@@ -329,8 +329,10 @@ export class IOSProjectService extends projectServiceBaseLib.PlatformProjectServ
 
 		// build only for device specific architecture
 		if (!buildConfig.release && !buildConfig.architectures) {
-			await this.$devicesService.initialize({ platform: this.$devicePlatformsConstants.iOS.toLowerCase(), deviceId: buildConfig.device,
-				isBuildForDevice: true });
+			await this.$devicesService.initialize({
+				platform: this.$devicePlatformsConstants.iOS.toLowerCase(), deviceId: buildConfig.device,
+				isBuildForDevice: true
+			});
 			let instances = this.$devicesService.getDeviceInstances();
 			let devicesArchitectures = _(instances)
 				.filter(d => this.$mobileHelper.isiOSPlatform(d.deviceInfo.platform) && d.deviceInfo.activeArchitecture)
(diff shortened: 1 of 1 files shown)


lib/services/ios-project-service.ts
```diff
@@ -704,7 +704,7 @@ We will now place an empty obsolete compatability white screen LauncScreen.xib f
 		return Promise.resolve();
 	}
 
-	private async mergeInfoPlists(projectData: IProjectData): Promise<void> {
+	private async mergeInfoPlists(release: boolean, projectData: IProjectData): Promise<void> {
```
rosen-vladimirov: According to our coding guidelines, we prefer passing objects with the required boolean properties, which will make the calling of the function more readable. In this case I suggest you to use the `IRelease` interface, for example:
```TypeScript
private async mergeInfoPlists(buildOption: IRelease, projectData: IProjectData): Promise<void> {
```
In the method body you'll check if `buildOptions.release` is true. The caller will pass `{ release: <value> `

rosen-vladimirov: Also I prefer having the projectData as a first argument... but I'm not sure why I prefer it

lib/services/ios-project-service.ts
```diff
@@ -672,7 +672,7 @@ We will now place an empty obsolete compatability white screen LauncScreen.xib f
 	}
 
 	public async processConfigurationFilesFromAppResources(release: boolean, projectData: IProjectData): Promise<void> {
-		await this.mergeInfoPlists(projectData);
+		await this.mergeInfoPlists({release}, projectData);
```
Mitko-Kerezov: Could you format this (either put spaces around the curly brackets or use VS Code's formatting)

lib/services/ios-project-service.ts
```diff
@@ -704,7 +704,7 @@ We will now place an empty obsolete compatability white screen LauncScreen.xib f
 		return Promise.resolve();
 	}
 
-	private async mergeInfoPlists(projectData: IProjectData): Promise<void> {
+	private async mergeInfoPlists(buildOption: IRelease, projectData: IProjectData): Promise<void> {
```
Mitko-Kerezov: I'd rather we name `buildOption` argument to `buildOptions` (plural) as we could put more options in it in the future

PR #2864, opened 2017-05-30
Make project-changes-service stateless

Mitko-Kerezov:
Better to be stateless in a long living process.


Mitko-Kerezov:
Making `project-changes-service` stateless has proven to be a tad more difficult than initially though. 
As this functionality is not essential for us at the moment I'm closing this PR but will leave the branch so that we can take the work up in the future and finish it.

PR #2868, opened 2017-06-01
Expose loadExtension to fix double loading of packages

rosen-vladimirov:
In case you install extension, the method automatically loads it. However calling `loadExtensions` after that loads it again, so our `$injector` fails.
In order to fix this, remove the loading from the `installExtension` method and expose `loadExtension(name)` method.
So in the future in case someone wants to load a new an extension after calling `loadExtensions`, they'll be able to use the new method.

Update ios-device-lib
Update `ios-device-lib` to 0.4.3. The new version includes a fix for long living processes where detaching iOS device causes a huge memory consuption and core dump.

### DIFF
--- a/PublicAPI.md
+++ b/PublicAPI.md
@@ -94,12 +94,12 @@ interface IExtensionData {
 ```
 
 ### installExtension
-Installs specified extension and loads it in the current process, so the functionality that it adds can be used immediately.
+Installs specified extension.
 
 * Definition:
 ```TypeScript
 /**
- * Installs and loads specified extension.
+ * Installs a specified extension.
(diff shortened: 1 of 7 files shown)


lib/definitions/extensibility.d.ts
```diff
@@ -36,6 +36,14 @@ interface IExtensibilityService {
 	loadExtensions(): Promise<IExtensionData>[];
 
 	/**
+	 * Loads a single extension, so their methods and commands can be used from CLI.
```
Mitko-Kerezov: their -> its

PublicAPI.md
```diff
@@ -180,6 +180,31 @@ for (let promise of loadExtensionsPromises) {
 }
 ```
 
+### loadExtension
+Loads a specified extension.
+
+* Definition
+```TypeScript
+/**
+ * Loads a single extension, so their methods and commands can be used from CLI.
```
Mitko-Kerezov: their  -> its

Issue #2874, opened 2017-06-05
Unable to build and deploy app to iTunes: Unable to connect to iTunes Connect


dtopuzov:
_From @briandilley on June 3, 2017 22:53_

`    /usr/bin/touch -c /Users/distiller/Library/Developer/Xcode/DerivedData/quickridensmobile-bnzhogcsicpkjudhwfepmetihxvs/Build/Intermediates/ArchiveIntermediates/quickridensmobile/BuildProductsPath/Release-iphoneos/quickridensmobile.app.dSYM

** ARCHIVE SUCCEEDED **

Archive at: /Users/distiller/quickride-ns-mobile/platforms/ios/build/archive/quickridensmobile.xcarchive
2017-06-03 15:36:48.606 xcodebuild[17633:31256] [MT] IDEDistribution: -[IDEDistributionLogging _createLoggingBundleAtPath:]: Created bundle at path '/var/folders/jm/fw86rxds0xn69sk40d18y69m0000gp/T/quickridensmobile_2017-06-03_15-36-48.605.xcdistributionlogs'.
Exported quickridensmobile.xcarchive to: /Users/distiller/quickride-ns-mobile/platforms/ios/build/archive
** EXPORT SUCCEEDED **

Export at: /Users/distiller/quickride-ns-mobile/platforms/ios/build/archive/quickridensmobile.ipa
Unable to connect to iTunes Connect
An error occurred while processing the json request. (1268)`

_Copied from original issue: NativeScript/NativeScript#4310_

OscarLopezArnaiz:
+1 Having this issue since Friday, thought it was iTunes having some issues but today got same error message when trying to publish a build.

dtopuzov:
@OscarLopezArnaiz @briandilley we can confirm this is valid issue and we will try to provide fix soon.

OscarLopezArnaiz:
Cheers @dtopuzov  

As a temporal workout [application loader worked for me](https://developer.apple.com/library/content/documentation/LanguagesUtilities/Conceptual/iTunesConnect_Guide/Chapters/UploadingBinariesforanApp.html).

To publish a build with Application Loader

1. Launch Xcode.
2. Choose Xcode > Open Developer Tool > Application Loader.
3. [Upload your application binary files](http://help.apple.com/itc/apploader/e3#/apdATD1E12-D1E1A1303-D1E12A1126)

vchimev:
[...]

PR #2882, opened 2017-06-08
Fix the `tns publish ios` command

PanayotCankov:
The http-client was refactored and was sending the requests in chunks. Added Content-Length header to prevent the "request" module from using chunks.


dtopuzov:
:1st_place_medal:

vchimev:
Fixes: https://github.com/NativeScript/nativescript-cli/issues/2874.

PanayotCankov:
Fixes: #2874

PR #2895, opened 2017-06-16
Pass mobileProvisionData to prepare

rosen-vladimirov:
On Windows and Linux we are unable to execute macOS specific commands. During project preparation for iOS we need to know the data from specified provison. In this scenario, pass the data from it as a non mandatory parameter.
This way the default behavior will be kept, but we'll have a way to skip the `security` command which fails on Windows.


rosen-vladimirov:
run ci

PR #2897, opened 2017-06-16
Fix debug device/web socket closing

TsvetanMilanov:
When we close the websocket we have to destroy the device socket as well.
We need to wait the close event on the device socket instead of end.


TsvetanMilanov:
@zahhak I created an issue -> https://github.com/NativeScript/ios-runtime/issues/832 with some additional information.

PR #2899, opened 2017-06-16
Return information about the error if loadExtension fails

nadyaA:
In the CLI the error is visible, but not in clients that are using it.

http://teampulse.telerik.com/view#item/345796

### DIFF
--- a/lib/services/extensibility-service.ts
+++ b/lib/services/extensibility-service.ts
@@ -84,7 +84,7 @@ export class ExtensibilityService implements IExtensibilityService {
 			return { extensionName };
 		} catch (error) {
 			this.$logger.warn(`Error while loading ${extensionName} is: ${error.message}`);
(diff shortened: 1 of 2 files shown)


lib/services/extensibility-service.ts
```diff
@@ -84,7 +84,7 @@ export class ExtensibilityService implements IExtensibilityService {
 			return { extensionName };
 		} catch (error) {
 			this.$logger.warn(`Error while loading ${extensionName} is: ${error.message}`);
-			const err = <IExtensionLoadingError>new Error(`Unable to load extension ${extensionName}. You will not be able to use the functionality that it adds.`);
+			const err = <IExtensionLoadingError>new Error(`Unable to load extension ${extensionName}. You will not be able to use the functionality that it adds. ${error.message}`);
```
rosen-vladimirov: This looks strange, can we add something like `...Error is: ${error.message}`

dtopuzov:
run ci

rosen-vladimirov:
run ci

PR #2900, opened 2017-06-16
Emit liveSyncStopped for every device

Mitko-Kerezov:
In case one wants to stop livesync for all devices emit events for each device instead of one event for them all.

Ping @rosen-vladimirov


dtopuzov:
run ci

rosen-vladimirov:
run ci

rosen-vladimirov:
run ci

PR #2904, opened 2017-06-19
Allow prepare for Android without ANDROID_HOME

rosen-vladimirov:
Allow preparing Android project without the requirement to have ANDROID_HOME set.
The change is in the required services, which should work without ANDROID_HOME in some cases.
However the commands should not allow this, so move the validation in the commands.

Also addresses: https://github.com/NativeScript/nativescript-cli/issues/2905

### DIFF
--- a/lib/commands/prepare.ts
+++ b/lib/commands/prepare.ts
@@ -4,7 +4,8 @@ export class PrepareCommand implements ICommand {
 	constructor(private $options: IOptions,
 		private $platformService: IPlatformService,
 		private $projectData: IProjectData,
-		private $platformCommandParameter: ICommandParameter) {
+		private $platformCommandParameter: ICommandParameter,
+		private $platformsData: IPlatformsData) {
 		this.$projectData.initializeProjectData();
 	}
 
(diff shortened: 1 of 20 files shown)


lib/commands/prepare.ts
```diff
@@ -14,7 +15,12 @@ export class PrepareCommand implements ICommand {
 	}
 
 	public async canExecute(args: string[]): Promise<boolean> {
-		return await this.$platformCommandParameter.validate(args[0]) && await this.$platformService.validateOptions(this.$options.provision, this.$projectData, args[0]);
+		const platform = args[0];
+		const result = await this.$platformCommandParameter.validate(platform) && await this.$platformService.validateOptions(this.$options.provision, this.$projectData, platform);
+		const platformData = this.$platformsData.getPlatformData(platform, this.$projectData);
+		const platformProjectService = platformData.platformProjectService;
+		await platformProjectService.validate(this.$projectData);
```
TsvetanMilanov: Can we skip this check if `!result`?

rosen-vladimirov:
@Mitko-Kerezov we'll handle this later, for the moment I'll merge it this way.
We have to find the best way to have validation for iOS and Android.

lib/commands/prepare.ts
```diff
@@ -17,9 +17,12 @@ export class PrepareCommand implements ICommand {
 	public async canExecute(args: string[]): Promise<boolean> {
 		const platform = args[0];
 		const result = await this.$platformCommandParameter.validate(platform) && await this.$platformService.validateOptions(this.$options.provision, this.$projectData, platform);
-		const platformData = this.$platformsData.getPlatformData(platform, this.$projectData);
-		const platformProjectService = platformData.platformProjectService;
-		await platformProjectService.validate(this.$projectData);
+		if (!result) {
```
TsvetanMilanov: I think we should check `if (result)`

PR #2907, opened 2017-06-20
Fix adding specific platform

rosen-vladimirov:
When trying to execute `tns platform add <platform>@<version>` we fail with message: `Cannot read property 'platformProjectService' of undefined`
The reason is incorrect check when getting platforms-data where we do not expect the version. Split the incoming string by `@` and use the real platform when trying to get the platforms data.


rosen-vladimirov:
run ci

PR #2910, opened 2017-06-21
Fix platform update command

rosen-vladimirov:
`tns platform update` command should install the latest platform in case it is not installed. Currently this fails (regression from previous commit).
Fix this by calling correct validate method.


dtopuzov:
run ci

Issue #2916, opened 2017-06-23
If no device or emulator is attached `tns debug android` kills the commandline process and doesn't start an emulator


pkoleva:
### Did you verify this is a real problem by searching [Stack Overflow](http://stackoverflow.com/questions/tagged/nativescript) and the [other open issues in this repo](https://github.com/NativeScript/nativescript/issues)?
Yes

### Tell us about the problem
`tns debug android` kills the CLI process if no emulator or device is connected. It should start an emulator, deploy the app, start it and attach the debugger

### Which platform(s) does your issue occur on?
Android

### Please provide the following version numbers that your issue occurs with:
- CLI: 3.1.0

### Please tell us how to recreate the issue in as much detail as possible.
1. Create an app
2. Execute `tns debug android`

PR #2917, opened 2017-06-23
Fix not starting emulator during debug command

rosen-vladimirov:
`tns debug <platform>` command should start emulator in case device is not attached. The current command does not start emulator as we have passed `skipInferPlatform` - when this option is used, devicesService does not know for which platform to start the emulator.
Fix this by passing correct options.

Fixes https://github.com/NativeScript/nativescript-cli/issues/2916


rosen-vladimirov:
I'll take a different approach to fix this. Closing this PR.

PR #2918, opened 2017-06-23
The dev webpack does prepare + run, but run always provided bundle: false, and triggered incorrect re-prepare

PanayotCankov:


rosen-vladimirov:
Why do you need webpack + liveSync - currently webpack is used for release builds, where livesync will not work?

PanayotCankov:
Because I need to fix the webpack for an android app but I haven't published it yet and I don't have keystore keys to use, and I also need to collect proper logs to figure out the problems so I don't want to use --release?

PanayotCankov:
See also: https://github.com/NativeScript/android-runtime/pull/788
It will prevent the `gradle clean` to clean the `configurations` in `platforms/android`.

rosen-vladimirov:
Okay, but will the livesync work with webpack, I assume the workflow is:
 - prepare the project
 - webpack's after prepare triggers some additional changes
 - application is build and installed on device
 - FileSystem watcher is started
 - a change in `.ts` file is applied
 - CLI prepares the project
 - webpack's after prepare triggers some additional changes that will remove the transpiled `.js` from `platforms/android/...`
 - CLI tries to upload the file from `platforms/android/.../`  to device and fails

Am I correct? How is this supposed to work?

PanayotCankov:
Amended.

PR #2925, opened 2017-06-26
Fix android with webpack

PanayotCankov:
`gradle clean` cleans the configurations made by the CLI. This PR will recreate them when running with bundle.

### DIFF
--- a/lib/services/platform-service.ts
+++ b/lib/services/platform-service.ts
@@ -39,7 +39,8 @@ export class PlatformService extends EventEmitter implements IPlatformService {
 		private $npm: INodePackageManager,
 		private $devicePlatformsConstants: Mobile.IDevicePlatformsConstants,
 		private $projectChangesService: IProjectChangesService,
-		private $analyticsService: IAnalyticsService) {
+		private $analyticsService: IAnalyticsService,
+		private $nodeModulesDependenciesBuilder: INodeModulesDependenciesBuilder) {
 		super();
 	}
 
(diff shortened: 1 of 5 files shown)


PanayotCankov:
run ci

dtopuzov:
run ci

lib/services/platform-service.ts
```diff
@@ -223,14 +224,20 @@ export class PlatformService extends EventEmitter implements IPlatformService {
...
 
 		this.$logger.trace("Changes info in prepare platform:", changesInfo);
 
 		if (changesInfo.hasChanges) {
 			await this.cleanProject(platform, appFilesUpdaterOptions, platformData, projectData);
+		}
+		if (changesInfo.hasChanges || bundle) {
 			await this.preparePlatformCore(platform, appFilesUpdaterOptions, projectData, config, changesInfo, filesToSync);
+		}
+		if (changesInfo.hasChanges) {
```
rosen-vladimirov: So in case `--bundle` is passed, we'll not update the prepareInfo file. Is this expected?

lib/services/platform-service.ts
```diff
@@ -302,6 +309,16 @@ export class PlatformService extends EventEmitter implements IPlatformService {
...
 		if (!changesInfo || changesInfo.modulesChanged) {
 			await this.copyTnsModules(platform, projectData);
+		} else if (appFilesUpdaterOptions.bundle) {
+			let dependencies = this.$nodeModulesDependenciesBuilder.getProductionDependencies(projectData.projectDir);
+			for (let dependencyKey in dependencies) {
+				const dependency = dependencies[dependencyKey];
+				let isPlugin = !!dependency.nativescript;
+				if (isPlugin) {
+					let pluginData = this.$pluginsService.convertToPluginData(dependency, projectData.projectDir);
+					await this.$pluginsService.preparePluginNativeCode(pluginData, platform, projectData);
```
rosen-vladimirov: instead of exposing `preparePluginNativeCode` and `convertToPluginData` can't we just call `this.$pluginsService.prepare...` ?

PanayotCankov: No, it will move scripts to app/tns_modules  I think, and we need to avoid that :(

lib/services/platform-service.ts
PanayotCankov: I strongly agree the whole --bundle if-fing is wrong.

PR #2928, opened 2017-06-27
Update TS to latest release

nadyaA:
Fix corresponding tests

https://github.com/telerik/mobile-cli-lib/pull/980

### DIFF
--- a/lib/commands/debug.ts
+++ b/lib/commands/debug.ts
@@ -24,15 +24,15 @@
 		await this.$platformService.trackProjectType(this.$projectData);
 
 		if (this.$options.start) {
-			return this.$debugLiveSyncService.printDebugInformation(await this.debugService.debug<string[]>(debugData, debugOptions));
+			return this.$debugLiveSyncService.printDebugInformation(await this.debugService.debug(debugData, debugOptions));
 		}
 
 		this.$config.debugLivesync = true;
 
 		await this.$devicesService.detectCurrentlyAttachedDevices();
 
-		const devices = this.$devicesService.getDeviceInstances();
 		// Now let's take data for each device:
+		const devices = this.$devicesService.getDeviceInstances();
 		const deviceDescriptors: ILiveSyncDeviceInfo[] = devices.filter(d => !this.platform || d.deviceInfo.platform === this.platform)
 			.map(d => {
 				const info: ILiveSyncDeviceInfo = {
@@ -91,10 +91,7 @@
 		}
 
 		if (this.$devicesService.deviceCount > 1) {
-			// Starting debugger on emulator.
-			this.$options.emulator = true;
-
-			this.$logger.warn("Multiple devices found! Starting debugger on emulator. If you want to debug on specific device please select device with --device option.".yellow.bold);
+			this.$errors.failWithoutHelp("Multiple devices found! To debug on specific device please select device with --device option.");
 		}
 
 		return true;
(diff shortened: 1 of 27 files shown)


lib/services/ios-debug-service.ts
```diff
@@ -171,20 +171,20 @@ class IOSDebugService extends DebugServiceBase implements IPlatformDebugService
...
 	}
 
-	private async deviceStart(debugData: IDebugData, debugOptions: IDebugOptions): Promise<string[]> {
+	private async deviceStart(debugData: IDebugData, debugOptions: IDebugOptions): Promise<string> {
 		await this.$devicesService.initialize({ platform: this.platform, deviceId: debugData.deviceIdentifier });
 		const action = async (device: Mobile.IiOSDevice) => device.isEmulator ? await this.emulatorStart(debugData, debugOptions) : await this.deviceStartCore(device, debugData, debugOptions);
-		const results = await this.$devicesService.execute(action, this.getCanExecuteAction(debugData.deviceIdentifier));
-		return _.map(results, r => r.result);
+		const deviceActionResult = await this.$devicesService.execute(action, this.getCanExecuteAction(debugData.deviceIdentifier));
+		return deviceActionResult[0].result;
```
rosen-vladimirov: previously this has not been failing when the `deviceActionResult` is `null/undefined` or empty array, but now it will fail with error - `Cannot read property result of undefined.` Is this expected? In case not, you can use:
```TypeScript
return deviceActionResult && deviceActionResult[0] && deviceActionResult[0].result;
```

test/plugins-service.ts
```diff
@@ -403,17 +391,15 @@ describe("Plugins service", () => {
 						}
 					},
 					"devDependencies": {
-						"grunt": "0.4.2"
+						"grunt": "0.4.`2`"
```
rosen-vladimirov: ?

lib/services/livesync/debug-livesync-service.ts
```diff
@@ -66,7 +66,7 @@ export class DebugLiveSyncService extends LiveSyncService implements IDebugLiveS
 		const buildConfig: IBuildConfig = _.merge({ buildForDevice: !deviceAppData.device.isEmulator }, deployOptions);
 		debugData.pathToAppPackage = this.$platformService.lastOutputPath(debugService.platform, buildConfig, projectData);
 
-		this.printDebugInformation(await debugService.debug<string[]>(debugData, debugOptions));
+		this.printDebugInformation([await debugService.debug(debugData, debugOptions)]);
```
rosen-vladimirov: we always call `printDebugInformation` by manually converting some string to array, can't we change the `printDebugInformation` method to has `string` for parameter instead of `string[]`

lib/services/ios-debug-service.ts
rosen-vladimirov: Agreed that this will never happen. Disregard the comment,

lib/commands/debug.ts
```diff
@@ -22,16 +22,16 @@
 		let debugData = this.$debugDataService.createDebugData(this.$projectData, this.$options);
 
 		await this.$platformService.trackProjectType(this.$projectData);
+		const devices = this.$devicesService.getDeviceInstances();
```
rosen-vladimirov: why is this line moved? By definition the `const`s should be declared where they are used, not before that.

nadyaA:
run ci

PR #2932, opened 2017-06-28
[npm 5] Fix copping of local dependencies to the tns_modules folder.

TomaNikolov:
Npm 5 create a symlink of the local dependencies.
In order to copy these dependencies in the tns_modules folder,
we need to get there real path, not the symlink path.


rosen-vladimirov:
Closing this PR as the fix is not required - shelljs already handles this case correctly.

PR #2938, opened 2017-06-30
Error handling fixes.

TomaNikolov:
* The parameter of `failWithoutHelp` should be a string, not an Error object.
* Remove console.error().
* Move check for tgz.

If the path to the tgz is wrong we will try to get a latest compatible version, which will exec `npm view dist-tag --json`. However, npm view over tgz file has different behavior with npm 3, 4, and 5.
npm 3 will return a parse error.
npm 4 return 500 internal server error.
npm 5 try to find package.json in the current working directory and return information about this package.


dtopuzov:
run ci

PR #2944, opened 2017-07-03
Allow running postinstall scripts when adding a plugin

### LINKED ISSUE #2521: How to request user input during plugin installation?

EddyVerbruggen:
Fixes #2521 again.

### DIFF
--- a/lib/node-package-manager.ts
+++ b/lib/node-package-manager.ts
@@ -238,6 +238,8 @@ export class NodePackageManager implements INodePackageManager {
 					};
 
 					resolve(result);
+					// this makes sure postinstall scripts run after a 'tns plugin add' command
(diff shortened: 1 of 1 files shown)


dtopuzov:
run ci

rosen-vladimirov:
Hey @EddyVerbruggen ,
Please excuse me for the late reply. Thank you for this PR. The current fix will trigger double `npm install`, which is not exactly the desired effect.
However, I've found another way to handle this and I've created a PR, you can take a look at it [here](https://github.com/NativeScript/nativescript-cli/pull/2949).
I'll close this  PR, again thank you for your efforts.

We'll release the fix as soon as possible,

EddyVerbruggen:
You're the best, thanks a bunch for the swift response! 💯

PR #2948, opened 2017-07-05
Track exceptions in separate Analytics project

rosen-vladimirov:
Track all exceptions from CLI in a separate analytics project.
Also track the result of `tns error-reporting enable/disable` in the same project where the exceptions are tracked.

Update common-lib where the following changes had been applied:
Allow tracking exceptions in a different Analytics project. In case the specified CLI wants to track exceptions to a separeate project, it can do it by setting the new property in staticConfig - `ANALYTICS_EXCEPTIONS_API_KEY`.
Allow tracking the result of `tns error-reporting enable/disable` to the same project where exceptions tracking is sent.
Speed up check if current Analytics request is sent - we check on 1000ms, while in fact it takes between 150 and 400ms. So make the check on 100ms.


dtopuzov:
run ci

PR #2957, opened 2017-07-06
Debug on emulator by default when multiple devices/emulators attached

rosen-vladimirov:
When there are multiple devices/emulators, calling `tns debug android` will fail that multiple devices are attached.
This break VS Code as previous code of CLI has been using emulators by default.
Fix this by using the most recent version (highest API Level) of running emulators, i.e. in case you have three Android emulators with version 5.1, 6.0 and 7.0, the one with 7.0 will be used for debugging.


lib/constants.ts
```diff
@@ -87,3 +87,8 @@ export const CONNECTION_ERROR_EVENT_NAME = "connectionError";
 export const VERSION_STRING = "version";
 export const INSPECTOR_CACHE_DIRNAME = "ios-inspector";
 export const POST_INSTALL_COMMAND_NAME = "post-install-cli";
+
+export class DebugCommandErrors {
+	public static UNABLE_TO_USE_FOR_DEVICE_AND_EMULATOR = "The options --for-device and --emulator cannot be used simultaneously. Please use only one of them.";
+	public static NO_DEVICES_EMULATORS_FOUND_FOR_OPTIONS = "Unable to find device or emulator for specified options.";
+}
```
nadyaA: May be we can add all new errors here? For consistency, not a merge stopper, though.

lib/commands/debug.ts
```diff
@@ -70,6 +73,62 @@
...
+				} else {
+					if (emulators.length) {
+						selectedInstance = _.last(emulators);
+					} else {
+						selectedInstance = _.last(devices);
+					}
+				}
+
+				this.$logger.warn(`Multiple devices/emulators found. Starting debugger on ${selectedInstance.deviceInfo.identifier}.` +
+					`If you want to debug on specific device/emulator, you can specify it with --device option.`);
```
TomaNikolov: This is not a template strings so you can use "" instead of ``

lib/constants.ts
rosen-vladimirov: the only error remaining in debug command is the:
`Applications for platform ${this.platform} can not be built on this OS` - it is used in several places other than debug command, so it will be handled in a separate PR.
But I totally agree that we should have the errors in such constants.

rosen-vladimirov:
Implements https://github.com/NativeScript/nativescript-cli/issues/2966

Issue #2934, opened 2017-06-29
Running 'tns build android --release ...' uses *.debug.* files in build output


rowandh:
### Did you verify this is a real problem by searching [Stack Overflow](http://stackoverflow.com/questions/tagged/nativescript) and the [other open issues in this repo](https://github.com/NativeScript/nativescript/issues)?
Yes

### Tell us about the problem

I have a solution with config.debug.json and config.release.json. Running 'tns build android --release' copies config.debug.json -> config.json. I expected it to copy config.release.json -> config.json

### Which platform(s) does your issue occur on?
Only able to test on Android, but I'd imagine it occurs on both.

### Please provide the following version numbers that your issue occurs with:
- CLI: (run `tns --version` to fetch it)
3.1.1

- Cross-platform modules: (check the 'version' attribute in the
`node_modules/tns-core-modules/package.json` file in your project)
next (3.2.0-2017-06-28-7092)

- Runtime(s): (look for the `"tns-android"` and `"tns-ios"` properties in the
`package.json` file of your project)
    "tns-ios": {
      "version": "3.0.1"
    },
    "tns-android": {
      "version": "3.1.1"
    }

- Plugin(s): (look for the version number in the `package.json` file of your
project)
See https://github.com/rowandh/sample-Groceries/blob/debug-release-bug/package.json

### Please tell us how to recreate the issue in as much detail as possible.
* Clone https://github.com/rowandh/sample-Groceries/tree/debug-release-bug
* Run 'tns build android --release --key-store-path debug.keystore --key-store-password nativescript --key-store-alias debug --key-store-alias-password nativescript'
* Unzip the compiled apk. Open assets/app/config.json and verify that it says config: "debug"

My own compiled apk is in the 'apk' folder.

### Is there code involved? If so, please share the minimal amount of code needed to recreate the problem.
See https://github.com/rowandh/sample-Groceries/tree/debug-release-bug

rowandh:
Updated because I realised the original config.js files were ignored by the project's .gitignore.

fmsaraujo:
Can confirm this also happens for iOS builds.

nadyaA:
Hello @rowandh, @fmsaraujo,
The fix is merged and you can use it with `nativescript@next`. 

[...]

PR #2958, opened 2017-07-06
Respect configuration when preparing project files

nadyaA:
Implement processing configuration specific files on build, run and debug. There was some initial logic to process debug configuration by default, but when configuration is explicitly passed we did not respect it.  For example:
`tns build android --release`

Now if the passed argument is valid configuration we process specific files, taking the parameter into consideration.

https://github.com/NativeScript/nativescript-cli/issues/2934

common lib PR: https://github.com/telerik/mobile-cli-lib/pull/986

### DIFF
--- a/lib/services/platform-service.ts
+++ b/lib/services/platform-service.ts
@@ -294,10 +294,11 @@ export class PlatformService extends EventEmitter implements IPlatformService {
 		this.$logger.out("Preparing project...");
 
 		let platformData = this.$platformsData.getPlatformData(platform, projectData);
-		await this.preparePlatformCoreJS(platform, platformData, appFilesUpdaterOptions, projectData, platformSpecificData, changesInfo);
+		const projectFilesConfig = helpers.getProjectFilesConfig({ isReleaseBuild: appFilesUpdaterOptions.release });
+		await this.preparePlatformCoreJS(platform, platformData, appFilesUpdaterOptions, projectData, platformSpecificData, changesInfo, filesToSync, projectFilesConfig);
 
 		if (!nativePrepare || !nativePrepare.skipNativePrepare) {
-			await this.preparePlatformCoreNative(platform, platformData, appFilesUpdaterOptions, projectData, platformSpecificData, changesInfo);
+			await this.preparePlatformCoreNative(platform, platformData, appFilesUpdaterOptions, projectData, platformSpecificData, changesInfo, projectFilesConfig);
 		}
 
 		let directoryPath = path.join(platformData.appDestinationDirectoryPath, constants.APP_FOLDER_NAME);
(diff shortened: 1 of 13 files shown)


lib/commands/debug.ts
```diff
@@ -66,8 +68,10 @@
 			skipWatcher: !this.$options.watch || this.$options.justlaunch,
 			watchAllFiles: this.$options.syncAllFiles
 		};
-
-		await this.$debugLiveSyncService.liveSync(deviceDescriptors, liveSyncInfo);
+const projectFilesConfig: IProjectFilesConfig = {
```
rosen-vladimirov: incorrect indentation

lib/commands/run.ts
```diff
@@ -116,8 +116,10 @@ export class RunCommandBase implements ICommand {
 			watchAllFiles: this.$options.syncAllFiles,
 			clean: this.$options.clean
 		};
-
-		await this.$liveSyncService.liveSync(deviceDescriptors, liveSyncInfo);
+		const projectFilesConfig: IProjectFilesConfig = {
```
rosen-vladimirov: these lines are introduced on a lot of places, maybe you can introduce a method in [helpers](https://github.com/telerik/mobile-cli-lib/blob/master/helpers.ts)

lib/commands/run.ts
```diff
@@ -116,8 +116,10 @@ export class RunCommandBase implements ICommand {
...
+			configuration: this.$options.release ? Configurations.Release.toLowerCase() : Configurations.Debug.toLowerCase()
+		};
+		await this.$liveSyncService.liveSync(deviceDescriptors, liveSyncInfo, projectFilesConfig);
```
rosen-vladimirov: Do you need to pass `projectFilesConfig` to `liveSync` method - `liveSync` is called only for debug configuration (you can check some lines above that in case `--release` is passed, we'll never get here. Isn't the default behavior of file mappings (using the debug configuration) working fine in this case?

lib/definitions/livesync.d.ts
```diff
@@ -154,7 +154,7 @@ interface ILiveSyncService {
 	 * @param {ILiveSyncInfo} liveSyncData Describes the LiveSync operation - for which project directory is the operation and other settings.
 	 * @returns {Promise<void>}
 	 */
-	liveSync(deviceDescriptors: ILiveSyncDeviceInfo[], liveSyncData: ILiveSyncInfo): Promise<void>;
+	liveSync(deviceDescriptors: ILiveSyncDeviceInfo[], liveSyncData: ILiveSyncInfo, projectFilesConfig: IProjectFilesConfig): Promise<void>;
```
rosen-vladimirov: This method is exposed publicly when CLI is used as a library, is this really required?

lib/services/platform-service.ts
```diff
@@ -325,7 +329,11 @@ export class PlatformService extends EventEmitter implements IPlatformService {
 			excludedDirs.push(constants.TNS_MODULES_FOLDER_NAME);
 		}
 
-		this.$projectFilesManager.processPlatformSpecificFiles(directoryPath, platform, excludedDirs);
+		const projectFilesConfig: IProjectFilesConfig = {
```
rosen-vladimirov: this code is already available in this method - some lines above in the first `if`, can you declare it at a single place and use it in both methods where it is required.

dtopuzov:
run ci

lib/commands/run.ts
nadyaA: 🎣

lib/tools/node-modules/node-modules-builder.ts
```diff
@@ -43,6 +43,6 @@ export class NodeModulesBuilder implements INodeModulesBuilder {
 
 		return productionDependencies;
 	}
-}
+	}
```
rosen-vladimirov: Why is this changed?

PR #2959, opened 2017-07-07
Fix remove of non-production dependencies.

TomaNikolov:
If the scope dependency is not at the root of `node_modules` it will be deleted. We need to traverse the scope dependency directory and append all subdirectory to its name. Then we will get the real name of the dependency like is written in the `package.json` and we can use `productionDependencies.hasOwnProperty(dir)`


TomaNikolov:
run-ci

Issue #2936, opened 2017-06-30
CFBundleURLTypes cannot be overridden from a plugin


darind:
### Did you verify this is a real problem by searching [Stack Overflow]
No

### Tell us about the problem
I developed a custom {N} plugin and added a `platform/ios/Info.plist` file with the following contents:

```<xml>
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
<plist version="1.0">
<dict>
    <key>CFBundleURLTypes</key>
    <array>
        <dict>
            <key>CFBundleTypeRole</key>
            <string>Editor</string>
            <key>CFBundleURLName</key>
            <string>org.nativescript.foo</string>
            <key>CFBundleURLSchemes</key>
            <array>
                <string>bar</string>
            </array>
        </dict>
    </array>
    <key>NSAppTransportSecurity</key>
    <dict>
        <key>NSAllowsArbitraryLoads</key>
        <true/>
    </dict>
</dict>
</plist>
```
Then I installed the plugin into a newly created {N} app with:

```
tns create myapp
cd my app
tns plugin add /work/myplugin
tns build ios
```

The resulting application doesn't appear to take into account the `CFBundleURLTypes` setting from the plugin. On the other hand the `NSAppTransportSecurity` works fine.

### Which platform(s) does your issue occur on?

iOS

### Please provide the following version numbers that your issue occurs with:
- CLI: 3.1.1
- Cross-platform modules: 3.1.0
- Runtime(s): tns-android: 3.0.0 and tns-ios: 3.0.0
- Plugin(s): 

### Please tell us how to recreate the issue in as much detail as possible.

```
mkdir myplugin
cd myplugin/
mkdir platforms
mkdir platforms/ios

echo '{
    "name": "MyPlugin",
    "version": "1.0.0",
    "nativescript": {
        "platforms": {
            "android": "3.0.0",
            "ios": "3.0.0"
        }
    }
}' > package.json

echo '<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
<plist version="1.0">
<dict>
    <key>CFBundleURLTypes</key>
    <array>
        <dict>
[...]

PR #2961, opened 2017-07-07
Fix CFBundleURLTypes merging.

TomaNikolov:
Remove `updateCFBundleURLSchemes`  method since we fixed session.patch logic for merging  CFBundleURLTypes.

Fix [#2936](https://github.com/NativeScript/nativescript-cli/issues/2936)
* Note merge after new version of plist-merge-patch is released. Until this happened npm install will fail.

### DIFF
--- a/lib/services/ios-project-service.ts
+++ b/lib/services/ios-project-service.ts
@@ -6,7 +6,7 @@ import * as constants from "../constants";
 import * as helpers from "../common/helpers";
 import { attachAwaitDetach } from "../common/helpers";
 import * as projectServiceBaseLib from "./platform-project-service-base";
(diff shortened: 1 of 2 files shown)


TomaNikolov:
run-ci

lib/services/ios-project-service.ts
```diff
@@ -770,42 +766,36 @@ We will now place an empty obsolete compatability white screen LauncScreen.xib f
 			});
 		}
 
+		if (!buildOptions.release && projectData.projectId) {
+			session.patch({
+				name: "CFBundleURLTypes from package.json nativescript.id",
```
rosen-vladimirov: this name looks really strange

Issue #2975, opened 2017-07-13
CLI Process hangs on native build failure


Pip3r4o:
### Tell us about the problem
The CLI run/build process will not terminate (exit) if the gradle script throws an error, instead a warning message will be printed:

`Unable to apply changes on device: 192.168.50.101:5555. Error is: Command gradlew.bat failed with exit code 1.`

### Which platform(s) does your issue occur on?
Tested on Android, have not checked if its present for iOS builds

### Please provide the following version numbers that your issue occurs with:
- CLI: 3.1.2
- Runtime(s): 3.1

### Please tell us how to recreate the issue in as much detail as possible.
```shell
tns create testclicrash
cd testclicrash
tns plugin add nativescript-social-login
tns run android
```
Replace `tns plugin add nativescript-social-login` with any other possible method to crash the runtime build. The nativescript-social-login in particular crashes the build because it requires user configuration as explained in its respective README.

PR #2976, opened 2017-07-13
Fix recursive calling of actions.

TomaNikolov:
In case we are in the action chain lambda and some error occurs we don't need to wait for the same action in `stopLiveSync` method.

[#2975](https://github.com/NativeScript/nativescript-cli/issues/2975)


lib/definitions/livesync.d.ts
```diff
@@ -160,9 +160,10 @@ interface ILiveSyncService {
 	 * Stops LiveSync operation for specified directory.
 	 * @param {string} projectDir The directory for which to stop the operation.
 	 * @param {string[]} @optional deviceIdentifiers Device ids for which to stop the application. In case nothing is passed, LiveSync operation will be stopped for all devices.
+	 *  @param { shouldAwaitAllActions: boolean } @optional stopOptions Specifies whether we should await all actions.
```
rosen-vladimirov: incorrect space before `@param`

Issue #2470, opened 2017-02-01
Establish a recommended workflow for simultaneous Android & iOS development


tjvantoll:
I commonly develop on Android & iOS at the same time, and I believe a lot of our users do as well. Before the 2.5 release I ran `tns run android` and `tns run ios`, and then ran `tns livesync --watch` to watch both platforms simultaneously.

Since 2.5 that workflow no longer really works. The `livesync` command is now marked as deprecated, and executing two separate `tns run` commands is problematic.

With that in mind we need to establish a recommended workflow for cross-platform simultaneous development. Let’s list ideas in this issue; I’ll start with an idea I’m stealing from @bundyo.

What if we made `tns run` with no platform work. On macOS, `tns run` could build and deploy for iOS and Android, and watch each. On Windows and Linux `tns run` could build and deploy for just Android, and watch there as well.

Thoughts? Other ideas? Regardless of what we come up with I do think it’s important to have an established workflow for this common scenario for our next release.

Plamen5kov:
related issue: https://github.com/NativeScript/nativescript-cli/issues/2445

bundyo:
I don't think *tns run* is error prone - we can detect the connected devices and emulators and build&deploy to them. I actually expected that to work, as it is the default behavior in Cordova CLI.

Plamen5kov:
Cordova CLI is different and I don't think it's a good measure. We can make anything work, that's a fact. I prefer taking the safest rout and making the commands stable before anything else.

jlooper:
Just adding a vote for `tns run`. Right now if I start emulating iOS, and then start Genymotion to view the app on Android, the iOS emulator starts looping crazily and vice versa if I start on Android.

ejsuncy:
Looking forward to a resolution on this--I'm demoing NS 3.0.0-rc.1 to my dev team at work and was hoping to show simultaneous android & ios live reloading.

rosen-vladimirov:
Hey @jlooper , @tjvantoll , @ejsuncy ,
In the last weeks we've worked on implementing this feature and we are now ready to ship it. You can try it now with CLI's `next` version while we are preparing official release, where it will be included.
The idea is to use `$ tns run` command and the command will work with all running devices and emulators. In case there's not running devices/emulators, the command will fail.

Please give it a try and inform us in case you have any questions or concerns.

tjvantoll:
@rosen-vladimirov Just tested this out and it works great!

---

[...]

PR #2982, opened 2017-07-18
Fix run without platform on ios.

KristianDD:
Last part of implementing https://github.com/NativeScript/nativescript-cli/issues/2470

### DIFF
--- a/lib/commands/run.ts
+++ b/lib/commands/run.ts
@@ -33,7 +33,7 @@ export class RunCommandBase implements ICommand {
 			this.platform = this.$devicePlatformsConstants.Android;
 		}
 
-		const availablePlatforms = this.platform ? [this.platform] : _.values<string>(this.$platformsData.availablePlatforms);
+		const availablePlatforms = this.$liveSyncCommandHelper.getPlatformsForOperation(this.platform);
 		for (let platform of availablePlatforms) {
 			const platformData = this.$platformsData.getPlatformData(platform, this.$projectData);
 			const platformProjectService = platformData.platformProjectService;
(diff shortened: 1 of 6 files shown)


lib/commands/run.ts
```diff
@@ -32,9 +32,9 @@ export class RunCommandBase implements ICommand {
 			this.platform = this.$devicePlatformsConstants.Android;
 		}
 
-		const availablePlatforms = this.platform ? [this.platform] : this.$platformsData.availablePlatforms;
-		for (let platform of availablePlatforms) {
-			const platformData = this.$platformsData.getPlatformData(platform, this.$projectData);
+		const availablePlatforms = this.platform ? {[this.platform]: this.platform} : this.$platformsData.availablePlatforms;
+		for (var key in availablePlatforms) {
```
Mitko-Kerezov: We like to stick with the ES2015 scoped variables keywords.
Semantically it is more correct to use `const` here instead of `var` or `let`.

lib/commands/run.ts
```diff
@@ -32,9 +32,9 @@ export class RunCommandBase implements ICommand {
 			this.platform = this.$devicePlatformsConstants.Android;
 		}
 
-		const availablePlatforms = this.platform ? [this.platform] : this.$platformsData.availablePlatforms;
-		for (let platform of availablePlatforms) {
-			const platformData = this.$platformsData.getPlatformData(platform, this.$projectData);
+		const availablePlatforms = this.platform ? {[this.platform]: this.platform} : this.$platformsData.availablePlatforms;
```
Mitko-Kerezov: I'd like to suggest a different approach to the problem - see if it suits you:
Instead of changing the array `[this.platform]` into an object why not change the object into an array.
That way the code would look like this:
```TypeScript
const availablePlatforms = this.platform ? [this.platform] : _.values(this.$platformsData.availablePlatforms);

for (const platform of availablePlatforms) {
```

That way you can retain the `for ... of` loop and it would be a bit more readable than using an object.
@KristianDD what do you think?

KristianDD: The main reasoning for leaving it an object is that availablePlatforms is widely used inside the code base and the change will have bigger impact for little value.

lib/commands/run.ts
KristianDD: Good point...

PR #2983, opened 2017-07-19
 Separate preparation of JS from native code

rosen-vladimirov:
* Introduce new parameter prepareNative to indicate that native code preparation is explicitly required. 
* Cache additional data in .nsprepareinfo file:   
* nativePlatformStatus - indicates whether native platform should be added, prepared or there's no need to do so;   
* projectFileHash - saves package.json file's contents hash for the current platform. We use this for better heuristics for checking whether package.json is changed. 
* Re-factoring i.e. extract common code to methods

Fixes https://github.com/NativeScript/nativescript-cli/issues/2986

### DIFF
--- a/lib/bootstrap.ts
+++ b/lib/bootstrap.ts
@@ -105,6 +105,7 @@ $injector.require("devicePathProvider", "./device-path-provider");
 $injector.requireCommand("platform|clean", "./commands/platform-clean");
 
 $injector.requirePublicClass("liveSyncService", "./services/livesync/livesync-service");
+$injector.requirePublicClass("liveSyncCommandHelper", "./services/livesync/livesync-command-helper");
 $injector.require("debugLiveSyncService", "./services/livesync/debug-livesync-service");
 $injector.require("androidLiveSyncService", "./services/livesync/android-livesync-service");
 $injector.require("iOSLiveSyncService", "./services/livesync/ios-livesync-service");
--- a/lib/bootstrap.ts
+++ b/lib/bootstrap.ts
@@ -105,7 +105,7 @@ $injector.require("devicePathProvider", "./device-path-provider");
 $injector.requireCommand("platform|clean", "./commands/platform-clean");
 
 $injector.requirePublicClass("liveSyncService", "./services/livesync/livesync-service");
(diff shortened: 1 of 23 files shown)


lib/bootstrap.ts
```diff
@@ -105,6 +105,7 @@ $injector.require("devicePathProvider", "./device-path-provider");
 $injector.requireCommand("platform|clean", "./commands/platform-clean");
 
 $injector.requirePublicClass("liveSyncService", "./services/livesync/livesync-service");
+$injector.requirePublicClass("liveSyncCommandHelper", "./services/livesync/livesync-command-helper");
```
rosen-vladimirov: this should be `require`

lib/definitions/project.d.ts
```diff
@@ -132,6 +132,10 @@ interface IBuildForDevice {
 	buildForDevice: boolean;
 }
 
+interface INativePrepare {
+	skip: boolean;
```
rosen-vladimirov: we need better name for this - skipNativePrepare for example

lib/definitions/project-changes.d.ts
```diff
@@ -22,12 +22,19 @@ interface IProjectChangesInfo {
 	readonly changesRequirePrepare: boolean;
 }
 
-interface IProjectChangesOptions extends IAppFilesUpdaterOptions, IProvision {}
+interface IProjectChangesOptions extends IAppFilesUpdaterOptions, IProvision {
+	nativePlatformStatus?: "1" | "2" | "3";
```
Mitko-Kerezov: We could add a comment above this stating the enumeration's values
```TypeScript
export const enum NativePlatformStatus {
	requiresPlatformAdd = "1",
	requiresPrepare = "2",
	alreadyPrepared = "3"
}
```

test/stubs.ts
```diff
@@ -574,6 +574,10 @@ export class ProjectChangesService implements IProjectChangesService {
 	public get currentChanges(): IProjectChangesInfo {
 		return <IProjectChangesInfo>{};
 	}
+
+	public setNativePlatformStatus(platform: string, projectData: IProjectData, nativePlatformStatus: IAddedNativePlatform): boolean {
```
rosen-vladimirov: this method should be void

PR #2990, opened 2017-07-20
Move skipNativePrepare to deviceDescriptors

rosen-vladimirov:
Each device can use different build types - cloud or local. So the `skipNativePrepare` option should be passed per each device, not for the whole livesync operation.


rosen-vladimirov:
run ci

PR #3002, opened 2017-07-25
Fix rebuilding application every time when a change is applied

rosen-vladimirov:
In case you have built your application and you add a new npm package as dependency, the LiveSync will always rebuild the app, no matter of the change applied afterwards. The problem is in our project-changes-service that persists shasum of the package.json. However it is never reset, so we always have the hashed shasum of the package.json when the applcation has been build for the first time. Fix this by setting the correct value for projectFileHash - this way the checks are correct and applying change in .js (.xml, .css) files will not trigger rebuild of app.


davecoffin:
On iOS, my application has started building every time I save a file. I am running all the latest versions. Is this a bug in nativescript cli? Is there a release coming?

PR #3008, opened 2017-07-26
move device discovery to debug command as it is done for run command

Plamen5kov:
* minor cli refactoring (keeping method API consistent between run and debug)
* method comments for a better world

### DIFF
--- a/lib/services/livesync/livesync-command-helper.ts
+++ b/lib/services/livesync/livesync-command-helper.ts
@@ -15,12 +15,16 @@ export class LiveSyncCommandHelper implements ILiveSyncCommandHelper {
 		return availablePlatforms;
 	}
 
+	/**
+	 * Method sets up configuration, before calling livesync and expects that devices are already discovered.
+	 * @param devices List of discovered devices
+	 * @param liveSyncService Service expected to do the actual livesyncing
+	 * @param platform The platform for which the livesync will be ran
+	 */
 	public async executeLiveSyncOperation(devices: Mobile.IDevice[], liveSyncService: ILiveSyncService, platform: string): Promise<void> {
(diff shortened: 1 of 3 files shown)


lib/services/livesync/livesync-command-helper.ts
```diff
@@ -15,13 +15,17 @@ export class LiveSyncCommandHelper implements ILiveSyncCommandHelper {
 		return availablePlatforms;
 	}
 
+	/**
+	 * Method sets up configuration, before calling livesync and expects that devices are already discovered.
+	 * @param devices List of discovered devices
+	 * @param liveSyncService Service expected to do the actual livesyncing
+	 * @param platform The platform for which the livesync will be ran
+	 */
```
rosen-vladimirov: The JSDocs are more applicable for the interface (`ILiveSyncCommandHelper`) as we are calling the method through it. In order to have autocomplete and JSDocs when using the injected dependency ("liveSyncCommandHelper"), the JSDocs must be in the interface itself.

Also you have to define the return type of the method:
```
* @returns {Promise<void>}
```

lib/services/livesync/livesync-command-helper.ts
```diff
@@ -15,13 +15,17 @@ export class LiveSyncCommandHelper implements ILiveSyncCommandHelper {
 		return availablePlatforms;
 	}
 
+	/**
+	 * Method sets up configuration, before calling livesync and expects that devices are already discovered.
+	 * @param devices List of discovered devices
```
rosen-vladimirov: you have to define the type of the parameter:
https://github.com/Microsoft/TypeScript/wiki/JsDoc-support-in-JavaScript

PR #3011, opened 2017-07-27
Emit liveSyncStopped event when LiveSync operation is really stopped

rosen-vladimirov:
In case `stopLiveSync` is called during livesync, we should await the current action and emit `liveSyncStopped` event. At the moment, we have the following behavior:
1. In case stopLiveSync is called with deviceIdentifiers, we emit `liveSyncStopped` immediately for all of them. However there may be pending actions, for example build + deploy for current device. We cannot stop the actions, so you receive `liveSyncStopped` and application is installed/updated on device after that.
Fix this by persisting the current action and await it before emitting `liveSyncStopped` event.
2. In case we are currently rebuilding the application and some changes are applied during build, we'll update the `actionsChain`. At this point we'll have several actions in the chain. In case stopLiveSync method is called (without deviceIdentifiers), we will await all pending actions, which will result in awaiting the build, deploy and all other actions and we'll emit the `liveSyncStopped` after that. However we should not execute all pending actions - we should just execute the current one and skip the rest.
In order to fix this, just set isStopped to true before awaiting the actions. This way only the current action will be executed.

### DIFF
--- a/lib/services/livesync/livesync-service.ts
+++ b/lib/services/livesync/livesync-service.ts
@@ -44,15 +44,15 @@ export class LiveSyncService extends EventEmitter implements ILiveSyncService {
 		const liveSyncProcessInfo = this.liveSyncProcessesInfo[projectDir];
 
 		if (liveSyncProcessInfo) {
-			_.each(deviceIdentifiers, deviceId => {
-				_.remove(liveSyncProcessInfo.deviceDescriptors, descriptor => {
-					const shouldRemove = descriptor.identifier === deviceId;
-					if (shouldRemove) {
-						this.emit(LiveSyncEvents.liveSyncStopped, { projectDir, deviceIdentifier: descriptor.identifier });
-					}
+			// In case we are coming from error during livesync, the current action is the one that erred (but we are still executing it),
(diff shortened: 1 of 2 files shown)


lib/services/livesync/livesync-service.ts
```diff
@@ -44,15 +44,15 @@ export class LiveSyncService extends EventEmitter implements ILiveSyncService {
 		const liveSyncProcessInfo = this.liveSyncProcessesInfo[projectDir];
 
 		if (liveSyncProcessInfo) {
-			_.each(deviceIdentifiers, deviceId => {
-				_.remove(liveSyncProcessInfo.deviceDescriptors, descriptor => {
-					const shouldRemove = descriptor.identifier === deviceId;
-					if (shouldRemove) {
-						this.emit(LiveSyncEvents.liveSyncStopped, { projectDir, deviceIdentifier: descriptor.identifier });
-					}
+			// In case we are coming from error during livesync, the current action is the one that errored (but we are still executing it),
```
Mitko-Kerezov: `the one that errored` should be `the one that erred`

PR #3013, opened 2017-07-28
"This" is not the correct property invoker in current context

nadyaA:
Options are passed as function argument and should not be called with this.


nadyaA:
run ci

Issue #2657, opened 2017-03-26
`tns run android/ios` does not remove folders correctly


dtopuzov:
### Which platform(s) does your issue occur on?
Android and iOS

### Please provide the following version numbers that your issue occurs with:
- CLI: 3.0.0-2017.3.24.5

### Please tell us how to recreate the issue in as much detail as possible.
```
tns create MyApp
code MyApp (open the app in VSCode)
tns run android --path MyApp
Create new folder, add some js file -> File is there.
Delete the folder -> file is removed, but folder is available on still on device
```

Notes:

How to files on device: `adb shell run-as org.nativescript.MyApp ls /data/data/org.nativescript.MyApp/files/app/`

I'm not able to break livesync or crash the app.
It is just some garbage that we do not clean.

Plamen5kov:
related: https://github.com/NativeScript/nativescript-cli/issues/2890#issuecomment-311017175

dtopuzov:
Sill valid.
Tested with Api23 default emulator on macOS -> no issue.
Tested with the same Api23 default emulator on Linux and Windows -> The issue still exists.

miroslavaivanova:
This issue is fixed.

Issue #2485, opened 2017-02-03
Nativescript 2.5.0 can't synchronize changes into device, ENOENT and EACCES errors


NickIliev:
_From @S4M37 on February 3, 2017 14:27_

I'm using NativeScript 2.5.0 with Angular2, using android platform with building tool 25.0.2, I'm testing on a device, and `tns doctor` no issues are found. So my problem is when running 'tns run android' the app is launched but no changes are applied, here is the log

> Project successfully prepared (android)
Transferring project files...
Unable to sync files. Error is: ENOENT: no such file or directory, stat 'C:\Users\S4M37\Desktop\CrossPlatform\project\platforms\android\src\main\assets\app\modules\feed\item\feed.item.css___jb_tm
p___'
02-03 15:16:18.512  5693  5693 W System.err: remove failed: EACCES (Permission denied) : /data/local/tmp/org.nativescript.project/sync/tns_modules/rxjs/src/MiscJSDoc.ts
02-03 15:16:18.532  5335  5335 E audit   : type=1400 msg=audit(1486131378.512:526786): avc:  denied  { unlink } for  pid=5693 comm="tivescript.project" name="MiscJSDoc.ts" dev="dm-1" ino=335142 s
context=u:r:untrusted_app:s0:c512,c768 tcontext=u:object_r:shell_data_file:s0 tclass=file permissive=0
02-03 15:16:18.562  5693  5693 W System.err: remove failed: EACCES (Permission denied) : /data/local/tmp/org.nativescript.project/sync/tns_modules/rxjs/src/observable/dom/MiscJSDoc.ts
02-03 15:16:18.582  5335  5335 E audit   : type=1400 msg=audit(1486131378.562:526982): avc:  denied  { unlink } for  pid=5693 comm="tivescript.project" name="MiscJSDoc.ts" dev="dm-1" ino=335340 s
context=u:r:untrusted_app:s0:c512,c768 tcontext=u:object_r:shell_data_file:s0 tclass=file permissive=0
JS: Angular 2 is running in the development mode. Call enableProdMode() to enable the production mode.
02-03 15:16:22.322  5693  5693 W ExifInterface:         at com.tns.Runtime.callJSMethodNative(Native Method)

_Copied from original issue: NativeScript/NativeScript#3569_

danieljohngomez:
Also experiencing this error after updating to 2.5

alpinparker:
[...]

PR #3027, opened 2017-08-02
livesync deletes files and cleans temp folder afterwards

Plamen5kov:
_problem_
Related issue: https://github.com/NativeScript/nativescript-cli/issues/2657
Possibly related to https://github.com/NativeScript/nativescript-cli/issues/2485
* emulators with API level >= 23 the installed application doesn't have the permissions to write in `/data/local/tmp/<application_name>/` and disrupts livesync in some scenarios
* on devices the problem with permissions is even worse and even on API level 22, the application still doesn't have permissions to write in `/data/local/tmp/<application_name>/`

_solution_
The CLI will take care of notifying the runtime and deleting all files after livesync is done, regardless of whether the application needs to be restarted or not.

Edit: tested on iOS and Android: new/delete/modify work, but renaming only works on android, I'll fix the problem on iOS in a different PR. We don't have tests for renaming a file, so it won't affect the tests.

### DIFF
--- a/lib/services/livesync/android-device-livesync-service.ts
+++ b/lib/services/livesync/android-device-livesync-service.ts
@@ -36,14 +36,25 @@ export class AndroidDeviceLiveSyncService extends DeviceLiveSyncServiceBase impl
 				`${deviceProjectRootDirname}/sync`]
 		);
 
+		this.reloadResources(deviceAppData, localToDevicePaths);
+
 		const canExecuteFastSync = !liveSyncInfo.isFullSync && !_.some(localToDevicePaths,
 			(localToDevicePath: Mobile.ILocalToDevicePathData) => !this.canExecuteFastSync(localToDevicePath.getLocalPath(), projectData, this.device.deviceInfo.platform));
 
-		if (canExecuteFastSync) {
-			return this.reloadPage(deviceAppData, localToDevicePaths);
+		if (!canExecuteFastSync) {
+			return this.restartApplication(deviceAppData);
 		}
+	}
 
-		return this.restartApplication(deviceAppData);
+	private async cleanLivesyncDirectories(deviceAppData: Mobile.IDeviceAppData): Promise<void> {
+		const deviceRootPath = await this.$devicePathProvider.getDeviceProjectRootPath(deviceAppData.device, {
+			appIdentifier: deviceAppData.appIdentifier,
+			getDirname: true
+		});
+
+		await this.device.adb.executeShellCommand(["rm", "-rf", await this.$mobileHelper.buildDevicePath(deviceRootPath, LiveSyncPaths.FULLSYNC_DIR_NAME),
+			this.$mobileHelper.buildDevicePath(deviceRootPath, LiveSyncPaths.SYNC_DIR_NAME),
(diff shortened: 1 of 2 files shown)


Plamen5kov:
run CI

Plamen5kov:
run CI

Plamen5kov:
run CI

rosen-vladimirov:
The logic until this PR was:
0. Transfer files.
1. Check if we can execute fast sync.
1.1 In case yes - try sending page reload message
1.1.1 In case page reload succeeds - finish
1.1.2 in case page reload fails - restart the app
1.2 In case we cannot execute fast sync, the application is restarted.

After this PR, the logic is:
0. Transfer files
1. Send page reload message. 
1.1 In case page reload returns true - clean resources
1.2 In case page reload returns false - restart application.
2. Check if we can execute fast sync
2.1 In case not - restart the application.

So what will happen in case `.css` and `.js` files are changed, but we are unable to send page reload message - the application will be restarted twice.

In the same scenario - sync `.css` and `.js` - the page reload message succeeds, the view is refreshed on device (change is visible) and after that application is restarted.

@Plamen5kov  are those two behaviours expected and correct?

Plamen5kov:
In case we're unable to send "page reload" message, I didn't see what the idea behind restarting the application was in the first place because all we would do is give the runtime a second chance to load the changes, but I kept it for consistency with previous behavior. IMO the app should crash if the runtime socket returns an error or closes prematurely.
The second scenario is expected.

rosen-vladimirov:
@Plamen5kov I believe the restart in case page reload fails is for cases where the application is not running on the device. For example the user may stop it manually during livesync and then to change `.xml` file.
The other case is when the user had made some kind of a mistake in the code that causes app crash, application is not running, user changes something in `.xml` , page reload fails, so we restart the application.

The second scenario is a change from the previous behavior, but I think we can live with it.
Ping @dtopuzov just to be informed for the current discussion.

Plamen5kov:
@rosen-vladimirov in the case where the user stops the application in the middle of a livesync I see a couple of scenarios:

* stops the application _before_ CLI manages to sync file on device - in this case we're fine because user won't see `Successfully transferred <file_name>` and will know the sync couldn't finish, and he'll know the file is not synced (expected)
* stops the application _after_ CLI manages to sync file on the device and:
  *  _before_ runtime is notified of change - on next run of `tns run <platform>` or when user starts the application manually, runtime will take the files and the app will work with updated files (expected)
  * _after_ runtime is notified of change:
    * before or during file transfers (undefined behavior, unexpected)
    * after file transfer - like before, everything is ok if `tns run <platform>` is ran or application is started manually
   * _after_ runtime is notified of the change and _before_ CLI cleans files from the respective temp folder - (undefined behavior) possible solutions: run `tns run..` again so we clean the temp folders in `beforeLivesync` and go on with workflow.
  * _after_ runtime is notified of change and_after_ CLI cleans the files from the respective temp folder - we're good because all operations of livesync have passed

We can discuss the cases where there's unexpected behavior if livesync is stopped and come up with a statemet like: "we work in these cases and we don't work or have undefined behavior in the following cases" and possible workarounds in the docs.

PR #3032, opened 2017-08-03
Stop copying android plugin code, stop writing configurations dir

Pip3r4o:
**Problem**:
The CLI will execute gradle clean when a plugin with a native counterpart is added/removed, that will force removal of all build artifacts, including the infamous `configurations` directory, where the CLI copies all plugin `include.gradle` scripts. The user may call `gradle clean` on their own however, that will remove `configurations` and the CLI will not regenerate it on a consecutive build, because it would not have detected a change in the plugins in order to copy the gradle scripts over. As a result builds that follow will not include any of the android plugins.

**Proposed Solution**:
Hand over the control over android plugins to the Gradle script, it will generate its own configurations and copy files where necessary as it sees fit. NativeScript CLI's only responsibility towards android plugins will now be to include them in a `dependencies.json` file which is written inside the `platforms/android` directory of a NS project.

**Implementation implications**:
- The plugin variables interpolation logic is removed, as the CLI will no longer take care of copying android plugin files inside platforms/android. There are slightly more elegant ways to handle this. For one - plugin variables could be moved inside the `ns-project/app/package.json` directory either by hand or automatically. The `package.json` will be available on the target device and is readable by plugins.
- Versions of the android runtime prior to 3.2 (or whenever we end up releasing this change) will not be compatible with previous versions of the CLI, and vice-versa. An error message in the android build script will notify the user of the version discrepancies should they occur.

Related issues/PRs: 
 - android runtime - https://github.com/NativeScript/android-runtime/issues/790
 - meta QoL improvements issue - https://github.com/NativeScript/NativeScript/issues/4459

---

CC: @PanayotCankov @Plamen5kov

### DIFF
--- a/lib/services/android-project-service.ts
+++ b/lib/services/android-project-service.ts
@@ -12,7 +12,6 @@ import { SpawnOptions } from "child_process";
 export class AndroidProjectService extends projectServiceBaseLib.PlatformProjectServiceBase implements IPlatformProjectService {
 	private static VALUES_DIRNAME = "values";
 	private static VALUES_VERSION_DIRNAME_PREFIX = AndroidProjectService.VALUES_DIRNAME + "-v";
-	private static ANDROID_PLATFORM_NAME = "android";
 	private static MIN_RUNTIME_VERSION_WITH_GRADLE = "1.5.0";
 	private static REQUIRED_DEV_DEPENDENCIES = [
 		{ name: "babel-traverse", version: "^6.4.5" },
@@ -33,7 +32,6 @@ export class AndroidProjectService extends projectServiceBaseLib.PlatformProject
 		$projectDataService: IProjectDataService,
 		private $sysInfo: ISysInfo,
 		private $injector: IInjector,
-		private $pluginVariablesService: IPluginVariablesService,
 		private $devicePlatformsConstants: Mobile.IDevicePlatformsConstants,
 		private $npm: INodePackageManager) {
 		super($fs, $projectDataService);
@@ -348,42 +346,11 @@ export class AndroidProjectService extends projectServiceBaseLib.PlatformProject
 	}
 
 	public async preparePluginNativeCode(pluginData: IPluginData, projectData: IProjectData): Promise<void> {
-		const pluginPlatformsFolderPath = this.getPluginPlatformsFolderPath(pluginData, AndroidProjectService.ANDROID_PLATFORM_NAME);
-		await this.processResourcesFromPlugin(pluginData, pluginPlatformsFolderPath, projectData);
+		// Add to platforms/android/dependencies.json
 	}
 
 	public async processConfigurationFilesFromAppResources(): Promise<void> {
 		return;
-	}
-
-	private async processResourcesFromPlugin(pluginData: IPluginData, pluginPlatformsFolderPath: string, projectData: IProjectData): Promise<void> {
-		const configurationsDirectoryPath = path.join(this.getPlatformData(projectData).projectRoot, "configurations");
-		this.$fs.ensureDirectoryExists(configurationsDirectoryPath);
-
-		const pluginConfigurationDirectoryPath = path.join(configurationsDirectoryPath, pluginData.name);
-		if (this.$fs.exists(pluginPlatformsFolderPath)) {
-			this.$fs.ensureDirectoryExists(pluginConfigurationDirectoryPath);
-
-			const isScoped = pluginData.name.indexOf("@") === 0;
-			const flattenedDependencyName = isScoped ? pluginData.name.replace("/", "_") : pluginData.name;
-
-			// Copy all resources from plugin
-			const resourcesDestinationDirectoryPath = path.join(this.getPlatformData(projectData).projectRoot, "src", flattenedDependencyName);
-			this.$fs.ensureDirectoryExists(resourcesDestinationDirectoryPath);
-			shell.cp("-Rf", path.join(pluginPlatformsFolderPath, "*"), resourcesDestinationDirectoryPath);
-
-			const filesForInterpolation = this.$fs.enumerateFilesInDirectorySync(resourcesDestinationDirectoryPath, file => this.$fs.getFsStats(file).isDirectory() || path.extname(file) === constants.XML_FILE_EXTENSION) || [];
-			for (const file of filesForInterpolation) {
-				this.$logger.trace(`Interpolate data for plugin file: ${file}`);
-				await this.$pluginVariablesService.interpolate(pluginData, file, projectData);
-			}
-		}
-
-		// Copy include.gradle file
-		const includeGradleFilePath = path.join(pluginPlatformsFolderPath, "include.gradle");
-		if (this.$fs.exists(includeGradleFilePath)) {
-			shell.cp("-f", includeGradleFilePath, pluginConfigurationDirectoryPath);
-		}
 	}
 
 	public async removePluginNativeCode(pluginData: IPluginData, projectData: IProjectData): Promise<void> {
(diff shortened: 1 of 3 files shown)


lib/tools/node-modules/node-modules-dependencies-builder.ts
```diff
@@ -47,7 +47,18 @@ export class NodeModulesDependenciesBuilder implements INodeModulesDependenciesB
...
+	}
+
+	private filterUniqueDependencies(dependencies: IDependencyData[]): IDependencyData[] {
+		const depsDictionary = dependencies.reduce((dict, dep) => {
+			const collision = dict[dep.name];
+			if (!collision || collision.depth > dep.depth) {
+				dict[dep.name] = dep;
+			}
+			return dict;
+		}, <{ [key: string]: IDependencyData}>{});
```
rosen-vladimirov: the type is basically  `IDictionary<IDependencyData>`

lib/tools/node-modules/node-modules-dependencies-builder.ts
```diff
@@ -47,7 +47,18 @@ export class NodeModulesDependenciesBuilder implements INodeModulesDependenciesB
...
 		}
 
-		return resolvedDependencies;
+		return this.filterUniqueDependencies(resolvedDependencies);
+	}
+
+	private filterUniqueDependencies(dependencies: IDependencyData[]): IDependencyData[] {
+		const depsDictionary = dependencies.reduce((dict, dep) => {
+			const collision = dict[dep.name];
+			if (!collision || collision.depth > dep.depth) {
```
rosen-vladimirov: I'm not sure I understand what's the idea here, can you please add comment in the code

dtopuzov:
run ci

dtopuzov:
run ci

dtopuzov:
run ci

dtopuzov:
run ci

Pip3r4o:
run ci

Pip3r4o:
@rosen-vladimirov Changes have been made in regards to the backwards compatibility concerns that we had. Please review.

lib/services/android-project-service.ts
```diff
@@ -79,6 +79,26 @@ export class AndroidProjectService extends projectServiceBaseLib.PlatformProject
 		return this._platformData;
 	}
 
+	// TODO: Remove prior to the 4.0 CLI release @Pip3r4o @PanayotCankov
+	// Similar to the private method of the same name in platform-service.
+	private getCurrentPlatformVersion(platformData: IPlatformData, projectData: IProjectData): string {
+		const currentPlatformData: any = this.$projectDataService.getNSValue(projectData.projectDir, platformData.frameworkPackageName);
+		let version: string;
+		if (currentPlatformData && currentPlatformData[constants.VERSION_STRING]) {
```
rosen-vladimirov: you can directly return this value:
```TypeScript
return currentPlatformData && currentPlatformData[constants.VERSION_STRING];
```

lib/services/android-project-service.ts
```diff
@@ -79,6 +79,26 @@ export class AndroidProjectService extends projectServiceBaseLib.PlatformProject
...
+			version = currentPlatformData[constants.VERSION_STRING];
+		}
+
+		return version;
+	}
+
+	// TODO: Remove prior to the 4.0 CLI release @Pip3r4o @PanayotCankov
+	private shouldUseNewRuntimeGradleRoutine(projectData: IProjectData): boolean {
+		const platformVersion = this.getCurrentPlatformVersion(this._platformData, projectData);
+		const NEW_RUNTIME_GRADLE_ROUTINE_VERSION = "3.3.0";
```
rosen-vladimirov: the name of the variables inside methods should be in camel case

lib/services/android-project-service.ts
```diff
@@ -348,8 +368,12 @@ export class AndroidProjectService extends projectServiceBaseLib.PlatformProject
...
 
 	public async preparePluginNativeCode(pluginData: IPluginData, projectData: IProjectData): Promise<void> {
-		const pluginPlatformsFolderPath = this.getPluginPlatformsFolderPath(pluginData, AndroidProjectService.ANDROID_PLATFORM_NAME);
-		await this.processResourcesFromPlugin(pluginData, pluginPlatformsFolderPath, projectData);
+		if (!this.shouldUseNewRuntimeGradleRoutine(projectData)) {
+			const pluginPlatformsFolderPath = this.getPluginPlatformsFolderPath(pluginData, AndroidProjectService.ANDROID_PLATFORM_NAME);
+			await this.processResourcesFromPlugin(pluginData, pluginPlatformsFolderPath, projectData);
+		}
+
+		// Do nothing, the Android Gradle script will configure itself based on the input dependencies.json
```
rosen-vladimirov: this comment looks strange - in case we enter the `if`, we'll still get here so it leaves me wondering what does `do nothing` mean

lib/services/android-project-service.ts
```diff
@@ -407,23 +435,61 @@ export class AndroidProjectService extends projectServiceBaseLib.PlatformProject
...
+	private filterUniqueDependencies(dependencies: IDependencyData[]): IDependencyData[] {
+		const depsDictionary = dependencies.reduce((dict, dep) => {
+			const collision = dict[dep.name];
+			// in case there are multiple dependencies to the same module, the one declared in the package.json takes precedence
+			if (!collision || collision.depth > dep.depth) {
+				dict[dep.name] = dep;
+			}
+			return dict;
+		}, <IDictionary<IDependencyData>>{});
+		return Object.keys(depsDictionary).map(key => depsDictionary[key]);
```
rosen-vladimirov: this seems like you are trying to return array of the values in the dictionary. If that's the case, you can just use:
```TypeScript
_.values(depsDictionary)
```

test/debug.ts
```diff
@@ -333,24 +333,4 @@ describe("debug command tests", () => {
...
 	});
-
-	describe("Debugger tests", () => {
-		let testInjector: IInjector;
-
-		beforeEach(() => {
-			testInjector = createTestInjector();
-		});
-
-		it("Ensures that beforePrepareAllPlugins will call gradle with clean option when *NOT* livesyncing", async () => {
```
rosen-vladimirov: I believe we should keep this test - we still have a case where we want to call gradle clean (using the old runtime logic).

lib/services/android-project-service.ts
```diff
@@ -79,6 +79,26 @@ export class AndroidProjectService extends projectServiceBaseLib.PlatformProject
 		return this._platformData;
 	}
 
+	// TODO: Remove prior to the 4.0 CLI release @Pip3r4o @PanayotCankov
+	// Similar to the private method of the same name in platform-service.
+	private getCurrentPlatformVersion(platformData: IPlatformData, projectData: IProjectData): string {
+		const currentPlatformData: any = this.$projectDataService.getNSValue(projectData.projectDir, platformData.frameworkPackageName);
```
rosen-vladimirov: the type should be `IDictionary<any>` maybe

test/debug.ts
Pip3r4o: Yes. you are right, do you think there is a straightforward way to disable it/always assert true when a more recent runtime is used? Or should I use the same logic as in - https://github.com/NativeScript/nativescript-cli/pull/3032/files/f4a47a233f86e8d499825247342ce98b1404d525#diff-b64860922c31aeed973aef6663b2a536R95? Thank you!

lib/services/android-project-service.ts
Pip3r4o: The method wouldn't perform any work after the changes in the android build script, and would be completely empty after 4.0. I thought that maybe the comment would be relevant

test/debug.ts
rosen-vladimirov: I would modify the test and make it always use runtime version 3.2.0. This way you'll not have to handle the new scenario, just ensure we call clean.

lib/services/android-project-service.ts
rosen-vladimirov: Ahh, I've missed we have this method in the interface and it is implemented for iOS. Ignore my comment :)

Pip3r4o:
@rosen-vladimirov I believe I addressed your comments, please review the test, and whether it's done properly - https://github.com/NativeScript/nativescript-cli/pull/3032/files#diff-0f742b54cfee0b7c871e095194f60c84.

Thanks!

lib/services/android-project-service.ts
```diff
@@ -79,6 +79,22 @@ export class AndroidProjectService extends projectServiceBaseLib.PlatformProject
...
+	// TODO: Remove prior to the 4.0 CLI release @Pip3r4o @PanayotCankov
+	// Similar to the private method of the same name in platform-service.
+	private getCurrentPlatformVersion(platformData: IPlatformData, projectData: IProjectData): string {
+		const currentPlatformData: IDictionary<any> = this.$projectDataService.getNSValue(projectData.projectDir, platformData.frameworkPackageName);
+
+		return currentPlatformData && currentPlatformData[constants.VERSION_STRING];
+	}
+
+	// TODO: Remove prior to the 4.0 CLI release @Pip3r4o @PanayotCankov
+	private shouldUseNewRuntimeGradleRoutine(projectData: IProjectData): boolean {
```
rosen-vladimirov: The private methods should be below the public/protected ones

PR #3036, opened 2017-08-04
Fix IPA export

### LINKED ISSUE #3020: iOS Archive Export unexpected behavior when using AdHoc or AppStore provisioning
From Xcode 8.3+ the way IPA is built changed.
The generation of the IPA is a two step process now:
 1. `xcodebuild archive`
 2. `xcodebuild export`

#### Archive
This makes an archive folder that has app, debug symbols, etc. It has all the artefacts necessary to export for Development, Ad-Hoc or App Store distribution. The app is signed with the certificate and provisioning profiles set in the .xcconfig or in the xcodeproj (using the --provision option).

#### Export
This takes the archive and exports IPA for Development, Ad-Hoc or AppStore. During the export, the app is often resigned with a distribution certificate and an ad-hoc or AppStore distribution provisioning profile.

## Problems
Currently the CLI does archive then export for development when `tns build|run` is issued. This leads to underwhelming results when developers try to publish at the store or export for ad-hoc using `tns build`. Why? During the export session the xcodebuild command accepts an `-exportOptionsPlist` that can contain keys listed in [`xcodebuild --help`](https://gist.github.com/PanayotCankov/3c81cf106c8b7bb5efa3ad05d9731abe). Most importantly the: `method : String`:

> Describes how Xcode should export the archive. Available options: app-store, package, ad-hoc, enterprise, development, and developer-id. The list of options varies based on the type of archive. Defaults to development.

The default is "development", for exportOptionsPlist currently the app/App_Resources/Info.plist is currently provided (IMO incorrectly) and it has no `method` key, so `development` is used as fallback. When the developers configure the .xcconfig to use Ad-Hoc or AppStore provisioning profile, the export is still performed as development, the result IPA is then incorrectly set to still use development settings such as entitlements.

Here are the related issues to exactly this problem:
 - https://github.com/NativeScript/nativescript-cli/issues/2888
 - https://github.com/NativeScript/nativescript-cli/issues/2999 
    (perhaps incorrectly?) tries to provide distribution certificate for archive, while archive should be performed with development certificate and export should sign with distribution certificate
 - https://github.com/NativeScript/nativescript-cli/issues/2922
    Signes multitarget app with the AppStore distribution, there is a bug in old Ruby and Cocoapods that leads to this
 - https://github.com/NativeScript/nativescript-cli/issues/2888
    Building with --release as mentioned still exports with method: development
 - https://github.com/NativeScript/nativescript-cli/issues/2716
 - https://github.com/NativeScript/nativescript-cli/issues/2702
    Similar to 2922 - resolved by ruby and cocoapods update but also the missing exportOptionsPlist is noted in warnings

We have discussed internally that adding `tns archive` + `tns export` would diverge too much from the current and Android build commands, also people shouldn't care how the build is implemented internally.

Here are some clues how this can be addressed:
The export during `tns build ios` is done here:
https://github.com/NativeScript/nativescript-cli/blob/1955028ac3ded6a012813476f08d0f4f0d89dddd/lib/services/ios-project-service.ts#L259-L263
It incorrectly gets the app/App_Resources/Info.plist as exportOptionsPlist. Instead such export.plist files should be generated somewhere in the platforms/ios/build similar to the exportArchive that is currently used only for distribution:
https://github.com/NativeScript/nativescript-cli/blob/1955028ac3ded6a012813476f08d0f4f0d89dddd/lib/services/ios-project-service.ts#L212-L229
The method - Ad-Hoc or AppStore distribution that should be used during `tns build ios` should either be provided through an option, or inferred from the provisioning profile used during the build.

PeterStaev:
Fixes #3020 and related issues by reading the `embedded.mobileprovision` from the built `.app`, detecting the type of the provision that is used and then sending the correct export method when exporting the `.ipa`

### DIFF
--- a/lib/services/ios-project-service.ts
+++ b/lib/services/ios-project-service.ts
@@ -251,6 +251,20 @@ export class IOSProjectService extends projectServiceBaseLib.PlatformProjectServ
 		let projectRoot = platformData.projectRoot;
 		let archivePath = options.archivePath;
 		let buildOutputPath = path.join(projectRoot, "build", "device");
+		let exportOptionsMethod = await this.getExportOptionsMethod(projectData);
+		let plistTemplate = `<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+    <key>method</key>
+    <string>${exportOptionsMethod}</string>
+</dict>
+</plist>`;
+
+		// Save the options...
+		temp.track();
+		let exportOptionsPlist = temp.path({ prefix: "export-", suffix: ".plist" });
+		this.$fs.writeFile(exportOptionsPlist, plistTemplate);
 
 		// The xcodebuild exportPath expects directory and writes the <project-name>.ipa at that directory.
 		let exportPath = path.resolve(options.exportDir || buildOutputPath);
@@ -259,7 +273,7 @@ export class IOSProjectService extends projectServiceBaseLib.PlatformProjectServ
 		let args = ["-exportArchive",
 			"-archivePath", archivePath,
 			"-exportPath", exportPath,
-			"-exportOptionsPlist", platformData.configurationFilePath
+			"-exportOptionsPlist", exportOptionsPlist
 		];
 		await this.$childProcess.spawnFromEvent("xcodebuild", args, "exit",
 			{ stdio: buildConfig.buildOutputStdio || 'inherit', cwd: this.getPlatformData(projectData).projectRoot },
@@ -1329,6 +1343,32 @@ We will now place an empty obsolete compatability white screen LauncScreen.xib f
 
 		if (infoPlist.CFBundleIdentifier && infoPlist.CFBundleIdentifier !== mergedPlist.CFBundleIdentifier) {
 			this.$logger.warnWithLabel("The CFBundleIdentifier key inside the 'Info.plist' will be overriden by the 'id' inside 'package.json'.");
+		}
+	}
+
+	private async getEmbeddedPlistData(projectData: IProjectData): Promise<IDictionary<any>> {
+		let embeddedPlistContents = "";
+		const embeddedMobileProvisionPath = path.join(this.getPlatformData(projectData).deviceBuildOutputPath, `${projectData.projectName}.app`, "embedded.mobileprovision");
+
+		embeddedPlistContents = await this.$childProcess.exec(`security cms -D -i ${embeddedMobileProvisionPath}`);
+
+		return plist.parse(embeddedPlistContents);
+	}
+
+	private async getExportOptionsMethod(projectData: IProjectData): Promise<"app-store" | "ad-hoc" | "enterprise" | "development"> {
+		const embeddedPlistData = await this.getEmbeddedPlistData(projectData);
+
+		if (embeddedPlistData.ProvisionsAllDevices) {
+			return "enterprise";
+		} else if (embeddedPlistData.ProvisionedDevices && embeddedPlistData.ProvisionedDevices.length) {
+			const entitlements = embeddedPlistData.Entitlements;
+			if (entitlements["get-task-allow"]) {
+				return "development";
+			} else {
+				return "ad-hoc";
+			}
+		} else {
+			return "app-store";
 		}
 	}
 }
--- a/lib/services/ios-project-service.ts
+++ b/lib/services/ios-project-service.ts
@@ -258,6 +258,8 @@ export class IOSProjectService extends projectServiceBaseLib.PlatformProjectServ
 <dict>
     <key>method</key>
     <string>${exportOptionsMethod}</string>
+    <key>uploadBitcode</key>
+    <false/>
 </dict>
 </plist>`;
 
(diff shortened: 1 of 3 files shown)


dtopuzov:
run ci

lib/services/ios-project-service.ts
```diff
@@ -1312,6 +1326,32 @@ We will now place an empty obsolete compatability white screen LauncScreen.xib f
...
+	private async getEmbeddedPlistData(projectData: IProjectData): Promise<IDictionary<any>> {
+		let embeddedPlistContents = "";
+		const embeddedMobileProvisionPath = path.join(this.getPlatformData(projectData).deviceBuildOutputPath, `${projectData.projectName}.app`, "embedded.mobileprovision");
+
+		embeddedPlistContents = await this.$childProcess.exec(`security cms -D -i ${embeddedMobileProvisionPath}`);
+
+		return plist.parse(embeddedPlistContents);
+	}
+
+	private async getExportOptionsMethod(projectData: IProjectData): Promise<"app-store" | "ad-hoc" | "enterprise" | "development"> {
```
PanayotCankov: Working with provisioning profiles.
This belongs to the https://github.com/NativeScript/nativescript-cli/blob/master/lib/services/ios-provision-service.ts#L9
Add a method there that by given path the mobileprovision can be read.

PanayotCankov: Extract this block:
https://github.com/NativeScript/ios-mobileprovision-finder/blob/master/index.ts#L88
onto a "read from file" function and export that function in the ios-mobileprovision-finder
Use it instead of plist parse to obtain the information about the mobileprovision,
If necessary fix/extend the type here: https://github.com/NativeScript/ios-mobileprovision-finder/blob/master/index.ts#L101
then map that type to mode in the current method.

dtopuzov:
run ci

vbresults:
Isn't exportOptions.plist included in the new App_Resources iOS template and copied during the prepare step?  Why are we creating a temporary one here?

PanayotCankov:
@vbresults I don't think they are, or at least I haven't noticed. I don't see them here:
https://github.com/NativeScript/template-hello-world/tree/master/App_Resources/iOS
Imo @PeterStaev is on the right track. Just the code should be reorganised. And if possible reused with the existing stuff in the provision service.

PeterStaev:
@vbresults I do not see those being used in the CLI code even if they are added to the template, but I also do not see it there. Also it does not make sense to make them part of the template as those options change depending on the build/provision type.

lib/services/ios-project-service.ts
PanayotCankov: Just disable the bitcode:
```
    <key>uploadBitcode</key>
    <false/>
```
later, if it is necessary, we may add flags to specify path to a custom plist file.

vbresults:
I feel like hardcoding will cause problems again, the way it did the first time, leading to this PR.  Why not put in a regular plist variable `${EXPORT_METHOD}` and then replace it during the prepare step?

We could have it as part of the `tns-ios` project template by default, and if defined in `App_Resources` it overwrites with that; this helps avoid surprises later or imposing decisions made.

p.s. Turns out that I edited that script to use my exportOptions.plist.

PanayotCankov:
This is good to go: https://www.npmjs.com/package/ios-mobileprovision-finder 1.0.10

lib/services/ios-project-service.ts
```diff
@@ -1312,6 +1328,18 @@ We will now place an empty obsolete compatability white screen LauncScreen.xib f
 		}
 		return teamId;
 	}
+
+	private getExportOptionsMethod(projectData: IProjectData): string {
+		const embeddedMobileProvisionPath = path.join(this.getPlatformData(projectData).deviceBuildOutputPath, `${projectData.projectName}.app`, "embedded.mobileprovision");
+		const provision = mobileprovision.provision.readFromFile(embeddedMobileProvisionPath);
+
+		return {
```
PanayotCankov: I love this way of mapping big time!
Definitely better than switches!

dtopuzov:
run ci

PanayotCankov:
I just resolved a trivial merge conflict.

lib/services/ios-project-service.ts
PeterStaev: Actually took the idea from the `ios-mobileprovision-finder` 😄

dtopuzov:
run ci

dtopuzov:
@PanayotCankov before we get QAs on this we should fix Travis and Jenkins checks.

Currently builds fails with:
```
[33mCompiling...[39m
Using tsc v2.4.1
lib/services/ios-project-service.ts(1336,47): error TS2339: Property 'readFromFile' does not exist on type 'typeof provision'.
lib/services/ios-project-service.ts(1338,10): error TS7017: Element implicitly has an 'any' type because type '{ "Development": string; "AdHoc": string; "Distribution": string; "Enterprise": string; }' has no index signature.
```

PeterStaev:
@dtopuzov , for some reason the CI build does not install the new version of `ios-mobileprovision-finder` and the package continues to be 1.0.9 instead of 1.0.10: https://travis-ci.org/NativeScript/nativescript-cli/builds/262575088#L720. Not sure why as in the `package.json` I've put the new version.

rosen-vladimirov:
@PeterStaev , this is due to the existence of `shrinkwrap` file in the repository. I can help you to resolve this issue by regenerating the shrinkwrap in a separate commit.

PeterStaev:
That would be great @rosen-vladimirov ! I've tried regenerating it using `npm shrinkwrap` but the new file is totally different...

rosen-vladimirov:
run ci

rosen-vladimirov:
@PeterStaev I've updated the shrinkwrap json (in fact I've removed the ios-mobileprovision-finder from package.json and executed `npm install --save --save-exact ios-mobileprovision-finder@1.0.10` with npm5. This updated only this entry in the shrinkwrap file.
I've also rebased your branch on latest master, so I had to push force it, please excuse me for this.

The functional tests are running and we'll see the results ;)

dtopuzov:
run ci

dtopuzov:
run ci

dtopuzov:
run ci

vbresults:
How much time before this is available in @next?

dtopuzov:
I hope it will be live later today :)

rosen-vladimirov:
Hey @PeterStaev ,  we've merged this PR. Thanks again for your great contribution.

@vbresults this change is already available in `@next` of CLI.

PR #3038, opened 2017-08-04
Expose debugLiveSyncService

Mitko-Kerezov:
* Expose debugLiveSyncService so it can be used when CLI is `require`-d as a library
* Respect `outputPath` if passed in `refreshApplication`

### DIFF
--- a/PublicAPI.md
+++ b/PublicAPI.md
@@ -702,6 +702,12 @@ tns.liveSyncService.on("notify", data => {
 });
 ```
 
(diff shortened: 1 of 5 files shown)


rosen-vladimirov:
I believe these changes are no longer necessary, @Mitko-Kerezov can you close the PR or update it?

Mitko-Kerezov:
I think I'd rather close this as I need to make some additional changes. 
Will open a new PR when ready

PR #3039, opened 2017-08-04
ignore generated .js and .js.map files

Plamen5kov:
Vs code project configuration update.

_problem_
* `.vscode/launch.json` configures the local workspace for every developer, so this file can be different on each machine and IMO can be gitignored. 
* when the project is opened by vs code, all generated `.js` and `.js.map` files are showed in explorer, which is distracting and makes navigation difficult.

_solution_
* remove `.vscode/launch.json` leaving every developer to have his own configuration of the workspace
* `settings.json` on the other hand makes the project more readable by hiding generated `.js` and `.js.map` files.

_Recommended steps for update after this branch is merged_
* copy your `.vscode/launch.json` configuration
* `git pull` master branch
* paste your `.vscode/launch.json`


.vscode/launch.json
```diff
@@ -1,76 +0,0 @@
...
-			// "args": [ "plugin", "add", "nativescript-barcodescanner", "--path", "cliapp"]
-			// "args": [ "plugin", "remove", "nativescript-barcodescanner", "--path", "cliapp"]
-			// "args": [ "build", "android", "--path", "cliapp"]
-			// "args": [ "run", "android", "--path", "cliapp"]
-			// "args": [ "debug", "android", "--path", "cliapp"]
-			// "args": [ "livesync", "android", "--path", "cliapp"]
-			// "args": [ "livesync", "android", "--watch", "--path", "cliapp"]
-		},
-		{
-			// in case you want to debug a single test, modify it's code to be `it.only(...` instead of `it(...`
```
Pip3r4o: Can we just leave the Test configurations? I am not sure how smart VSCode is, but it may not be able to regenerate the same test configuration

Plamen5kov: it won't be that's the idea, every dev can have his own configuration. 
You'll still have your configuration, but it won't be in github.
Maybe I should have mentioned that you should, save your `launch.json` configuration somewhere, then pull master, then put bach the `launch.json` configuration.

Plamen5kov:
run CI

rosen-vladimirov:
👎 
I strongly disagree with these changes. `launch.json` should be used for debugging purposes, it has nothing specific for dev's environment. It's main purpose is to make it easier for new people, who are trying to contribute to CLI, to debug its code and unit tests. The setup (current `launch.json`) is not trivial, that's why it is in the `git` repo.
In case we merge this PR, how will you provide the `launch.json` to external contributors? Currently we have it on our machines, but newcomers will not have it when they clone the repository. How will they debug the unit tests?

On the other side - settings.json is exactly based on the user's experience. For example I prefer to see both `.js` and `.ts` files in the tree, but in case this PR is merged, I'll have to explicitly exclude this change whenever I want to commit something (as I'll definitely go back to the `settings.json` and remove the ignorance of `.js` files). Also, settings.json purpose is to be workspace config, that overrides user configuration, i.e. in case I have defined my user configuration to hide `.js` files when `.ts` files exist and there's no `settings.json` in the repository (current state of {N} CLI), I'll not see `.js` files. But in case there's `settings.json` inside the repo, its setting will override mine.

Plamen5kov:
I thought about putting the launch configuration, somewhere in the docs, but I'm closing the PR since I'm the only one who thinks it's for the best.

PR #3042, opened 2017-08-08
make livesync work with renamed files and folders

Plamen5kov:
_problem_
On Mac, renaming isn't caught as a change by the incremental prepare.

_effect_
When livesync is ran and a file in the `app/` folder is renamed, the change is not properly applied in the application.

_solution_
Add check for access time so renamed files are prepared.

_Edit_: Checking folder access times before traversing files in it can save us a lot of time checking files.

### DIFF
--- a/lib/services/project-changes-service.ts
+++ b/lib/services/project-changes-service.ts
@@ -219,6 +219,10 @@ export class ProjectChangesService implements IProjectChangesService {
 	}
 
 	private containsNewerFiles(dir: string, skipDir: string, projectData: IProjectData, processFunc?: (filePath: string, projectData: IProjectData) => boolean): boolean {
+		if (this.isDirectoryModified(dir)) {
+			return true;
+		}
+
 		let files = this.$fs.readDirectory(dir);
 		for (let file of files) {
 			let filePath = path.join(dir, file);
@@ -244,10 +248,6 @@ export class ProjectChangesService implements IProjectChangesService {
 			}
 
 			if (fileStats.isDirectory()) {
-				if (this.isDirectoryModified(dir)) {
-					return true;
-				}
-
 				if (this.containsNewerFiles(filePath, skipDir, projectData, processFunc)) {
 					return true;
 				}
--- a/lib/services/project-changes-service.ts
+++ b/lib/services/project-changes-service.ts
@@ -219,7 +219,8 @@ export class ProjectChangesService implements IProjectChangesService {
 	}
 
 	private containsNewerFiles(dir: string, skipDir: string, projectData: IProjectData, processFunc?: (filePath: string, projectData: IProjectData) => boolean): boolean {
-		if (this.isDirectoryModified(dir)) {
+		const dirFileStat = this.$fs.getFsStats(dir);
+		if (this.isFileModified(dirFileStat)) {
 			return true;
 		}
 
@@ -230,10 +231,8 @@ export class ProjectChangesService implements IProjectChangesService {
 				continue;
 			}
 
-			let fileStats = this.$fs.getFsStats(filePath);
-
-			let changed =  fileStats.mtime.getTime() > this._outputProjectMtime ||
-				fileStats.ctime.getTime() >= this._outputProjectCTime;
+			const fileStats = this.$fs.getFsStats(filePath);
+			let changed = this.isFileModified(fileStats);
 
 			if (changed) {
 				if (processFunc) {
@@ -257,10 +256,9 @@ export class ProjectChangesService implements IProjectChangesService {
 		return false;
 	}
 
-	private isDirectoryModified(dirPath: string): boolean {
-		const dirPathStat = this.$fs.getFsStats(dirPath);
-		return  dirPathStat.mtime.getTime() > this._outputProjectMtime ||
-					dirPathStat.ctime.getTime() >= this._outputProjectCTime;
+	private isFileModified(filePathStat: IFsStats): boolean {
+		return  filePathStat.mtime.getTime() >= this._outputProjectMtime ||
+					filePathStat.ctime.getTime() >= this._outputProjectCTime;
 	}
 
 	private fileChangeRequiresBuild(file: string, projectData: IProjectData) {
--- a/lib/services/project-changes-service.ts
+++ b/lib/services/project-changes-service.ts
@@ -233,6 +233,10 @@ export class ProjectChangesService implements IProjectChangesService {
 
 			const fileStats = this.$fs.getFsStats(filePath);
 			let changed = this.isFileModified(fileStats);
+			if (!changed) {
+				let lFileStats = this.$fs.getLsStats(filePath);
+				changed = this.isFileModified(lFileStats);
+			}
 
 			if (changed) {
 				if (processFunc) {
--- a/lib/services/project-changes-service.ts
+++ b/lib/services/project-changes-service.ts
@@ -220,7 +220,7 @@ export class ProjectChangesService implements IProjectChangesService {
 
 	private containsNewerFiles(dir: string, skipDir: string, projectData: IProjectData, processFunc?: (filePath: string, projectData: IProjectData) => boolean): boolean {
 		const dirFileStat = this.$fs.getFsStats(dir);
-		if (this.isFileModified(dirFileStat)) {
+		if(this.isFileModified(dirFileStat, dir)) {
 			return true;
 		}
 
@@ -232,11 +232,7 @@ export class ProjectChangesService implements IProjectChangesService {
 			}
 
 			const fileStats = this.$fs.getFsStats(filePath);
-			let changed = this.isFileModified(fileStats);
-			if (!changed) {
-				let lFileStats = this.$fs.getLsStats(filePath);
-				changed = this.isFileModified(lFileStats);
-			}
+			let changed = this.isFileModified(fileStats, filePath);
 
 			if (changed) {
 				if (processFunc) {
@@ -260,9 +256,17 @@ export class ProjectChangesService implements IProjectChangesService {
 		return false;
 	}
 
-	private isFileModified(filePathStat: IFsStats): boolean {
-		return  filePathStat.mtime.getTime() >= this._outputProjectMtime ||
+	private isFileModified(filePathStat: IFsStats, filePath: string): boolean {
+		let changed = filePathStat.mtime.getTime() >= this._outputProjectMtime ||
 					filePathStat.ctime.getTime() >= this._outputProjectCTime;
+
+		if (!changed) {
+			let lFileStats = this.$fs.getLsStats(filePath);
+			changed = lFileStats.mtime.getTime() >= this._outputProjectMtime ||
+					lFileStats.ctime.getTime() >= this._outputProjectCTime;
+		}
+
+		return changed;
 	}
 
 	private fileChangeRequiresBuild(file: string, projectData: IProjectData) {
--- a/lib/services/project-changes-service.ts
+++ b/lib/services/project-changes-service.ts
@@ -220,7 +220,7 @@ export class ProjectChangesService implements IProjectChangesService {
 
 	private containsNewerFiles(dir: string, skipDir: string, projectData: IProjectData, processFunc?: (filePath: string, projectData: IProjectData) => boolean): boolean {
 		const dirFileStat = this.$fs.getFsStats(dir);
-		if(this.isFileModified(dirFileStat, dir)) {
+		if (this.isFileModified(dirFileStat, dir)) {
 			return true;
 		}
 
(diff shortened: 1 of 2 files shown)


lib/services/project-changes-service.ts
```diff
@@ -160,6 +161,7 @@ export class ProjectChangesService implements IProjectChangesService {
 			let prepareInfoFile = path.join(platformData.projectRoot, prepareInfoFileName);
 			this._outputProjectMtime = this.$fs.getFsStats(prepareInfoFile).mtime.getTime();
 			this._outputProjectCTime = this.$fs.getFsStats(prepareInfoFile).ctime.getTime();
+			this._outputProjectATime = this.$fs.getFsStats(prepareInfoFile).atime.getTime();
 			return false;
```
Pip3r4o: Why is `false` returned here?

lib/services/project-changes-service.ts
```diff
@@ -227,11 +230,9 @@ export class ProjectChangesService implements IProjectChangesService {
 
 			let fileStats = this.$fs.getFsStats(filePath);
 
-			let changed = fileStats.mtime.getTime() >= this._outputProjectMtime || fileStats.ctime.getTime() >= this._outputProjectCTime;
-			if (!changed) {
-				let lFileStats = this.$fs.getLsStats(filePath);
-				changed = lFileStats.mtime.getTime() >= this._outputProjectMtime || lFileStats.ctime.getTime() >= this._outputProjectCTime;
-			}
+			let changed =  fileStats.mtime.getTime() >= this._outputProjectMtime ||
```
Pip3r4o: Why is a `greater than or equal` comparison made here? Wouldn't `changed` flip to true if the 2 sides were equal?

Plamen5kov: You got a point, but we're safe because the API is synchronous. I'll change it anyway, though, just in case.

lib/services/project-changes-service.ts
Plamen5kov: so we start checking for changes: https://github.com/NativeScript/nativescript-cli/blob/fe95da72b8f3c93e9a124aa673fb25a7d8c10995/lib/services/project-changes-service.ts#L61

rosen-vladimirov:
Access time is change when the [file is read](https://nodejs.org/dist/latest-v6.x/docs/api/fs.html#fs_stat_time_values), so in case you just open it in and IDE, we'll trigger new LiveSync operation, which is not required.

Whenever a file is renamed, the `mtime` and `ctime` of the parent directory should be changed. Based on the file system, the `mtime` and the `ctime` of the file might not change. So basically, I believe there's some problem with the check of `app` dir's `mtime` and `ctime`. Can you check this?

Plamen5kov:
On mac:
```
/////////// created file alabala.js
mcsofplpetkov:test plpetkov$ stat -x alabala.js 
  File: "alabala.js"
  Size: 0            FileType: Regular File
  Mode: (0644/-rw-r--r--)         Uid: (1592642667/plpetkov)  Gid: (157071322/PROGRESS\TelerikDomainUsers)
Device: 1,3   Inode: 4296408    Links: 1
Access: Tue Aug  8 11:18:58 2017
Modify: Tue Aug  8 11:18:58 2017
Change: Tue Aug  8 11:18:58 2017

/////////// renamed file alabala.js to asd.js with Mac finder
mcsofplpetkov:test plpetkov$ stat -x asd.js 
  File: "asd.js"
  Size: 0            FileType: Regular File
  Mode: (0644/-rw-r--r--)         Uid: (1592642667/plpetkov)  Gid: (157071322/PROGRESS\TelerikDomainUsers)
Device: 1,3   Inode: 4296408    Links: 1
Access: Tue Aug  8 17:06:03 2017
Modify: Tue Aug  8 11:18:58 2017
Change: Tue Aug  8 11:18:58 2017

/////////// renamed file asd.js to newFile.js with mv command
mcsofplpetkov:test plpetkov$ mv asd.js newName.js
mcsofplpetkov:test plpetkov$ stat -x newName.js 
  File: "newName.js"
  Size: 0            FileType: Regular File
  Mode: (0644/-rw-r--r--)         Uid: (1592642667/plpetkov)  Gid: (157071322/PROGRESS\TelerikDomainUsers)
Device: 1,3   Inode: 4296408    Links: 1
Access: Tue Aug  8 17:06:39 2017
Modify: Tue Aug  8 11:18:58 2017
Change: Tue Aug  8 11:18:58 2017

```

Seems only access time is changed on rename through finder or `mv` command.

rosen-vladimirov:
@Plamen5kov , thanks for the details. As I've pointed - on some File Systems, the ctime and mtime of a file, will not be changed when it is renamed. However, the `ctime` and `mtime` of the directory in which this file is located, will be changed.
During checking for changes we iterate over all files and dirs inside app dir (recursively) and compare there `ctime` and `mtime` with the last prepared time. But maybe we are not checking the `app` dir's `mtime` and `ctime` itself?

Plamen5kov:
Yes I just thought it would be interesting information. I'll check if that's the issue and fix it if necessary.

Plamen5kov:
@rosen-vladimirov turns out you were right, when I added the check for the parent directory, everything started running smoothly. Rename of files and folders is running on mac, and soon on android.

lib/services/project-changes-service.ts
```diff
@@ -227,10 +228,15 @@ export class ProjectChangesService implements IProjectChangesService {
...
 			let fileStats = this.$fs.getFsStats(filePath);
 
-			let changed = fileStats.mtime.getTime() >= this._outputProjectMtime || fileStats.ctime.getTime() >= this._outputProjectCTime;
-			if (!changed) {
-				let lFileStats = this.$fs.getLsStats(filePath);
-				changed = lFileStats.mtime.getTime() >= this._outputProjectMtime || lFileStats.ctime.getTime() >= this._outputProjectCTime;
+			let changed =  fileStats.mtime.getTime() > this._outputProjectMtime ||
+				fileStats.ctime.getTime() >= this._outputProjectCTime;
+
+			if (this.$hostInfo.isDarwin && !fileStats.isDirectory() && !changed) {
```
rosen-vladimirov: Why is this only for darwin? The case is for the File System itself, not the OS, so it may happen on different OSes.

lib/services/project-changes-service.ts
```diff
@@ -227,10 +228,15 @@ export class ProjectChangesService implements IProjectChangesService {
...
+				const parentDirPath: string = path.dirname(filePath);
+				const parentDirStat = this.$fs.getFsStats(parentDirPath);
```
rosen-vladimirov: In case we have 50 files in the current dir and nothing is changed (including the parent dir), we'll get the stats of the parent directory 50 times. 
Also, consider the following structure:
```
-- app
   |
   -- components
       |
        -- file1
        -- file 2
```

We'll call `containsNewerFiles` for `app` dir. It will list the contents of the app dir and for each of them it will check the `ctime` and `mtime`, i.e. it will check the times of `components` directory.
In case there's no change, the method will recursively check the contents of the `components` directory. At this point we'll check file1 and file2 and for each of them we'll get the parent dir (`components`) and check its `ctime` and `mtime`. But we've already done this in the previous step, why should we do it again?

dtopuzov:
run ci

lib/services/project-changes-service.ts
Plamen5kov: I agree with you, but I don't know a trivial way of finding out what operating system is the code running on. That's why I check for darwin, because that's where the problem presents itself.

lib/services/project-changes-service.ts
Plamen5kov: @rosen-vladimirov I did this as a temporary solution because I needed to test the fix on Mac and I needed to pull the code from it. I'm sorry if I inconvenienced you. I'm currently working on an optimized, permanent solution and will ping you when ready.

Plamen5kov:
ping @rosen-vladimirov

lib/services/project-changes-service.ts
```diff
@@ -227,11 +232,8 @@ export class ProjectChangesService implements IProjectChangesService {
 
 			let fileStats = this.$fs.getFsStats(filePath);
 
-			let changed = fileStats.mtime.getTime() >= this._outputProjectMtime || fileStats.ctime.getTime() >= this._outputProjectCTime;
-			if (!changed) {
-				let lFileStats = this.$fs.getLsStats(filePath);
```
rosen-vladimirov: why is this code deleted?

lib/services/project-changes-service.ts
```diff
@@ -250,10 +252,17 @@ export class ProjectChangesService implements IProjectChangesService {
 					return true;
 				}
 			}
+
 		}
 		return false;
 	}
 
+	private isDirectoryModified(dirPath: string): boolean {
+		const dirPathStat = this.$fs.getFsStats(dirPath);
```
rosen-vladimirov: The code is the same as the one used on line 235. Also the method is called `isDirectortyModified`, but the implementation does not have any code specific to directories only.
I suggest renaming the method and use it on line 235.

test/project-changes-service.ts
```diff
@@ -26,7 +28,8 @@ class ProjectChangesServiceTest extends BaseServiceTest {
 		this.injector.register("platformsData", PlatformsData);
 		this.injector.register("androidProjectService", {});
 		this.injector.register("iOSProjectService", {});
-
+		this.injector.register("hostInfo", HostInfoLib.HostInfo);
```
rosen-vladimirov: this is not used and shouldn't be added

test/project-changes-service.ts
```diff
@@ -26,7 +28,8 @@ class ProjectChangesServiceTest extends BaseServiceTest {
...
+		this.injector.register("errors", ErrorsLib.Errors);
```
rosen-vladimirov: this is not used and shouldn't be added

lib/services/project-changes-service.ts
```diff
@@ -227,11 +232,8 @@ export class ProjectChangesService implements IProjectChangesService {
 
 			let fileStats = this.$fs.getFsStats(filePath);
 
-			let changed = fileStats.mtime.getTime() >= this._outputProjectMtime || fileStats.ctime.getTime() >= this._outputProjectCTime;
-			if (!changed) {
-				let lFileStats = this.$fs.getLsStats(filePath);
-				changed = lFileStats.mtime.getTime() >= this._outputProjectMtime || lFileStats.ctime.getTime() >= this._outputProjectCTime;
-			}
+			let changed =  fileStats.mtime.getTime() > this._outputProjectMtime ||
```
rosen-vladimirov: why is this without `=` sign, but the next check has it? I believe you need the `=` in both cases, as the `_outputProjectMtime ` is set after the prepare. In case it is equal with a file's mtime, this means the file has been modified when the prepare is finishing (exactly at the same time), so we need to prepare the project again.

lib/services/project-changes-service.ts
Plamen5kov: it's unnecessary to do `this.$fs.getLsStats(filePath)` twice and the logic is kept the same in the refactored version

rosen-vladimirov: `getLstStats` and `getFsStats` are different. This code is used when the file/dir is symlink - we need to check the symlink itself (for example if it is recreated) and this is done by `getLsStats`.

Plamen5kov:
run CI

lib/services/project-changes-service.ts
rosen-vladimirov: ping @Plamen5kov

dtopuzov:
run ci

lib/services/project-changes-service.ts
Plamen5kov: I will bring back the code, so we continue working with synlinked files.

lib/services/project-changes-service.ts
```diff
@@ -218,19 +219,23 @@ export class ProjectChangesService implements IProjectChangesService {
 	}
 
 	private containsNewerFiles(dir: string, skipDir: string, projectData: IProjectData, processFunc?: (filePath: string, projectData: IProjectData) => boolean): boolean {
+		const dirFileStat = this.$fs.getFsStats(dir);
```
rosen-vladimirov: what if the lstat of the dir has been changed?

Plamen5kov: Well spotted, I don't know how I missed that!

lib/services/project-changes-service.ts
```diff
@@ -218,20 +219,20 @@ export class ProjectChangesService implements IProjectChangesService {
...
+		if(this.isFileModified(dirFileStat, dir)) {
```
rosen-vladimirov: space is required between `if` and `(`

lib/services/project-changes-service.ts
```diff
@@ -218,20 +219,20 @@ export class ProjectChangesService implements IProjectChangesService {
...
 			}
 
-			let fileStats = this.$fs.getFsStats(filePath);
-
-			let changed = fileStats.mtime.getTime() >= this._outputProjectMtime || fileStats.ctime.getTime() >= this._outputProjectCTime;
-			if (!changed) {
-				let lFileStats = this.$fs.getLsStats(filePath);
-				changed = lFileStats.mtime.getTime() >= this._outputProjectMtime || lFileStats.ctime.getTime() >= this._outputProjectCTime;
-			}
+			const fileStats = this.$fs.getFsStats(filePath);
```
rosen-vladimirov: the `fileStats` object is not used anywhere outside of the `isFileModified` method. So I believe its place is inside it. At the moment on line 222 we have:
1. get fstat 
2. call `isFileModified`

Here we are doing the same:
1. get fstat
2. call `isFileModified`

I suggest passing only the path to `isFileModified` and getting the fstat in the method itself

Plamen5kov: It's used [here](https://github.com/NativeScript/nativescript-cli/pull/3042/files/9020a8648f41416cec30c7c45456033c33b96db7#diff-919912d52b0e0303e76febe7c84cacfdR249)

rosen-vladimirov:
run ci

Plamen5kov:
run CI

dtopuzov:
run ci

dtopuzov:
run ci

Issue #2810, opened 2017-05-12
Cannot read property 'match' of null error when installing tns cli


teobugslayer:
### Tell us about the problem
I am installing tns on clean Mac Sierra, without any dependencies installed.
in this case, the following issue appears:

> Warning: CocoaPods is not installed or is not configured properly
> ....
> Cannot read property 'match' of null
> Do you want to help us improve NativeScript by automatically ....

### Which platform(s) does your issue occur on?
macOS

### Please provide the following version numbers that your issue occurs with:
- CLI: 2.5.0

### Please tell us how to recreate the issue in as much detail as possible.
On a freshly install macOS Sierra, install homebrew, node and then attempt to install NativeScript CLI

vchimev:
Perhaps, java has not been installed. Related to: https://github.com/NativeScript/nativescript-cli/issues/1922.

dtopuzov:
It looks more like https://github.com/NativeScript/nativescript-cli/issues/1245#issuecomment-159562131

dtopuzov:
Actually those are two separate issues:
1. Setup script do not install cocoa pods properly
2. Tns doctor do not show appropriate error message

Plamen5kov:
Can't reproduce the problem.

Plamen5kov:
@teobugslayer can you provide the output from the command:
`tns doctor --log trace` so we can have a more detailed look at the computer configuration, because there's a possibility xCode is not installed or `xcodebuild` tool is not in the PATH.

Plamen5kov:
@teobugslayer if your xcode is not installed at all or not configured properly, you should see something like:
```
mcsofplpetkov:nativescript-cli plpetkov$ tns doctor
WARNING: Xcode is not installed or is not configured properly.
You will not be able to build your projects for iOS or run them in the iOS Simulator.
To be able to build for iOS and run apps in the native emulator, verify that you have installed Xcode.

WARNING: CocoaPods is not installed or is not configured properly.
You will not be able to build your projects for iOS if they contain plugin with CocoaPod file.
To be able to build such projects, verify that you have installed CocoaPods.
Cannot read property 'match' of null
# doctor
```
If that's the case, we'll work on a better error message.

PR #3043, opened 2017-08-08
doctor doesn't crash on missing xcode tools

Plamen5kov:
Doctor crashes because of invalid environment.
Related issue: NativeScript/nativescript-cli#2810

_problem_
* doctor crashes when `xcodebuild` can't be found (when xCode build tools aren't installed).
* provide a more detailed message for: "Your components are up to date"

_solution_
* don't verify xCode project if xcodeVersion and cocoaPods aren't verified
* list the components that are up to date for the user

### DIFF
--- a/lib/declarations.d.ts
+++ b/lib/declarations.d.ts
@@ -604,10 +604,9 @@ interface IVersionsService {
 	getRuntimesVersions(): Promise<IVersionInformation[]>;
 
 	/**
-	 * Gets versions information about the nativescript components with new.
-	 * @return {Promise<IVersionInformation[]>} The version information.
+	 * Checks version information about the nativescript components and prints available updates if any.
 	 */
-	getComponentsForUpdate(): Promise<void>;
+	checkComponentsForUpdate(): Promise<void>;
 
 	/**
 	 * Gets versions information about all nativescript components.
--- a/lib/declarations.d.ts
+++ b/lib/declarations.d.ts
@@ -607,7 +607,7 @@ interface IVersionsService {
 	 * Gets versions information about the nativescript components with new.
 	 * @return {Promise<IVersionInformation[]>} The version information.
 	 */
-	getComponentsForUpdate(): Promise<IVersionInformation[]>;
+	getComponentsForUpdate(): Promise<void>;
 
 	/**
 	 * Gets versions information about all nativescript components.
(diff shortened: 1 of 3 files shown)


Plamen5kov:
run CI

Plamen5kov:
run CI

dtopuzov:
run ci

lib/declarations.d.ts
```diff
@@ -607,7 +607,7 @@ interface IVersionsService {
 	 * Gets versions information about the nativescript components with new.
 	 * @return {Promise<IVersionInformation[]>} The version information.
 	 */
-	getComponentsForUpdate(): Promise<IVersionInformation[]>;
+	getComponentsForUpdate(): Promise<void>;
```
rosen-vladimirov: In case the method is void, its name is incorrect as it does not get anything

lib/services/doctor-service.ts
```diff
@@ -95,7 +95,7 @@ class DoctorService implements IDoctorService {
 				result = true;
 			}
 
-			if (await this.$xcprojService.verifyXcproj(false)) {
+			if (sysInfo.cocoapodVer && sysInfo.cocoapodVer && await this.$xcprojService.verifyXcproj(false)) {
```
rosen-vladimirov: maybe this check should be:
```TypeScript
if (sysInfo.xcodeVer && sysInfo.cocoapodVer && ...)
```

lib/services/versions-service.ts
```diff
@@ -84,7 +86,7 @@ class VersionsService implements IVersionsService {
 		return runtimesVersions;
 	}
 
-	public async getComponentsForUpdate(): Promise<IVersionInformation[]> {
+	public async getComponentsForUpdate() {
```
rosen-vladimirov: this method needs return type

lib/services/versions-service.ts
```diff
@@ -94,7 +96,18 @@ class VersionsService implements IVersionsService {
 			}
 		});
 
-		return componentsForUpdate;
+		this.printVersionsInformation(componentsForUpdate, allComponents);
+	}
+
+	private printVersionsInformation(versionsInformation: IVersionInformation[], allComponents: IVersionInformation[]) {
```
rosen-vladimirov: this method needs return type

PR #3051, opened 2017-08-11
don't restart application twice on crash

Plamen5kov:
(android) Don't restart application twice after it crashes for any reason.

_problem_
scenario:
* livesync is running and the user crashes the application (e.g. delete `app.js` in js template)
* user realizes mistake and fixes it (e.g undo delete of `app.js`)
* livesync starts the application twice

_solution_
pass the unsuccessful reload check to the provided process for checking if an application should be restarted

_effect_
The expected behavior after this PR is, after a crash or in other cases where the runtime socket is not started or crashing, there won't be two restarts of the device.

>Note: I tested these scenarios on iOS, and there seem to be no such issues.

### DIFF
--- a/lib/services/livesync/android-device-livesync-service.ts
+++ b/lib/services/livesync/android-device-livesync-service.ts
@@ -36,9 +36,9 @@ export class AndroidDeviceLiveSyncService extends DeviceLiveSyncServiceBase impl
 				`${deviceProjectRootDirname}/sync`]
 		);
 
-		await this.reloadResources(deviceAppData, localToDevicePaths);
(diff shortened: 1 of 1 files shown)


Plamen5kov:
The app will be restarted in case you change a `.js` file, because the other checks still hold. You are right that `reloadedSuccessfully` variable will be `true`, but the `canExecuteFastSync` will be false because of the `.js` extension, so the app will be restarted and the changes will be applied.
After this PR all previous behavior is kept, except for the double restart.

rosen-vladimirov:
My face when I think && is || :
image

PR #2800, opened 2017-05-10
Improve get latest compatible version to search for major.minor.*any*…

yyosifov:
… versions

PR #3052, opened 2017-08-11
Improve get latest compatible version to search for major.minor.*x* versions

rosen-vladimirov:
This has been fixed in CLI 2.5.5, but it never get into our master or release branch:
Not merged PR: https://github.com/NativeScript/nativescript-cli/pull/2800
Merged PR in 2.5.5: https://github.com/NativeScript/nativescript-cli/pull/2793

I've just cherry-picked the commit from PR 2800.

The problem arises when we have CLI 3.1.3 for example, but a template has versions 3.1.0, 3.1.1, 3.2.0, 3.2.1...
In this case CLI should install template 3.1.1, but instead it uses latest (3.2.1). The problem is that we have been using semver comparison based on CLI's version (~3.1.3), while in fact we should get only major and minor numbers from it (~3.1.0)

### DIFF
--- a/lib/npm-installation-manager.ts
+++ b/lib/npm-installation-manager.ts
@@ -20,15 +20,23 @@ export class NpmInstallationManager implements INpmInstallationManager {
 	}
 
 	public async getLatestCompatibleVersion(packageName: string): Promise<string> {
+		const configVersion = this.$staticConfig.version;
+		const isPreReleaseVersion = semver.prerelease(configVersion) !== null;
+		let cliVersionRange = `~${semver.major(configVersion)}.${semver.minor(configVersion)}.0`;
+		if (isPreReleaseVersion) {
+			// if the user has some 0-19 pre-release version, include pre-release versions in the search query.
+			cliVersionRange = `~${configVersion}`;
+		}
 
-		let cliVersionRange = `~${this.$staticConfig.version}`;
-		let latestVersion = await this.getLatestVersion(packageName);
+		const latestVersion = await this.getLatestVersion(packageName);
 		if (semver.satisfies(latestVersion, cliVersionRange)) {
 			return latestVersion;
 		}
-		let data = await this.$npm.view(packageName, { "versions": true });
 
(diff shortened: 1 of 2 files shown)


lib/npm-installation-manager.ts
```diff
@@ -20,15 +20,22 @@ export class NpmInstallationManager implements INpmInstallationManager {
 	}
 
 	public async getLatestCompatibleVersion(packageName: string): Promise<string> {
+		const configVersion = this.$staticConfig.version;
+		const isPreReleaseVersion = (<any>semver).prerelease(configVersion) !== null;
```
Mitko-Kerezov: I do believe our definitions for `semver` already include a definition for `prerelease` so you can lose the `<any>` cast

lib/npm-installation-manager.ts
```diff
@@ -20,15 +20,22 @@ export class NpmInstallationManager implements INpmInstallationManager {
...
+		let cliVersionRange = `~${semver.major(configVersion)}.${(<any>semver).minor(configVersion)}.0`;
```
Mitko-Kerezov: Same remark about casting `semver` to `<any>` when calling `minor`

lib/npm-installation-manager.ts
```diff
@@ -20,15 +20,22 @@ export class NpmInstallationManager implements INpmInstallationManager {
...
 
-		let cliVersionRange = `~${this.$staticConfig.version}`;
 		let latestVersion = await this.getLatestVersion(packageName);
 		if (semver.satisfies(latestVersion, cliVersionRange)) {
 			return latestVersion;
 		}
 		let data = await this.$npm.view(packageName, { "versions": true });
 
-		return semver.maxSatisfying(data, cliVersionRange) || latestVersion;
+		let maxSatisfying = semver.maxSatisfying(data, cliVersionRange);
```
Mitko-Kerezov: This could be const

rosen-vladimirov:
run ci

dtopuzov:
run ci

rosen-vladimirov:
run ci

Issue #2728, opened 2017-04-20
 `tns run ios --device fakeID` start iOS Simulator 


dtopuzov:
Steps to reproduce: `tns run ios --device fake --path TestApp`

Log:
```
Executing before-prepare hook from /Users/topuzov/Git/nativescript-cli-tests/TestApp/hooks/before-prepare/nativescript-dev-android-snapshot.js
Preparing project...
Project successfully prepared (ios)
Executing after-prepare hook from /Users/topuzov/Git/nativescript-cli-tests/TestApp/hooks/after-prepare/nativescript-dev-android-snapshot.js
Searching for devices...
Could not find device by specified identifier 'fake'. To list currently connected devices and verify that the specified identifier exists, run 'tns device'.
```
...log is fine, but iOS Simulator is also started.

Tested with {N} CLI 3.0.0-2017.4.13.3

PR #3060, opened 2017-08-14
Update ios-sim-portable to 3.1.0

rosen-vladimirov:
This will fix issue with starting iOS Simulator when a fake device id is passed. Instead we will fail with correct error from now on.

As we are now using shrinkwrap file, the version of ios-sim-portable is always strict, so set it to strict in package.json as well.
Related to: https://github.com/NativeScript/nativescript-cli/issues/2728

Merge after https://github.com/telerik/mobile-cli-lib/pull/1000

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -2926,9 +2926,9 @@
       }
     },
     "ios-sim-portable": {
-      "version": "3.0.0",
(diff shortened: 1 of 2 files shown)


rosen-vladimirov:
run ci

rosen-vladimirov:
run ci

rosen-vladimirov:
run ci

PR #3064, opened 2017-08-15
Fix tns run ios when no simulator is running

rosen-vladimirov:
In case when there's no device attached and no simulator running, `tns run ios` should start default iOS Simulator. However, this fails. The problem is resolved in `ios-sim-portable` 3.1.1.
Bump the version, so the issue will be resolved.


rosen-vladimirov:
run ci

PR #3069, opened 2017-08-17
Fix passing provision to `tns run ios` and passsing `--teamId`

rosen-vladimirov:
Passing `--provision` to `tns run <ios>` is not respected. Fix this by including additional option in the deviceDescriptors. This allows us to pass different provisions for each device, however in CLI's commands we'll set one provision for all devices.

Also fix passing `--teamId` option - during previous refactoring we've broken the code for it and it has not been respected. Remove the unused alias `teamIdentifier` and use `teamId` in all cases.

Rename `IAddPlatformCoreOptions` to `IPlatformOptions` as it is better name of the purpose of the interface.

### DIFF
--- a/lib/definitions/livesync.d.ts
+++ b/lib/definitions/livesync.d.ts
@@ -94,6 +94,11 @@ interface ILiveSyncDeviceInfo {
 	 * Whether to skip preparing the native platform.
 	 */
 	skipNativePrepare?: boolean;
+
+	/**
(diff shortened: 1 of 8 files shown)


rosen-vladimirov:
run ci

dtopuzov:
run ci

rosen-vladimirov:
run ci

PR #3070, opened 2017-08-17
Fix xcode9 provision, provide --quiet by default, enable -allowProvisioningUpdates

PanayotCankov:
Depends on:
https://github.com/NativeScript/nativescript-cli/pull/3069

This PR will:

 - Update the export options plist and list the provisioning profile so Xcode9 can export when working with manual signing style and the --provision switch.
 - Provide --quiet to xcodebuild and -quiet to the gradle tools so they log only errors by default. The command line args with which these commands are called will be printed in place of the verbose output. You can switch to --log trace for full output.
 - Provide -allowProvisioningUpdates to xcodebuild when working with Xcode version > 9.0 so when passing team, xcodebuild may generate automatically managed provisioning profiles for you  
  (THIS ONE IS MY FAVOURITE!!!)
 - --teamId now behaves like --provision, it will require prepare if the last time other teamId was issued, it will print all teams if no arg is provided, it will switch to automatic signing
 - using --teamId and --provision will throw that they are mutually exclusive
 - teamId should work with team name, if we find a provision that has the team name and id  
  (we should improve that by somehow looking in the accounts registered in Xcode)

### DIFF
--- a/lib/services/android-project-service.ts
+++ b/lib/services/android-project-service.ts
@@ -517,12 +517,18 @@ export class AndroidProjectService extends projectServiceBaseLib.PlatformProject
 		if (this.$androidToolsInfo.getToolsInfo().androidHomeEnvVar) {
 			const gradlew = this.$hostInfo.isWindows ? "gradlew.bat" : "./gradlew";
 
+			const localArgs = [...gradleArgs];
+			if (this.$logger.getLevel() === "INFO") {
+				localArgs.push("--quiet");
+				this.$logger.info(`${gradlew} ${localArgs.join(" ")}`);
+			}
+
 			childProcessOpts = childProcessOpts || {};
 			childProcessOpts.cwd = childProcessOpts.cwd || projectRoot;
 			childProcessOpts.stdio = childProcessOpts.stdio || "inherit";
 
 			return await this.spawn(gradlew,
-				gradleArgs,
+				localArgs,
 				childProcessOpts,
 				spawnFromEventOptions);
 		}
--- a/lib/services/android-project-service.ts
+++ b/lib/services/android-project-service.ts
@@ -447,7 +447,7 @@ export class AndroidProjectService extends projectServiceBaseLib.PlatformProject
 		await adb.executeShellCommand(["rm", "-rf", deviceRootPath]);
 	}
 
-	public checkForChanges(changesInfo: IProjectChangesInfo, options: IProjectChangesOptions, projectData: IProjectData): void {
+	public async checkForChanges(changesInfo: IProjectChangesInfo, options: IProjectChangesOptions, projectData: IProjectData): Promise<void> {
 		// Nothing android specific to check yet.
 	}
 
--- a/lib/services/android-project-service.ts
+++ b/lib/services/android-project-service.ts
@@ -520,7 +520,7 @@ export class AndroidProjectService extends projectServiceBaseLib.PlatformProject
 			const localArgs = [...gradleArgs];
 			if (this.$logger.getLevel() === "INFO") {
 				localArgs.push("--quiet");
-				this.$logger.info(`${gradlew} ${localArgs.join(" ")}`);
+				this.$logger.info("Gradle build...");
 			}
 
 			childProcessOpts = childProcessOpts || {};
--- a/lib/services/ios-project-service.ts
+++ b/lib/services/ios-project-service.ts
@@ -433,8 +433,12 @@ export class IOSProjectService extends projectServiceBaseLib.PlatformProjectServ
 	private async xcodebuild(args: string[], cwd: string, stdio: any = "inherit"): Promise<ISpawnResult> {
 		const localArgs = [...args];
 		const xcodeBuildVersion = await this.getXcodeVersion();
-		if (helpers.versionCompare(xcodeBuildVersion, "9.0") >= 0) {
-			localArgs.push("-allowProvisioningUpdates");
+		try {
+			if (helpers.versionCompare(xcodeBuildVersion, "9.0") >= 0) {
+				localArgs.push("-allowProvisioningUpdates");
+			}
+		} catch (e) {
+			console.warn("Failed to use xcodebuild with -allowProvisioningUpdates due to error: " + e);
 		}
 		if (this.$logger.getLevel() === "INFO") {
 			localArgs.push("-quiet");
@@ -1283,9 +1287,7 @@ We will now place an empty obsolete compatability white screen LauncScreen.xib f
 		}
 
 		let splitedXcodeBuildVersion = xcodeBuildVersion.split(".");
-		if (splitedXcodeBuildVersion.length === 3) {
-			xcodeBuildVersion = `${splitedXcodeBuildVersion[0]}.${splitedXcodeBuildVersion[1]}`;
-		}
+		xcodeBuildVersion = `${splitedXcodeBuildVersion[0] || 0}.${splitedXcodeBuildVersion[1] || 0}`;
 
 		return xcodeBuildVersion;
 	}
--- a/lib/services/ios-project-service.ts
+++ b/lib/services/ios-project-service.ts
@@ -429,7 +429,7 @@ export class IOSProjectService extends projectServiceBaseLib.PlatformProjectServ
 		}
 		if (this.$logger.getLevel() === "INFO") {
 			localArgs.push("-quiet");
-			this.$logger.info(`xcodebuild ${localArgs.join(" ")}`);
+			this.$logger.info("Xcode build...");
 		}
 		return this.$childProcess.spawnFromEvent("xcodebuild",
 			localArgs,
--- a/lib/services/ios-project-service.ts
+++ b/lib/services/ios-project-service.ts
@@ -93,10 +93,20 @@ export class IOSProjectService extends projectServiceBaseLib.PlatformProjectServ
 		return this._platformData;
 	}
 
-	public async validateOptions(projectId: string, provision: true | string): Promise<boolean> {
+	public async validateOptions(projectId: string, provision: true | string, teamId: true | string): Promise<boolean> {
+		if (provision && teamId) {
+			this.$errors.failWithoutHelp("The options --provision and --teamId are mutually exclusive.");
+		}
+
 		if (provision === true) {
-			await this.$iOSProvisionService.list(projectId);
+			await this.$iOSProvisionService.listProvisions(projectId);
 			this.$errors.failWithoutHelp("Please provide provisioning profile uuid or name with the --provision option.");
+			return false;
+		}
+
+		if (teamId === true) {
+			await this.$iOSProvisionService.listTeams();
+			this.$errors.failWithoutHelp("Please provide team id or team name with the --teamId options.");
 			return false;
 		}
 
@@ -194,7 +204,7 @@ export class IOSProjectService extends projectServiceBaseLib.PlatformProjectServ
 		let args = ["archive", "-archivePath", archivePath, "-configuration",
 			(!buildConfig || buildConfig.release) ? "Release" : "Debug"]
 			.concat(this.xcbuildProjectArgs(projectRoot, projectData, "scheme"));
-		await this.xcodebuild(args, projectRoot, buildConfig.buildOutputStdio);
+		await this.xcodebuild(args, projectRoot, buildConfig && buildConfig.buildOutputStdio);
 		return archivePath;
 	}
 
@@ -272,7 +282,7 @@ export class IOSProjectService extends projectServiceBaseLib.PlatformProjectServ
 	<key>${projectData.projectId}</key>
 	<string>${options.provision}</string>
 </dict>`;
-	}	
+	}
 	plistTemplate += `
     <key>uploadBitcode</key>
     <false/>
@@ -403,17 +413,18 @@ export class IOSProjectService extends projectServiceBaseLib.PlatformProjectServ
 			await this.setupSigningForDevice(projectRoot, buildConfig, projectData);
 		}
 
-		if (buildConfig && buildConfig.codeSignIdentity) {
-			args.push(`CODE_SIGN_IDENTITY=${buildConfig.codeSignIdentity}`);
-		}
+		// These are now set in the .pbxproj, avoid passing them as arguments!
+		// if (buildConfig && buildConfig.codeSignIdentity) {
+		// 	args.push(`CODE_SIGN_IDENTITY=${buildConfig.codeSignIdentity}`);
+		// }
 
-		if (buildConfig && buildConfig.mobileProvisionIdentifier) {
-			args.push(`PROVISIONING_PROFILE=${buildConfig.mobileProvisionIdentifier}`);
-		}
+		// if (buildConfig && buildConfig.mobileProvisionIdentifier) {
+		// 	args.push(`PROVISIONING_PROFILE=${buildConfig.mobileProvisionIdentifier}`);
+		// }
 
-		if (buildConfig && buildConfig.teamId) {
-			args.push(`DEVELOPMENT_TEAM=${buildConfig.teamId}`);
-		}
+		// if (buildConfig && buildConfig.teamId) {
+		// 	args.push(`DEVELOPMENT_TEAM=${buildConfig.teamId}`);
+		// }
 
 		await this.xcodebuild(args, projectRoot, buildConfig.buildOutputStdio);
 		await this.createIpa(projectRoot, projectData, buildConfig);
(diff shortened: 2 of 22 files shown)


dtopuzov:
run ci

dtopuzov:
It looks ok, but we need to update e2e functional tests to respect changed output.
Will review the PR later today.

lib/services/ios-project-service.ts
```diff
@@ -219,6 +229,13 @@ export class IOSProjectService extends projectServiceBaseLib.PlatformProjectServ
     <string>${options.teamID}</string>
 `;
 		}
+		if (options && options.provision) {
+			plistTemplate += `    <key>provisioningProfiles</key>
```
Mitko-Kerezov: A separate service/helper function could be extracted for getting the `exportOptions.plist` file and be reused here and [here](https://github.com/NativeScript/nativescript-cli/pull/3070/files#diff-b673f137deb4afccccc0bd5f1572c2a0R280).
This could be done in another PR though

lib/services/ios-project-service.ts
```diff
@@ -388,27 +413,75 @@ export class IOSProjectService extends projectServiceBaseLib.PlatformProjectServ
...
+			localArgs,
 			"exit",
-			{ stdio: buildConfig.buildOutputStdio || "inherit", cwd: this.getPlatformData(projectData).projectRoot },
+			{ stdio: stdio || "inherit", cwd },
 			{ emitOptions: { eventName: constants.BUILD_OUTPUT_EVENT_NAME }, throwError: true });
-		// this.$logger.out("xcodebuild build succeded.");
+	}
 
-		await this.createIpa(projectRoot, projectData, buildConfig);
+	private async setupSigningFromTeam(projectRoot: string, projectData: IProjectData, teamId?: string) {
```
Mitko-Kerezov: Is the `teamId` argument really optional here - what will happen if it is not passed - `shouldUpdateXcode` would be set to `true` and then `Automatic` signing will be set without a team.

lib/services/ios-project-service.ts
```diff
@@ -388,27 +413,75 @@ export class IOSProjectService extends projectServiceBaseLib.PlatformProjectServ
...
+		const xcode = this.$pbxprojDomXcode.Xcode.open(this.getPbxProjPath(projectData));
+		const signing = xcode.getSigning(projectData.projectName);
+
+		let shouldUpdateXcode = false;
+		if (signing && signing.style === "Automatic") {
+			if (signing.team !== teamId) {
+				// Maybe the provided team is name such as "Telerik AD" and we need to convert it to CH******37
```
Mitko-Kerezov: What if the team is passed as an id instead of a name - we'd still try and look for it as if it was a name

lib/services/android-project-service.ts
```diff
@@ -517,12 +517,18 @@ export class AndroidProjectService extends projectServiceBaseLib.PlatformProject
 		if (this.$androidToolsInfo.getToolsInfo().androidHomeEnvVar) {
 			const gradlew = this.$hostInfo.isWindows ? "gradlew.bat" : "./gradlew";
 
+			const localArgs = [...gradleArgs];
+			if (this.$logger.getLevel() === "INFO") {
```
rosen-vladimirov: let's keep it this way, `$loggingLevels` has a different purpose. I'll handle the logging levels in a separate PR and will change this code with a constant in it.

lib/services/android-project-service.ts
```diff
@@ -517,12 +517,18 @@ export class AndroidProjectService extends projectServiceBaseLib.PlatformProject
...
+				localArgs.push("--quiet");
+				this.$logger.info(`${gradlew} ${localArgs.join(" ")}`);
```
rosen-vladimirov: why do we want to print the command that we execute in the info level?

lib/services/ios-project-service.ts
```diff
@@ -257,7 +275,15 @@ export class IOSProjectService extends projectServiceBaseLib.PlatformProjectServ
 <plist version="1.0">
 <dict>
     <key>method</key>
-    <string>${exportOptionsMethod}</string>
+	<string>${exportOptionsMethod}</string>`;
+	if (options && options.provision) {
+		plistTemplate += `    <key>provisioningProfiles</key>
+<dict>
```
rosen-vladimirov: this plist seems duplicate of the one used on line 234, can we have it on a single place and reuse it

lib/services/ios-project-service.ts
```diff
@@ -388,27 +413,75 @@ export class IOSProjectService extends projectServiceBaseLib.PlatformProjectServ
 			await this.setupSigningForDevice(projectRoot, buildConfig, projectData);
 		}
 
-		if (buildConfig && buildConfig.codeSignIdentity) {
-			args.push(`CODE_SIGN_IDENTITY=${buildConfig.codeSignIdentity}`);
-		}
+		// These are now set in the .pbxproj, avoid passing them as arguments!
+		// if (buildConfig && buildConfig.codeSignIdentity) {
```
rosen-vladimirov: I vote for deleting it

lib/services/ios-project-service.ts
```diff
@@ -388,27 +413,75 @@ export class IOSProjectService extends projectServiceBaseLib.PlatformProjectServ
...
-			args,
+	private async xcodebuild(args: string[], cwd: string, stdio: any = "inherit"): Promise<ISpawnResult> {
+		const localArgs = [...args];
+		const xcodeBuildVersion = await this.getXcodeVersion();
+		try {
+			if (helpers.versionCompare(xcodeBuildVersion, "9.0") >= 0) {
+				localArgs.push("-allowProvisioningUpdates");
+			}
+		} catch (e) {
+			console.warn("Failed to use xcodebuild with -allowProvisioningUpdates due to error: " + e);
```
rosen-vladimirov: this should be `$logger.warn`

lib/services/ios-project-service.ts
```diff
@@ -388,27 +413,75 @@ export class IOSProjectService extends projectServiceBaseLib.PlatformProjectServ
...
+		}
+		if (this.$logger.getLevel() === "INFO") {
```
rosen-vladimirov: Same as above:
let's keep it this way, `$loggingLevels` has a different purpose. I'll handle the logging levels in a separate PR and will change this code with a constant in it.

lib/services/ios-project-service.ts
```diff
@@ -388,27 +413,75 @@ export class IOSProjectService extends projectServiceBaseLib.PlatformProjectServ
...
+			localArgs.push("-quiet");
+			this.$logger.info(`xcodebuild ${localArgs.join(" ")}`);
```
rosen-vladimirov: and again, is it important for the user to see the command we are executing. Until now this was not shown, can we show it in trace level only?

lib/services/ios-provision-service.ts
```diff
@@ -1,11 +1,19 @@
...
+const months = ["Jan", "Feb", "Mar", "Apr", "May", "Jun", "Jul", "Aug", "Sep", "Oct", "Nov", "Dec"];
 function formatDate(date: Date): string {
 	return `${date.getDay()} ${months[date.getMonth()]} ${date.getFullYear()}`;
 }
 
+/**
+ * Formats the argument so it can easily be copied from the terminal and provided as value for an option.
+ * @param arg The string to format.
+ */
+function cmdEscape(arg: string): string {
```
rosen-vladimirov: Maybe [quoteString](https://github.com/telerik/mobile-cli-lib/blob/master/helpers.ts#L99-L105) can do the same work for you?

lib/services/ios-provision-service.ts
```diff
@@ -90,6 +103,31 @@ export class IOSProvisionService {
 
 		return { devices, match };
 	}
+
+	public async getDevelopmentTeams(): Promise<{ id: string, name: string }[]> {
```
PanayotCankov: This is used once, if I have to write it out in 2-3 places perhaps I will make it a separate interface. For now type inference works so well, this type is never written out anywhere else in the code.

lib/services/ios-provision-service.ts
```diff
@@ -90,6 +103,31 @@ export class IOSProvisionService {
...
+		const teams: { [teamName: string]: Set<string> } = {};
```
PanayotCankov: This is used locally and just once. I don't want to pollute the module with interfaces. It could be IDictionary<Set<string>> but then it would be even harder to figure out what the key would be.

lib/services/ios-project-service.ts
PanayotCankov: The time is tight, let's do this in a separate PR if we ever touch these two again.

lib/services/ios-project-service.ts
```diff
@@ -388,27 +413,75 @@ export class IOSProjectService extends projectServiceBaseLib.PlatformProjectServ
...
+		}
+		return await this.$childProcess.spawnFromEvent("xcodebuild",
+			localArgs,
 			"exit",
-			{ stdio: buildConfig.buildOutputStdio || "inherit", cwd: this.getPlatformData(projectData).projectRoot },
+			{ stdio: stdio || "inherit", cwd },
```
PanayotCankov: It is passed on few occasions from things that get here as args. If you pass null, it won't be converted to "inherit" I think. I don't have time to do excessive regression tests on this one.

lib/services/ios-project-service.ts
PanayotCankov: Yes. I think it is highly unlikely to have a team name matching the randomly generated team ids on the same machine. teamId probably should be just team like the provision switch is not provisionIdentifier or provisionUUID. Being able to pass "Telerik AD" for team however is really convenient.

lib/services/ios-project-service.ts
```diff
@@ -655,12 +711,16 @@ We will now place an empty obsolete compatability white screen LauncScreen.xib f
 	}
 
 	public async prepareProject(projectData: IProjectData, platformSpecificData: IPlatformSpecificData): Promise<void> {
-		let provision = platformSpecificData && platformSpecificData.provision;
+		const projectRoot = path.join(projectData.platformsDir, "ios");
```
PanayotCankov: `const projectRoot = path.join(projectData.platformsDir, $devicePlatformsConstants.iOS.toLowerCase());`
Much better!

PanayotCankov: Or can I use the IOS_PLATFORM_NAME

PanayotCankov: The "ios" here is the name of the directory in "platforms" where the native project is created. Do we have a constant in the project service or something that is not somewhat related with devices? As I look into the IDevicePlatformsConstants I wonder why there is a "Device" in the name?

lib/services/ios-project-service.ts
```diff
@@ -910,28 +970,44 @@ We will now place an empty obsolete compatability white screen LauncScreen.xib f
...
-				if (signing && signing.style === "Manual") {
-					for (let name in signing.configurations) {
-						let config = signing.configurations[name];
-						if (config.uuid !== provision && config.name !== provision) {
-							changesInfo.signingChanged = true;
-							break;
+
+				if (provision !== undefined) {
+					if (signing && signing.style === "Manual") {
+						for (let name in signing.configurations) {
```
PanayotCankov: Can we have a lint rule for these?

lib/services/android-project-service.ts
PanayotCankov: Otherwise nothing is printed. I wanted to print something that indicates that a native build had started. I am ok to go for just "Xcode build..." and "Gradle build...", should I change these?

lib/services/ios-project-service.ts
PanayotCankov: First Write, Second Copy, Third Refactor

lib/services/ios-project-service.ts
PanayotCankov: Same. Now nothing is shown.
xcodebuild will not output a single line if run with -quiet and no errors or warning were present. So far the user had seen 10 screens of output in place of this with all the flags and environment variables passed to Xcode, various subtools that compile storyboards or copy PNGs. All of it is now collapsed to this line. I think this line is important otherwise the CLI will stall for several seconds and the user won't know that the native build actually started.

Again I can change the message if you provide me a better one, but I think it is important to appear in the default logging mode when -quiet is passed.

lib/services/ios-provision-service.ts
PanayotCankov: No. It does cmdEscape. Its purpose is to make the string passable as argument to an option. Telerik A D can be escaped as "Telerik A D" or as Telerik\ A\ D and I wouldn't care less which one would be the result of the escape. On the other hand the escaping is flawed, it would fail on Telerik "AD" as the commas will go nuts with "Telerik "AD"" :( but I don't have time to figure out better escaping.

lib/services/ios-project-service.ts
```diff
@@ -93,13 +93,23 @@ export class IOSProjectService extends projectServiceBaseLib.PlatformProjectServ
...
-			await this.$iOSProvisionService.list(projectId);
+			await this.$iOSProvisionService.listProvisions(projectId);
 			this.$errors.failWithoutHelp("Please provide provisioning profile uuid or name with the --provision option.");
 			return false;
 		}
 
+		if (teamId === true) {
+			await this.$iOSProvisionService.listTeams();
+			this.$errors.failWithoutHelp("Please provide team id or team name with the --teamId options.");
+			return false;
```
PanayotCankov: Omg. It returns void. :D I will make it "never".

PanayotCankov: `failWithoutHelp(message: string, ...args: any[]): never;`
please use the `never` type for each method that terminate unconditionally the process or throw.

lib/services/ios-provision-service.ts
PanayotCankov: omg :D I clicked the link, you have cmdQuote :D I will use that one :)

lib/services/ios-project-service.ts
Mitko-Kerezov: You are indeed correct - passing `null` will not assign the default value from the method signature to the `stdio` variable. However, there's still no need to default twice so you can safely delete the default value from the method signature and leave the code here as it is

lib/services/ios-project-service.ts
Mitko-Kerezov: My bad, I didn't manage to express my concern correctly - according to the docs you've written for `--team-id` one can either pass a team name or a team id.
So what of the case where the user knows the team id and passes it as a value to `--team-id` - we'd still look for it and validate it as if it was a team name

lib/services/ios-project-service.ts
```diff
@@ -388,27 +413,75 @@ export class IOSProjectService extends projectServiceBaseLib.PlatformProjectServ
...
+		await this.createIpa(projectRoot, projectData, buildConfig);
+	}
 
-		// this.$logger.out("xcodebuild...");
-		await this.$childProcess.spawnFromEvent("xcodebuild",
-			args,
+	private async xcodebuild(args: string[], cwd: string, stdio: any = "inherit"): Promise<ISpawnResult> {
+		const localArgs = [...args];
+		const xcodeBuildVersion = await this.getXcodeVersion();
+		try {
```
PanayotCankov: I wouldn't do it if it didn't fail on CI once. I will change the message to:
```
Failed to detect whether -allowProvisioningUpdates can be used with your xcodebuild version due to error: 
```

lib/services/android-project-service.ts
rosen-vladimirov: I prefer using `Gradle build...`
In case you want to print dots, you can use:
```TypeScript
const action = () => this.spawn(gradlew, localArgs, childProcessOpts, spawnFromEventOptions);
this.$logger.printInfoMessageOnSameLine("Gradle build...");
await this.$progressIndicator.showProgressIndicator(action(), 2000);
```

lib/services/ios-project-service.ts
rosen-vladimirov: As above

In case you want to print dots, you can use:
```TypeScript
const action = () => this.$childProcess.spawnFromEvent("xcodebuild", ... );
this.$logger.printInfoMessageOnSameLine("Building application...");
await this.$progressIndicator.showProgressIndicator(action(), 2000);
```

lib/services/ios-project-service.ts
Mitko-Kerezov: Seems reasonable

lib/services/ios-project-service.ts
Mitko-Kerezov: I don't believe there is a lowercase `"ios"` global constant anywhere. Feel free to use `IOS_PLATFORM_NAME` from this file - it seems reasonable

lib/services/ios-project-service.ts
Mitko-Kerezov: We can, but enabling it leads to 1400 compile errors, so I advise against this in this PR

lib/services/ios-project-service.ts
PanayotCankov: I made a PR for 'never' but it is completely unrelated (I think) with this current PR:
https://github.com/telerik/mobile-cli-lib/pull/1006

PanayotCankov:
@rosen-vladimirov @Mitko-Kerezov 
I've addressed most of the important comments in a separate commit.
Could you take a look again?

lib/services/ios-project-service.ts
rosen-vladimirov: Nice one! Yes, it is unrelated

lib/services/ios-project-service.ts
rosen-vladimirov: Agreed to handle this in a separate PR

lib/services/ios-provision-service.ts
rosen-vladimirov: We have a scenario where we'll need interface, but let's discuss it personally and handle it in a separate PR once it is needed.

lib/services/ios-project-service.ts
PanayotCankov: The code here will do the following:
It will check if the arg passed to team-id matches the team in the Automatic signing style in Xcode. If it does, the native project and prepare are up-to-date.
If it does not - it will consider the team-id to be actually a team name, it will check for a team with name matching the provided team-id and if it finds one it will check if Xcode is configured to sign with that team's id.

This can fail if you have a team that has team id matching one of your other team's names, that's highly unlikely. If this raises as an issue we can split a team-id and team-name switches.

lib/services/android-project-service.ts
PanayotCankov: Doesn't this throw garbage progress indicator symbols around if the xcodebuild or gradle build actually output stuff? I've seen rogue npm progress bars printed at random.

PanayotCankov: This actually sounds reasonable I'll give it a try.

PanayotCankov: Ok we've agreed to use "Gradle build..." and "Xcode build...".
Bear in mind that we do 3 xcodebuild executions on clean "tns run ios" on device.

PR #3071, opened 2017-08-17
Add option to use chrome-devtools fronted from appspot

rosen-vladimirov:
URLs starting with `chrome-devtools://devtools/` cannot be opened in chrome with code.
However, we can open the urls in https://chrome-devtools-frontend.appspot.com/, so add new option in Public API that allows using the remote version.

Keep the current default behavior, i.e. for Android we are using the `bundled` DevTools, for iOS we are using a commit pointing to Chrome 55.
In case both `useBundledDevTools` and `useHttpUrl` are passed, the value of `useBundledDevTools` is disregarded.

Add unit tests for all of the cases.

### Add `devToolsCommit` to debug options (and public API)

Currently CLI uses specific SHA of dev tools, that corresponds to Chrome 55. However, we may need different version for Android or iOS, so add new option. This will also allow specifying it from public API, so external calls may pass the commit SHA.

### DIFF
--- a/lib/services/ios-debug-service.ts
+++ b/lib/services/ios-debug-service.ts
@@ -14,7 +14,7 @@ const inspectorAppName = "NativeScript Inspector.app";
 const inspectorNpmPackageName = "tns-ios-inspector";
 const inspectorUiDir = "WebInspectorUI/";
 
-class IOSDebugService extends DebugServiceBase implements IPlatformDebugService {
+export class IOSDebugService extends DebugServiceBase implements IPlatformDebugService {
 	private _lldbProcess: ChildProcess;
 	private _sockets: net.Socket[] = [];
 	private _childProcess: ChildProcess;
@@ -91,6 +91,15 @@ class IOSDebugService extends DebugServiceBase implements IPlatformDebugService 
 			await this.killProcess(this._childProcess);
 			this._childProcess = undefined;
 		}
+	}
+
+	protected getChromeDebugUrl(debugOptions: IDebugOptions, port: number): string {
+		const debugOpts = _.cloneDeep(debugOptions);
+		debugOpts.useBundledDevTools = debugOpts.useBundledDevTools === undefined ? false : debugOpts.useBundledDevTools;
+		debugOpts.useHttpUrl = debugOpts.useHttpUrl === undefined ? false : debugOpts.useHttpUrl;
+
+		const chromeDebugUrl = super.getChromeDebugUrl(debugOpts, port);
+		return chromeDebugUrl;
 	}
 
 	private async killProcess(childProcess: ChildProcess): Promise<void> {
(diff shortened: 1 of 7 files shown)


test/services/android-debug-service.ts
```diff
@@ -0,0 +1,161 @@
...
+			{
+				scenarioName: "useBundledDevTools is not passed and useHttpUrl is true",
+				debugOptions: {
+					useHttpUrl: true
+				},
+				expectedChromeUrl: `https://chrome-devtools-frontend.appspot.com/serve_file/@02e6bde1bbe34e43b309d4ef774b1168d25fd024/inspector.html?experiments=true&ws=localhost:${expectedPort}`,
+			},
+
+			// devToolsCommit tests
+
```
TsvetanMilanov: We don't need this new line here.

test/services/android-debug-service.ts
```diff
@@ -0,0 +1,161 @@
...
+				},
+				expectedChromeUrl: `chrome-devtools://devtools/bundled/inspector.html?experiments=true&ws=localhost:${expectedPort}`,
+			},
+			{
+				scenarioName: "useBundledDevTools is true and useHttpUrl is true",
+				debugOptions: {
+					useBundledDevTools: true,
+					useHttpUrl: true
+				},
+				expectedChromeUrl: `https://chrome-devtools-frontend.appspot.com/serve_file/@02e6bde1bbe34e43b309d4ef774b1168d25fd024/inspector.html?experiments=true&ws=localhost:${expectedPort}`,
```
TsvetanMilanov: You can extract the chrome devtools default commit to a constant. This way we won't use our favorite `Ctrl + Shift + H` 😄

lib/services/ios-debug-service.ts
```diff
@@ -93,6 +93,15 @@ class IOSDebugService extends DebugServiceBase implements IPlatformDebugService
 		}
 	}
 
+	protected getChromeDebugUrl(debugOptions: IDebugOptions, port: number): string {
+		const debugOpts = _.cloneDeep(debugOptions);
+		debugOpts.useBundledDevTools = debugOpts.useBundledDevTools === undefined ? false : debugOpts.useBundledDevTools;
+		debugOpts.useHttpUrl = debugOpts.useHttpUrl === undefined ? false : debugOpts.useHttpUrl;
```
rosen-vladimirov: The idea is that we may have many different cases for the debug options in both platforms. That's why I prefer using two separate methods. Also I really want to assure that when ios-debug-service is used, we are using correct options, so testing only the base class is not an option.
I'll move this line to the `super` method, but I'll keep using the same logic.

test/services/android-debug-service.ts
```diff
@@ -0,0 +1,161 @@
...
+			{
+				scenarioName: "devToolsCommit defaults to 02e6bde1bbe34e43b309d4ef774b1168d25fd024 when useBundledDevTools is set to false",
+				debugOptions: {
+					useBundledDevTools: false
+				},
+				expectedChromeUrl: `chrome-devtools://devtools/remote/serve_file/@02e6bde1bbe34e43b309d4ef774b1168d25fd024/inspector.html?experiments=true&ws=localhost:${expectedPort}`,
+			},
+			{
+				scenarioName: "devToolsCommit is disregarded when useBundledDevTools is not passed",
+				debugOptions: {},
```
rosen-vladimirov: The idea was to keep a separate test case for this scenario. This way, we may change the test you've mentioned in the future, but this one should be kept this way as it's clear what is its purpose.

rosen-vladimirov:
Merging as the failed test is caused by a problem with iOS Simulator on the machine.

Issue #3073, opened 2017-08-21
Saving two platform-specific files during livesync causes an exception


Mitko-Kerezov:
### Tell us about the problem
Saving two platform-specific files in `app/App_Resources` simultaneously after `tns run` causes an exception.

### Which platform(s) does your issue occur on?
Both

### Please provide the following version numbers that your issue occurs with:
- CLI: 3.1.3

### Please tell us how to recreate the issue in as much detail as possible.
1. Create a new project
1. Execute `tns run android` for that project
1. Open the project with an editor, make a minor change in `app/App_Resources/Android/AndroidManifest.xml` and in `app/App_Resources/iOS/build.xcconfig`. Save the files simultaneously.
1. CLI errors with `Unable to apply changes for device: <deviceId>. Error is: Cannot read property 'match' of null.`

PR #3074, opened 2017-08-21
Fix simultaneous livesync error

Mitko-Kerezov:
Whenever two platform-specific files from `App_Resources` which are of different platforms need to be synced, one of them gets filtered out because it is for the wrong platform. However, instead of it being filtered out its path remains `null`. This causes a null-reference exception when trying to define whether the file should be synced or not.

Fix for https://github.com/NativeScript/nativescript-cli/issues/3073

Merge after https://github.com/telerik/mobile-cli-lib/pull/1003


rosen-vladimirov:
run ci

Mitko-Kerezov:
Merging this with one failing test - `run_ios_tests.RunIOSSimulatorTests.test_001_tns_run_ios_js_css_xml`
The test fails due to infrastructure problems

PR #3075, opened 2017-08-21
Fix native platform add after cloud build.

TomaNikolov:
When we prepare the platform for cloud build we skip the native prepare.
After this partial prepare, the platform directory contains only .js files.
In order to make a local build we should add the native part of the platform
which fails with `ENOTEMPTY: directory not empty, rename`.


TomaNikolov:
run ci

rosen-vladimirov:
run ci

PR #3077, opened 2017-08-22
Fix stopLiveSync method

rosen-vladimirov:
### Fix stopLiveSync method
Currently stopLiveSync method does not emit correct events in some cases. For example, when there's only one attached device and its identifier is passed to `stopLiveSync` method, we do not emit `liveSyncStopped` event. Fix the code and add unit tests to assure `liveSyncStopped` event is raised correctly.

### Expose getLiveSyncDeviceDescriptors method

Add new public method `getLiveSyncDeviceDescriptors` to `liveSyncService` - its purpose is to get information for current LiveSync operation.

### DIFF
--- a/lib/services/livesync/livesync-service.ts
+++ b/lib/services/livesync/livesync-service.ts
@@ -50,7 +50,7 @@ export class LiveSyncService extends EventEmitter implements ILiveSyncService {
 
 			const deviceIdentifiersToRemove = deviceIdentifiers || _.map(liveSyncProcessInfo.deviceDescriptors, d => d.identifier);
 
-			const removedDeviceIdentifiers = _.remove(liveSyncProcessInfo.deviceDescriptors, descriptor => _.indexOf(deviceIdentifiersToRemove, descriptor.identifier) !== -1)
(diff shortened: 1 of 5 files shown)


lib/services/livesync/livesync-service.ts
```diff
@@ -48,12 +48,10 @@ export class LiveSyncService extends EventEmitter implements ILiveSyncService {
...
 			const shouldAwaitPendingOperation = !stopOptions || stopOptions.shouldAwaitAllActions;
 
-			let removedDeviceIdentifiers: string[] = deviceIdentifiers || [];
+			const deviceIdentifiersToRemove = deviceIdentifiers || _.map(liveSyncProcessInfo.deviceDescriptors, d => d.identifier);
 
-			_.each(deviceIdentifiers, deviceId => {
-				removedDeviceIdentifiers = _.remove(liveSyncProcessInfo.deviceDescriptors, descriptor => descriptor.identifier === deviceId)
-					.map(deviceDescriptor => deviceDescriptor.identifier);
-			});
+			const removedDeviceIdentifiers = _.remove(liveSyncProcessInfo.deviceDescriptors, descriptor => _.indexOf(deviceIdentifiersToRemove, descriptor.identifier) !== -1)
```
Mitko-Kerezov: You can use `_.includes` instead of `_.indexOf` for slightly better readability

PR #3079, opened 2017-08-22
Improve iOS log filter regex.

TomaNikolov:
Our regex doesn't match `console.log` from app deployed on a iOS simulator. The reason that we have logs from this kind of application is that the cli set PID to the installed application. However, when the cli is required, the app is installed with `justlaunch` options the PID is not set which cause missing logging.

Merge after https://github.com/telerik/mobile-cli-lib/pull/1005 and update sha


rosen-vladimirov:
NOTE: Same can be observed in case when liveSync is running, but application is restarted manually on device/simulator.

rosen-vladimirov:
run ci

dtopuzov:
run ci

PR #3086, opened 2017-08-24
Fix unit-tests on Windows and Linux

rosen-vladimirov:
Due to recent changes in `ios-project-service`, some unit tests cannot work on Windows and Linux, so ignore them in case the OS is not macOS.

### DIFF
--- a/test/ios-project-service.ts
+++ b/test/ios-project-service.ts
@@ -199,16 +199,21 @@ describe("iOSProjectService", () => {
 				}
 			};
 		}
-		it("by default exports xcodearchive to platforms/ios/build/archive/<projname>.xcarchive", async () => {
-			let setup = await setupArchive();
-			await setup.run();
-			setup.assert();
-		});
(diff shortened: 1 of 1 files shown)


test/ios-project-service.ts
```diff
@@ -199,16 +199,21 @@ describe("iOSProjectService", () => {
...
-			await setup.run();
-			setup.assert();
-		});
-		it("can pass archivePath to xcodebuild -archivePath", async () => {
-			let setup = await setupArchive({ archivePath: "myarchive.xcarchive" });
-			await setup.run();
-			setup.assert();
-		});
+
+		if (require("os").platform() !== "darwin") {
```
nadyaA: Duplicated code?

rosen-vladimirov: in many places ;) 
These are tests, so I'll live with it for the moment. In the future we may introduce a property of `it` and `define` and handle it automatically.

PR #3088, opened 2017-08-24
Improve analytics

rosen-vladimirov:
We have a requirement to track information in several different providers. However, this may slow down CLI's operations as each track (at least for some of the providers) may last around 100-300ms.
In order to track in multiple analytics provider without slowing down CLI commands, move the tracking to a new process. The process is detached from CLI, so when CLI finishes its work it will exit, while the child process will remain alive until it sends all required information. The new process is called `analytics-broker-process`, it is detached from CLI, but it there's IPC channel between them, so CLI can send information to be tracked. Just before CLI dies, it will send "finish" message to the `broker` process and the IPC channel will be closed.

In CLI implement a new `analytics-service`, so when the methods `track`, `trackFeature` and `trackException` are called, instead of tracking directly in CLI's process (i.e. the old implementation), to send information to the broker process.
The only exception is tracking of the answer if user allows us to send anonymous statistics. As this action requires user interaction, it is executed in CLI's process, a new eqatec analytics monitor is started and the answer is sent to it. After that the analytics is stopped. In case the answer is "yes" and there is more information that we want to track in the current process, it will be send to broker and tracked in all providers. In case the answer is "no", no other data will be tracked.

The broker process is started from analytics process once per process, i.e. on each CLI process (which have something to track) there's a corresponding `analytics-broker-process`. Once it is initialized, it sends BrokerReadyToReceive message to CLI and CLI knows that it can start sending tracking information. Initialization phase is creating a new instance of AnalyticsBroker and attaching handlers for `message` and `disconnect` events. The AnalyitcsBroker itself has a single purpose - sends tracking data to all available prociders. Currently the only provider is `eqatecAnalyticsProvider`. Once CLI sends "finish" message (or `analytics-broker-process` receives `disconnect` event), the process sends information to all providers that there's nothing more to be tracked. When all providers finish their work, they should disocnnect themselves from the broker process and it will exit gracefully.

`EqatecAnalyticsProvider` is used to track features and exceptions in Progress Telerik Analytics. Our requirement is to track features in two projects (same data should be send to both of them) and exceptions in another project. Each of them should be in a separate child process, so failure in one of them will not reflect the others. So the provider starts all required processes. When broker sends information to the provider, it sends it to all required child processes, for example if a feature should be tracked, the provider will send information for it to the two processes used for feature tracking.

Each eqatec child process is a new Node.js  process runnign `eqatec-analytics-process`. Again - this is a detached process. Once it is initialized, it sends "EqatecAnalyticsReadyToReceive" message to the `EqatecAnalyitcsProvider` and it knows it can send information to the new child process. The initialization phase is creation of new instance of `EqatectAnalyticsService`, attaching to "message" and "disconnect" events. Once `EqatecAnalyitcsProvider` receives `EqatecAnalyticsReadyToReceive` message, it sends initialization data for the eqatec monitor. It includes the analytics project API key, user id, current sessions count, etc. The information is persisted in a file on user's machine, so this file should be updated from a single process. As we might have 3 eqatec analytics processes, the provider is the only place where the file should be read and updated.

When `eqatec-analytics-process` receives initialization data, it creates a new instance of the `EqatectAnalyticsService` class with it. However a monitor is not started yet. It will be started when there's something to be tracked, i.e. once `eqatec-analytics-process` receives feature or exception to be tracked.

When `eqatec-analytics-process` receives "finish" message, it awaits all pending requests and after that disconnets itself from the parent. In case this operation fails, the process just exits, which will allow the parent to exit gracefully.

Remove the public `startEqatecMonitor` method from API - it has not been used.

A simple scheme of the processes is shown below:
analytics_1

## UPDATE
After latest discussion and commits, all EqatecAnalytics are ran in the analytics-broker-process directly. Also AcceptTrackFeatureUsage is moved to this process, so no tracking should happen in CLI itself.
untitled diagram

### Add tracking to Google Analytics

Add new methods to track information in Google Analytics(GA). As the data that we want to track is different than the one we track in Eqatec Analytics, call the new methods whenever we want to send some information to GA.
Change the broker process to send information to Eqatec Analytics when we want to track feature or exception and to GA only when the new method for tracking there is called.

In Google Analytics (GA) we track two types of data - pages (called pageviews) and events. Each command is tracked as a page - we only track the name of the command as a page:
* `tns build android` will be tracked as visited page - `build android`
* `tns run` will be tracked as visited page - `run`

When we want to track some additional actions that happen during execution of each command, we send events to GA. Each event has the following data:
* category - currently we do not have a specific category for each action, so it is hardcoded to "CLI".
* action - this is the real action that will be tracked, for example `Build`, `LiveSync`, `Deploy`, etc.
* label - contains specific inforamation for the current action, for example what is the platform, device type, etc. In many cases we send labels with a lot of information, but as the value must be string, we use `_` as a separator. For example:
  * Action `Build` may have the following label: `Android_Debug_Incremental`. This is `<platform>_<build configuration>_<build type>`.
  * Action `LiveSync` may have the following label: `Android_Emulator_5.1`. This is `<platform>_<device type>_<device os version>`.

#### Custom dimensions
For each additional data may be send to GA via Custom Dimensions. More about them can be found [here](https://support.google.com/analytics/answer/2709828). We are using several custom dimensions, most of them have hit scope, i.e. their values are saved for the current hit (event or page), but may be changed for the next time. The only exclusion is `session id` - it has session scope, so whenever it is sent, the new value overwrites the values of the "session id" property in all hits of the current session.
One interesting custom dimension is `Client` - it will be set to the value of `--analyticsClient` in case it is passed. In case not, we'll set it to "CLI" when terminal is interactive and "Unknown" when terminal is not interactive.

#### Sessions - what is session for GA
In Eqatec Analytics we call `stopMonitor` at the end of each command. So the session there is one command. In GA we've decided to keep the default way of measuring a session - a session is stopped when there are 30 inactive minutes, i.e. no data is sent from 30 minutes. This allows us to investigate the behavior and the flow of commands a user is executing in a single session (during the day for example).

#### Client ID and User ID:
Currently we do not send User ID to GA, we use Client ID instead as sending User IDs requires different logic and agreements. For us all users are anonymous.

#### User-Agent
When sending requests to GA, we need to set correct User-Agent, so GA will be able to understand the Operating System. For each OS the string is painfully specific... Current implementation works fine for Windows, Linux and macOS.

#### Implementation
In order to send data to GA, we are using a `universal-analytics` npm package. We cannot use Google's default libraries (`analytics.js`, `gtag.js`) as they are written for browser. So the only way to track data is to use the Measurement Protocol (directly send http requests). However, `universal-analytics` package provides a very good abstraction over the Measurement Protocol, so we are using it instead of implementing each call.

More information about Measurement protocol:
https://developers.google.com/analytics/devguides/collection/protocol/v1/parameters


lib/services/analytics/eqatec-analytics.ts
```diff
@@ -0,0 +1,52 @@
...
+	public async trackInformation(data: IFeatureTrackingInformation): Promise<void> {
+		try {
+			await this.startEqatecMonitor(this.analyticsSettings);
+
+			if (this._eqatecMonitor) {
+				this._eqatecMonitor.trackFeature(`${data.featureName}.${data.featureValue}`);
+				await this.waitForSending();
+			}
+		} catch (e) {
+			this.$logger.trace("Analytics exception: '%s'", e.toString());
```
Mitko-Kerezov: I don't think `e.toString()` is relevant here - maybe just `this.$logger.trace("Analytics exception: ", e);`?

lib/services/analytics/eqatec-analytics.ts
```diff
@@ -0,0 +1,52 @@
...
+	public async trackError(data: IExceptionsTrackingInformation): Promise<void> {
+		try {
+			await this.startEqatecMonitor(this.analyticsSettings);
+
+			if (this._eqatecMonitor) {
+				this._eqatecMonitor.trackException(data.exception, data.message);
+				await this.waitForSending();
+			}
+		} catch (e) {
+			this.$logger.trace("Analytics exception: '%s'", e.toString());
```
Mitko-Kerezov: Same remark about the `e.toString()` - calling `toString()` might lead to `[object Object]`

lib/services/analytics/analytics-service.ts
```diff
@@ -0,0 +1,148 @@
...
+
+		}
+
+	}
+
+	private async sendExceptionForTracking(exception: Error, message: string): Promise<void> {
+		await this.initAnalyticsStatuses();
+
+		if (this.analyticsStatuses[this.$staticConfig.ERROR_REPORT_SETTING_NAME] === AnalyticsStatus.enabled) {
+			const broker = await this.getAnalyticsBroker();
```
Mitko-Kerezov: You can extract this functionality into a separate private method:
```TypeScript
private async sendMessageToBroker(message: any): Promise<void> {
    const broker = await this.getAnalyticsBroker();
    return new Promise<void>((resolve, reject) => {
         broker.send(message, () => { resolve(); });
    };
}
```

lib/services/analytics/analytics-broker.ts
```diff
@@ -0,0 +1,39 @@
...
+	@cache()
+	private get $eqatecAnalyticsProvider(): IAnalyticsProvider {
+		return this.$injector.resolve("eqatecAnalyticsProvider", { pathToBootstrap: this.pathToBootstrap });
+	}
+
+	constructor(private pathToBootstrap: string,
+		private $injector: IInjector) { }
+
+	private get analyticsProviders(): IAnalyticsProvider[] {
+		return [
```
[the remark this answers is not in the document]
rosen-vladimirov: easier debugging, also I'll add more providers soon

lib/services/analytics/analytics-broker.ts
```diff
@@ -0,0 +1,39 @@
...
+					await provider.trackException(<IExceptionsTrackingInformation>trackInfo);
+					break;
+				case TrackingTypes.Feature:
+					await provider.trackFeature(<IFeatureTrackingInformation>trackInfo);
+					break;
+				case TrackingTypes.Finish:
+					await provider.finishTracking();
+					break;
+				default:
+					throw new Error(`Invalid tracking type: ${trackInfo.type}`);
```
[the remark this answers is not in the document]
rosen-vladimirov: You are correct, but I'm not sure what to do here. Maybe we can try sending the incorrect information to parent (CLI) in case it is still alive.

dtopuzov:
run ci

lib/services/analytics/analytics-service.ts
```diff
@@ -0,0 +1,123 @@
...
+			this.$processService.attachToProcessExitSignals(this, () => {
+				broker.send({
+					type: TrackingTypes.Finish
+				});
+			});
+		});
+	}
+
+	private async sendDataForTracking(featureName: string, featureValue: string): Promise<void> {
+		await this.initAnalyticsStatuses();
```
Mitko-Kerezov: `sendDataForTracking` and `sendExceptionForTracking`'s implementations could be extracted into a single method:
```TypeScript
private async sendInfoForTracking(trackingInfo: IExceptionsTrackingInformation|IFeatureTrackingInformation, settingName: string): Promise<void> {
    await this.initAnalyticsStatuses();
    
    if (this.analyticsStatuses[settingName] === AnalyticsStatus.enabled) {
         return this.sendMessageToBroker(trackingInfo);
    }
}
```

lib/services/analytics/analytics-service.ts
```diff
@@ -0,0 +1,123 @@
...
+					clearTimeout(timeoutId);
+
+					if (!isSettled) {
+						isSettled = true;
+						resolve(broker);
+					}
+				}
+			});
+
+			this.$processService.attachToProcessExitSignals(this, () => {
```
Plamen5kov: In case broker process fails we would try to send it `TrackingTypes.Finish` if CLI has not crashed.
How about we `attachToProcessExitSignals` after `AnalyticsMessages.BrokerReadyToReceive`?

Plamen5kov:
when commands:
`tns usage-reporting enable` or `tns usage-reporting disable` are ran process hangs and can't be escaped with `ctrl + c`

rosen-vladimirov:
run ci

rosen-vladimirov:
@Plamen5kov thanks for pointing this, I've fixed the issue.

lib/services/analytics/google-analytics-provider.ts
```diff
@@ -0,0 +1,120 @@
...
+		}
+
+		const userAgent = `tnsCli/${this.$staticConfig.version} (${osString}; ${this.$osInfo.arch()})`;
+
+		return userAgent;
+	}
+
+	private getMacOSReleaseVersion(osRelease: string): string {
+		// https://en.wikipedia.org/wiki/Darwin_(operating_system)#Release_history
+		// Each macOS version is labeled 10.<version>, where it looks like <versions> is taken from the major version returned by os.release() (16.x.x for example) and substituting 4 from it.
```
Mitko-Kerezov: I think `substituting` here should be `subtracting` 💭

PR #3094, opened 2017-08-29
Debug improvements

Mitko-Kerezov:
Includes but is not limited to:
* `enableDebugging` - each promise is now resolved with an object
* added docs for `debuggerAttached` event
* added `debuggerDetached` event and docs for it

Ping @rosen-vladimirov @TsvetanMilanov @nadyaA

### DIFF
--- a/lib/services/debug-service.ts
+++ b/lib/services/debug-service.ts
@@ -1,4 +1,5 @@
 import { platform } from "os";
+import { parse } from "url";
 import { EventEmitter } from "events";
 import { CONNECTION_ERROR_EVENT_NAME, DebugCommandErrors } from "../constants";
 import { CONNECTED_STATUS } from "../common/constants";
@@ -14,7 +15,7 @@ export class DebugService extends EventEmitter implements IDebugService {
 		this._platformDebugServices = {};
 	}
 
-	public async debug(debugData: IDebugData, options: IDebugOptions): Promise<string> {
+	public async debug(debugData: IDebugData, options: IDebugOptions): Promise<IDebugInformation> {
 		const device = this.$devicesService.getDeviceByIdentifier(debugData.deviceIdentifier);
 
 		if (!device) {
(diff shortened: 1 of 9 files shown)


dtopuzov:
run ci

Mitko-Kerezov:
run ci

test/services/debug-service.ts
```diff
@@ -214,7 +215,8 @@ describe("debugService", () => {
 					const debugData = getDebugData();
 					const url = await debugService.debug(debugData, null);
 
-					assert.deepEqual(url, fakeChromeDebugUrl);
+					assert.deepEqual(url.url, fakeChromeDebugUrl);
```
rosen-vladimirov: why not asserting the whole object? Currently the object may have additional properties and the test will pass. I prefer asserting equality of the whole object, so in case we add new properties, we'll have to add/update tests for them.

lib/services/debug-service.ts
```diff
@@ -92,6 +93,16 @@ export class DebugService extends EventEmitter implements IDebugService {
...
 		platformDebugService.on(CONNECTION_ERROR_EVENT_NAME, connectionErrorHandler);
 	}
+
+	private getDebugInformation(fullUrl: string): IDebugInformation {
+		const parseQueryString = true;
+		const wsQueryParam = parse(fullUrl, parseQueryString).query.ws;
+		const hostPortSplit = wsQueryParam && wsQueryParam.split(":");
+		return {
+			url: fullUrl,
+			port: hostPortSplit && +hostPortSplit[1]
```
rosen-vladimirov: in case the wsQueryParam does not contain `:`, you'll end up with port = NaN, is this okay?

Mitko-Kerezov: I think so - in case the `ws` query param isn't in the correct format I can't really tell the port so a `null`-like value seems okay

rosen-vladimirov:
run ci

PR #3103, opened 2017-09-05
Only return debugInfo if url is passed

Mitko-Kerezov:
Whenever iOS inspector is used in favor of chrome, `null` is passed to `getDebugInformation` which causes `url.parse` to fail.

Add a `null` check in order to safeguard the method.


Mitko-Kerezov:
run ci

Mitko-Kerezov:
run ci

PR #3109, opened 2017-09-07
Minor LiveSync+Debug fixes

Mitko-Kerezov:
Fixes:
1. When debugging is enabled no fast sync should be performed - we should always full sync
1. We shouldn't swallow the exception from the first time when trying to attach a debugger the second time


lib/services/livesync/livesync-service.ts
```diff
@@ -166,6 +166,7 @@ export class LiveSyncService extends EventEmitter implements IDebugLiveSyncServi
 		} catch (err) {
 			this.$logger.trace("Could not stop application during debug livesync. Will try to restart app instead.", err);
 			if ((err.message || err) === "Could not find developer disk image") {
+				liveSyncResultInfo.isFullSync = true;
```
rosen-vladimirov: can you set comment here why you are setting isFullSync to true? It does not make any sense to do it.

Issue #3040, opened 2017-08-04
No option to have different app identifiers for iOS and Android and livesync working


KristianDD:
### Tell us about the problem
Currently livesync woks with the application identifier specified inside "package.json". In order to have different application identifiers for iOS and Android currently, the user has to edit the "applicationId" key inside "app.gradle" file. This change will cause the livesync functionality to not work properly. 

A solution could be to add the option for "id" key inside "package.json" to have values for iOS and Andoid. Something like:
```JavaScript
"nativescript": {
    "id": {
        "ios": "{{appID1}}" ,
        "android": "{{appID2}}"
    }
  }
```
### Which platform(s) does your issue occur on?
iOS/Android/Both

### Please provide the following version numbers that your issue occurs with:
- CLI: 3.0.1
- Runtime(s): 3.0.0
- Plugin(s): 3.0.0

### Please tell us how to recreate the issue in as much detail as possible.
Change the application Id in `/App_Resources/android/app.gradle`

android {  
  defaultConfig {  
    generatedDensities = []
    applicationId = "{{appID1}}" 

Change the application Id in `package.json` to something different.

  "nativescript": {
    "id": "{{appID2}}",

### Solution
1. Set in `package.json`
```JavaScript
"nativescript": {
    "id": {
        "ios": "{{appID1}}" ,
        "android": "{{appID2}}"
    }
  }
```
2. If you are using old project, remove the `applicationId` key from `{AppResources}/Android/app.gradle`
3. If you have opened the the project in NS Sidekick remove the `{AppResources}/Android/settings.json` file
4. Run the build

KristianDD:
Related to #2825

frapaa:
Is it possible to also make this available during a build step? I have a case where the same app will be built and deployed for multiple customers on their Apple and Google accounts, so they all need a unique app id. If we change this in the files (package.json for iOS and app.gradle + shortcuts.xml for Android) as part of the build process, the release build will have uncommitted changes and be "dirty".

Gamadril:
```
"nativescript": {
    "id": {
        "ios": "{{appID1}}" ,
        "android": "{{appID2}}"
    }
  }
```
This really useful feature should be documented in https://docs.nativescript.org/core-concepts/project-structure
It really helps if you must have "-" in your iOS app ID, but are not allowed to use it in android app ID...

PR #3120, opened 2017-09-13
Kddimitrov/add platform application identifier setting

KristianDD:
DO NOT MERGE

### Description
While still overwrite of applicationId from app.gradle is possible, user is able to specify a platform specific identifier inside package.json. This addresses problems related to livesync not working when applicationId is changed inside app.gradle.

### Does your commit message include the wording below to reference a specific issue in this repo?

Fixes https://github.com/NativeScript/nativescript-cli/issues/3040.

### Related Pull Requests
https://github.com/NativeScript/android-runtime/pull/848

### DIFF
--- a/lib/definitions/livesync.d.ts
+++ b/lib/definitions/livesync.d.ts
@@ -187,7 +187,7 @@ interface ILiveSyncEventData {
 	deviceIdentifier: string,
 	applicationIdentifier?: string,
 	projectDir: string,
-	syncedFiles?: Object
+	syncedFiles?: string[],
 	error? : Error,
 	notification?: string,
 	isFullSync?: boolean
--- a/lib/definitions/livesync.d.ts
+++ b/lib/definitions/livesync.d.ts
@@ -183,6 +183,16 @@ interface IHasUseHotModuleReloadOption {
 	useHotModuleReload: boolean;
 }
 
+interface ILiveSyncEventData {
+	deviceIdentifier: string,
+	applicationIdentifier?: string,
+	projectDir: string,
+	syncedFiles?: Object
+	error? : Error,
+	notification?: string,
+	isFullSync?: boolean
+}
+
 interface ILatestAppPackageInstalledSettings extends IDictionary<IDictionary<boolean>> { /* empty */ }
 
 interface IIsEmulator {
--- a/lib/project-data.ts
+++ b/lib/project-data.ts
@@ -2,6 +2,7 @@ import * as constants from "./constants";
 import * as path from "path";
 import { parseJson } from "./common/helpers";
 import { EOL } from "os";
+import { cache } from "./common/decorators";
 
 interface IProjectType {
 	type: string;
@@ -35,7 +36,17 @@ export class ProjectData implements IProjectData {
 	public projectDir: string;
 	public platformsDir: string;
 	public projectFilePath: string;
-	public projectId: string;
+	public projectIdentifiers: Mobile.IProjectIdentifier;
+	get projectId(): string {
+		this.warnProjectId();
+		return this.projectIdentifiers.ios;
+	}
+	//just in case hook/extension modifies it.
+	set projectId(identifier: string) {
+		this.warnProjectId();
+		this.projectIdentifiers.ios = identifier;
+		this.projectIdentifiers.android = identifier;
+	}
 	public projectName: string;
 	public nsConfig: INsConfig;
 	public appDirectoryPath: string;
@@ -108,7 +119,7 @@ export class ProjectData implements IProjectData {
 			this.projectName = this.$projectHelper.sanitizeName(path.basename(projectDir));
 			this.platformsDir = path.join(projectDir, constants.PLATFORMS_DIR_NAME);
 			this.projectFilePath = projectFilePath;
-			this.projectId = nsData.id;
+			this.projectIdentifiers = this.initializeProjectIdentifiers(nsData.id);
 			this.dependencies = packageJsonData.dependencies;
 			this.devDependencies = packageJsonData.devDependencies;
 			this.projectType = this.getProjectType();
@@ -206,6 +217,25 @@ export class ProjectData implements IProjectData {
 		return path.resolve(projectDir, pathToResolve);
 	}
 
+	private initializeProjectIdentifiers(data: string | Mobile.IProjectIdentifier): Mobile.IProjectIdentifier {
+		let identifier: Mobile.IProjectIdentifier;
+		data = data || "";
+
+		if (typeof data === "string") {
+			identifier = {
+				android: data,
+				ios: data
+			};
+		} else {
+			identifier = {
+				android: data.android || "",
+				ios: data.ios || ""
+			};
+		}
+
+		return identifier;
+	}
+
 	private getProjectType(): string {
 		let detectedProjectType = _.find(ProjectData.PROJECT_TYPES, (projectType) => projectType.isDefaultProjectType).type;
 
@@ -219,6 +249,11 @@ export class ProjectData implements IProjectData {
 		});
 
 		return detectedProjectType;
+	}
+
+	@cache()
+	private warnProjectId(): void {
+		this.$logger.warnWithLabel("IProjectData.projectId is deprecated. Please use IProjectData.projectIdentifiers[platform].");
 	}
 }
 $injector.register("projectData", ProjectData);
--- a/lib/services/android-project-service.ts
+++ b/lib/services/android-project-service.ts
@@ -120,7 +120,7 @@ export class AndroidProjectService extends projectServiceBaseLib.PlatformProject
 	}
 
 	public async validate(projectData: IProjectData, options: IOptions, notConfiguredEnvOptions?: INotConfiguredEnvOptions): Promise<IValidatePlatformOutput> {
-		this.validatePackageName(projectData.projectId);
+		this.validatePackageName(projectData.projectIdentifiers.android);
 		this.validateProjectName(projectData.projectName);
 
 		const checkEnvironmentRequirementsOutput = await this.$platformEnvironmentRequirements.checkEnvironmentRequirements({
@@ -269,16 +269,17 @@ export class AndroidProjectService extends projectServiceBaseLib.PlatformProject
 			// will replace applicationId in app/App_Resources/Android/app.gradle if it has not been edited by the user
 			const appGradleContent = this.$fs.readText(projectData.appGradlePath);
 			if (appGradleContent.indexOf(constants.PACKAGE_PLACEHOLDER_NAME) !== -1) {
-				shell.sed('-i', new RegExp(constants.PACKAGE_PLACEHOLDER_NAME), projectData.projectId, projectData.appGradlePath);
+				//TODO: For compatibility with old templates. Once all templates are updated should delete.
+				shell.sed('-i', new RegExp(constants.PACKAGE_PLACEHOLDER_NAME), projectData.projectIdentifiers.android, projectData.appGradlePath);
 			}
 		} catch (e) {
-			this.$logger.warn(`\n${e}.\nCheck if you're using an outdated template and update it.`);
+			this.$logger.trace(`Templates updated and no need for replace in app.gradle.`);
 		}
 	}
 
 	public interpolateConfigurationFile(projectData: IProjectData, platformSpecificData: IPlatformSpecificData): void {
 		const manifestPath = this.getPlatformData(projectData).configurationFilePath;
-		shell.sed('-i', /__PACKAGE__/, projectData.projectId, manifestPath);
+		shell.sed('-i', /__PACKAGE__/, projectData.projectIdentifiers.android, manifestPath);
 		if (this.$androidToolsInfo.getToolsInfo().androidHomeEnvVar) {
 			const sdk = (platformSpecificData && platformSpecificData.sdk) || (this.$androidToolsInfo.getToolsInfo().compileSdkVersion || "").toString();
 			shell.sed('-i', /__APILEVEL__/, sdk, manifestPath);
@@ -287,8 +288,8 @@ export class AndroidProjectService extends projectServiceBaseLib.PlatformProject
 
 	private getProjectNameFromId(projectData: IProjectData): string {
 		let id: string;
-		if (projectData && projectData.projectId) {
-			const idParts = projectData.projectId.split(".");
+		if (projectData && projectData.projectIdentifiers && projectData.projectIdentifiers.android) {
+			const idParts = projectData.projectIdentifiers.android.split(".");
 			id = idParts[idParts.length - 1];
 		}
 
@@ -532,7 +533,7 @@ export class AndroidProjectService extends projectServiceBaseLib.PlatformProject
 			const filesForInterpolation = this.$fs.enumerateFilesInDirectorySync(resourcesDestinationDirectoryPath, file => this.$fs.getFsStats(file).isDirectory() || path.extname(file) === constants.XML_FILE_EXTENSION) || [];
 			for (const file of filesForInterpolation) {
 				this.$logger.trace(`Interpolate data for plugin file: ${file}`);
-				await this.$pluginVariablesService.interpolate(pluginData, file, projectData);
+				await this.$pluginVariablesService.interpolate(pluginData, file, projectData.projectDir, projectData.projectIdentifiers.android);
 			}
 		}
 
@@ -642,7 +643,7 @@ export class AndroidProjectService extends projectServiceBaseLib.PlatformProject
 
 	public async cleanDeviceTempFolder(deviceIdentifier: string, projectData: IProjectData): Promise<void> {
 		const adb = this.$injector.resolve(DeviceAndroidDebugBridge, { identifier: deviceIdentifier });
-		const deviceRootPath = `${LiveSyncPaths.ANDROID_TMP_DIR_NAME}/${projectData.projectId}`;
+		const deviceRootPath = `${LiveSyncPaths.ANDROID_TMP_DIR_NAME}/${projectData.projectIdentifiers.android}`;
 		await adb.executeShellCommand(["rm", "-rf", deviceRootPath]);
 	}
 
--- a/lib/services/livesync/livesync-service.ts
+++ b/lib/services/livesync/livesync-service.ts
@@ -100,7 +100,7 @@ export class LiveSyncService extends EventEmitter implements IDebugLiveSyncServi
 
 			// Emit LiveSync stopped when we've really stopped.
 			_.each(removedDeviceIdentifiers, deviceIdentifier => {
-				this.emit(LiveSyncEvents.liveSyncStopped, { projectDir, deviceIdentifier });
+				this.emitLivesyncEvent(LiveSyncEvents.liveSyncStopped, { projectDir, deviceIdentifier });
 			});
 		}
 	}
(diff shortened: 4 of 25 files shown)


KristianDD:
run ci

lib/services/project-service.ts
```diff
@@ -80,7 +80,7 @@ export class ProjectService implements IProjectService {
 	public isValidNativeScriptProject(pathToProject?: string): boolean {
 		try {
 			this.$projectData.initializeProjectData(pathToProject);
-			return !!this.$projectData.projectDir && !!this.$projectData.projectId;
+			return !!this.$projectData.projectDir && !!this.$projectData.projectIdentifiers;
```
KristianDD: Should check this. Not sure if we should verify for both iOS and android identifiers here or only for at least one.

lib/project-data.ts
```diff
@@ -30,7 +30,16 @@ export class ProjectData implements IProjectData {
 	public projectDir: string;
 	public platformsDir: string;
 	public projectFilePath: string;
-	public projectId: string;
+	public projectIdentifiers: IProjectIdentifier;
+	get projectId(): string {
+		this.warnProjectId();
+		return this.projectIdentifiers.ios;
```
Plamen5kov: why does a common module `project-data.ts` return only an iOS project identifier?

lib/project-data.ts
```diff
@@ -30,7 +30,16 @@ export class ProjectData implements IProjectData {
...
+	}
+	set projectId(identifier: string) {
+		this.warnProjectId();
+		this.projectIdentifiers.ios = identifier;
```
Plamen5kov: is there a validation for the identifier somewhere before this line? (check if valid ios identifier)

lib/project-data.ts
```diff
@@ -30,7 +30,16 @@ export class ProjectData implements IProjectData {
...
+		this.projectIdentifiers.android = identifier;
```
Plamen5kov: is this a valid android identifier?

lib/project-data.ts
```diff
@@ -30,7 +30,16 @@ export class ProjectData implements IProjectData {
 	public projectDir: string;
 	public platformsDir: string;
 	public projectFilePath: string;
-	public projectId: string;
+	public projectIdentifiers: IProjectIdentifier;
+	get projectId(): string {
+		this.warnProjectId();
+		return this.projectIdentifiers.ios;
+	}
+	set projectId(identifier: string) {
```
Plamen5kov: is setting project id used only in `initializeProjectData`?

lib/project-data.ts
```diff
@@ -87,6 +97,25 @@ export class ProjectData implements IProjectData {
...
+	private getProjectIdentifiers(data :any): IProjectIdentifier {
+		let identifier: IProjectIdentifier;
+		data = data || "";
+
+		if (typeof data === "string") {
+			identifier = {
+				android: data,
+				ios: data
+			};
+		} else {
```
Plamen5kov: What's the case when the android and ios package identifiers should be set to an empty string? Shouldn't we throw an error/warning here?

lib/project-data.ts
```diff
@@ -87,6 +97,25 @@ export class ProjectData implements IProjectData {
 		this.$errors.fail("No project found at or above '%s' and neither was a --path specified.", projectDir || this.$options.path || currentDir);
 	}
 
+	private getProjectIdentifiers(data :any): IProjectIdentifier {
```
Plamen5kov: since this is a private function, used only once, why don't we change the signature to:
`private getProjectIdentifiers(projectIdentifier :string): IProjectIdentifier`

lib/services/android-project-service.ts
```diff
@@ -209,16 +209,17 @@ export class AndroidProjectService extends projectServiceBaseLib.PlatformProject
 		// will replace applicationId in app/App_Resources/Android/app.gradle if it has not been edited by the user
 		const userAppGradleFilePath = path.join(projectData.appResourcesDirectoryPath, this.$devicePlatformsConstants.Android, "app.gradle");
 
+		//TODO: kddimitrov: For compatability with old templates. Once all templates are updated should delete.
 		try {
-			shell.sed('-i', /__PACKAGE__/, projectData.projectId, userAppGradleFilePath);
+			shell.sed('-i', /__PACKAGE__/, projectData.projectIdentifiers.android, userAppGradleFilePath);
 		} catch (e) {
-			this.$logger.warn(`\n${e}.\nCheck if you're using an outdated template and update it.`);
+			// Ignore
```
Plamen5kov: I think we should at least trace here, so we don't forget to remove this code later on

lib/project-data.ts
```diff
@@ -70,7 +80,7 @@ export class ProjectData implements IProjectData {
 					this.projectFilePath = projectFilePath;
 					this.appDirectoryPath = path.join(projectDir, constants.APP_FOLDER_NAME);
 					this.appResourcesDirectoryPath = path.join(projectDir, constants.APP_FOLDER_NAME, constants.APP_RESOURCES_FOLDER_NAME);
-					this.projectId = data.id;
+					this.projectIdentifiers = this.getProjectIdentifiers(data.id);
```
Plamen5kov: please rename getProjectIdentifiers to something that better describes what the method does

lib/services/android-project-service.ts
```diff
@@ -375,7 +376,7 @@ export class AndroidProjectService extends projectServiceBaseLib.PlatformProject
 			const filesForInterpolation = this.$fs.enumerateFilesInDirectorySync(resourcesDestinationDirectoryPath, file => this.$fs.getFsStats(file).isDirectory() || path.extname(file) === constants.XML_FILE_EXTENSION) || [];
 			for (const file of filesForInterpolation) {
 				this.$logger.trace(`Interpolate data for plugin file: ${file}`);
-				await this.$pluginVariablesService.interpolate(pluginData, file, projectData);
+				await this.$pluginVariablesService.interpolate(pluginData, file, projectData, projectData.projectIdentifiers.android);
```
Plamen5kov: why do you pass `projectData.projectIdentifiers.android` as an additional parameter when you already pass `projectData`?

lib/services/android-project-service.ts
```diff
@@ -227,8 +228,8 @@ export class AndroidProjectService extends projectServiceBaseLib.PlatformProject
 
 	private getProjectNameFromId(projectData: IProjectData): string {
 		let id: string;
-		if (projectData && projectData.projectId) {
-			const idParts = projectData.projectId.split(".");
+		if (projectData && projectData.projectIdentifiers && projectData.projectIdentifiers.android) {
+			const idParts = projectData.projectIdentifiers.android.split(".");
```
Plamen5kov: split by "." would imply we're sure there is a ".", is there a validation of the package identifier upstream?

lib/services/debug-service.ts
```diff
@@ -35,8 +35,10 @@ export class DebugService extends EventEmitter implements IDebugService {
 			projectDir: debugData.projectDir
 		});
 
-		if (!(await device.applicationManager.isApplicationInstalled(debugData.applicationIdentifier))) {
-			this.$errors.failWithoutHelp(`The application ${debugData.applicationIdentifier} is not installed on device with identifier ${debugData.deviceIdentifier}.`);
+		const devicePlatform = device.deviceInfo.platform;
+
+		if (!(await device.applicationManager.isApplicationInstalled(debugData.applicationIdentifier[devicePlatform.toLowerCase()]))) {
```
Plamen5kov: please change all the `toLowerCase()` all over the PR to a commonplace.

lib/services/ios-debug-service.ts
```diff
@@ -118,7 +118,7 @@ export class IOSDebugService extends DebugServiceBase implements IPlatformDebugS
 			waitForDebugger: true,
 			captureStdin: true,
 			args: args,
-			appId: debugData.applicationIdentifier,
+			appId: debugData.applicationIdentifier.ios,
```
Plamen5kov: ideally this should be named `applicationIdentifiers` plural

lib/services/livesync/livesync-service.ts
```diff
@@ -671,6 +672,10 @@ export class LiveSyncService extends EventEmitter implements IDebugLiveSyncServi
 		}
 	}
 
+	public emitLivesyncEvent (event: string, livesyncData: ILivesyncEventData): boolean {
```
Plamen5kov: is this only a rename of the function?

lib/services/platform-service.ts
```diff
@@ -236,13 +236,21 @@ export class PlatformService extends EventEmitter implements IPlatformService {
 			platform = this.$mobileHelper.normalizePlatformName(platform);
 			this.$logger.trace("Validate options for platform: " + platform);
 			const platformData = this.$platformsData.getPlatformData(platform, projectData);
-			return await platformData.platformProjectService.validateOptions(projectData.projectId, provision, teamId);
+			return await platformData.platformProjectService.validateOptions(
+				projectData.projectIdentifiers[platform.toLowerCase()],
```
Plamen5kov: why don't we just validate both project identifiers?

lib/services/platform-service.ts
```diff
@@ -236,13 +236,21 @@ export class PlatformService extends EventEmitter implements IPlatformService {
...
+				provision,
```
Plamen5kov: a bit strange to pass `provision` in a common file like `platform-service.ts` when it's iOS specific

lib/services/platform-service.ts
```diff
@@ -236,13 +236,21 @@ export class PlatformService extends EventEmitter implements IPlatformService {
...
+				teamId
+			);
 		} else {
 			let valid = true;
 			for (const availablePlatform in this.$platformsData.availablePlatforms) {
 				this.$logger.trace("Validate options for platform: " + availablePlatform);
 				const platformData = this.$platformsData.getPlatformData(availablePlatform, projectData);
-				valid = valid && await platformData.platformProjectService.validateOptions(projectData.projectId, provision, teamId);
+				valid = valid && await platformData.platformProjectService.validateOptions(
+					projectData.projectIdentifiers[availablePlatform.toLowerCase()],
```
Plamen5kov: same comments as above

lib/definitions/debug.d.ts
```diff
@@ -5,7 +5,7 @@ interface IDebugData extends Mobile.IDeviceIdentifier {
 	/**
 	 * Application identifier of the app that it will be debugged.
 	 */
-	applicationIdentifier: string;
+	applicationIdentifier: IProjectIdentifier;
```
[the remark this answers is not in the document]
KristianDD: Will check what this affects and whether we will need to add new property and change this to setter and getter.

lib/project-data.ts
KristianDD: This is intended to let hooks and extension that use the old property to behave as before.
Before if the user changed the application identifier for Android he must use the gradle file and the rest of the code will still use the one inside the package.json (the one for iOS).

lib/project-data.ts
KristianDD: Its not used but its intended for hooks and extensions that might try to modify it.

lib/project-data.ts
KristianDD: No the validation happens at a later stage.

lib/project-data.ts
KristianDD: No the validation happens at a later stage.

lib/project-data.ts
KristianDD: Rosen is right that the argument might be a string and an object. This depends on the use case of the user - if he needs different identifiers for Android and iOS the argument will be object.

lib/project-data.ts
KristianDD: Not sure I understand what the naming of the method should be. I am open to suggestions.

lib/services/android-project-service.ts
KristianDD: Well I left a comment to remind us to remove this peace of code. If you consider tracing is better I can remove the comment. Do you prefer both the comment and trace?

lib/services/android-project-service.ts
KristianDD: @Plamen5kov(accidently deleted the comment) Yes there is validation for the identifier.

lib/services/debug-service.ts
KristianDD: Well I discussed the options with @PanayotCankov and @rosen-vladimirov and this was the least bad way to handle it.

The other was to make a getter and setter for the lowercase properties inside an implementation of IProjectIdentifier.

lib/services/livesync/livesync-service.ts
KristianDD: Not exactly. There is still a calls to `this.emit` that made the introduction of ILivesyncEventData very ugly(all of the properties had to be optional).

lib/services/platform-service.ts
KristianDD: Because its not relevant to validate android identifier when trying to run the application on iOS device,

lib/services/platform-service.ts
KristianDD: Will check it, but will probably fix in separate PR.

lib/project-data.ts
Plamen5kov: maybe put a comment in the code, because it looks a bit strange returning only ios identifier in a common file

KristianDD: Will do that.

lib/project-data.ts
Plamen5kov: I think validation should happen as early as possible, but it's up to you.

lib/project-data.ts
Plamen5kov: it's up to you, but as I mentioned earlier validation should happen as early as possible so other modules can be sure validation is already done

lib/project-data.ts
Plamen5kov: please add a comment as this is not easily deducted

lib/project-data.ts
Plamen5kov: my suggestion is `initializeProjectIdentifiers`

lib/services/debug-service.ts
Plamen5kov: Given the platforms and other constants are living only in our CLI code, I don't understand the issue of not constantly using `toLowerCase`?

lib/services/livesync/livesync-service.ts
Plamen5kov: and why was the introduction of ILivesyncEventData necessary?

lib/services/platform-service.ts
Plamen5kov: Codewise it kind of makes sense, but I get what you mean.

lib/project-data.ts
Plamen5kov: In this case please rename `data: any` to `projectIdentifierInfo: any` or something better, but let it not be just `data`

lib/services/android-project-service.ts
Plamen5kov: Personally, I prefer trace, because it see it anytime I'm debugging as opposed to looking at the code (not so often)

miroslavaivanova:
run ci

lib/services/android-project-service.ts
KristianDD: Because pluginVariablesService doesn't have the platform context to get the proper identifier for the case.

lib/services/livesync/livesync-service.ts
KristianDD: So that we have types of the properties of the data object. I was going to miss it and break Sidekick.

KristianDD:
run ci

lib/services/livesync/livesync-service.ts
```diff
@@ -117,26 +117,27 @@ export class LiveSyncService extends EventEmitter implements IDebugLiveSyncServi
...
-			const msg = `Unable to start application ${projectData.projectId} on device ${liveSyncResultInfo.deviceAppData.device.deviceInfo.identifier}. Try starting it manually.`;
+			this.$logger.info(`Error while trying to start application ${projectData.projectIdentifiers[platform.toLowerCase()]} on device ${liveSyncResultInfo.deviceAppData.device.deviceInfo.identifier}. Error is: ${err.message || err}`);
+			const msg = `Unable to start application ${projectData.projectIdentifiers[platform.toLowerCase()]} on device ${liveSyncResultInfo.deviceAppData.device.deviceInfo.identifier}. Try starting it manually.`;
 			this.$logger.warn(msg);
 			if (!settings || !settings.shouldSkipEmitLiveSyncNotification) {
-				this.emit(LiveSyncEvents.liveSyncNotification, {
+				this.emitLivesyncEvent(LiveSyncEvents.liveSyncNotification, {
 					projectDir: projectData.projectDir,
-					applicationIdentifier: projectData.projectId,
+					applicationIdentifier: projectData.projectIdentifiers[platform.toLowerCase()],
```
rosen-vladimirov: you can extract the applicationIdentifier in a variable above and use it everywhere
```TypeScript
const applicationIdentifier = projectData.projectIdentifiers[platform.toLowerCase()]
```

lib/services/livesync/platform-livesync-service-base.ts
```diff
@@ -118,9 +121,10 @@ export abstract class PlatformLiveSyncServiceBase {
 	}
 
 	protected async getAppData(syncInfo: IFullSyncInfo): Promise<Mobile.IDeviceAppData> {
-		const deviceProjectRootOptions: IDeviceProjectRootOptions = _.assign({ appIdentifier: syncInfo.projectData.projectId }, syncInfo);
+		const platform = syncInfo.device.deviceInfo.platform.toLowerCase();
+		const deviceProjectRootOptions: IDeviceProjectRootOptions = _.assign({ appIdentifier: syncInfo.projectData.projectIdentifiers[platform] }, syncInfo);
```
rosen-vladimirov: you can extract `appIdentifier` in a separate variable:
```
const appIdentifier = syncInfo.projectData.projectIdentifiers[platform];
```

KristianDD:
run ci

KristianDD:
run ci

KristianDD:
run ci

mpochron:
@KristianDD what status now?

miroslavaivanova:
run ci

miroslavaivanova:
run ci

miroslavaivanova:
run ci

miroslavaivanova:
run ci

miroslavaivanova:
run ci

miroslavaivanova:
run ci

miroslavaivanova:
run ci

miroslavaivanova:
run ci

lib/services/livesync/livesync-service.ts
```diff
@@ -792,6 +795,10 @@ export class LiveSyncService extends EventEmitter implements IDebugLiveSyncServi
 		}
 	}
 
+	public emitLivesyncEvent (event: string, livesyncData: ILiveSyncEventData): boolean {
+		return this.emit(event, livesyncData);
```
rosen-vladimirov: Maybe we can add some trace message before emitting the event.

lib/definitions/livesync.d.ts
```diff
@@ -182,6 +182,16 @@ interface IHasUseHotModuleReloadOption {
 	useHotModuleReload?: boolean;
 }
 
+interface ILiveSyncEventData {
+	deviceIdentifier: string,
+	applicationIdentifier?: string,
+	projectDir: string,
+	syncedFiles?: Object
```
rosen-vladimirov: is this really an object?

lib/services/livesync/livesync-service.ts
KristianDD: Will tracing the event name be enough or should we `JSON.stringify` the livesyncData?

rosen-vladimirov: You can use:
```TypeScript
this.$logger.trace(`Will emit event ${event} with data`, livesyncData);
```

lib/definitions/livesync.d.ts
KristianDD: My bad. It is a `string[]`.

KristianDD:
run ci

KristianDD:
run ci

kspearrin:
Any idea what version this will land in?

rosen-vladimirov:
Hey @kspearrin , 
We'll release this in NativeScript 5.0.0, but you can try it even now by installing CLI's next version:
```
npm i -g nativescript@next
```

PR #3128, opened 2017-09-15
Introduce interface for localBuildService

Mitko-Kerezov:
* Introduce `ILocalBuildService` so that it can be reused
* Register `localBuildService` so that other services can use it
* Fail in `localBuildService` if building for `android` in `release` but not all of the `--key*` flags have been passed.

Merge after https://github.com/telerik/mobile-cli-lib/pull/1014


Mitko-Kerezov:
run ci

Mitko-Kerezov:
run ci

PR #3140, opened 2017-09-29
Fix execution of before-prepare hooks in a long living process

rosen-vladimirov:
When LiveSync operation is started we set `usbLiveSyncService.isInitialized` to true. This variable is used in several `before-prepare` hooks.
For example `nativescript-dev-typescript` has the following logic:
- When before-prepare hook is executed it checks if the `usbLiveSyncService.isInitialized` property is set to true. In this case it does nothing. The idea is that when LiveSync watch process is started, the plugin has started `tsc --watch` process, so there's no need to do anything on `before-prepare`.
However, in a long living process, where the project may be changed, whenever we stop LiveSync process, we must set usbLiveSyncService to false. In case we do not do it, the property remains true and the before-prepare hook of the other project does nothing and the project is not transpiled.


rosen-vladimirov:
run ci

PR #3143, opened 2017-10-02
Update ios-device-lib to 0.4.10

rosen-vladimirov:
Update ios-device-lib to latest version - 0.4.10. The main fix there is a case where CLI may hang while awaiting for notification from iOS application.

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -2144,13 +2144,6 @@
             }
           }
         },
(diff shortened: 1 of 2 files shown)


rosen-vladimirov:
run ci

rosen-vladimirov:
run ci

rosen-vladimirov:
Tests are green, the network share is the problem, so merging this PR.

Issue #3141, opened 2017-09-29
No console.log output Xcode 9 iOS 11


terreb:
### Did you verify this is a real problem by searching [Stack Overflow]
I took time to search, but haven't found anything mentioning this issue.

### Tell us about the problem
After upgrade to Xcode 9 and using iOS 11 simulator I'm not seeing console.log output after "tns run ios" any more.

### Which platform(s) does your issue occur on?
iOS

### Please provide the following version numbers that your issue occurs with:
- CLI: 3.2.1
- Cross-platform modules: tns-core-modules 3.2.0
- Runtime(s): tns-ios 3.2.0

Xcode 9.0 (9A235)
Simulator iPhone 6 iOS 11.0 (15A372)

KristianDD:
Thank you for the report. I can confirm that this is indeed an issue and we are currently working on a resolution.

As a temporary workaround open the Console app on Mac and from the Devices list select the connected device. Now in the right pane you can see all messages coming from the selected device. In the Search field type `proc:YourAppName` to see only the log coming from yours app process.

terreb:
@KristianDD, ok, thank you for the tip!

justindujardin:
Another workaround is to use the `tns debug ios` and the console output will be shown in the inspector.

terreb:
@justindujardin, yeah, I know that, but "tns debug" became so slow and hangs too often after update to Xcode 9 / iOS 11 that I would use "run" instead if I could.

briosheje:
@terreb If you use the watch you shouldn't have huge issues with `tns debug`. I've had to use it aswell since Xcode 9, it's just a bit slow when you run the instance, but as long as you use what you should be able to work with it, it shouldn't take a significant time to load everything (about 5 to 10 seconds in my case).

justindujardin:
@terreb Agreed that the inspector is not a great workaround. I ended up downgrading because I didn't need XCode 9, it was installed as part of the automatic update process.

The console being gone really hindered my development, and as you point out the inspector makes for really slow iterations, so it made sense to downgrade XCode until a fix is published. Now my console is back, and I only have to worry about the inspector slowly eating itself when I have to debug something. 😅 

If it makes sense for your project, and you have a valid developer account with Apple, you can downgrade like this:
- [Download the 8.x version of your choosing](https://developer.apple.com/downloads/index.action) 
- Drag your current XCode app into the trash.
- Extract the downloaded XCode and drag into `Applications`

[...]

Issue #3105, opened 2017-09-05
TNS 3.1.3 eating exceptions


NathanaelA:
## Please, provide the details below:
Doing a standard `tns run ios` will eat exceptions that occur.   

### Did you verify this is a real problem by searching 
yes

### Tell us about the problem
For some reason the newer versions of TNS seems to eat exceptions.  It logged my `CONSOLE` logs fine then the app appeared to freeze.  No output, it would allow updates to restart the app; but no exception/crash was logged to the screen.  Finally after a lot of wasted time, running it in xcode; I got to see the actual exception report so I could fix the issue.   

### Which platform(s) does your issue occur on?
IOS

### Please provide the following version numbers that your issue occurs with:
- CLI: 3.1.3

### Please tell us how to recreate the issue in as much detail as possible.
I'm not 100% sure if it eliminates all exceptions, but NONE of this printed in TNS:

```
2017-09-05 13:46:10.998030 demo[2304:1299789] *** JavaScript call stack:
(
    0   UIApplicationMain@[native code]
    1   start@file:///app/tns_modules/tns-core-modules/application/application.js:211:26
    2   anonymous@file:///app/app.js:10:18
    3   evaluate@[native code]
    4   moduleEvaluation@:1:11
    5   @:7:50
    6   promiseReactionJob@:1:11
)
2017-09-05 13:46:10.998102 demo[2304:1299789] *** Terminating app due to uncaught exception 'UIApplicationInvalidInterfaceOrientation', reason: 'Supported orientations has no common orientation with the application, and [GVRRendererViewController shouldAutorotate] is returning YES'
*** First throw call stack:
[...]

PR #3104, opened 2017-09-05
Emit deviceIdentifier in debuggerAttached event

rosen-vladimirov:
`debuggerAttached` event should emit full information required for debugging. However, the deviceIdentifier is missing, so add it.
Update PublicAPI.md according to latest changes.

PR #3145, opened 2017-10-03
DO NOT MERGE Fix iOS11 Simulator logs

KristianDD:
Part of the fix for https://github.com/NativeScript/nativescript-cli/issues/3141#issuecomment-333449119.

Log format for iOS 10.3: "Oct  3 19:07:55 machinename cliapp[15339]: CONSOLE LOG file:///app/main-view-model.js:18:20: Test Console fix..."

Log format for iOS 11: "2017-10-03 19:09:14.681822+0300 0x511df    Default     0x30f33             14407  cliapp: (NativeScript) CONSOLE LOG file:///app/main-view-model.js:18:20: Test Console fix.."

Needs https://github.com/telerik/ios-sim-portable/pull/95 and https://github.com/telerik/mobile-cli-lib/pull/1016


lib/services/ios-log-filter.ts
```diff
@@ -16,7 +16,7 @@ export class IOSLogFilter extends iOSLogFilterBase.IOSLogFilter implements Mobil
 
 	public filterData(data: string, logLevel: string, pid?: string): string {
 		data = super.filterData(data, logLevel, pid);
-		if (pid && data && data.indexOf(`[${pid}]`) === -1) {
+		if (pid && data && data.search(new RegExp(`[\\[\\s]${pid}[\\]\\s]`)) === -1) {
```
Plamen5kov: IMO either choose RegExp or indexOf for consistency [in the code](https://github.com/telerik/mobile-cli-lib/pull/1016/files#diff-6970fd71154619adb4a5ac4fc61de7fcR11)

lib/services/ios-log-filter.ts
```diff
@@ -41,14 +41,16 @@ export class IOSLogFilter extends iOSLogFilterBase.IOSLogFilter implements Mobil
 				// This code removes unnecessary information from log messages. The output looks like:
 				// CONSOLE LOG file:///location:row:column: <actual message goes here>
 				if (pid) {
-					const searchString = "[" + pid + "]: ";
-					const pidIndex = line.indexOf(searchString);
-					if (pidIndex > 0) {
-						line = line.substring(pidIndex + searchString.length, line.length);
+					const pidIndex = line.indexOf(`[${pid}]`);
+					const ios11PidIndex = line.indexOf(` ${pid} `);
+					if (pidIndex > 0 || ios11PidIndex > 0) {
```
Plamen5kov: again choose either `!= 1` or `> 0` for consistency [in the code](https://github.com/telerik/mobile-cli-lib/pull/1016/files#diff-6970fd71154619adb4a5ac4fc61de7fcR11)

lib/services/ios-log-filter.ts
KristianDD: I used regex to avoid nested or complex "if" statements:
```javascript
if(data && pid){
  if(data.indexOf(` ${pid} `) ||  data.indexOf(`[${pid}]`)) {
  }
}
```
or
```javascript
if(data && pid && (data.indexOf(` ${pid} `) ||  data.indexOf(`[${pid}]`)) {
}
```

lib/services/ios-log-filter.ts
KristianDD: Could change it to `!= -1`

KristianDD: I chose to use `indexOf` because its a bit more fast in my opinion(didn't really compared performance tests though).

Plamen5kov: It is a bit faster, but I made the suggestions just for consistency sake because IMO it's more important

NathanaelA:
Just pass `--predicate 'senderImagePath contains "NativeScript"'` and the  `log` will automatically filter it to just message from the NativeScript application.     Not only is that faster; but it eliminates potential code for breaking and potentially filtering out valid messages which is a painful issue with the old logging.  #3105  #3104

lib/services/ios-log-filter.ts
Mitko-Kerezov: The statement `` (data.indexOf(` ${pid} `) ||  data.indexOf(`[${pid}]`) `` can be extracted into a separate method and called everywhere needed.
We'd avoid complexity and gain readability IMO

lib/services/ios-log-filter.ts
```diff
@@ -41,14 +41,16 @@ export class IOSLogFilter extends iOSLogFilterBase.IOSLogFilter implements Mobil
...
+						const pidRegex = new RegExp(`.*[\\[\\s]{1}${pid}[\\s\\]]{1}.*?:(?:.*?\\(NativeScript\\) )*`);
```
Mitko-Kerezov: Why do we need the part after `:` in this `RegExp` - I'm talking about the ignored capture group `` (?:.*?\\(NativeScript\\) )* ``

If one should wish to `console.log` something like `Loving (NativeScript) right now` in their application the log would appear as
`"Aug 22 10:59:20 MCSOFAPPBLD TestApp[52946]: CONSOLE LOG file:///app/home/home-view-model.js:6:20: Loving (NativeScript) right now"`

and the `RegExp` would strip it to `right now`

lib/services/ios-log-filter.ts
```diff
@@ -41,14 +41,16 @@ export class IOSLogFilter extends iOSLogFilterBase.IOSLogFilter implements Mobil
...
+						line = line.replace(pidRegex, "").trim();
 						this.getOriginalFileLocation(line);
 						result += this.getOriginalFileLocation(line) + "\n";
-						continue;
 					}
+
+					continue;
```
Mitko-Kerezov: If we don't match with neither `pidIndex` nor `ios11PidIndex` I do not believe we'd want to `continue`.
I think this `continue` statement should be returned in the previous block

lib/services/ios-log-filter.ts
KristianDD: You are right will fix the regex.

lib/services/ios-log-filter.ts
KristianDD: Taking `continue` outside the `if` statement prevents lines that don't contain the pid to be logged.

NathanaelA:
@KristianDD & @Mitko-Kerezov - Is there some reason you are adding all this crazy code to filter; why not Just pass `--predicate 'senderImagePath contains "NativeScript"'` and the log will automatically filter it to just message from the NativeScript application.   Is their some reason to over complicate and slow down the solution with indexof and regex's?

KristianDD:
@NathanaelA we are passing the predicate [here](https://github.com/telerik/mobile-cli-lib/pull/1016/files#diff-b88cb245cddf36be1404a81c0fb9b469R13), but we are still supporting older iOS versions as you can see [here](https://github.com/telerik/ios-sim-portable/pull/95/files#diff-24d886fc42817807be20588a822623e0R170). For the older iOS versions we still need the filtration code. The benefit of the predicate is that the code that filters messages will be invoked only for NS messages, Moreover the regex is used only if the line contains the "pid" of the application. We haven't noticed any performance issues so far.

The filtration of exception messages is not reproducible on Simulators, at least on my side. The real device logging and filtering is handled by different branch of the code.

lib/services/ios-log-filter.ts
```diff
@@ -41,14 +41,14 @@ export class IOSLogFilter extends iOSLogFilterBase.IOSLogFilter implements Mobil
 				// This code removes unnecessary information from log messages. The output looks like:
 				// CONSOLE LOG file:///location:row:column: <actual message goes here>
 				if (pid) {
-					const searchString = "[" + pid + "]: ";
-					const pidIndex = line.indexOf(searchString);
-					if (pidIndex > 0) {
-						line = line.substring(pidIndex + searchString.length, line.length);
+					if (line.indexOf(`[${pid}]: `) !== -1) {
+						const pidRegex = new RegExp(`.*[\\[]{1}${pid}[\\]]{1}(?::\\s(:?\\(NativeScript\\)\\s)*)`);
```
Mitko-Kerezov: The `RegExp` can be shortened to:
`` const pidRegex = new RegExp(`.*[\\[]${pid}[\\]]:\\s(:?\\(NativeScript\\)\\s)*`); ``

KristianDD: How does this look as a final version:
```javascript
const pidRegex = new RegExp(`^.*[\\[]${pid}[\\]]:\\s(?:\\(NativeScript\\)\\s)?\`);
```

lib/services/ios-log-filter.ts
```diff
@@ -41,14 +41,14 @@ export class IOSLogFilter extends iOSLogFilterBase.IOSLogFilter implements Mobil
 				// This code removes unnecessary information from log messages. The output looks like:
 				// CONSOLE LOG file:///location:row:column: <actual message goes here>
 				if (pid) {
-					const searchString = "[" + pid + "]: ";
-					const pidIndex = line.indexOf(searchString);
-					if (pidIndex > 0) {
-						line = line.substring(pidIndex + searchString.length, line.length);
+					if (line.indexOf(`[${pid}]: `) !== -1) {
+						const pidRegex = new RegExp(`^.*[\\[]${pid}[\\]]:\\s(?:\\(NativeScript\\)\\s)?`);
```
Mitko-Kerezov: There is actually no need for `[\\[]` and `[\\]]` it can be further shortened to:
`` const pidRegex = new RegExp(`^.*\\[${pid}\\]:\\s(?:\\(NativeScript\\)\\s)?`); ``

dtopuzov:
run ci

dtopuzov:
run ci

Issue #3016, opened 2017-07-29
tns_modules randomly appears in app folder and breaks build


thaoula:
Hi Guys,

A folder called tns_modules is randomly added to the app folder. The contents seem to be a subset of node_modules.

Once the folder appears, you cannot use tns run / build etc. Deleting the folder fixes these issues.

Why does tns_modules get created?

I am using TNS 3.1.3 on a Mac, this happened in 3.1.0 as well.

Regards,
Tarek

dtopuzov:
Hi @thaoula 

Can you specify version of `node` and `npm` on your machines?
If you can also isolate workflow that reproduce the issue it will be great.

thaoula:
Hi @dtopuzov,

I am using npm 5.3.0 and node 8.

I am still trying to identify the workflow that causes tns_modules to be created. Unfortunately, i have been unable to work out any repeatable steps.

However, it happened several times yesterday.

Regards,
Tarek

dtopuzov:
Hi @thaoula 

Thanks for the info.
Two more questions:
- Do you use `npm link` in your development workflow?
- Do you use some dev plugins like `nativescript-dev-sass` or `nativescript-dev-less` (something that modify files in the project while you develop)?

thaoula:
Hi @dtopuzov,

Thanks for getting back to me -
I use nativescript-dev-less and I also use npm link for our model project. 

Is their anyway to debounce watches or ignore files?

Regards,
Tarek

dtopuzov:
Hi @thaoula 

We do not have such feature available at the moment, but it is in the backlog: https://github.com/NativeScript/nativescript-cli/issues/887

The issue seems to happen only if you have `npm5` and use `npm link`.
We will try to fix it for next release(s) and until then you can try to workaround it via downgrade to `npm4`.

dtopuzov:
It looks other users also hit the same issue: https://github.com/NativeScript/nativescript-angular/issues/916

Plamen5kov:
@thaoula If you could provide us with the steps to reproduce the issue, that will contribute to how fast it will be fixed. Thank you!

kmuthukrishnan:
Hi @dtopuzov 
Am having the same problem and am not using npm link in the development workflow.
Also not using dev plugins like nativescript-dev-sass or nativescript-dev-less.

Once I delete the tns_modules folder everything works.
Thanks

mikebranstein:
[...]

PR #3154, opened 2017-10-10
Fix tns_modules added to app dir

rosen-vladimirov:
In some cases `tns_modules` dir may appear in project's app dir. The issue is caused by sending Ctrl + C while `ios-sim-portable` is executing `xcrun simctl` command.
In this case, it returns empty string for applicationPath instead of breaking the CLI process. Add new check if the projectRoot on device is not empty string and update ios-sim-portable.

https://github.com/NativeScript/nativescript-cli/issues/3016


rosen-vladimirov:
run ci

dtopuzov:
run ci

rosen-vladimirov:
run ci

dtopuzov:
run ci

PR #3170, opened 2017-10-19
Add timeout before starting the app-inspector on tns debug ios

Pip3r4o:
The changes in the PR introduce a delay before the AppInspector debugger for iOS applications is started, in order to prevent a timing issue in the `AppInspector - iOS Runtime - tns-core-modules inspector implementation` communication, which is yet to be identified.

The PR is intended as a temporary workaround to allow shipping of 3.3 without breaking existing functionality. Investigation of the underlying problem will continue after the official 3.3 release, and ideally this PR's changes will be rolled back if they are no longer necessary in 3.4.

CC @etabakov @mbektchiev

### DIFF
--- a/lib/services/ios-debug-service.ts
+++ b/lib/services/ios-debug-service.ts
@@ -207,7 +207,17 @@ export class IOSDebugService extends DebugServiceBase implements IPlatformDebugS
 			return this.getChromeDebugUrl(debugOptions, this._socketProxy.options.port);
 		} else {
 			this._socketProxy = await this.$socketProxyFactory.createTCPSocketProxy(this.getSocketFactory(device));
-			await this.openAppInspector(this._socketProxy.address(), debugData, debugOptions);
+			// delay opening the AppInspector until after the runtime has finished initializing the
+			// dynamically registered Chrome DevTools agent implementations in JavaScript using the tns-core-modules
+			// The workaround is necessary to prevent an unexpected timing crash introduced in the 3.3 modules
+			const inspectorStartDelay = 6000;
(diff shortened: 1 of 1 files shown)


etabakov:
run ci

lib/services/ios-debug-service.ts
```diff
@@ -207,7 +207,13 @@ export class IOSDebugService extends DebugServiceBase implements IPlatformDebugS
 			return this.getChromeDebugUrl(debugOptions, this._socketProxy.options.port);
 		} else {
 			this._socketProxy = await this.$socketProxyFactory.createTCPSocketProxy(this.getSocketFactory(device));
-			await this.openAppInspector(this._socketProxy.address(), debugData, debugOptions);
+			// delay opening the AppInspector until after the runtime has finished initializing the
+			// dynamically registered Chrome DevTools agent implementations in JavaScript using the tns-core-modules
+			// The workaround is necessary to prevent an unexpected timing crash introduced in the 3.3 modules
+			const inspectorStartDelay = 6000;
+			setTimeout(async () => {
+				await this.openAppInspector(this._socketProxy.address(), debugData, debugOptions);
```
rosen-vladimirov: In case this operation fails, you'll receive unhandled rejection, can you please add `try { } catch()` around the call and log the error with $logger.

Pip3r4o: Certainly, will also rebase to release.

Pip3r4o:
Closing this in favor of a workaround in the tns-core-modules.

PR #3192, opened 2017-11-02
Improve create project performance

ivanovit:
In general case, we will perform only one npm install instead of two.

 **Before:**
  npm install <template>
  copy dependencies from the installed template package.json to project package.json
  npm install on project

**Now:**
npm view <template>
create project package.json with all dependencies
npm install (only once)


lib/services/project-service.ts
```diff
@@ -195,5 +163,21 @@ export class ProjectService implements IProjectService {
...
+		const packageJson = this.$fs.readJson(projectFilePath);
+
+		packageJson.dependencies = _.extend({}, packageJson.dependencies, dependencies) ;
+		packageJson.devDependencies = _.extend({}, packageJson.devDependencies, devDependencies);
+
+		this.$fs.writeJson(projectFilePath, packageJson);
+	}
+
+	private async installAndGetInfo(projectDir: string, template: string) {
+		const result = await this.$npm.install(template, projectDir, <any>{  silent: true, save: true });
```
rosen-vladimirov: I believe you should use `npmInstallationManager` here.

Plamen5kov:
Hi @ivanovit, you can't do `npm view` on `.tgz` packages, that's why we originally do npm install and then check the `package.json` version. In case you were wandering about why we do npm install twice.

ivanovit:
Closing this PR because with npm5 there is no significant performance improvement.

Issue #3118, opened 2017-09-13
Console.log has no output on terminal


NickIliev:
_From @bigradish on September 13, 2017 2:41_

When I follow the steps in the "Debugging apps" section of your tutorial https://docs.nativescript.org/tutorial/chapter-1 , both console.log and console.dir have no output on the terminal.

More info:
1. development machine: Windows 10 64bit
2. target platform: android
3. cli version: 3.2.0
4. Cross-platform modules: 3.1.1
5. Runtime: 3.2.0
6. Plugin: no plugin used, just the HelloWorld project in the tutorial.

Thank you.

_Copied from original issue: NativeScript/NativeScript#4835_

christian-kolb:
I have the same issue on the same CLI and runtime version but on Mac and with iOS as target platform.

christian-kolb:
If I run the app with `tns debug ios --watch` I get console output in the Nativescript Inspector. But only there and this seems to be the only way to get to the logging. The problem is for all logging information including errors.

christian-kolb:
@bigradish and @NickIliev Did one of you try to use an older version of the CLI yet? I first thought it might be related to the XCode update and the new iOS 11, but as you're using Android on Windows those two things don't seem to be related.

glen-stephens:
I just upgraded to XCode 9 and iOS 11 and my console.log quit outputting to the terminal.  Was working fine previous to updating XCode. Have not been able to get it working again. ` tns debug ios --watch` does output to the nativescript console, but not the terminal.  Any help or suggestions would be awesome!

thegermin8er:
Same as glen-stephens

etabakov:
Hey guys,
Regarding the problem with the missing logs with XCode9 / iOS 11, this is caused by changes in the latest iOS and how event logging is handled. We are working on resolution and it should be available in the next minor release.

Regarding the problem with Windows and Android - this is not related to the above issue. We will take a look and let you know.

NathanaelA:
For those who need logging back; here is the way I'm doing it now... 
fluentreports.com/blog/?p=545

glen-stephens:
@etabakov If it helps, console.log does work to the terminal when running on an actual device, just not from the simulator.

KristianDD:
Here is a separate issue for iOS11 Simulator and Xcode9: https://github.com/NativeScript/nativescript-cli/issues/3141. Lets continue the discussion there.

bigradish:
[...]

PR #3198, opened 2017-11-06
Fix missing console.logs on Android when CLI is used as library

rosen-vladimirov:
In case CLI is used as a library, the device logs are no longer captured after application is restarted. The problem is the newly added logic for Android devices that stops the logcat process when CLI calls `stopApplication` and starts it when `startApplication` is called.
However, in `startApplication` there's a check if `$options.justlaunch` is false - only then the device log operation is started again.
When CLI is used as a library, we have hardcoded the `justlaunch` option to true in order to prevent duplicate logs for iOS devices. So the solution to resolve Android logs issues is to remove the setting of justlaunch option.
However, this caused the duplicate logs issue for iOS devices. Investigating it further, it turned out to be a memory leak when iOS device is detached. It's fixed in the submodule:

### Fix leaking handlers for iOS device logs
In case iOS Device is detached, the `IOSDevice` instance is kept alive. The reason is the handler for device logs, that `IOSDevice` passes to `iOSDeviceOperations`.
As `iOSDeviceOperations` is still alive, the `IOSDevice` instance is also alive. Reattaching the same device will cause duplicate logs. Detaching and attaching it again will lead to additional logs.

In order to fix the issue, convert `iOSDeviceOpertations` to event emitter and emit event when there's data for logging. Each `IOSDevice` will add handler for the event and will do its logic.
In case device is detached (i.e. `DeviceLost` event is fired in `DevicesService`), call a newly added method to the specific `IDevice` instance. It's purpose is to clean the used resource.
For iOS device, this method will remove the handler for `devceLogData` event of `iOSDeviceOperations`.

### DIFF
--- a/lib/nativescript-cli-lib-bootstrap.ts
+++ b/lib/nativescript-cli-lib-bootstrap.ts
@@ -9,7 +9,4 @@ $injector.requirePublic("companionAppsService", "./common/appbuilder/services/li
 $injector.requirePublicClass("deviceEmitter", "./common/appbuilder/device-emitter");
 $injector.requirePublicClass("deviceLogProvider", "./common/appbuilder/device-log-provider");
 
-// We need this because some services check if (!$options.justlaunch) to start the device log after some operation.
-// We don't want this behaviour when the CLI is required as library.
-$injector.resolve("options").justlaunch = true;
 $injector.resolve<IStaticConfig>("staticConfig").disableAnalytics = true;


dtopuzov:
run ci

bigradish:
@rosen-vladimirov , console.log also does not work on android when CLI is NOT used as a library. You can refer to issue #3118 .Thanks.

Issue #3164, opened 2017-10-18
build-*-bundle gets stuck at nativescript-unit-test-runner hook in NS 3.2 


dbenninger:
### Tell us about the problem
The commands `npm run build-android-bundle` and `npm run build-ios-bundle` get stuck at the nativescript-unit-test-runner hook and never fail or complete in NS 3.2:

```
npm run build-ios-bundle

> @ build-ios-bundle /Users/davidbenninger/Desktop/Test
> npm run ns-bundle --ios --build-app


> @ ns-bundle /Users/davidbenninger/Desktop/Test
> ns-bundle

Running tns prepare...
Executing before-prepare hook from /Users/davidbenninger/Desktop/Test/hooks/before-prepare/nativescript-dev-typescript.js
Preparing project...
Project successfully prepared (ios)
Executing after-prepare hook from /Users/davidbenninger/Desktop/Test/hooks/after-prepare/nativescript-unit-test-runner.js
```

a normal `tns run ...` seems to work fine.

### Which platform(s) does your issue occur on?
with iOS and Android

### Please provide the following version numbers that your issue occurs with:
- CLI: 3.2.1
- Cross-platform modules: 3.2.0
- Runtime(s): 3.2.0
- Plugin(s): "nativescript-unit-test-runner": "^0.3.4"

### To reproduce:

1. tns create Test --template ng
2. cd Test
3. npm install --save-dev nativescript-dev-webpack
4. npm install
5. tns test init -> select "jasmine"
6. npm run build-android-bundle

NickIliev:
@dbenninger thank you for reporting this issue - confirming this as a bug.
Steps to reproduce: use the ones described in the thread above!

dbenninger:
@NickIliev this blocks us from upgrading. Any news on this?

NickIliev:
@dbenninger we are currently looking into it but meanwhile as the plugin nativescript-unit-test-runner is used mainly in development, perhaps you can remove the plugin (and the hooks) before bundling with WebPack (temporary solution)

renestalder:
Anyone can post the concrete "workaround" here, maybe with code examples?

dbenninger:
@renestalder Our workaround is to remove the unit test plugin before running the bundle build:

tns plugin remove nativescript-unit-test-runner

renestalder:
Thank you. :)

I'm not that used to the whole tns cli possibilities yet.

jevenson:
[...]

PR #3212, opened 2017-11-10
Fix hanging commands when webpack and unit tests are used

rosen-vladimirov:
In case you try using npm scripts that come with `nativescript-dev-webpack` and you also have `nativescript-unit-test-runner` in your project, the scripts hang. The problem is that `nativescript-unit-test-runner` has a hook that resolves `testExecutionService`. In its constructor it sets that analytics broker should not be disposed. However, this is incorrect - it should not be disposed only in case we execute tests with livesync enabled. In order to fix this, move the logic in the commands. This way the hook will not change the disposing of the analytics broker, so once CLI process finishes, analytics broker process will be disconnected and CLI's process will die gracefully.

Fixes: https://github.com/NativeScript/nativescript-cli/issues/3164

Also update submodule, where the following change is applied:
Fix execution of hooks outside of process

In case you need a hook to be executed outside of process, you should place correct shebang at the beginning of the file and CLI should spawn it.
However, this is not working at the moment, as when CLI detects that the hook should be spawned, it just does nothing. The code is incorrectly placed inside the `if` for in process execution. This causes another issue - in case a hook returns a function, which returns falsey value, we decide the hook should be executed outside of process and spawn a new Node.js process to execute the hook. So we execute it twice.

Fix the if-else logic, so the hooks will work correctly.

### DIFF
--- a/lib/commands/test.ts
+++ b/lib/commands/test.ts
@@ -4,8 +4,10 @@ function RunTestCommandFactory(platform: string) {
 	return function RunTestCommand(
 		$options: IOptions,
 		$testExecutionService: ITestExecutionService,
-		$projectData: IProjectData) {
+		$projectData: IProjectData,
+		$analyticsService: IAnalyticsService) {
 		$projectData.initializeProjectData();
+		$analyticsService.setShouldDispose($options.justlaunch || !$options.watch);
 		const projectFilesConfig = helpers.getProjectFilesConfig({ isReleaseBuild: $options.release });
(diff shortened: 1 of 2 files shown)


rosen-vladimirov:
run ci

rosen-vladimirov:
run ci

rosen-vladimirov:
run ci

rosen-vladimirov:
Merging as the failing test is due to infrastructure issues.

Issue #3217, opened 2017-11-13
node-package-manager assumes that platform package is always the first in the list of updated ones


Buuhuu:
In [node-package-manager.ts#L158](
https://github.com/NativeScript/nativescript-cli/blob/d4817bc24be8b2f08e1101d22a53b65e6d3e9d8c/lib/node-package-manager.ts#L158) the assumption is made that the installed platform package is always the first entry in the list of returned updated dependencies. In my setup I install a couple of `file:...` dependencies which are listed in the same place:

```
[
  {
    "action": "update",
    "name": "nativescript-adobe-marketing-cloud",
    "version": "file:nativescript-adobe-marketing-cloud-1.3.1.tgz",
    "path": "...",
    "previousVersion": ""
  },
  {
    "action": "update",
    "name": "nativescript-angular",
    "version": "file:nativescript-angular-3.1.3.tgz",
    "path": "...",
    "previousVersion": ""
  },
  {
    "action": "update",
    "name": "nativescript-google-maps-sdk",
    "version": "file:nativescript-google-maps-sdk-2.3.1.tgz",
    "path": "...",
    "previousVersion": ""
  },
  {
    "action": "update",
    "name": "nativescript-image-swipe",
    "version": "file:nativescript-image-swipe-2.0.3.tgz",
    "path": "...",
    "previousVersion": ""
  },
  {
    "action": "update",
    "name": "tns-android",
    "version": "3.1.1",
    "path": "...",
    "previousVersion": "3.1.1"
  }
]
```

And tns-android in that case it the last entry.

rosen-vladimirov:
Hey @Buuhuu ,
Can you send us the output of the following commands:
```
$ node -v
$ npm -v 
```

rosen-vladimirov:
Also which command fails on your side? Is it `tns build android` or something else?

Buuhuu:
node 8.9.1, npm 5.5.1 and its `tns prepare android` and any other like that - working on a PR right now.

PR #3213, opened 2017-11-10
parse npm install results correctly 

dnadoba:
Solves Issue https://github.com/NativeScript/nativescript-cli/issues/3217

This pull request changes the implementation to search for the correct object by name. It does no longer depend on the position and should be more reliable especially as the `--json` flag is experimental as described in the npm documentation: https://docs.npmjs.com/misc/config#json

As described in the comment of the current implementation of `parseNpmInstallResults` we just use the first object of the `updated` array.

https://github.com/NativeScript/nativescript-cli/blob/e181f034264d82f40d3cbbfbdfd6599f5b2064c7/lib/node-package-manager.ts#L158-L160

But when I execute the following command:
```
npm install tns-ios --json --dry-run
```
`tns-ios` is not the first element in the `updated` array. The output looks something like this:
```json
{
  "added": [],
  "removed": [],
  "updated": [
    {
      "action": "update",
      "name": "colors",
      "version": "https://registry.npmjs.org/colors/-/colors-1.1.2.tgz",
      "path": "<path to project>/node_modules/colors",
      "previousVersion": ""
    },
    {
      "action": "update",
      "name": "inherits",
      "version": "https://registry.npmjs.org/inherits/-/inherits-2.0.3.tgz",
      "path": "<path to project>/node_modules/inherits",
      "previousVersion": ""
    },
    {
      "action": "update",
      "name": "tns-ios",
      "version": "3.3.0",
      "path": "<path to project>/node_modules/tns-ios",
      "previousVersion": "3.3.0"
    }
  ],
  "moved": [],
  "failed": [],
  "elapsed": 3113
}
```
Instead of the first element, `tns-ios` is the last element. 

I had an issue with the following command:
```
tns platform add ios
```
It always failed with the following error:
```
cp: no such file or directory: <path to project>/node_modules/<some random updated module>/framework/*
``` 
Now it works like a charm.

PR #3218, opened 2017-11-13
Fix broken prepare platform on NPM 5.5.1

Buuhuu:
Instead of taking the first updated dependency, we resolve the one passed by userSpecifiedPackageName) (#3217)

### DIFF
--- a/lib/node-package-manager.ts
+++ b/lib/node-package-manager.ts
@@ -159,7 +159,7 @@ export class NodePackageManager implements INodePackageManager {
 			// Considering that the dependency is already installed we should
 			// find it in the `updated` key as a first element of the array.
 			if (!name && npm5Output.updated) {
-				const updatedDependency = npm5Output.updated[0];
+				const updatedDependency = npm5Output.updated.find(dependencyInfo => dependencyInfo.name === userSpecifiedPackageName);
 				return {
 					name: updatedDependency.name,
 					originalOutput,


dtopuzov:
run ci

rosen-vladimirov:
Hey @Buuhuu ,
Thank you very much for your efforts. Another community member already prepared a PR for a similar issue and we are currently checking the implementation. https://github.com/NativeScript/nativescript-cli/pull/3213
Can you help us by verifying the other solution works on your side as well?

We'll see which solution covers all scenarios and we'll decide which PR to merge. 
Once again, thanks a lot for making this PR!

Buuhuu:
Your are right, this is exactly the same and i think, without checking my project, this will resolve my issue as well. Though I don’t see the need of the split by @ nor the need of using _ for that in #3213. 

The split should not be necessary as the parameter, from its name, is not a package descriptor but a package name. The find can be done as shown in this PR.

rosen-vladimirov:
Hey @Buuhuu ,
The idea of the split is that `userSpecifiedPackageName` may contain both package name and version, for example:
```
$ tns plugin add nativescript-imagepicker@4.0.0
$ tns platform add android@3.3.0
```

The other scenario where the `userSpecifiedPackageName` may contain different informatio (not the real package name) is when a path to local package or .tgz is installed:
```
$ tns plugin add nativescript-imagepicker-packed.tgz
$ tns platform add android --frameworkPath package.tgz
```
In these cases we still have to default to something (as done in the other PR).

I've also checked the failing tests in both PRs (sorry you are unable to see them, we will do some work to show the results in the PR itself in the future) - we have some integration tests that install runtime packages with `--frameworkPath` and they fail on this PR , but pass on the other one.

So I'm going to close this one and continue checking the other PR.

Thanks again for taking your time to create this PR, we really appreciate it and we'll be glad to assist you further in your next contributions to NativeScript!

rosen-vladimirov:
Closing in favor of https://github.com/NativeScript/nativescript-cli/pull/3213

Buuhuu:
Thanks for the explanation, that completely makes sense.

PR #3220, opened 2017-11-15
README.md - CLI Command --help to load info in the console

dean-huczok:
Added a single line to document this small and simple but important feature `tns --help` to load HELP info in the console rather than launching in a browser or text editor window.

Read a few issues where users have asked about preventing HELP loading in the browser (including a `-nobrowser` request) or text editor (when users computer has set a text editor as default app to open HTML files) and found the `--help` command hidden in a response from @Fatme, so I thought it would be great to add that little command into the CLI command docs :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,7 +115,7 @@ Quick Start
 
 ## The Commands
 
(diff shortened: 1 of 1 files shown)


rosen-vladimirov:
run ci

Issue #3202, opened 2017-11-07
`tns update <version>` does not work


dtopuzov:
Help says:
```
tns update --help
# platform update

│ Usage                          │ Synopsis               │
│ Update with the latest version │ $ tns update           │
│ Update with specific version   │ $ tns update <version> │

Updates a NativeScript project to the latest (or specified) version.
```

`tns update 3.3.0` results in:
```
Invalid platform 3.3.0. Valid platforms are ios or android.
```

erkanarslan:
I cannot upgrade to version 3.3.0. I hope this is fixed soon.

erkanarslan:
Is this fixed or not?

KristianDD:
Hi @erkanarslan,
The fix for the issue is merged in master branch and is being verified by our QA team. The fix should be included in the next official release.

marcocasadio:
I still have the issue to update nativescript to 3.4.3.... Any suggest?
Now I can't use Sidekick because it needs the 3.4.3 of nativescript cli to works

Fatme:
hey @marcocasadio,

Can you give us more information about the issue you have? Are you using `update` command or `npm install`? Please tell us how to recreate the issue in as much detail as possible.

marcocasadio:
Hi @Fatme!
I run "tns update 3.4.3" and I get "Could not update the project!"
I've a project with angular + typescript.
I've tried to remove nativescript and reinstall with the commands wrote by @KristianDD but I always get the 3.4.2 version and I can't use Sidekick.

sroy93:
I have this same issue trying to update to 3.4.3.  I also cannot use Sidekick and I need it asap.

rosen-vladimirov:
Hey @marcocasadio , @sroy93 , 
Can you try the following commands:
```
$ npm cache clean --force
$ npm uninstall -g nativescript
$ which tns // This command will show you if you have installed NativeScript CLI somewhere else - it should return empty result in case the package is not installed anywhere
$ npm install -g nativescript@latest
```
This should install the latest version of NativeScript CLI on your side.

marcocasadio:
@rosen-vladimirov now it seems ok! (@sroy93)
after the uninstall I launched these commands: https://docs.npmjs.com/misc/removing-npm

thanks

sroy93:
It seems as though nativescript will not uninstall.. Even after I run the uninstall command, I can do 
tns --version and get a result back..

marcocasadio:
[...]

PR #3227, opened 2017-11-22
Fix update command and add tests.

KristianDD:
Fixes https://github.com/NativeScript/nativescript-cli/issues/3202


KristianDD:
run ci

KristianDD:
run ci

KristianDD:
run ci

KristianDD:
run ci

lib/commands/update.ts
```diff
@@ -66,42 +51,79 @@ export class UpdateCommand implements ICommand {
...
 
-	private async executeCore(args: string[], folders: string[]): Promise<void> {
-		let platforms = this.$platformService.getInstalledPlatforms(this.$projectData);
-		const availablePlatforms = this.$platformService.getAvailablePlatforms(this.$projectData);
-		const packagePlatforms: string[] = [];
+	private async executeCore(args: string[]): Promise<void> {
+		const platforms = this.getPlatforms();
 
-		for (const platform of availablePlatforms) {
+		for (const platform of _.xor(platforms.installed, platforms.packagePlatforms)) {
```
Mitko-Kerezov: The logic has changed a bit here - instead of iterating over all `availablePlatforms` we are basically iterating over `packagePlatforms` because `platforms.packagePlatforms` is `platforms.installed` concatenated with `packagePlatforms` and the `xor` will basically only return `packagePlatforms`.
Is this expected

KristianDD: Yes this is intentional, because we are missing the following part:
```typescript
if (platformVersion) {
 	packagePlatforms.push(platform);
 	this.$projectDataService.removeNSProperty(this.$projectData.projectDir, platformData.frameworkPackageName);
}
```

The if statement is moved to `getPlatforms` and we want to call `removeNSProperty` only for the ones that passed the condition.

The behavior should be the same.

KristianDD:
run ci

lib/commands/update.ts
```diff
@@ -15,49 +15,34 @@ export class UpdateCommand implements ICommand {
 		this.$projectData.initializeProjectData();
 	}
 
+	static readonly folders: string[] = ["lib", "hooks", "platforms", "node_modules"];
```
Plamen5kov: you have constants required anyway, please add `lib` and `hooks` there and replace all strings with constants

lib/commands/update.ts
```diff
@@ -15,49 +15,34 @@ export class UpdateCommand implements ICommand {
...
-			shelljs.rm("-fr", tmpDir);
-			shelljs.mkdir(tmpDir);
-			shelljs.cp(path.join(this.$projectData.projectDir, "package.json"), tmpDir);
-			for (const folder of folders) {
-				const folderToCopy = path.join(this.$projectData.projectDir, folder);
-				if (this.$fs.exists(folderToCopy)) {
-					shelljs.cp("-rf", folderToCopy, tmpDir);
-				}
-			}
+			this.backup(tmpDir);
```
Plamen5kov: better API to understand is this.backup(source, destination), or if you use only source, why do you need a temp folder?

test/update.ts
```diff
@@ -0,0 +1,247 @@
...
+			return updateCommand.execute(["3.3.0"]).then(() => {
+				assert.isTrue(copyFileStub.calledWith(path.join(tempDir, "package.json"), projectFolder));
+				for (const folder of UpdateCommand.folders) {
+					assert.isTrue(deleteDirectoryStub.calledWith(path.join(projectFolder, folder)));
+					assert.isTrue(copyFileStub.calledWith(path.join(tempDir, folder), projectFolder));
+				}
+			});
+		});
+
+		it("calls remove for all falders", async () => {
```
Plamen5kov: typo: falders

lib/commands/update.ts
KristianDD: I send the parameter just to avoid repetition of:
`const tmpDir = path.join(this.$projectData.projectDir, UpdateCommand.tempFolder);`

PR #3230, opened 2017-11-24
tns commands will print error sources with line and column information pointing to .ts files

PanayotCankov:


bin/tns
```diff
@@ -1,6 +1,9 @@
 #!/usr/bin/env node
 
 "use strict";
+
+require("source-map-support/register");
```
rosen-vladimirov: I believe this will fail if CLI is packed, as we do not have dependency to `source-map-support`

package.json
```diff
@@ -105,7 +105,7 @@
     "istanbul": "0.4.5",
     "mocha": "3.1.2",
     "should": "7.0.2",
-    "source-map-support": "^0.4.14",
+    "source-map-support": "^0.4.18",
```
rosen-vladimirov: this should be strict version and maybe this should not be devDependency

PanayotCankov: So moving this to dependencies should also resolve the previous comment right?

Plamen5kov:
Please add description to show why this is necessary.

dtopuzov:
run ci

dtopuzov:
run ci

dtopuzov:
run ci

dtopuzov:
run ci

dtopuzov:
run ci

rosen-vladimirov:
I'm closing this PR as the proposed change will break CLI's execution when installed from npm. We do not distribute source-maps, so even if we move the source-map plugin to dependencies section, the execution will still fail at runtime.

PR #3232, opened 2017-11-27
Check for changes fails on infinite deep directories after symlink

PanayotCankov:
Check for changes fails where:
```
nativescript-dev-plugin/
  demo/SampleApp/
    node_modules/nativescript-dev-plugin (symlink to the root dir)
    app/...
```
The check for changes follows to a infinite deep directory:
`naitvescript-dev-plugin/demo/SampleApp/node_modules/nativescript-dev-plugin/demo/SampleApp/node_modules...`
It will now stop the DFS when it finds it have visited already the file or directory's real path.

### DIFF
--- a/lib/services/project-changes-service.ts
+++ b/lib/services/project-changes-service.ts
@@ -64,11 +64,14 @@ export class ProjectChangesService implements IProjectChangesService {
 			this._changesInfo.packageChanged = this.isProjectFileChanged(projectData, platform);
 			this._changesInfo.appResourcesChanged = this.containsNewerFiles(projectData.appResourcesDirectoryPath, null, projectData);
 			/*done because currently all node_modules are traversed, a possible improvement could be traversing only the production dependencies*/
+			const visitedPaths = new Set<string>();
+			visitedPaths.add(projectData.projectDir);
 			this._changesInfo.nativeChanged = this.containsNewerFiles(
 				path.join(projectData.projectDir, NODE_MODULES_FOLDER_NAME),
 				path.join(projectData.projectDir, NODE_MODULES_FOLDER_NAME, "tns-ios-inspector"),
 				projectData,
-				this.fileChangeRequiresBuild);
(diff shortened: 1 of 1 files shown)


Plamen5kov:
Please add an issue if one is present or create one for this PR to explain, what problem are we fixing.

dtopuzov:
run ci

dtopuzov:
run ci

dtopuzov:
run ci

dtopuzov:
run ci

dtopuzov:
run ci

dtopuzov:
run ci

KristianDD:
@Mitko-Kerezov @rosen-vladimirov

lib/services/project-changes-service.ts
```diff
@@ -292,7 +308,7 @@ export class ProjectChangesService implements IProjectChangesService {
 				const fullFilePath = path.join(projectDir, path.join(filePath, "package.json"));
 				if (this.$fs.exists(fullFilePath)) {
 					const json = this.$fs.readJson(fullFilePath);
-					if (json["nativescript"] && _.startsWith(file, path.join(filePath, "platforms"))) {
+					if (json.nativescript && json.nativescript.platforms && _.startsWith(file, path.join(filePath, "platforms"))) {
```
sis0k0: Is that still necessary after we added the project directory to the map with visited paths?

KristianDD: Well not "necessary" but still sounds like the proper way to detect plugins. Can revert the change if you prefer so.

Mitko-Kerezov:
@KristianDD 
This PR seems to have a significant performance impact on projects which have a large amount of files.
Currently on average `checkForChanges` executes for about 3.5 seconds whereas with this PR it is around 17 seconds.

rosen-vladimirov:
Closing this PR as the suggested solution causes too much issues.

PR #3233, opened 2017-11-27
Expose API for getting installed applications

rosen-vladimirov:
Expose API for getting installed applications for a specific device when CLI is required as library. Add the new method to `devicesService`.
Add required tests for exposing methods from `devicesService`.


rosen-vladimirov:
run ci

PR #3237, opened 2017-12-01
ignore .(anything) files when checking for changes

Plamen5kov:
When checking if files are changed we should ignore all files ending in `.(something)` as these files are hidden on mac and usually shouldn't be modified anyway.
This will make livesync work a bit faster when `--sync-all-files` flag is used, because there are a number of `.(something)` directories inside.

### DIFF
--- a/lib/services/project-changes-service.ts
+++ b/lib/services/project-changes-service.ts
@@ -219,6 +219,12 @@ export class ProjectChangesService implements IProjectChangesService {
 	}
 
 	private containsNewerFiles(dir: string, skipDir: string, projectData: IProjectData, processFunc?: (filePath: string, projectData: IProjectData) => boolean): boolean {
+
+		const dirName = path.basename(dir);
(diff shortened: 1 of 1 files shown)


lib/services/project-changes-service.ts
```diff
@@ -219,6 +219,12 @@ export class ProjectChangesService implements IProjectChangesService {
 	}
 
 	private containsNewerFiles(dir: string, skipDir: string, projectData: IProjectData, processFunc?: (filePath: string, projectData: IProjectData) => boolean): boolean {
+
+		// ignore directories and files ending with .something (.bin, .git ...)
+		if (/.*[/\\]\.[^/\\]*$/.test(dir)) {
```
Mitko-Kerezov: While this code is correct I would prefer a non-regex approach. For example:
```TypeScript
const dirName = path.basename(dir);
if (_.startsWith(dirName, '.')) {
    return false;
}
```

I think that a bit more readable.

PR #3253, opened 2017-12-10
Use proxy-lib package when working behind proxy

rosen-vladimirov:
The `proxy-lib` package provides required methods for setting and getting proxy. Its implementation is more secure when stroing credentials, so switch to it instead of current implementation.
Remove the credentials-manager from CLI as it is no longer needed.
Remove the CredentialsManager.exe as well and its source code.

Fix the error `Cannot read property ExceptionMessage of null` when wokring behind proxy - add correct null checks.

Fix tracking to google analytics behind proxy - pass the proxy URL to Universal Analytics module.

### DIFF
--- a/lib/declarations.d.ts
+++ b/lib/declarations.d.ts
@@ -433,14 +433,6 @@ interface IInfoService {
 }
 
 /**
- * Describes standard username/password type credentials.
- */
-interface ICredentials {
-	username: string;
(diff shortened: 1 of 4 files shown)


rosen-vladimirov:
run ci

dtopuzov:
run ci

PR #3265, opened 2017-12-12
Clean app directory when bundle flag is switched

Mitko-Kerezov:
When switching the value of `--bundle` (i.e. passing it or not passing it) between two builds the latter build should clear the app directory inside platforms prior to building.


Mitko-Kerezov:
run ci

PR #3269, opened 2017-12-13
Always clean platforms/app when --bundle is passed

Mitko-Kerezov:
Includes:
* Always clean platforms/app when `--bundle` is passed because we may end up with duplicate symbols
* Do not use `$options` in `$nodeModulesBuilder` as CLI may be `require`d as a library where `$options` is not used at all

Ping @rosen-vladimirov @sis0k0

### DIFF
--- a/lib/definitions/platform.d.ts
+++ b/lib/definitions/platform.d.ts
@@ -271,15 +271,15 @@ interface IAppFilesUpdaterOptionsComposition {
 	appFilesUpdaterOptions: IAppFilesUpdaterOptions;
 }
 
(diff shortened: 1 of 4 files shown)


Plamen5kov:
Tested with: 
"tns-core-modules": "^4.0.0-2017-12-12-2"
"tns-android": "3.4.0-2017-12-6-3"
"nativescript-dev-webpack": https://github.com/NativeScript/nativescript-dev-webpack/pull/350

Tested commands in the following order:
```
tns run android
tns run android --bundle --env.snapshot
tns run android --bundle
tns run android
tns run android --bundle
tns run android --bundle --env.snapshot
tns run android
```

and it builds every and runs as expected with no excess files in `platforms/android/app/src/main/assets/app`.

Issue #3183, opened 2017-10-27
tns debug ios --chrome keeps changing port on every livesync update


recon472:
Hi,

### Tell us about the problem
When I run "tns debug ios --chrome" or run the debug via sidekick, everytime livesync updates the app when the code is changed, the chrome devtools port changes. So you have to copy the url again after every change. When debugging android, the port is always the same.

### Which platform(s) does your issue occur on?
iOS

### Please provide the following version numbers that your issue occurs with:
- CLI: 3.3.0
- Cross-platform modules: 3.3.0
- Runtime(s): 3.3.0
- Plugin(s): 
    "nativescript-pro-ui": "3.1.4",
    "nativescript-sqlite": "1.1.11",
    "nativescript-theme-core": "1.0.4",
    "tns-core-modules": "3.3.0"
- Xcode: 9.0.1

### Please tell us how to recreate the issue in as much detail as possible.
run "tns debug ios --chrome"
open chrome with the provided url
change someting in the code
chrome devtools stops working until reloaded with new url from terminal

Best regards
Jacklin

Pip3r4o:
Hey @recon472 there is already a feature request open at https://github.com/NativeScript/nativescript-cli/issues/3171. Please join the discussion at #3171

Pip3r4o:
@recon472 I can confirm that this indeed the current behavior when debugging with Chrome on iOS. 

Unfortunately currently there isn't a workaround, but we'll let you know when we've addressed it!

snovos:
I updated cli to 3.4.1 and still getting same issue

Pip3r4o:
@snovos could you please share the output of the `tns debug ios` command?

snovos:
@Pip3r4o I was able to spin up my app and it works, but when I paste that debug chrome url im getting this 
`Frontend client connected.
Backend socket created.
Backend socket closed!
Frontend socket closed!
Frontend client connected.
Backend socket created.
Backend socket closed!`

Pip3r4o:
@snovos can you make sure that you are also running the 3.4.1 ios runtime, as the CLI and the runtime need to be the latest versions to benefit from the latest inspector fixes.

snovos:
@Pip3r4o  

> "nativescript": {
    "tns-android": {
      "version": "3.4.0"
    },
    "hooks": [
      {
        "type": "before-prepare",
        "script": "nativescript-config.js"
      }
    ],
    "tns-ios": {
      "version": "3.4.0"
[...]

Issue #3261, opened 2017-12-11
`tns debug ios` do not play nicely with livesync


dtopuzov:
`tns debug ios` now by default starts chrome debugging.

Issues with `tns debug ios` in nativescript@3.4.0:

1. Debug URL is not automatically opened in browser (not so critical)
2. Debug URL changes on each file change (critical)

PR #3275, opened 2017-12-18
Generate debug port in range for Chrome DevTools websocket - iOS

Pip3r4o:
When generating a port to start a websocket on, use the same logic as is present in the android debug service - get the first available available port in a range, and reuse it for the next debug session of the same application if it is still available

Fixes https://github.com/NativeScript/nativescript-cli/issues/3183 and https://github.com/NativeScript/nativescript-cli/issues/3261

### DIFF
--- a/test/services/ios-debug-service.ts
+++ b/test/services/ios-debug-service.ts
@@ -48,6 +48,10 @@ const createTestInjector = (): IInjector => {
 		on: (event: string | symbol, listener: Function): any => undefined
 	});
 
+	testInjector.register("net", {
+		getAvailablePortInRange: async (startPort: number, endPort?: number): Promise<number> => 41000
+	});
+
 	return testInjector;
 };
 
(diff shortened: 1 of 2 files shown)


test/services/ios-debug-service.ts
```diff
@@ -48,6 +49,12 @@ const createTestInjector = (): IInjector => {
 		on: (event: string | symbol, listener: Function): any => undefined
 	});
 
+	testInjector.register("net", {
+		getAvailablePortInRange: function (startPort: number, endPort?: number): any {
```
rosen-vladimirov: You should return Promise here. 
And also you can use lambda syntax here:
```TypeScript
testInjector.register("net", {
    getAvailablePortInRange: async (startPort: number, endPort?: number): Promise<number> => 41000
}
```

lib/services/ios-debug-service.ts
```diff
@@ -212,7 +213,7 @@ export class IOSDebugService extends DebugServiceBase implements IPlatformDebugS
 				this.$logger.info("'--chrome' is the default behavior. Use --inspector to debug iOS applications using the Safari Web Inspector.");
 			}
 
-			this._socketProxy = await this.$socketProxyFactory.createWebSocketProxy(this.getSocketFactory(device));
+			this._socketProxy = await this.$net.getAvailablePortInRange(41000);
```
KristianDD: This bypasses a lot of logic, please check it.

Pip3r4o: Right, must have pushed the wrong thing

Issue #3268, opened 2017-12-12
tns run android - EMFILE: too many open files


JILeXanDR:
`tns --version` **3.3.1**

Command's `tns run android` output is:
```
Searching for devices...
Skipping prepare.
Building project...
Gradle build...
         + applying user-defined configuration from /home/alexandr/projects/MyApp/app/App_Resources/Android/app.gradle
         + creating include.gradle for plugin /home/alexandr/projects/MyApp/node_modules/tns-core-modules-widgets
         + applying plugin configuration from /home/alexandr/projects/MyApp/platforms/android/build/configurations/tns-core-modules-widgets/include.gradle
         + adding nativescript runtime package dependency: nativescript-optimized
         + adding aar plugin dependency: /home/alexandr/projects/MyApp/node_modules/tns-core-modules-widgets/platforms/android/widgets-release.aar
Running full build
Project successfully built.
Installing...
Successfully installed on device with identifier 'd5a7d6b70404'.
Unable to apply changes on device: d5a7d6b70404. Error is: EMFILE: too many open files, open '/home/alexandr/projects/MyApp/platforms/android/src/main/assets/app/tns_modules/rxjs/_esm5/operators/retry.js.map'.
The project file undefined is corrupted. 
Consider restoring an earlier version from your source control or backup.
Additional technical info: Error: EMFILE: too many open files, open '/home/alexandr/projects/MyApp/package.json'
Failed to display command help { Error: EMFILE: too many open files, scandir '/home/alexandr/.nvm/versions/node/v7.5.0/lib/node_modules/nativescript/docs/man_pages'
    at Object.fs.readdirSync (fs.js:871:18)
    at FileSystem.readDirectory (/home/alexandr/.nvm/versions/node/v7.5.0/lib/node_modules/nativescript/lib/common/file-system.js:153:19)
    at FileSystem.enumerateFilesInDirectorySync (/home/alexandr/.nvm/versions/node/v7.5.0/lib/node_modules/nativescript/lib/common/file-system.js:275:31)
    at HelpService.<anonymous> (/home/alexandr/.nvm/versions/node/v7.5.0/lib/node_modules/nativescript/lib/common/services/help-service.js:136:50)
    at Generator.next (<anonymous>)
    at fulfilled (/home/alexandr/.nvm/versions/node/v7.5.0/lib/node_modules/nativescript/lib/common/services/help-service.js:4:58)
  errno: -24,
  code: 'EMFILE',
[...]

PR #3277, opened 2017-12-19
Fix EMFILE error when running on Android

rosen-vladimirov:
During running on Android, CLI tries to take smart decision which files to upload. This is done by checking the shasum of all project files and comparing it with the latest one uploaded to device. However, in huge projects, this leads to error with code EMFILE as the OS have limits for opened file handles.
This happens as CLI executes the operation simultaneously for all files. In order to fix this behavior, execute the actions by chunks (currently set to 100).
Introduce new method in helpers to execute this action and add tests for it.

Fixes https://github.com/NativeScript/nativescript-angular/issues/1079 and https://github.com/NativeScript/nativescript-cli/issues/3268

> NOTE: Merge after https://github.com/telerik/mobile-cli-lib/pull/1036


dtopuzov:
run ci

PR #3281, opened 2017-12-21
Generate debug port in range for Chrome DevTools websocket - iOS (master branch)

Pip3r4o:
Same as #3275, targets master

When generating a port to start a websocket on, use the same logic as is present in the android debug service - get the first available available port in a range, and reuse it for the next debug session of the same application if it is still available


Pip3r4o:
run ci

PR #3294, opened 2017-12-29
Add check for JAVA 9

rosen-vladimirov:
Currently Gradle cannot work with JAVA 9, so detect if it has been used and break the build for Android. Also the check will print warning when `tns doctor` is called.

### Handle case when Javac version is a single number (9 for example)

 In some cases javac version is a single number, for example 9. In this case our validation fails to detect the correct version and to check if we support this version.
In order to resolve this issue, use the `appendZeroesToVersion` method in order to make the versin semver valid.
Change the return type of `validateJavacVersion` method - it does not require to return a Promise.
Add unit tests for `validateJavacVersion` method.

Also update submodule, where the following change is applied:

### Fix detection of Javac version

The command `javac -version` prints result to stderr when JAVA 8 is used and to stdout when JAVA 9 is used. Current check in CLI uses the stderr output, so when JAVA 9 is installed it fails to detect the correct version.
In order to support both JAVA 8 and JAVA 9, capture both stdout and stderr and get the version from there.
Also remove unneeded check for Java version - we care about JAVA Compiler, which is included in JDK.

Merge after https://github.com/telerik/mobile-cli-lib/pull/1037

### DIFF
--- a/lib/android-tools-info.ts
+++ b/lib/android-tools-info.ts
@@ -2,6 +2,7 @@ import * as path from "path";
 import * as semver from "semver";
 import { EOL } from "os";
 import { cache } from "./common/decorators";
+import { appendZeroesToVersion } from './common/helpers';
 
 export class AndroidToolsInfo implements IAndroidToolsInfo {
 	private static ANDROID_TARGET_PREFIX = "android";
@@ -102,7 +103,7 @@ export class AndroidToolsInfo implements IAndroidToolsInfo {
 		return detectedErrors || !isAndroidHomeValid;
 	}
 
-	public async validateJavacVersion(installedJavaVersion: string, options?: { showWarningsAsErrors: boolean }): Promise<boolean> {
+	public validateJavacVersion(installedJavacVersion: string, options?: { showWarningsAsErrors: boolean }): boolean {
 		let hasProblemWithJavaVersion = false;
 		if (options) {
 			this.showWarningsAsErrors = options.showWarningsAsErrors;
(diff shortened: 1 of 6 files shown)


lib/android-tools-info.ts
```diff
@@ -111,10 +112,14 @@ export class AndroidToolsInfo implements IAndroidToolsInfo {
...
 			" described in " + this.$staticConfig.SYS_REQUIREMENTS_LINK;
 		const matchingVersion = (installedJavaVersion || "").match(AndroidToolsInfo.VERSION_REGEX);
-		if (matchingVersion && matchingVersion[1]) {
-			if (semver.lt(matchingVersion[1], AndroidToolsInfo.MIN_JAVA_VERSION)) {
+		const installedJavaCompilerVersion = matchingVersion && matchingVersion[1];
+		if (installedJavaCompilerVersion) {
+			if (semver.lt(installedJavaCompilerVersion, AndroidToolsInfo.MIN_JAVA_VERSION)) {
 				hasProblemWithJavaVersion = true;
 				this.printMessage(`Javac version ${installedJavaVersion} is not supported. You have to install at least ${AndroidToolsInfo.MIN_JAVA_VERSION}.`, additionalMessage);
+			} else if (semver.gt(installedJavaCompilerVersion, AndroidToolsInfo.MAX_JAVA_VERSION)) {
```
rosen-vladimirov: this should be `gte`

test/android-tools-info.ts
```diff
@@ -0,0 +1,113 @@
...
+			}
+		];
+
+		_.each(testData, ({ javacVersion, expectedResult, warnings }) => {
+			it(`returns ${expectedResult} when version is ${javacVersion}`, () => {
+				const testInjector = createTestInjector();
+				const androidToolsInfo = testInjector.resolve<IAndroidToolsInfo>(AndroidToolsInfo);
+				assert.deepEqual(androidToolsInfo.validateJavacVersion(javacVersion), expectedResult);
+				if (warnings && warnings.length) {
+
```
Mitko-Kerezov: We could do without this new line

PR #3302, opened 2018-01-05
fix(ios-inspector): Update cached inspector package to latest compatible version

mbektchiev:
* Currently it is never updated and is left at the first version ever installed in `~/.local`.
As a result newly published inspector packages are not used unless the user installs
them explicitly in the project or deletes the cache manually.
* The iOS inspector package version must be determined according to project's iOS runtime
version instead of the version of CLI as it has been till now

### DIFF
--- a/test/npm-installation-manager.ts
+++ b/test/npm-installation-manager.ts
@@ -10,6 +10,7 @@ import * as StaticConfigLib from "../lib/config";
 import * as yok from "../lib/common/yok";
 import ChildProcessLib = require("../lib/common/child-process");
 import { SettingsService } from "../lib/common/test/unit-tests/stubs";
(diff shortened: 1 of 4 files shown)


test/npm-installation-manager.ts
```diff
@@ -10,6 +10,7 @@ import * as StaticConfigLib from "../lib/config";
 import * as yok from "../lib/common/yok";
 import ChildProcessLib = require("../lib/common/child-process");
 import { SettingsService } from "../lib/common/test/unit-tests/stubs";
+import { ProjectDataService } from "../lib//services/project-data-service";
```
rosen-vladimirov: `../lib//services/` should be `../lib/services/`

PR #3307, opened 2018-01-08
Revert version to 3.4.1

ginev:
Revert CLI version to 3.4.1.


rosen-vladimirov:
Why do we want this change?

PR #3311, opened 2018-01-09
Update proxy-lib to 0.4.0

rosen-vladimirov:
Update proxy-lib to latest version - 0.4.0 - the major change in this version is that it allows using it without having Visual C++ redistributable installed on the users Windows machine.


rosen-vladimirov:
run ci

PR #3313, opened 2018-01-09
Allow using Android SDK 27

rosen-vladimirov:
Add version 27 of Android SDK as verified, so users will use it by default when it is installed.

### Use latest available appCompat version

The current CLI logic finds appCompat version that matches the selected SDK version. However, the latest appCompat version is 26.0.0-alpha, so in case we have Android SDK 27, we are unable to find matchin appCompat.
In order to resolve the issue and allow using Android SDK 27, get latest available appCompat version when there's no matching one.


mbektchiev:
run ci

miroslavaivanova:
run ci

PR #3329, opened 2018-01-22
feat: make app_resources's path configurable

Pip3r4o:
Refactor the project-data service to read a config.json for the 'appResources' property entry, which points relatively to the App_Resources directory.

* All references to app/App_Resources now retrieve the path from the said service.

* To keep the service backward compatible, if no config.json is present, or the `appResources` entry isn't available, location defaults to `projectDir/app/App_Resources`.

* Change the app-files-updater class to not collect the App_Resources directory when enumerating through the user's source files. Instead a separate method will enumerate the App_Resources directory and its files, and will copy it in `platforms/**/app`. I find that behavior to be more predictable, otherwise enumerating just the `myProject/app` directory, could skip enumerating the App_Resources if they are located someplace else, as declared in the nsconfig.json file.

### DIFF
--- a/lib/constants.ts
+++ b/lib/constants.ts
@@ -29,7 +29,7 @@ export const OUTPUTS_DIR = "outputs";
 export const APK_DIR = "apk";
 export const RESOURCES_DIR = "res";
 export const CONFIG_NS_FILE_NAME = "nsconfig.json";
-export const CONFIG_NS_APP_RESOURCES_ENTRY = "app_resources";
+export const CONFIG_NS_APP_RESOURCES_ENTRY = "appResourcesPath";
 
 export class PackageVersion {
 	static NEXT = "next";
--- a/lib/constants.ts
+++ b/lib/constants.ts
@@ -28,6 +28,8 @@ export const BUILD_DIR = "build";
 export const OUTPUTS_DIR = "outputs";
 export const APK_DIR = "apk";
 export const RESOURCES_DIR = "res";
+export const CONFIG_NS_FILE_NAME = "nsconfig.json";
+export const CONFIG_NS_APP_RESOURCES_ENTRY = "app_resources";
 
 export class PackageVersion {
 	static NEXT = "next";
--- a/lib/project-data.ts
+++ b/lib/project-data.ts
@@ -69,7 +69,6 @@ export class ProjectData implements IProjectData {
 					this.platformsDir = path.join(projectDir, constants.PLATFORMS_DIR_NAME);
 					this.projectFilePath = projectFilePath;
 					this.appDirectoryPath = path.join(projectDir, constants.APP_FOLDER_NAME);
-					this.appResourcesDirectoryPath = path.join(projectDir, constants.APP_FOLDER_NAME, constants.APP_RESOURCES_FOLDER_NAME);
 					this.projectId = data.id;
 					this.dependencies = fileContent.dependencies;
 					this.devDependencies = fileContent.devDependencies;
@@ -85,6 +84,27 @@ export class ProjectData implements IProjectData {
 
 		// This is the case when no project file found
 		this.$errors.fail("No project found at or above '%s' and neither was a --path specified.", projectDir || this.$options.path || currentDir);
+	}
+
+	public getAppResourcesDirectoryPath(projectDir?: string): string {
+		if (!projectDir) {
+			projectDir = this.projectDir;
+		}
+
+		const configNSFilePath = path.join(projectDir, constants.CONFIG_NS_FILE_NAME);
+		let absoluteAppResourcesDirPath: string;
+
+		if (this.$fs.exists(configNSFilePath)) {
+			const configNS = this.$fs.readJson(configNSFilePath);
+
+			if (configNS && configNS[constants.CONFIG_NS_APP_RESOURCES_ENTRY]) {
+				const appResourcesDirPath = configNS[constants.CONFIG_NS_APP_RESOURCES_ENTRY];
+
+				absoluteAppResourcesDirPath = path.resolve(projectDir, appResourcesDirPath);
+			}
+		}
+
+		return absoluteAppResourcesDirPath || path.join(projectDir, constants.APP_FOLDER_NAME, constants.APP_RESOURCES_FOLDER_NAME);
 	}
 
 	private getProjectType(): string {
--- a/lib/project-data.ts
+++ b/lib/project-data.ts
@@ -33,7 +33,9 @@ export class ProjectData implements IProjectData {
 	public projectId: string;
 	public projectName: string;
 	public appDirectoryPath: string;
-	public appResourcesDirectoryPath: string;
+	get appResourcesDirectoryPath(): string {
+		return this.getAppResourcesDirectoryPath();
+	}
 	public dependencies: any;
 	public devDependencies: IStringDictionary;
 	public projectType: string;
--- a/lib/project-data.ts
+++ b/lib/project-data.ts
@@ -93,6 +93,10 @@ export class ProjectData implements IProjectData {
 			projectDir = this.projectDir;
 		}
 
+		if (!projectDir) {
+			return null;
+		}
+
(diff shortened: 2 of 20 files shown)


KristianDD:
run ci

KristianDD:
run ci

dtopuzov:
run ci

lib/project-data.ts
```diff
@@ -87,6 +88,27 @@ export class ProjectData implements IProjectData {
...
+		const configNSFilePath = path.join(projectDir, constants.CONFIG_NS_FILE_NAME);
+		let absoluteAppResourcesDirPath: string;
+
+		if (this.$fs.exists(configNSFilePath)) {
+			const configNS = this.$fs.readJson(configNSFilePath);
+
+			if (configNS && configNS[constants.CONFIG_NS_APP_RESOURCES_ENTRY]) {
+				const appResourcesDirPath = configNS[constants.CONFIG_NS_APP_RESOURCES_ENTRY];
+
+				absoluteAppResourcesDirPath = path.resolve(projectDir, appResourcesDirPath);
```
rosen-vladimirov: is it worth checking if this path exists?

lib/project-data.ts
```diff
@@ -87,6 +88,27 @@ export class ProjectData implements IProjectData {
 		this.$errors.fail("No project found at or above '%s' and neither was a --path specified.", projectDir || this.$options.path || currentDir);
 	}
 
+	public getAppResourcesDirectoryPath(projectDir?: string): string {
+		if (!projectDir) {
+			projectDir = this.projectDir;
```
rosen-vladimirov: if `initializeProjectData` has not been called, the `this.projectDir` will also be undefined and the path.join below will fail.

lib/constants.ts
```diff
@@ -28,6 +28,8 @@ export const BUILD_DIR = "build";
 export const OUTPUTS_DIR = "outputs";
 export const APK_DIR = "apk";
 export const RESOURCES_DIR = "res";
+export const CONFIG_NS_FILE_NAME = "nsconfig.json";
+export const CONFIG_NS_APP_RESOURCES_ENTRY = "app_resources";
```
rosen-vladimirov: I prefer naming json properties in camelCase (appResources in this case), but I do not have a strong opinion here. @KristianDD , @Mitko-Kerezov  what do you think?

lib/services/prepare-platform-js-service.ts
```diff
@@ -36,6 +36,7 @@ export class PreparePlatformJSService extends PreparePlatformService implements
 	public async preparePlatform(config: IPreparePlatformJSInfo): Promise<void> {
 		if (!config.changesInfo || config.changesInfo.appFilesChanged || config.changesInfo.changesRequirePrepare) {
 			await this.copyAppFiles(config);
+			this.copyAppResourcesFiles(config);
```
rosen-vladimirov: I assume this is related to https://github.com/NativeScript/nativescript-cli/pull/3329/files#diff-d7315453ac31262b45d0c2530d69cd94R47 change. Can you describe a little bit more what is the idea

rosen-vladimirov:
@Pip3r4o I've tried the following commands with the current branch and some warnings are shown from gradle:
```
$ tns create myApp
$ cd myApp
$ tns build android
$ mv app/App_Resources my_app_resources
$ tns build android -
Warning is printed: 
No manifest found in `<path to project>\app\App_Resources\Android\AndroidManifest.xml`, but build continues.
$ vi nsconfig.json // set the following content { "app_resources": "my_app_resources" }
$ tns build android
Preparing project...
Project successfully prepared (android)
Building project...
Gradle build...
         + couldn't load user-defined configuration from D:\Work\nativescript-cli\scratch\app74\app\App_Resources\Android\app.gradle. File doesn't exist.
         + adding nativescript runtime package dependency: nativescript-optimized
         + adding aar plugin dependency: D:\Work\nativescript-cli\scratch\app74\node_modules\tns-core-modules-widgets\platforms\android\widgets-release.aar
false
Running full build
Project successfully built.

```

lib/project-data.ts
Pip3r4o: Correct, I am unsure how to tackle this. The need for the additional `projectDir` parameter arose when using the getter during project creation. What would you recommend I check here?

Pip3r4o:
@rosen-vladimirov the gradle build script from the android-runtime release and master branches currently reads the `app.gradle` from a hardcoded path to the app_resources. There's an additional PR that uses a method similar to `getAppResourcesDirectoryPath` in order to locate the App_Resources dir.

lib/project-data.ts
rosen-vladimirov: I would just return null here. The caller should decide what to do.

lib/constants.ts
Pip3r4o: I'll follow the convention seemingly established in package.json and tsconfig and go with camelCase.

test/stubs.ts
```diff
@@ -244,16 +248,40 @@ export class ProjectDataStub implements IProjectData {
...
 	devDependencies: IStringDictionary;
 	projectType: string;
-	initializeProjectData(projectDir?: string): void {
+	get appResourcesDirectoryPath(): string {
+		return this.getAppResourcesDirectoryPath();
+	}
+	public initializeProjectData(projectDir?: string): void {
 		this.projectDir = this.projectDir || projectDir;
 	}
+	public getAppResourcesDirectoryPath(projectDir?: string): string {
```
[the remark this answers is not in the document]
Pip3r4o: I found it easier to test this way, since I would always get the default path to the App_Resources. Ideally I could always resolve the path to the default app/App_Resources for the purposes of testing.

Pip3r4o:
@rosen-vladimirov @KristianDD I made some changes after taking your comments into consideration. Please review.

lib/project-data.ts
```diff
@@ -103,10 +107,17 @@ export class ProjectData implements IProjectData {
 				const appResourcesDirPath = configNS[constants.CONFIG_NS_APP_RESOURCES_ENTRY];
 
 				absoluteAppResourcesDirPath = path.resolve(projectDir, appResourcesDirPath);
+
+				if (this.$fs.exists(absoluteAppResourcesDirPath)) {
```
KristianDD: Check what happens if we go throu here https://github.com/NativeScript/nativescript-cli/blob/27e937b71f27234a62d02c3ace7d1896aaa7d05e/lib/services/project-service.ts#L113

lib/services/app-files-updater.ts
```diff
@@ -61,9 +65,12 @@ export class AppFilesUpdater {
 			constants.LIVESYNC_EXCLUDED_FILE_PATTERNS.forEach(pattern => sourceFiles = sourceFiles.filter(file => !minimatch(file, pattern, { nocase: true })));
 		}
 
-		if (this.options.bundle) {
-			sourceFiles = sourceFiles.filter(file => minimatch(file, "**/App_Resources/**", { nocase: true }));
-		}
+		// exclude the app_resources directory from being enumerated
+		// for copying if it is present in the application sources dir
+		const appResourcesPath = projectData.appResourcesDirectoryPath;
+		const appResourcesDirInSources = path.join(this.appSourceDirectoryPath, path.basename(appResourcesPath));
```
KristianDD: Check if it works with `appResourcesPath` directly.

lib/project-data.ts
Pip3r4o: @KristianDD 
Creating an ng/ts project, which would suggest that 2 templates will be downloaded - one before `ensureAppResourcesExist` is called, the other one - afterwards, doesn't make the process crash. I am not sure if the case would be the same if the template is shipped with nsconfig.json which points to App_Resources.

What do you think? Should I perhaps return the path, even if the directory doesn't exist, as that's where the App_Resources from the template should be extracted?

Pip3r4o:
Closing this PR in favor of https://github.com/NativeScript/nativescript-cli/pull/3356 which includes the changes presented here.

PR #3332, opened 2018-01-23
Fix multiple typescript/sass watchers

Mitko-Kerezov:
Whenever using CLI as a library, calling livesync multiple times leads to multiple typescript/sass watchers.
This happens due to the fact that `currentWatcherInfo.pattern` is an array and checking two arrays for equality with `!==` is bound to return `true` at all times. `toString()` both arrays and check the string literals instead.

Ping @KristianDD @rosen-vladimirov

### DIFF
--- a/lib/definitions/livesync.d.ts
+++ b/lib/definitions/livesync.d.ts
@@ -60,7 +60,7 @@ interface ILiveSyncProcessInfo {
 	timer: NodeJS.Timer;
 	watcherInfo: {
 		watcher: IFSWatcher,
-		pattern: string | string[]
+		pattern: string[]
 	};
 	actionsChain: Promise<any>;
 	isStopped: boolean;
--- a/lib/definitions/livesync.d.ts
+++ b/lib/definitions/livesync.d.ts
@@ -60,7 +60,7 @@ interface ILiveSyncProcessInfo {
 	timer: NodeJS.Timer;
 	watcherInfo: {
 		watcher: IFSWatcher,
-		pattern: string[]
+		patterns: string[]
 	};
 	actionsChain: Promise<any>;
 	isStopped: boolean;
--- a/lib/services/livesync/livesync-service.ts
+++ b/lib/services/livesync/livesync-service.ts
@@ -526,8 +526,8 @@ export class LiveSyncService extends EventEmitter implements IDebugLiveSyncServi
 		}
 
 		const currentWatcherInfo = this.liveSyncProcessesInfo[liveSyncData.projectDir].watcherInfo;
-
-		if (!currentWatcherInfo || currentWatcherInfo.pattern !== pattern) {
+		const areWatcherPatternsDifferent = () => _.difference(currentWatcherInfo.pattern, pattern).length || _.difference(pattern, currentWatcherInfo.pattern).length;
+		if (!currentWatcherInfo || areWatcherPatternsDifferent()) {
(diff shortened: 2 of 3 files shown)


lib/services/livesync/livesync-service.ts
```diff
@@ -527,7 +527,7 @@ export class LiveSyncService extends EventEmitter implements IDebugLiveSyncServi
 
 		const currentWatcherInfo = this.liveSyncProcessesInfo[liveSyncData.projectDir].watcherInfo;
 
-		if (!currentWatcherInfo || currentWatcherInfo.pattern !== pattern) {
+		if (!currentWatcherInfo || currentWatcherInfo.pattern.toString() !== pattern.toString()) {
```
KristianDD: It might be good to sort the array (if its an array) so that the order of patterns doesn't matter.

Mitko-Kerezov: Actually, `pattern` is always an array as it can be seen [here](https://github.com/NativeScript/nativescript-cli/blob/ab4e6e43d7be6e3d7b5aff5e1d83cf8a9be333db/lib/services/livesync/livesync-service.ts#L516). `currentWatcherInfo.pattern` on the other hand is always an array too as [it is actually the same `pattern` ](https://github.com/NativeScript/nativescript-cli/blob/ab4e6e43d7be6e3d7b5aff5e1d83cf8a9be333db/lib/services/livesync/livesync-service.ts#L653).

So there is no need to sort the array, as both arrays always have the same order.

@KristianDD what do you think?

KristianDD: But as seen [here](https://github.com/NativeScript/nativescript-cli/blob/013ed913482a441f30c24abfcae98a34dc6a0dd9/lib/definitions/livesync.d.ts#L63) it can be a string. And as I will add configurations for livesync the patterns might change.

Mitko-Kerezov:
Ping @KristianDD and @rosen-vladimirov. In light of @KristianDD's suggestion I decided to take a different approach to this.

lib/services/livesync/livesync-service.ts
```diff
@@ -526,8 +526,8 @@ export class LiveSyncService extends EventEmitter implements IDebugLiveSyncServi
 		}
 
 		const currentWatcherInfo = this.liveSyncProcessesInfo[liveSyncData.projectDir].watcherInfo;
-
-		if (!currentWatcherInfo || currentWatcherInfo.pattern !== pattern) {
+		const areWatcherPatternsDifferent = () => _.difference(currentWatcherInfo.pattern, pattern).length || _.difference(pattern, currentWatcherInfo.pattern).length;
```
KristianDD: You might use `_.xor(currentWatcherInfo.pattern, pattern).length` (not sure, should be checked).

lib/definitions/livesync.d.ts
```diff
@@ -60,7 +60,7 @@ interface ILiveSyncProcessInfo {
 	timer: NodeJS.Timer;
 	watcherInfo: {
 		watcher: IFSWatcher,
-		pattern: string | string[]
+		pattern: string[]
```
KristianDD: Optional: would look better if the property is called `patterns`.

lib/services/livesync/livesync-service.ts
Mitko-Kerezov: Was looking for this, thanks!

PR #3334, opened 2018-01-23
Fix livesync + debug on iOS Simulator

rosen-vladimirov:
When `tns debug ios` is used, each change of application's code will trigger restart of app and Chrome DevTools will be disconnected. CLI prints the "new" URL and in case you copy it and paste it immediately in the browser, you'll receive empty DevTools. The problem is that the URL is printed before application is launched successfully on simulator.
In order to resolve this issue, try connection on port 18181 - the connection will be successfull only when the application is running. Wait 10 seconds to connect and throw error in case we are unable to connect for this time.

Also fix incorrect behavior of connectionError event - the thrown error should contain the deviceIdentifier, but it was missing.

NOTE: Merge after https://github.com/telerik/mobile-cli-lib/pull/1041

### DIFF
--- a/lib/services/ios-debug-service.ts
+++ b/lib/services/ios-debug-service.ts
@@ -23,7 +23,7 @@ export class IOSDebugService extends DebugServiceBase implements IPlatformDebugS
 	constructor(protected device: Mobile.IDevice,
 		protected $devicesService: Mobile.IDevicesService,
 		private $platformService: IPlatformService,
-		private $iOSEmulatorServices: Mobile.IEmulatorPlatformServices,
+		private $iOSEmulatorServices: Mobile.IiOSSimulatorService,
 		private $childProcess: IChildProcess,
 		private $hostInfo: IHostInfo,
 		private $logger: ILogger,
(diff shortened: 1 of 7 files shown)


lib/services/ios-debug-service.ts
```diff
@@ -212,7 +225,7 @@ export class IOSDebugService extends DebugServiceBase implements IPlatformDebugS
 				this.$logger.info("'--chrome' is the default behavior. Use --inspector to debug iOS applications using the Safari Web Inspector.");
 			}
 
-			this._socketProxy = await this.$socketProxyFactory.createWebSocketProxy(this.getSocketFactory(device));
+			this._socketProxy = await this.$socketProxyFactory.createWebSocketProxy(this.getSocketFactory(device), device.deviceInfo.identifier);
```
rosen-vladimirov: device is not a mandatory argument, so you may receive `Cannot read property 'deviceInfo' of undefined.`

rosen-vladimirov:
run ci

PR #3339, opened 2018-01-26
feat(build): introduce cleanApp hook

Mitko-Kerezov:
Introduce a `cleanApp` hook which occurs whenever CLI needs to clean the app directory in `platforms`.

### DIFF
--- a/lib/services/platform-service.ts
+++ b/lib/services/platform-service.ts
@@ -560,6 +560,7 @@ export class PlatformService extends EventEmitter implements IPlatformService {
 		return null;
 	}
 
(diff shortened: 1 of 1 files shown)


KristianDD:
It might be good idea to document it as the other hooks. Just in case someone needs to use it(Vue.js integration for example).

rosen-vladimirov:
I agree with @KristianDD  - we need documentation for all hooks and how to use them. The current one is really poor, so I've added a new issue - lets handle it in a separate PR.
https://github.com/NativeScript/nativescript-cli/issues/3341

Issue #3338, opened 2018-01-25
`tns debug ios --chrome` can not stop on first line


dtopuzov:
There is no way to stop on first line with `tns debug ios` and ChromeDevTools.

### Please provide the following version numbers that your issue occurs with:
- CLI: 3.4.2@rc
- Cross-platform modules: 3.4.0
- Runtime(s): 3.4.1
- Plugin(s): No

### Please tell us how to recreate the issue in as much detail as possible.

Steps to reproduce:
```
tns create TestApp
tns debug ios --debug-brk --path TestApp
=> at this point app stops on launch screen
Open debug url in chrome browser
=> app continue and you can not hit the first line
```

Natalia-Hristova:
This issue is fixed and will be available in the next ```CLI``` official release.

PR #3340, opened 2018-01-29
fix(debug): debug-brk option does not work for iOS Simulator

rosen-vladimirov:
The `debug-brk` option was broken in a previous commit as we were trying to connect to the backend port (18181). The problem is that connecting to it makes the runtime thinks debugger is attached and it continues to next statements.
Instead of connecting to the port, check if it is in LISTEN state.
Fixes https://github.com/NativeScript/nativescript-cli/issues/3338

### DIFF
--- a/lib/services/ios-debug-service.ts
+++ b/lib/services/ios-debug-service.ts
@@ -32,7 +32,8 @@ export class IOSDebugService extends DebugServiceBase implements IPlatformDebugS
 		private $iOSNotification: IiOSNotification,
 		private $iOSSocketRequestExecutor: IiOSSocketRequestExecutor,
 		private $processService: IProcessService,
-		private $socketProxyFactory: ISocketProxyFactory) {
+		private $socketProxyFactory: ISocketProxyFactory,
+		private $net: INet) {
 		super(device, $devicesService);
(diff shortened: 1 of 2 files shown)


lib/services/ios-debug-service.ts
```diff
@@ -158,13 +158,13 @@ export class IOSDebugService extends DebugServiceBase implements IPlatformDebugS
...
-		await this.connectToApplicationOnEmulator(debugData.deviceIdentifier);
+		await this.waitForBackendPortToBeOpened(debugData.deviceIdentifier);
 		return result;
 	}
 
-	private async connectToApplicationOnEmulator(deviceIdentifier: string): Promise<void> {
-		const socket = await this.$iOSEmulatorServices.connectToPort({ port: inspectorBackendPort });
-		if (!socket) {
+	private async waitForBackendPortToBeOpened(deviceIdentifier: string): Promise<void> {
+		const portListens = await this.$net.waitForPortToListen(inspectorBackendPort, 10000, 200);
```
Mitko-Kerezov: Introducing an object instead of passing three separate values here would increase readability a lot IMO

Issue #3257, opened 2017-12-11
Make {N} project structure configurable


sis0k0:
Currently, a {N} project has to follow a few conventions in order to be built with {N} CLI. For example, the source code should be in a folder named `app`, the application resources should be in `app/App_Resources` and so on.
The project structure should be configurable per project. For example, a project can have a `.tns-cli.json` file inside the project where the following can be specified:
- app folder name (currently 'app')
- app resources folder name
- folders/files to be included in the app and to be watched during livesync (related to #887)
	- This is particularly important when building apps with Angular 5. The package `@angular/common` contains localization files (~17Mb) which will get deployed on the device even if they are not used.

dolanmiu:
Agreed.

I like to have my projects under `src/app`, just like how `@angular/cli` does it

KristianDD:
Currently a file named "nsconfig.json" will be respected. The valid values inside it are as follows:
```
{
    "appPath": "test/src",
    "appResourcesPath": "resources"
}
```

Both paths must be relative to the project root(where the "package.json" file and "platforms" directory are located) in order everything to work as expected. 

If `appResourcesPath` is omitted the CLI will assume that they are at their default location - folder called App_Resources inside the folder containing the rest of the app files. 

If `appPath` is omitted CLI will assume the application files are located inside a folder called "app" inside the project folder.

Examples:
Lets say the project is located at "/d/work/myapplication".

1 "nsconfig.json" content:
```
{
    "appPath": "test/src"
}
```
Will result in app located at "/d/work/myapplication/test/src" and resources located at "/d/work/myapplication/test/src/App_Resources".

2 "nsconfig.json" content:
```
{
    "appResourcesPath": "resources"
}
```
Will result in app located at "/d/work/myapplication/app" and resources located at "/d/work/myapplication/resources".

3 "nsconfig.json" content:
```
{
    "appPath": "test/src",
    "appResourcesPath": "resources"
}
```
Will result in app located at "/d/work/myapplication/test/src" and resources located at "/d/work/myapplication/resources".

4 If no "nsconfig.json" file is present the app folder will be  "/d/work/myapplication/app" and resources folder will be at "/d/work/myapplication/app/App_Resources"

The option folders/files to be included/excluded in the app and to be watched during livesync (related to #887) is still not implemented. Its discussion/implementation is moved in the separate issue.

dolanmiu:
Still does not work for me. It seems like the `nsconfig.json` has no effect:

[...]

PR #3356, opened 2018-02-08
Kddimitrov/nsconfig app folder

KristianDD:
Part of implementation for https://github.com/NativeScript/nativescript-cli/issues/3257

### DIFF
--- a/lib/bootstrap.ts
+++ b/lib/bootstrap.ts
@@ -6,7 +6,7 @@ $injector.require("options", "./options");
 $injector.require("nativescript-cli", "./nativescript-cli");
 
 $injector.require("projectData", "./project-data");
-$injector.require("projectDataService", "./services/project-data-service");
+$injector.requirePublic("projectDataService", "./services/project-data-service");
 $injector.requirePublic("projectService", "./services/project-service");
 $injector.require("androidProjectService", "./services/android-project-service");
 $injector.require("iOSEntitlementsService", "./services/ios-entitlements-service");
--- a/lib/constants.ts
+++ b/lib/constants.ts
@@ -29,7 +29,7 @@ export const OUTPUTS_DIR = "outputs";
 export const APK_DIR = "apk";
 export const RESOURCES_DIR = "res";
 export const CONFIG_NS_FILE_NAME = "nsconfig.json";
-export const CONFIG_NS_APP_RESOURCES_ENTRY = "app_resources";
+export const CONFIG_NS_APP_RESOURCES_ENTRY = "appResourcesPath";
 
 export class PackageVersion {
 	static NEXT = "next";
--- a/lib/constants.ts
+++ b/lib/constants.ts
@@ -28,6 +28,8 @@ export const BUILD_DIR = "build";
 export const OUTPUTS_DIR = "outputs";
 export const APK_DIR = "apk";
 export const RESOURCES_DIR = "res";
+export const CONFIG_NS_FILE_NAME = "nsconfig.json";
+export const CONFIG_NS_APP_RESOURCES_ENTRY = "app_resources";
 
 export class PackageVersion {
 	static NEXT = "next";
--- a/lib/constants.ts
+++ b/lib/constants.ts
@@ -30,6 +30,7 @@ export const APK_DIR = "apk";
 export const RESOURCES_DIR = "res";
 export const CONFIG_NS_FILE_NAME = "nsconfig.json";
 export const CONFIG_NS_APP_RESOURCES_ENTRY = "appResourcesPath";
+export const CONFIG_NS_APP_ENTRY = "appPath";
 
 export class PackageVersion {
 	static NEXT = "next";
--- a/lib/project-data.ts
+++ b/lib/project-data.ts
@@ -69,7 +69,6 @@ export class ProjectData implements IProjectData {
 					this.platformsDir = path.join(projectDir, constants.PLATFORMS_DIR_NAME);
 					this.projectFilePath = projectFilePath;
 					this.appDirectoryPath = path.join(projectDir, constants.APP_FOLDER_NAME);
-					this.appResourcesDirectoryPath = path.join(projectDir, constants.APP_FOLDER_NAME, constants.APP_RESOURCES_FOLDER_NAME);
 					this.projectId = data.id;
 					this.dependencies = fileContent.dependencies;
 					this.devDependencies = fileContent.devDependencies;
@@ -85,6 +84,27 @@ export class ProjectData implements IProjectData {
 
 		// This is the case when no project file found
 		this.$errors.fail("No project found at or above '%s' and neither was a --path specified.", projectDir || this.$options.path || currentDir);
+	}
+
+	public getAppResourcesDirectoryPath(projectDir?: string): string {
+		if (!projectDir) {
+			projectDir = this.projectDir;
+		}
+
+		const configNSFilePath = path.join(projectDir, constants.CONFIG_NS_FILE_NAME);
+		let absoluteAppResourcesDirPath: string;
+
+		if (this.$fs.exists(configNSFilePath)) {
+			const configNS = this.$fs.readJson(configNSFilePath);
+
+			if (configNS && configNS[constants.CONFIG_NS_APP_RESOURCES_ENTRY]) {
+				const appResourcesDirPath = configNS[constants.CONFIG_NS_APP_RESOURCES_ENTRY];
+
+				absoluteAppResourcesDirPath = path.resolve(projectDir, appResourcesDirPath);
+			}
+		}
+
+		return absoluteAppResourcesDirPath || path.join(projectDir, constants.APP_FOLDER_NAME, constants.APP_RESOURCES_FOLDER_NAME);
 	}
 
 	private getProjectType(): string {
--- a/lib/project-data.ts
+++ b/lib/project-data.ts
@@ -33,7 +33,9 @@ export class ProjectData implements IProjectData {
 	public projectId: string;
 	public projectName: string;
 	public appDirectoryPath: string;
-	public appResourcesDirectoryPath: string;
+	get appResourcesDirectoryPath(): string {
+		return this.getAppResourcesDirectoryPath();
+	}
 	public dependencies: any;
 	public devDependencies: IStringDictionary;
 	public projectType: string;
--- a/lib/project-data.ts
+++ b/lib/project-data.ts
@@ -55,9 +55,8 @@ export class ProjectData implements IProjectData {
 			const projectFilePath = this.getProjectFilePath(projectDir);
 
 			if (this.$fs.exists(projectFilePath)) {
-				let packageJsonContent: any = null;
-				packageJsonContent = this.$fs.readText(projectFilePath);
-				const nsConfigContent: any = this.getNsConfigContent(projectDir);
+				const packageJsonContent = this.$fs.readText(projectFilePath);
+				const nsConfigContent = this.getNsConfigContent(projectDir);
 
 				this.initializeProjectDataFromContent(packageJsonContent, nsConfigContent, projectDir);
 			}
@@ -72,9 +71,9 @@ export class ProjectData implements IProjectData {
 		projectDir = projectDir || this.$projectHelper.projectDir || "";
 		const projectFilePath = this.getProjectFilePath(projectDir);
 		// If no project found, projectDir should be null
-		let nsData: any = null;
-		let nsConfig: any = null;
-		let packageJsonData: any = null;
+		let nsData = null;
+		let nsConfig: INsConfig = null;
+		let packageJsonData = null;
 
 		try {
 			packageJsonData = parseJson(packageJsonContent);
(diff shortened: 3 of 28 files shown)


KristianDD:
run ci

lib/project-data.ts
```diff
@@ -87,6 +89,50 @@ export class ProjectData implements IProjectData {
...
+		if (configNS && configNS[constants.CONFIG_NS_APP_RESOURCES_ENTRY]) {
+			const appResourcesDirPath = configNS[constants.CONFIG_NS_APP_RESOURCES_ENTRY];
+
+			absoluteAppResourcesDirPath = path.resolve(projectDir, appResourcesDirPath);
+		}
+
+		return absoluteAppResourcesDirPath || path.join(this.getAppDirectoryPath(projectDir), constants.APP_RESOURCES_FOLDER_NAME);
+	}
+
+	public getAppDirectoryPath(projectDir?: string): string {
```
sis0k0: This method seems to be very similar to the one above (`getAppResourcesDirectoryPath`). Do you think it would be a good idea to make a common method that accepts the property name as an argument? Or maybe have a method that parses the config file and validates it?

dtopuzov:
run ci

lib/project-data.ts
```diff
@@ -47,46 +53,121 @@ export class ProjectData implements IProjectData {
...
+			this.projectType = this.getProjectType();
+			this.nsConfig = nsConfig;
+
+			return;
+		}
+
+		this.errorInvalidProject(projectDir);
+	}
+
+	private errorInvalidProject(projectDir: string) {
```
rosen-vladimirov: this method needs return type

lib/project-data.ts
```diff
@@ -47,46 +53,121 @@ export class ProjectData implements IProjectData {
...
+
+		return path.resolve(projectDir, this.getAppResourcesRelativeDirectoryPath());
+	}
+
+	public getAppResourcesRelativeDirectoryPath(): string {
+		if (this.nsConfig && this.nsConfig[constants.CONFIG_NS_APP_RESOURCES_ENTRY]) {
+			return this.nsConfig[constants.CONFIG_NS_APP_RESOURCES_ENTRY];
+		}
+
+		return path.join(this.getAppDirectoryRelativePath(), constants.APP_RESOURCES_FOLDER_NAME);
```
rosen-vladimirov: you can shorten this:
```TypeScript
return this.nsConfig && this.nsConfig[constants.CONFIG_NS_APP_RESOURCES_ENTRY]
           || path.join(this.getAppDirectoryRelativePath(), constants.APP_RESOURCES_FOLDER_NAME);
```

lib/project-data.ts
```diff
@@ -47,46 +53,121 @@ export class ProjectData implements IProjectData {
...
+
+		return path.resolve(projectDir, this.getAppDirectoryRelativePath());
+	}
+
+	public getAppDirectoryRelativePath(): string {
+		if (this.nsConfig && this.nsConfig[constants.CONFIG_NS_APP_ENTRY]) {
+			return this.nsConfig[constants.CONFIG_NS_APP_ENTRY];
+		}
+
+		return constants.APP_FOLDER_NAME;
```
rosen-vladimirov: same as above - you can shorten the code:
```TypeScript
return this.nsConfig && this.nsConfig[constants.CONFIG_NS_APP_ENTRY] || constants.APP_FOLDER_NAME;
```

lib/project-data.ts
```diff
@@ -47,46 +53,121 @@ export class ProjectData implements IProjectData {
...
+		if (!projectDir) {
+			projectDir = this.projectDir;
+		}
+
+		return path.resolve(projectDir, this.getAppDirectoryRelativePath());
+	}
+
+	public getAppDirectoryRelativePath(): string {
+		if (this.nsConfig && this.nsConfig[constants.CONFIG_NS_APP_ENTRY]) {
+			return this.nsConfig[constants.CONFIG_NS_APP_ENTRY];
```
rosen-vladimirov: what will happen in case the path in the nsconfig.json is not relative?

lib/project-data.ts
```diff
@@ -47,46 +53,121 @@ export class ProjectData implements IProjectData {
...
+				this.initializeProjectDataFromContent(packageJsonContent, nsConfigContent, projectDir);
 			}
+
+			return;
+		}
+
+		this.errorInvalidProject(projectDir);
+	}
+
+	public initializeProjectDataFromContent(packageJsonContent: string, nsconfigContent: string, projectDir?: string): void {
```
rosen-vladimirov: @sis0k0 is it ok to pass the content as string or you'll need a method that accepts the parsed json object (i.e. the parsed nsconfig.json) itself?

sis0k0: I think it's better to pass the content as string and let the project data service do the parsing/validation.

lib/project-data.ts
KristianDD: I consider mine more readable and structured. Will change it if you insist.

lib/project-data.ts
KristianDD: I think we should add validation for this as supporting paths not relative to the project will be very hard in cloud builds. Maybe we should also change the key inside `nsconfig.json` to appRelativePath or something in this lines.

lib/constants.ts
```diff
@@ -28,6 +28,9 @@ export const BUILD_DIR = "build";
 export const OUTPUTS_DIR = "outputs";
 export const APK_DIR = "apk";
 export const RESOURCES_DIR = "res";
+export const CONFIG_NS_FILE_NAME = "nsconfig.json";
+export const CONFIG_NS_APP_RESOURCES_ENTRY = "app_resources";
+export const CONFIG_NS_APP_ENTRY = "app_folder";
```
KristianDD: Change to camelCase

lib/project-data.ts
```diff
@@ -47,46 +53,121 @@ export class ProjectData implements IProjectData {
...
-					this.appResourcesDirectoryPath = path.join(projectDir, constants.APP_FOLDER_NAME, constants.APP_RESOURCES_FOLDER_NAME);
-					this.projectId = data.id;
-					this.dependencies = fileContent.dependencies;
-					this.devDependencies = fileContent.devDependencies;
-					this.projectType = this.getProjectType();
-
-					return;
-				}
+				let packageJsonContent: any = null;
+				packageJsonContent = this.$fs.readText(projectFilePath);
```
KristianDD: Check with invalid location (no package.json file)

lib/services/livesync/livesync-service.ts
```diff
@@ -527,6 +527,8 @@ export class LiveSyncService extends EventEmitter implements IDebugLiveSyncServi
 			}
 		}
 
+		patterns.push(projectData.appResourcesDirectoryPath);
```
KristianDD: Check what happens if app resources are still inside app folder.

lib/services/project-service.ts
```diff
@@ -101,7 +101,7 @@ export class ProjectService implements IProjectService {
 	private async extractTemplate(projectDir: string, realTemplatePath: string): Promise<void> {
 		this.$fs.ensureDirectoryExists(projectDir);
 
-		const appDestinationPath = path.join(projectDir, constants.APP_FOLDER_NAME);
+		const appDestinationPath = this.$projectData.getAppDirectoryPath(projectDir);
```
KristianDD: Check usages here. What happens if template has existing `nsconfig.json` inside.

lib/project-data.ts
```diff
@@ -47,46 +53,121 @@ export class ProjectData implements IProjectData {
...
+	public getAppResourcesRelativeDirectoryPath(): string {
+		if (this.nsConfig && this.nsConfig[constants.CONFIG_NS_APP_RESOURCES_ENTRY]) {
+			return this.nsConfig[constants.CONFIG_NS_APP_RESOURCES_ENTRY];
+		}
+
+		return path.join(this.getAppDirectoryRelativePath(), constants.APP_RESOURCES_FOLDER_NAME);
+	}
+
+	public getAppDirectoryPath(projectDir?: string): string {
+		if (!projectDir) {
```
KristianDD: return `null` if missing `projectDir`

lib/project-data.ts
rosen-vladimirov: no, not a problem ;)

lib/project-data.ts
```diff
@@ -47,46 +49,131 @@ export class ProjectData implements IProjectData {
...
+
+		this.errorInvalidProject(projectDir);
+	}
+
+	public initializeProjectDataFromContent(packageJsonContent: string, nsconfigContent: string, projectDir?: string): void {
+		projectDir = projectDir || this.$projectHelper.projectDir || "";
+		const projectFilePath = this.getProjectFilePath(projectDir);
+		// If no project found, projectDir should be null
+		let nsData: any = null;
+		let nsConfig: any = null;
```
rosen-vladimirov: we definitely need an interface for nsConfig and the data that can be included in it. Can you please add a new one and use it as a type here instead of `any`.

lib/services/app-files-updater.ts
```diff
@@ -60,9 +65,11 @@ export class AppFilesUpdater {
 			constants.LIVESYNC_EXCLUDED_FILE_PATTERNS.forEach(pattern => sourceFiles = sourceFiles.filter(file => !minimatch(file, pattern, { nocase: true })));
 		}
 
-		if (this.options.bundle) {
-			sourceFiles = sourceFiles.filter(file => minimatch(file, "**/App_Resources/**", { nocase: true }));
-		}
+		// exclude the app_resources directory from being enumerated
+		// for copying if it is present in the application sources dir
+		const appResourcesPath = projectData.appResourcesDirectoryPath;
+		sourceFiles = sourceFiles.filter(dirName => !path.normalize(dirName).startsWith(path.normalize(appResourcesPath)));
```
rosen-vladimirov: you can extract the `path.normalize(appResourcesPath)` in a separate variable before applying the filtering. This way you'll not calculate it on each iteration.

lib/bootstrap.ts
```diff
@@ -6,7 +6,7 @@ $injector.require("options", "./options");
 $injector.require("nativescript-cli", "./nativescript-cli");
 
 $injector.require("projectData", "./project-data");
-$injector.require("projectDataService", "./services/project-data-service");
+$injector.requirePublic("projectDataService", "./services/project-data-service");
```
rosen-vladimirov: When you add a new method to PublicApi, you have to:
1. Update the documentation: https://github.com/NativeScript/nativescript-cli/blob/master/PublicAPI.md
2. Add the service and the exposed methods in the [nativescript-cli-lib tests](https://github.com/NativeScript/nativescript-cli/blob/master/test/nativescript-cli-lib.ts)

NOTE: More information is available [here](https://github.com/NativeScript/nativescript-cli/blob/master/PublicAPI.md#how-to-add-a-new-method-to-public-api)

lib/services/project-data-service.ts
```diff
@@ -41,6 +42,13 @@ export class ProjectDataService implements IProjectDataService {
 		return projectDataInstance;
 	}
 
+	@exported("projectDataService")
```
rosen-vladimirov: can you export `getProjectData` as well - it will be used from Sidekick to get paths to app and App_Resources directories

lib/services/project-service.ts
KristianDD: Templates currently contain only the contents of the app folder. No option to add nsconfig in template for the moment.

lib/services/prepare-platform-js-service.ts
```diff
@@ -101,6 +102,13 @@ export class PreparePlatformJSService extends PreparePlatformService implements
 			this.$errors.failWithoutHelp(`Processing node_modules failed. ${error}`);
 		}
 	}
+
+	private copyAppResourcesFiles(config: IPreparePlatformJSInfo) {
+		const appDestinationDirectoryPath = path.join(config.platformData.appDestinationDirectoryPath, constants.APP_FOLDER_NAME);
+		const appResourcesSourcePath = config.projectData.appResourcesDirectoryPath;
+
+		shell.cp("-Rf", appResourcesSourcePath, appDestinationDirectoryPath);
```
KristianDD: Shouldn't we use fileSystem module?

lib/project-data.ts
```diff
@@ -47,46 +49,131 @@ export class ProjectData implements IProjectData {
...
+
+	public getAppResourcesRelativeDirectoryPath(): string {
+		if (this.nsConfig && this.nsConfig[constants.CONFIG_NS_APP_RESOURCES_ENTRY]) {
+			return this.nsConfig[constants.CONFIG_NS_APP_RESOURCES_ENTRY];
+		}
+
+		return path.join(this.getAppDirectoryRelativePath(), constants.APP_RESOURCES_FOLDER_NAME);
+	}
+
+	public getAppDirectoryPath(projectDir?: string): string {
```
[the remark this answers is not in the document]
KristianDD: done

lib/bootstrap.ts
KristianDD: done

lib/project-data.ts
```diff
@@ -47,46 +49,130 @@ export class ProjectData implements IProjectData {
...
+				this.initializeProjectDataFromContent(packageJsonContent, nsConfigContent, projectDir);
 			}
+
+			return;
+		}
+
+		this.errorInvalidProject(projectDir);
+	}
+
+	public initializeProjectDataFromContent(packageJsonContent: string, nsconfigContent: string, projectDir?: string): void {
```
sis0k0: Isn't `nsconfigContent` optional?

KristianDD:
run ci

lib/project-data.ts
KristianDD: Yes. I have missed to make it optional as we don't use it in the codebase.

PR #3362, opened 2018-02-09
refactor(livesync): speedup livesync

Mitko-Kerezov:
Speedup LiveSync with the following optimizations:
* Upon a file change **do not check all of node_modules for changes** unless `--syncAllFiles` is passed. This was unnecessary and leads to a **2 s delay** in angular projects
* With the previous implementation, whenever a file change was detected CLI used to delete all files in `platforms/.../app` directory, then **copy all files from the project's app directory** and then **only upload the changed ones**. Change the implementation so that **only changed files are copied and uploaded**.

### DIFF
--- a/lib/services/app-files-updater.ts
+++ b/lib/services/app-files-updater.ts
@@ -4,17 +4,15 @@ import * as constants from "../constants";
 import * as fs from "fs";
 
 export class AppFilesUpdater {
-	constructor(
-		private appSourceDirectoryPath: string,
+	constructor(private appSourceDirectoryPath: string,
 		private appDestinationDirectoryPath: string,
 		public options: IAppFilesUpdaterOptions,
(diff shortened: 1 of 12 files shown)


lib/services/app-files-updater.ts
```diff
@@ -81,6 +79,7 @@ export class AppFilesUpdater {
 				return this.fs.createDirectory(destinationPath);
 			}
 
+			console.log(`Copying ${source} to ${destinationPath}`);
```
rosen-vladimirov: meme_console_log

Mitko-Kerezov: This is actually a forgotten debug message, will remove

Mitko-Kerezov:
Ping @rosen-vladimirov with fixed comments

PR #3365, opened 2018-02-12
feat(extensions): Allow generation of help content for extension commands 

rosen-vladimirov:
Each CLI extension may add new commands to CLI. Allow showing command help for these commands in case:

Commmand execution fails.
Users need more information for command - tns <command> --help
Users need full html help for command - tns help <command>.
Each extension that wants to use this functionality will have to add nativescript key in its package.json and add docs key in it. The docs key must point to the directory where the help content (.md files) is located, relative to the root directory of the extension.
When CLI needs to show command line help for extension's command, it will search the docs directories of all extensions.
When CLI needs to show HTML help for the extension's commands, it will generate html directory right next to the docs dir in the extension.

Move extensibility.d.ts from {N} to mobile-cli-lib just to ensure correct transpilation.

Add different basic page for extensions html pages
When generating html pages for extensions commands, use a different html
page that has the extension name as header at the top.
Add a placeholder and replace it when generating the html page.


lib/services/extensibility-service.ts
```diff
@@ -35,7 +35,16 @@ export class ExtensibilityService implements IExtensibilityService {
 		const installResultInfo = await this.$npm.install(packageName, this.pathToExtensions, npmOpts);
 		this.$logger.trace(`Finished installation of extension '${extensionName}'. Trying to load it now.`);
 
-		return { extensionName: installResultInfo.name };
+		const packageJsonData = this.getExtensionPackageJsonData(installResultInfo.name);
```
Mitko-Kerezov: You can just return `getInstalledExtensionData` here

rosen-vladimirov:
Ping @Mitko-Kerezov - comments are fixed.

PR #3373, opened 2018-02-16
docs: update build instructions in readme and contributing guide

sis0k0:
The initial `npm install` in the repo is reporting errors because of the postinstall script, which expects that the 'ts' files are already transpiled. Maybe we should suggest running it with `--ignore-scripts`?
The other change is running grunt from the local installation of the package, because you may not have a global instance. Maybe the grunt invocation can be moved to a npm script?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -42,6 +42,10 @@ Before you submit a Pull Request, consider the following guidelines.
 ```bash
     git clone git@github.com:NativeScript/nativescript-cli.git
 ```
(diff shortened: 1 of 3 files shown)


rosen-vladimirov:
@sis0k0 we may add `npm run setup` script, which will execute `npm install --ignore-scripts && npm run grunt`? What do you think?

rosen-vladimirov:
@sis0k0 sure, adding the command `git submodule update --init` makes sense for me

Issue #3375, opened 2018-02-19
Failure to start LiveSync causes CLI to hang


Mitko-Kerezov:
### The problem
Whenever starting LiveSync on a new project if LiveSync is unable to start (for example a TypeScript error is present) then CLI tries to exit, but hangs because it has child processes attached to it.

### Steps to reproduce
On a Unix OS:
1. `tns create myApp --tsc`
1. `cd myApp`
1. `echo asd >> app/app.ts` // this will cause the transpile error
1. `tns run android` // could be ios just as well

At this point CLI fails to start LiveSync, because there is a syntax error in the main TypeScript file and the message `Unable to apply changes on device: XXXXXXXXXXXXXX. Error is: TypeScript compiler failed with exit code 1.` is displayed, however the process does not finish and the user is forced to break execution with Ctrl-C.

This issue is caused by the various child processes that CLI spawns like the one for communication with iOS devices or analytics.

PR #3376, opened 2018-02-19
fix(livesync): dispose when stopping livesync

Mitko-Kerezov:
Whenever LiveSync fails to launch for the first time CLI doesn't break process execution because there are hanging child processes (ios-device detection, analytics).

If this is the case, force dispose those resources.

Fixes: https://github.com/NativeScript/nativescript-cli/issues/3375

Ping @rosen-vladimirov

### DIFF
--- a/lib/services/analytics/analytics-service.ts
+++ b/lib/services/analytics/analytics-service.ts
@@ -104,8 +104,8 @@ export class AnalyticsService extends AnalyticsServiceBase {
 		await this.trackInGoogleAnalytics(googleAnalyticsEventData);
 	}
 
-	public dispose(): void {
-		if (this.brokerProcess && this.shouldDisposeInstance) {
+	public dispose(disposeOptions?: IForceOption): void {
+		if (this.brokerProcess && (disposeOptions && disposeOptions.force || this.shouldDisposeInstance)) {
 			this.brokerProcess.disconnect();
 		}
 	}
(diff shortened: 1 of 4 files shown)


Mitko-Kerezov:
run ci

lib/services/analytics/analytics-service.ts
```diff
@@ -104,8 +104,8 @@ export class AnalyticsService extends AnalyticsServiceBase {
 		await this.trackInGoogleAnalytics(googleAnalyticsEventData);
 	}
 
-	public dispose(): void {
-		if (this.brokerProcess && this.shouldDisposeInstance) {
+	public dispose(disposeOptions?: IForceOption): void {
+		if ((disposeOptions && disposeOptions.force) || (this.brokerProcess && this.shouldDisposeInstance)) {
```
rosen-vladimirov: In case you do not have a brokerProcess, but disposeOptions.force is true, you'll have an error on the next line.

dtopuzov:
run ci

lib/services/livesync/livesync-service.ts
```diff
@@ -91,6 +91,8 @@ export class LiveSyncService extends EventEmitter implements IDebugLiveSyncServi
 				// In case we are stopping the LiveSync we must set usbLiveSyncService.isInitialized to false,
 				// as in case we execute nativescript-dev-typescript's before-prepare hook again in the same process, it MUST transpile the files.
 				this.$usbLiveSyncService.isInitialized = false;
+				// After stopping LiveSync we need to dispose of everything in order for the process to end and release all resources
+				this.$injector.dispose({ force: true });
```
rosen-vladimirov: this will cause issues with device detection in Sidekick where we have a long living process. We shouldn't dispose the injector in this case, as this will kill iOS devices detection as well. We should find a better general solution for this case.

Mitko-Kerezov:
Closing this in regards to the issues @rosen-vladimirov pointed out, will rethink the approach an open a new PR

PR #3380, opened 2018-02-20
fix(log-filtering): Fix wrongly filtering out multiline output.

ginev:
`console.log` messages are wrongly filtered out when there is a `pid` (in cases when the app is run on a Simulator). The reason for this is that we split the data by `\n` and look for `pid` in each line to output it. When `console.log/console.dir` is called with a multiline output, the first line only is shown.

This is not the case when running the app on a device. We do not have a `pid` then and filter out only based on tokens matched by the `infoFilterRegex` regular expression.


ginev:
@rosen-vladimirov - I will actually redo this PR as it turned out we need a different approach to addressing this.

Issue #3372, opened 2018-02-16
Documentation about setting proxy is incorrect


rosen-vladimirov:
### Tell us about the problem
The [documentation](https://github.com/NativeScript/nativescript-cli#configure-proxy-usage) how to set proxy is outdated. It points to the old way of setting proxy. Now we have a separate command that should be used.
### Which platform(s) does your issue occur on?
Not applicable

### Please provide the following version numbers that your issue occurs with:
- CLI: Not applicable
- Cross-platform modules: Not applicable
- Runtime(s): Not applicable
- Plugin(s): Not applicable

### Please tell us how to recreate the issue in as much detail as possible.
Check CLI's README. 

### Tasks
* [ ] Describe how to set proxy via CLI's command - `tns proxy set`
* [ ] Describe commands parameters - url (and its formnat), username, password, --insecure
* [ ] Describe commands limitations.
* [ ] Describe the lifetime of the proxy settings - when will they be cleared, how to clear them manually.
* [ ] Describe how to check current proxy settings - `tns proxy` command.

ggarabedian:
https://github.com/NativeScript/nativescript-cli/pull/3387

Regarding the 4th task (describe the lifetime of the proxy settings - when will they be cleared, how to clear them manually), if the proxy can be cleared only with the command, I do not see the need to explain this to the users.

PR #3387, opened 2018-02-21
docs: update the configure proxy usage section

ggarabedian:
Related to #3372


ggarabedian:
1) Fix link in TOC
2) Specify that the proxy settings persist between NS CLI updates and the only way to clear them is with the **clear** command.

etabakov:
Should we add this information in the md files in some form as well? This is a new command that we adding in the CLI, so users will probably expect to be able to get help for is from the command line. This way, this information will be exposed on docs.nativescript.org as well (after we publish the CLI docs).

etabakov:
Just saw that they are already there - ignore my previous comment.

PR #3388, opened 2018-02-22
fix(livesync): delete files during livesync

Mitko-Kerezov:
https://github.com/NativeScript/nativescript-cli/pull/3362 regressed deletion of files during LiveSync.
Fix this by implementing the following logic:
1. In case this is the first ever LiveSync (no file system watcher) clean the `app` directory located inside `platforms` and prepare all javascript files anew
1. In case this is an incremental LiveSync (an event from within the file system watcher) only delete the files that were actually deleted (e. g. `filesToRemove`)

Ping @rosen-vladimirov

### DIFF
--- a/lib/services/app-files-updater.ts
+++ b/lib/services/app-files-updater.ts
@@ -11,20 +11,31 @@ export class AppFilesUpdater {
 	) {
 	}
 
-	public updateApp(beforeCopyAction: (sourceFiles: string[]) => void, filesToSync?: string[]): void {
-		const sourceFiles = filesToSync || this.resolveAppSourceFiles();
+	public updateApp(updateAppOptions: IUpdateAppOptions): void {
+		this.cleanDestinationApp(updateAppOptions);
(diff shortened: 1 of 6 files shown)


lib/services/app-files-updater.ts
```diff
@@ -11,20 +11,31 @@ export class AppFilesUpdater {
...
+			// We get here during the initial sync before the file system watcher is even started
+			// delete everything and prepare everything anew just to be sure
+			// Delete the destination app in order to prevent EEXIST errors when symlinks are used.
+			itemsToRemove = this.readDestinationDir();
+			itemsToRemove = itemsToRemove.filter(
+				(directoryName: string) => directoryName !== constants.TNS_MODULES_FOLDER_NAME);
+
+		}
+
+		_(itemsToRemove).each((directoryItem: string) => {
```
rosen-vladimirov: why not just `_.each(...`

PR #3389, opened 2018-02-22
update(cli): Improve filtering of iOS logs

ginev:
This PR improves the overall behaviour of the IOSLogFilter within the CLI. We now filter data based on the following core assumption:
- all messages coming from the related NS app contain the following artefact: `<app-name>(NativeScript)?[<pid>]`, where (NativeScript) might be missing
- all relevant multiline messages do not contain iOS artefacts like `May 24 15:54:52 Dragons-iPhone NativeScript250(NativeScript)[356] <Notice>:` and directly follow a message with the first artefact

Relying on these principles, we pass messages even if they are multiline (`console.dir`, stack-traces, etc).

Unit tests have been slightly updated to reflect the currently expected behaviour.

This PR should address this issue: https://github.com/NativeScript/nativescript-cli/issues/3105 and is also related to supporting this feature: https://github.com/NativeScript/ios-runtime/pull/884

Merge after https://github.com/telerik/mobile-cli-lib/pull/1062

### DIFF
--- a/lib/services/ios-log-filter.ts
+++ b/lib/services/ios-log-filter.ts
@@ -17,14 +17,15 @@ export class IOSLogFilter implements Mobile.IPlatformLogFilter {
 
 	private partialLine: string = null;
 
-	constructor(
+	constructor(private $logger: ILogger,
 		private $loggingLevels: Mobile.ILoggingLevels,
 		private $fs: IFileSystem,
 		private $projectData: IProjectData) {
 	}
 
-	public filterData(data: string, logLevel: string, pid?: string): string {
-		const specifiedLogLevel = (logLevel || '').toUpperCase();
+	public filterData(data: string, loggingOptions: Mobile.IDeviceLogOptions = <any>{}): string {
+		const specifiedLogLevel = (loggingOptions.logLevel || '').toUpperCase();
+		this.$logger.trace("Logging options", loggingOptions);
 
 		if (specifiedLogLevel !== this.$loggingLevels.info || !data) {
 			return data;
@@ -57,8 +58,8 @@ export class IOSLogFilter implements Mobile.IPlatformLogFilter {
 				// Check if the name of the app equals the name of the CLI project and turn on the filter if not.
 				// We call initializeProjectData in order to obtain the current project name as the instance
 				// of this filter may be used accross multiple projects.
-				this.$projectData.initializeProjectData();
-				this.filterActive = matchResult[1] !== this.$projectData.projectName;
+				const projectName = loggingOptions && loggingOptions.projectName;
+				this.filterActive = matchResult[1] !== projectName;
 			}
 
 			if (this.filterActive) {
--- a/lib/services/ios-log-filter.ts
+++ b/lib/services/ios-log-filter.ts
@@ -1,65 +1,86 @@
 const sourcemap = require("source-map");
 import * as path from "path";
 import { cache } from "../common/decorators";
-import * as iOSLogFilterBase from "../common/mobile/ios/ios-log-filter";
 
-export class IOSLogFilter extends iOSLogFilterBase.IOSLogFilter implements Mobile.IPlatformLogFilter {
-	protected infoFilterRegex = /^.*?((?:<Notice>:)?.*?(((?:CONSOLE|JS) (?:LOG|ERROR)).*?))$/im;
+export class IOSLogFilter implements Mobile.IPlatformLogFilter {
+	// Used to recognize output related to the current project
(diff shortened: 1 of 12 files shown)


test/services/ios-log-filter.ts
```diff
@@ -168,45 +175,55 @@ describe("iOSLogFilter", () => {
 	const infoLogLevel = "INFO";
 	const fullLogLevel = "FULL";
 
-	beforeEach(() => {
-		testInjector = createTestInjector();
-		logFilter = testInjector.resolve(IOSLogFilter);
-	});
+	// beforeEach(() => {
```
rosen-vladimirov: please remove these comments

test/services/ios-log-filter.ts
```diff
@@ -168,45 +175,55 @@ describe("iOSLogFilter", () => {
...
 			});
 
 			it(`returns correct data when logLevel is ${infoLogLevel} on iOS ${data.version} and data is passed one line at a time`, () => {
+				testInjector = createTestInjector(data.projectName);
+				logFilter = testInjector.resolve(IOSLogFilter);
 				data.originalDataArr.forEach((line, index) => {
 					const actualData = logFilter.filterData(line, infoLogLevel, null);
 					const expectedData = data.infoExpectedArr[index];
-					assert.deepEqual(actualData && actualData.trim(), expectedData && expectedData);
+					assert.equal(actualData && actualData.trim(), expectedData && expectedData);
```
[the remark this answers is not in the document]
ginev: I think `deepEqual` is not needed here as we are comparing strings.

ginev:
run ci

lib/services/ios-log-filter.ts
```diff
@@ -1,65 +1,86 @@
...
 
 	private partialLine: string = null;
 
-	constructor($loggingLevels: Mobile.ILoggingLevels,
+	constructor(
+		private $loggingLevels: Mobile.ILoggingLevels,
 		private $fs: IFileSystem,
 		private $projectData: IProjectData) {
-		super($loggingLevels);
+		this.projectName = this.$projectData.projectName;
```
KristianDD: I think this might not work in Sidekick as there only one instance is created of the filter. Try getting projectData from projectDataService.getProjectData and using the project name from there on evry invocation of the filterData function or pass it as an argument if available where the method gets called.

lib/services/ios-log-filter.ts
```diff
@@ -1,65 +1,86 @@
...
+				continue;
+			}
+
+			if (this.partialLine) {
+				currentLine = this.partialLine + currentLine;
+				this.partialLine = undefined;
+			}
+
+			if (i === chunkLines.length - 1 && skipLastLine) {
+				this.partialLine = currentLine;
```
KristianDD: Would be grate if we could persist this per device identifier so that we don't mix lines from different devices. (This is not mandatory as it was the same way before).

ginev: As far as I know, there is no way of receiving log messages from parallel processes with the latest version of CLI.

ginev:
run ci

rosen-vladimirov:
As @KristianDD mentioned, using `$projectData` will not work in Sidekick. Using `$projectDataService` is not an option, as you still have to know the projectDir every time when you filter the logs. I've been thinking how to resolve this issue and I think we can use the same approach as the one we have for Android log filters.
Currently, for Android applications, whenever CLI calls `startApplication` (i.e. runs application on device), we find the new PID of the application and set it to the filter. You can check the logic [here]( https://github.com/telerik/mobile-cli-lib/blob/6ec3a807b3d18399366f50d186266b9610c210d8/mobile/android/android-application-manager.ts#L75-L79)
The `$deviceLogProvider` later passes the value to Android log filter, as shown [here]( https://github.com/telerik/mobile-cli-lib/blob/6ec3a807b3d18399366f50d186266b9610c210d8/mobile/device-log-provider.ts#L10-L11)

So my suggestion is to use the same approach for iOS logs filtering - whenever CLI calls `startApplication`, we can set the projectName to the filter. The filtering method will receive it as an argument from the deviceLogProvider. So we have to add the logic of setting the projectName to the deviceLogProvider in both [ios-application-manager
(https://github.com/telerik/mobile-cli-lib/blob/6ec3a807b3d18399366f50d186266b9610c210d8/mobile/ios/device/ios-application-manager.ts#L83) and [ios-simulator-application-manager](https://github.com/telerik/mobile-cli-lib/blob/master/mobile/ios/simulator/ios-simulator-application-manager.ts#L43).
After that, in the deviceLogProvider, we have to get the already set projectName and pass it to the `filterData` method.

PR #3394, opened 2018-02-26
feat(extensibility): Add hint how to get access to a command from extension

rosen-vladimirov:
In case user executes any command that is registered in extension, but this extension is not installed, they'll receive error:
`Unknown command ...`. Instead of showing this, it is better to tell them how to get access to this command.
In order to achieve this, implement a new logic in `extensibilityService` to find commands registered in all extensions. CLI searches npm (via npms API) for packages that have the keyword `nativescript:extension`.
After that, CLI gets the package.json of the latest version of each dependency (via registry.npmjs.org) and searches for nativescript key in it. In case it has nativescript key, CLI reads the commands value inside it. The value should be string array containg all commands registered in extension's bootstrap.
When user writes down a command that is not registered in CLI, we check if there's such command and in case it is found, CLI will instruct the user how to install the respective extension.
Same will happen in case user tries to access the help content of a command that is registered in extension.

### DIFF
--- a/lib/node-package-manager.ts
+++ b/lib/node-package-manager.ts
@@ -10,7 +10,8 @@ export class NodePackageManager implements INodePackageManager {
 		private $hostInfo: IHostInfo,
 		private $errors: IErrors,
 		private $childProcess: IChildProcess,
-		private $logger: ILogger) { }
+		private $logger: ILogger,
+		private $httpClient: Server.IHttpClient) { }
 
 	@exported("npm")
 	public async install(packageName: string, pathToSave: string, config: INodePackageManagerInstallOptions): Promise<INpmInstallResultInfo> {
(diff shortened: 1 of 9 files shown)


lib/services/extensibility-service.ts
```diff
@@ -107,6 +107,49 @@ export class ExtensibilityService implements IExtensibilityService {
...
+				if (commands) {
+					const copyOfFullArgs = _.clone(inputOpts.inputStrings);
+					while (copyOfFullArgs.length) {
+						const currentCommand = copyOfFullArgs.join(inputOpts.commandSeparator).toLowerCase();
+
+						if (_.some(commands, c => c.toLowerCase() === currentCommand)) {
+							return {
+								extensionName,
+								registeredCommandName: currentCommand,
+								installationMessage: `The command ${currentCommand} is registered in extension ${extensionName}. You can install it by executing 'tns extension install ${extensionName}'`
```
rosen-vladimirov: beautify the command name (replace `|` with ` ` ).

lib/node-package-manager.ts
```diff
@@ -110,6 +111,23 @@ export class NodePackageManager implements INodePackageManager {
...
+		const httpRequestResult = await this.$httpClient.httpRequest(`https://api.npms.io/v2/search?q=keywords:${keyword}`);
+		const result: INpmsResult = JSON.parse(httpRequestResult.body);
+		return result;
+	}
+
+	public async getRegistryPackageData(packageName: string): Promise<any> {
+		const url = `https://registry.npmjs.org/${packageName}`;
+		this.$logger.trace(`Trying to get data from npm registry for package ${packageName}, url is: ${url}`);
+		const responseData = (await this.$httpClient.httpRequest(url)).body;
+		this.$logger.trace(`Successfully received data from npm registry for package ${packageName}.`);
```
Fatme: Shouldn't print `responseData` as well?

PR #3396, opened 2018-02-27
feat(Analytics): Respect playground key from package.json file

Fatme:
We need to track users that export their projects from playground and opens them in CLI or Sidekick. To support that we introduce playground key in nativescript key in package.json file. For example:
```
{
	"nativescript": {
		"playground": {
			"id": "some user quid",
			"usedTutorial": false // is not obligatory. In case it is present, can be true or false
		}
	}
}
```

In  case when package.json file contains playground key, {N} CLI reads playground data and saves it in userSettings file. If usedTutorial=true is already saved in userSettings file, {N} CLI does not overwrite it. After that {N} CLI deletes playground key from package.json file.

In case when package.json file does not contain playground key, {N} CLI checks if playground key is already saved in userSettings file. If this is the case, {N} CLI reads playground data from userSettings file.

### DIFF
--- a/PublicAPI.md
+++ b/PublicAPI.md
@@ -898,6 +898,29 @@ getUserAgentString(identifier: string): string;
 const userAgentString = tns.analyticsSettingsService.getUserAgentString("tns/3.3.0");
 ```
 
+### getPlaygroundInfo
+The `getPlaygroundInfo` method allows retrieving information for projects that are exported from playground
+
+* Definition:
+```TypeScript
+/**
+ * Gets information for projects that are exported from playground.
+ * Returns null in case when project does not have playground key in package.json file (e.g is not exported from playground) and no playground info is saved in userSettings file
+ * @param {string} projectDir The project directory.
+ * @returns {Promise<IPlaygroundInfo>} Playground info. { id: string, usedTutorial: boolean }
+ */
+getPlaygroundInfo(projectDir: string): Promise<IPlaygroundInfo>;
+```
+
+* Usage:
+```JavaScript
+tns.analyticsSettingsService.getPlaygroundInfo("/my/project/path")
+	.then(playgroundInfo => {
+		console.log(playgroundInfo.id);
+		console.log(playgroundInfo.usedTutorial);
+	});
+```
+
 ## How to add a new method to Public API
 CLI is designed as command line tool and when it is used as a library, it does not give you access to all of the methods. This is mainly implementation detail. Most of the CLI's code is created to work in command line, not as a library, so before adding method to public API, most probably it will require some modification.
 For example the `$options` injected module contains information about all `--` options passed on the terminal. When the CLI is used as a library, the options are not populated. Before adding method to public API, make sure its implementation does not rely on `$options`.
--- a/lib/services/analytics/google-analytics-provider.ts
+++ b/lib/services/analytics/google-analytics-provider.ts
@@ -46,7 +46,7 @@ export class GoogleAnalyticsProvider implements IGoogleAnalyticsProvider {
 				this.setCrossClientCustomDimensions(visitor, sessionId);
 				break;
 			default:
-				this.setCustomDimensions(visitor, trackInfo.customDimensions, sessionId);
+				await this.setCustomDimensions(visitor, trackInfo.customDimensions, sessionId);
 				break;
 		}
 
@@ -60,7 +60,7 @@ export class GoogleAnalyticsProvider implements IGoogleAnalyticsProvider {
 		}
 	}
 
-	private setCustomDimensions(visitor: ua.Visitor, customDimensions: IStringDictionary, sessionId: string): void {
+	private async setCustomDimensions(visitor: ua.Visitor, customDimensions: IStringDictionary, sessionId: string): Promise<void> {
 		const defaultValues: IStringDictionary = {
 			[GoogleAnalyticsCustomDimensions.cliVersion]: this.$staticConfig.version,
 			[GoogleAnalyticsCustomDimensions.nodeVersion]: process.version,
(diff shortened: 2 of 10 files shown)


lib/services/playground-service.ts
```diff
@@ -0,0 +1,50 @@
...
+
+				delete projectFileContent.nativescript.playground;
+				this.$fs.writeJson(projectData.projectFilePath, projectFileContent);
+
+				const result = { id , usedTutorial };
+				await this.$userSettingsService.saveSettings(<any>{playground: result});
+				return result;
+			}
+
+			return await this.getPlaygroundInfoFromUserSettingsFile();
```
rosen-vladimirov: I believe this line should be outside of the `if` statement as we want to return the settings from userSettingsFile even when the command is not executed in a project (i.e. when there's no projectData.

lib/services/playground-service.ts
```diff
@@ -0,0 +1,50 @@
...
+		} catch (e) { // in case command is executed in non-project folder
+			return null;
+		}
+	}
+
+	private hasPlaygroundKey(projectFileContent: any): boolean {
+		return projectFileContent && projectFileContent.nativescript && projectFileContent.nativescript.playground && projectFileContent.nativescript.playground.id;
+	}
+
+	private async getPlaygroundInfoFromUserSettingsFile() {
```
rosen-vladimirov: this method needs a return type

lib/services/analytics/google-analytics-provider.ts
```diff
@@ -60,16 +61,22 @@ export class GoogleAnalyticsProvider implements IGoogleAnalyticsProvider {
...
-	private setCustomDimensions(visitor: ua.Visitor, customDimensions: IStringDictionary, sessionId: string): void {
+	private async setCustomDimensions(visitor: ua.Visitor, customDimensions: IStringDictionary, sessionId: string): Promise<void> {
 		const defaultValues: IStringDictionary = {
 			[GoogleAnalyticsCustomDimensions.cliVersion]: this.$staticConfig.version,
 			[GoogleAnalyticsCustomDimensions.nodeVersion]: process.version,
 			[GoogleAnalyticsCustomDimensions.clientID]: this.clientId,
 			[GoogleAnalyticsCustomDimensions.projectType]: null,
 			[GoogleAnalyticsCustomDimensions.sessionID]: sessionId,
-			[GoogleAnalyticsCustomDimensions.client]: AnalyticsClients.Unknown
+			[GoogleAnalyticsCustomDimensions.client]: AnalyticsClients.Unknown,
```
rosen-vladimirov: you do not need this change

lib/services/playground-service.ts
```diff
@@ -0,0 +1,50 @@
...
+		const projectData = this.getProjectData(projectDir);
+		if (projectData) {
+			const projectFileContent = this.$fs.readJson(projectData.projectFilePath);
+			if (this.hasPlaygroundKey(projectFileContent)) {
+				const id = projectFileContent.nativescript.playground.id;
+				let usedTutorial = projectFileContent.nativescript.playground.usedTutorial || false;
+
+				// In case when usedTutorial=true is already saved in userSettings file, we shouldn't overwrite it
+				const playgroundInfo = await this.getPlaygroundInfoFromUserSettingsFile();
+				if (playgroundInfo && playgroundInfo.usedTutorial === true) {
```
rosen-vladimirov: just `if (playgroundInfo && playgroundInfo.usedTutorial)` is enough

lib/services/playground-service.ts
rosen-vladimirov: also you do not need `await` after return.

test/services/playground-service.ts
```diff
@@ -0,0 +1,199 @@
...
+	testInjector.register("projectDataService", {});
+	testInjector.register("userSettingsService", {});
+	testInjector.register("injector", testInjector);
+
+	return testInjector;
+}
+
+function mockPlaygroundService(testInjector: IInjector, data?: { projectData?: any, nativescriptKey?: any, userSettingsData?: any}) {
+	const projectDataService = testInjector.resolve<IProjectDataService>("projectDataService");
+	projectDataService.getProjectData = () => (data && data.projectData) ? data.projectData : <IProjectData>{};
```
rosen-vladimirov: you can shorten this:
```TypeScript
data && data.projectData || <IProjectData>{}
```

test/services/playground-service.ts
```diff
@@ -0,0 +1,199 @@
...
+
+	const userSettingsService = testInjector.resolve("userSettingsService");
+	userSettingsService.getSettingValue = async (keyName: string) => {
+		return data && data.userSettingsData ? data.userSettingsData[keyName] : null;
+	};
+	userSettingsService.saveSettings = async (settings: any) => { userSettings = settings; };
+
+	const fs = testInjector.resolve<IFileSystem>("fs");
+	fs.readJson = () => data && data.nativescriptKey ? data.nativescriptKey : {};
```
rosen-vladimirov: same as above - this can be shortend

test/services/playground-service.ts
```diff
@@ -0,0 +1,199 @@
...
+}
+
+describe("PlaygroundService", () => {
+	let  testInjector: IInjector = null;
```
rosen-vladimirov: one space between let and testInjector should be enough

test/services/playground-service.ts
```diff
@@ -0,0 +1,199 @@
...
+		testInjector = createTestInjector();
+		playgroundService = testInjector.resolve("playgroundService");
+	});
+
+	describe("getPlaygroundInfo", () => {
+		it("should return null when projectDir is not specified", async () => {
+			const result = await playgroundService.getPlaygroundInfo();
+			assert.equal(result, null);
+		});
+		it("should return null when projectDir has no nativescript key in package.json file", async () => {
```
rosen-vladimirov: In this case we would like to get the values from userSettings file.

PublicAPI.md
```diff
@@ -898,6 +898,28 @@ getUserAgentString(identifier: string): string;
 const userAgentString = tns.analyticsSettingsService.getUserAgentString("tns/3.3.0");
 ```
 
+### getPlaygroundInfo
+The `getPlaygroundInfo` method allows retrieving information for projects that are exported from playground
```
rosen-vladimirov: Please note when this method returns null or undefined.

Fatme: I do it here - https://github.com/NativeScript/nativescript-cli/pull/3396/files#diff-eaac680656b280b3360a73aac495df29R908. Is it ok?

dtopuzov:
run ci

PR #3398, opened 2018-02-27
Plamen5kov/rebuild aar

Plamen5kov:
Don't merge before: https://github.com/telerik/mobile-cli-lib/pull/1055
this PR is based on branch: `kddimitrov/nsconfig-app-folder`

_problem_
the CLI doesn't build aar files for the native plugins

_solution_
build aar when:
* there are AndroidManifest.xml or `res` folder in plugins' `platforms/android/` folder
* `--syncAllFiles` flag is passed (and a change is made in plugins' platforms/android/` folder

_related issues_
https://github.com/NativeScript/android-runtime/issues/890
https://github.com/NativeScript/nativescript-plugin-seed/issues/65


lib/services/android-plugin-build-service.ts
```diff
@@ -0,0 +1,385 @@
...
+			return (
+				getLsStats(src).isDirectory() &&
+				(dirName === "res" ||
+					dirName === "java" ||
+					dirName === "assets" ||
+					dirName === "jniLibs")
+			);
+		}
+
+		return this.getDirectories(source).filter(isAndroidSourceDirectory);
```
Fatme: Consider using the following approach
```
const directories = ["res", "java", "assets", "jniLibs"];
return this.$fs.enumerateFilesInDirectorySync(source, (file, stat) => stat.isDirectory() && _.includes(file, directoties));
```

lib/services/android-plugin-build-service.ts
```diff
@@ -0,0 +1,385 @@
...
+	}
+
+	private getManifest(platformsDir: string) {
```
Fatme: Can you use `platformsData.getPlatformData().configurationFilePath`?

lib/services/android-plugin-build-service.ts
```diff
@@ -0,0 +1,385 @@
...
+		);
+
+		return promise;
+	}
+
+	private getDirectories(source: string): any {
+		return this.$fs.readDirectory(source).map(name => path.join(source, name));
+	}
+
+	private createDirIfDoesntExist(dirPath: string, { isRelativeToScript = false } = {}) {
```
Fatme: Can you use `this.$fs.ensureDirectoryExists(dirPath)`?

rosen-vladimirov:
@Plamen5kov can you please rebase the branch correctly, so it will be easier to check the new commits. Also currently it conflicts and cannot be merged.

lib/services/android-plugin-build-service.ts
Plamen5kov: this is the plugin's platforms dir, not the project

Plamen5kov:
@rosen-vladimirov  I did

lib/services/android-plugin-build-service.ts
```diff
@@ -0,0 +1,365 @@
...
+		$: {
+			"xmlns:android": "http://schemas.android.com/apk/res/android"
+		}
+	};
+	private static ANDROID_PLUGIN_GRADLE_TEMPLATE = "gradle-plugin";
+
+	private getAndroidSourceDirectories(source: string): Array<string> {
+		const directories = ["res", "java", "assets", "jniLibs"];
+		const resultArr: Array<string> = [];
+		this.$fs.enumerateFilesInDirectorySync(source, (file, stat) => {
```
Fatme:
```
return this.$fs.enumerateFilesInDirectorySync(source, (file, stat) => stat.isDirectory() && _.some(directories, (element) => file.endsWith(element)))
```

lib/services/android-plugin-build-service.ts
```diff
@@ -0,0 +1,365 @@
...
+		// if a manifest OR/AND resource files are present - write files, build plugin
+		if (shouldBuildAar) {
+			this.$fs.ensureDirectoryExists(newPluginMainSrcDir);
+
+			if (manifestFilePath) {
+				let androidManifestContent;
+				try {
+					androidManifestContent = this.$fs.readFile(manifestFilePath).toString();
+				} catch (err) {
+					throw Error(
```
rosen-vladimirov: can you use the new keyword here - all over the codebase we are using `throw new Error` instead of `throw Error`
Same is valid for other places in the code where you are using the `throw Error` syntax

lib/services/android-plugin-build-service.ts
```diff
@@ -0,0 +1,365 @@
...
+				const destination = path.join(newPluginMainSrcDir, dirName);
+				this.$fs.ensureDirectoryExists(destination);
+
+				this.copyRecursive(path.join(dir, "*"), destination);
+			}
+
+			// copy the preconfigured gradle android library project template to the temporary android library
+			this.copyRecursive(path.join(path.resolve(path.join(__dirname, AndroidPluginBuildService.ANDROID_PLUGIN_GRADLE_TEMPLATE), "*")), newPluginDir);
+
+			// sometimes the AndroidManifest.xml or certain resources in /res may have a compile dependency to a lbirary referenced in include.gradle. Make sure to compile the plugin with a compile dependency to those libraries
```
rosen-vladimirov: lbirary -> library

lib/services/android-plugin-build-service.ts
```diff
@@ -0,0 +1,365 @@
...
+			const includeGradlePath = path.join(options.platformsAndroidDirPath, "include.gradle");
+			if (this.$fs.exists(includeGradlePath)) {
+				const includeGradleContent = this.$fs.readFile(includeGradlePath)
```
rosen-vladimirov: this.$fs.readText

lib/services/android-plugin-build-service.ts
```diff
@@ -0,0 +1,365 @@
...
+					.toString();
+				const repositoriesAndDependenciesScopes = this.getIncludeGradleCompileDependenciesScope(includeGradleContent);
+
+				// dependencies { } object was found - append dependencies scope
+				if (repositoriesAndDependenciesScopes.length > 0) {
+					const buildGradlePath = path.join(newPluginDir, "build.gradle");
+					this.$fs.appendFile(buildGradlePath, "\n" + repositoriesAndDependenciesScopes.join("\n")
+					);
```
rosen-vladimirov: I think this should be on the line above

lib/services/android-plugin-build-service.ts
```diff
@@ -0,0 +1,365 @@
...
+				const newIncludeGradleFileContent = includeGradleFileContent.replace(productFlavorsScope, "");
+				this.$fs.writeFile(includeGradleFilePath, newIncludeGradleFileContent);
+
+			} catch (e) {
+				throw Error(`Failed to write the updated include.gradle in - ${includeGradleFilePath}`);
+			}
+		}
+	}
+
+	private validateOptions(options: IBuildOptions) {
```
rosen-vladimirov: this method needs return type

lib/services/android-project-service.ts
```diff
@@ -413,14 +422,90 @@ export class AndroidProjectService extends projectServiceBaseLib.PlatformProject
...
+
+				this.prebuildNativePlugin(options);
+			}
 		}
 
 		// Do nothing, the Android Gradle script will configure itself based on the input dependencies.json
 	}
 
+	public async checkIfPluginsNeedBuild(projectData: IProjectData): Promise<Array<any>> {
+		const detectedPlugins: Array<any> = [];
```
rosen-vladimirov: you can introduce an interface in this class or at least inline the properties:
```TypeScript
public async checkIfPluginsNeedBuild(projectData: IProjectData): Promise<Array<{ platformsAndroidDirPath: string, pluginName: string }>> {
    const detectedPlugins: Array<{ platformsAndroidDirPath: string, pluginName: string }> = [];
```

lib/services/android-project-service.ts
```diff
@@ -413,14 +422,90 @@ export class AndroidProjectService extends projectServiceBaseLib.PlatformProject
...
+							pluginName: pluginName
+						});
+					}
+				}
+			}
+		}
+		return detectedPlugins;
+	}
+
+	private isAllowedFile(item: string) {
```
rosen-vladimirov: this method needs return type

lib/services/android-project-service.ts
```diff
@@ -413,14 +422,90 @@ export class AndroidProjectService extends projectServiceBaseLib.PlatformProject
...
+						return this.isAllowedFile(item);
+					});
+
+					if (hasGeneratedAar) {
+						const aarStat = this.$fs.getFsStats(generatedAarPath);
+						nativeFiles.forEach((item) => {
+							const currentItemStat = this.$fs.getFsStats(item);
+							if (currentItemStat.mtime > aarStat.mtime) {
+								detectedPlugins.push({
+									platformsAndroidDirPath: platformsAndroidDirPath,
```
rosen-vladimirov: you can use the short syntax here:
```TypeScript
detectedPlugins.push({
    platformsAndroidDirPath,
    pluginName
});
```

lib/services/android-project-service.ts
```diff
@@ -413,14 +422,90 @@ export class AndroidProjectService extends projectServiceBaseLib.PlatformProject
...
+									pluginName: pluginName
+								});
+							}
+						});
+					} else if (nativeFiles.length > 0) {
+						detectedPlugins.push({
+							platformsAndroidDirPath: platformsAndroidDirPath,
```
rosen-vladimirov: same as above:
you can use the short syntax here:
```TypeScript
detectedPlugins.push({
    platformsAndroidDirPath,
    pluginName
});
```

lib/services/android-plugin-build-service.ts
```diff
@@ -0,0 +1,365 @@
+import * as path from "path";
+import * as shell from "shelljs";
+
+const xml2js = require("xml2js");
```
Fatme: I suggest to add a separate service that is wrapper for `xml2js` package and to inject the new service in constructor. (It's not a merge stopper)

lib/services/android-plugin-build-service.ts
```diff
@@ -0,0 +1,365 @@
...
+			if (stat.isDirectory() && _.some(directories, (element) => file.endsWith(element))) {
+				resultArr.push(file);
+				return true;
+			}
+		});
+
+		return resultArr;
+	}
+
+	private getManifest(platformsDir: string) {
```
Fatme: Why we need to traverse the platforms dir to find the `manifest` file. Shoudn't it be in `platforms\android` folder. If this is the case, should'n we just return
```
path.join(platfromsDir, "android", AndroidPluginBuildService.ANDROID_MANIFEST_XML);
```

lib/services/android-plugin-build-service.ts
```diff
@@ -0,0 +1,365 @@
+import * as path from "path";
+import * as shell from "shelljs";
+
+const xml2js = require("xml2js");
+
+export class AndroidPluginBuildService implements IAndroidPluginBuildService {
```
Fatme: Consider to add some unit tests for this service.

lib/services/android-plugin-build-service.ts
Plamen5kov: doesn't work, returns true or false i need the array

lib/services/android-plugin-build-service.ts
Plamen5kov: The plugin's platform folder does not necessarily have an `AndroidManifest.xml` file

lib/services/android-plugin-build-service.ts
Plamen5kov: will do

lib/services/android-plugin-build-service.ts
Plamen5kov: IMO we can hold off on that because `xml2js` is used only in `AndroidPluginBuildService` and there's no publicly exposed API, that can be used when the CLI is required.

lib/services/android-plugin-build-service.ts
Pip3r4o: The migration tool started initially with the thought that it could also support flavors of the android libraries, in which cases there could be more than one manifests. Since I later realized that is a somewhat unrealistic expectation (that plugins would have additional flavors), I dropped the idea, but didn't modify the code as it kept working.

Yes, you can indeed use path.join followed by a check if `fs.existsSync()`

Plamen5kov:
Added unit tests and rebased on master branch.

lib/definitions/android-plugin-migrator.d.ts
```diff
@@ -0,0 +1,12 @@
+
+interface IBuildOptions {
```
rosen-vladimirov: As these options are specific for Android and for building plugins, I think the name should be more concrete: `IAndroidBuildOptions` or `IBuildAndroidPluginOptions`

lib/services/android-project-service.ts
```diff
@@ -413,14 +422,90 @@ export class AndroidProjectService extends projectServiceBaseLib.PlatformProject
...
+		const detectedPlugins: Array<{ platformsAndroidDirPath: string, pluginName: string }> = [];
+
+		const platformsAndroid = path.join(constants.PLATFORMS_DIR_NAME, "android");
+		const pathToPlatformsAndroid = path.join(projectData.projectDir, platformsAndroid);
+		const dependenciesJson = await this.$fs.readJson(path.join(pathToPlatformsAndroid, constants.DEPENDENCIES_JSON_NAME));
+		const productionDependencies = dependenciesJson.map((item: any) => {
+			return path.resolve(pathToPlatformsAndroid, item.directory);
+		});
+
+		for (const dependencyKey in productionDependencies) {
```
rosen-vladimirov: you can use directly:
```TypeScript
for (const dependency of productionDependencies) {
```

package.json
```diff
@@ -79,6 +79,7 @@
     "winreg": "0.0.17",
     "ws": "2.2.0",
     "xcode": "https://github.com/NativeScript/node-xcode/archive/1.4.0.tar.gz",
+    "xml2js": "^0.4.19",
```
rosen-vladimirov: Please use strict versions of both new dependencies.

test/ios-entitlements-service.ts
```diff
@@ -47,7 +47,7 @@ describe("IOSEntitlements Service Tests", () => {
 		injector = createTestInjector();
 
 		platformsData = injector.resolve("platformsData");
-		projectData = injector.resolve("projectData");
+		projectData = $injector.resolve<IProjectData>("projectData");
```
rosen-vladimirov: this way you'll not use the local `injector` variable, but the global `$injector` one. It is a bad practice to use the global variable in the tests. Can you elaborate why you needed to change this?

test/plugin-prepare.ts
```diff
@@ -32,20 +32,6 @@ describe("Plugin preparation", () => {
 		assert.deepEqual({}, pluginPrepare.preparedDependencies);
 	});
 
-	it("skips prepare if every plugin prepared", async () => {
```
rosen-vladimirov: does it mean that now we'll never skip prepare even if all plugins are okay?

test/services/android-plugin-build-service.ts
```diff
@@ -0,0 +1,184 @@
...
+				pluginName: "my-plugin",
+				aarOutputDir: tempFolder,
+				tempPluginDirPath: pluginFolder
+			};
+
+			const expectedAarName = "my_plugin.aar";
+			await androidBuildPluginService.buildAar(config);
+			const hasGeneratedAar = fs.exists(path.join(tempFolder, expectedAarName));
+
+			assert.equal(hasGeneratedAar, true);
```
rosen-vladimirov:
```TypeScript
assert.isTrue(hasGeneratedAar);
```

vendor/gradle-plugin/build.gradle
```diff
@@ -0,0 +1,36 @@
...
+        classpath 'com.android.tools.build:gradle:2.3.2'
+
+        // NOTE: Do not place your application dependencies here; they belong
+        // in the individual module build.gradle files
+    }
+}
+
+allprojects {
+    repositories {
+        jcenter()
```
rosen-vladimirov: shouldn't we add the google() repository as well?

test/plugin-prepare.ts
Plamen5kov: no, if the JS part of the plugins is prepared we will skill prepare, but we will always check for changes in the native part of a plugin for changes that need rebuilding.

vendor/gradle-plugin/build.gradle
Plamen5kov: no need, because we decide what dependencies to use, and we don't have any dependencies that would use google()

test/ios-entitlements-service.ts
Plamen5kov: probably a copy-paste error, but thanks for the explanation

lib/services/android-plugin-build-service.ts
```diff
@@ -0,0 +1,359 @@
...
+		const dependenciesScope = this.getScope("dependencies", includeGradleFileContent);
+		result.push(dependenciesScope);
+
+		return result;
+	}
+
+	private getScope(scopeName: string, content: string): string {
+		const indexOfScopeName = content.indexOf(scopeName);
+		let result = "";
+		const OPENING_BRACKET = "{";
```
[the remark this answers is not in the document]
Plamen5kov: Because the indexOfScopeName can be different unline the OPENING_BRACKET symbol, but if you like I could change it ... it's not really important to me.

Plamen5kov:
run ci

lib/services/android-project-service.ts
```diff
@@ -346,12 +347,28 @@ export class AndroidProjectService extends projectServiceBaseLib.PlatformProject
...
+			buildOptions.push(`-PksPath=${path.resolve(settings.keyStorePath)}`);
+			buildOptions.push(`-Palias=${settings.keyStoreAlias}`);
+			buildOptions.push(`-Ppassword=${settings.keyStoreAliasPassword}`);
+			buildOptions.push(`-PksPassword=${settings.keyStorePassword}`);
+		}
+
+		return buildOptions;
+	}
+
+	public getBuildOptions(configurationFilePath?: string): Array<string> {
```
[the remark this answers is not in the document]
rosen-vladimirov: Thanks for the explanation. As I cannot think of a better name, lets keep the current one.

lib/services/livesync/livesync-service.ts
```diff
@@ -658,16 +659,19 @@ export class LiveSyncService extends EventEmitter implements IDebugLiveSyncServi
...
 
 					filePath = path.join(liveSyncData.projectDir, filePath);
 
 					this.$logger.trace(`Chokidar raised event ${event} for ${filePath}.`);
 
-					if (event === "add" || event === "addDir" || event === "change" /* <--- what to do when change event is raised ? */) {
-						filesToSync.push(filePath);
-					} else if (event === "unlink" || event === "unlinkDir") {
-						filesToRemove.push(filePath);
+					if (!isAllowedFinalFile(filePath)) {
```
rosen-vladimirov: Can I place `.aar` file in App_Resources? In case yes, what should happen here - we'll not include the files in filesToSync and... we should rebuild the application? Will this happen?

Plamen5kov: We haven't talked about placing aar files in `App_Resources` so I wouldn't know what would happen.

Pip3r4o: @rosen-vladimirov App_Resources/Android/libs (where aars can be placed) is watched separately and should force a rebuild

lib/definitions/android-plugin-migrator.d.ts
```diff
@@ -0,0 +1,15 @@
+
+interface IBuildOptions extends IAndroidBuildOptions{
```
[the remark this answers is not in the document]
Plamen5kov: In one of your previous comments, you mentioned the properties are android specific so I moved them to an android specific interface.

lib/services/livesync/livesync-service.ts
Plamen5kov: @rosen-vladimirov I checked it out and when you have an `example.aar` file in the `App_Resources` and the `tns run android`  or `tns run android --syncAllFiles` is running and the `example.aar` is changed, the changes are being ignored. 
When the command is stopped and ran again, the `example.aar` forces a rebuild.

miroslavaivanova:
run ci

Issue #3401, opened 2018-02-27
Renaming App_Resources folder leads to failure in loading the app resources


Natalia-Hristova:
### Tell us about the problem
If you rename `App_Resources` folder and set the new path  in `nsconfig.json` - app builds and runs but the installed app has no icon - `Failed to find LaunchScreen.storyboard ...`

### Which platform(s) does your issue occur on?
iOS

### Please provide the following version numbers that your issue occurs with:
- CLI: [PR 3356 Kddimitrov/nsconfig app folder](https://github.com/NativeScript/nativescript-cli/pull/3356)
- Cross-platform modules: `3.4.1`
- Runtime(s): 
"tns-android" `3.4.1-2018-02-27-03`
"tns-ios" `3.4.1`

### Please tell us how to recreate the issue in as much detail as possible.
1. `tns create SomeApp`
2. Rename `App_Resources` folder
3. In project root folder create `nsconfig.json` with the new path to `App_Resources` folder like this:
4. tns run ios

```
{
    "appResourcesPath": "app/My_App_Resources"
}
```

### Is there code involved? If so, please share the minimal amount of code needed to recreate the problem.
Here is the log
```
Project successfully created.
Preparing project...
Successfully prepared plugin nativescript-theme-core for ios.
Successfully prepared plugin tns-core-modules for ios.
Successfully prepared plugin tns-core-modules-widgets for ios.
Failed to find LaunchScreen.storyboard but it was specified in the Info.plist.
Consider updating the resources in app/App_Resources/iOS/.
A good starting point would be to create a new project and diff the changes with your current one.
Also the following repo may be helpful: https://github.com/NativeScript/template-hello-world/tree/master/App_Resources/iOS
We will now place an empty obsolete compatability white screen LauncScreen.xib for you in platforms/ios/AppNSConfig3/Resources/LaunchScreen.xib so your app may appear as it did in pre v2.1.0 versions of the ios runtime.
Project successfully prepared (iOS)
Building project...
Xcode build...

```

Natalia-Hristova:
This is fixed in the current `master` branch and the `next` tag of `nativescript-cli`

Issue #3400, opened 2018-02-27
Android build fails when App_Resources folder is renamed


Natalia-Hristova:
### Tell us about the problem
If you rename `App_Resources` folder and set the new path  in `nsconfig.json` - build fails.

### Which platform(s) does your issue occur on?
Android

### Please provide the following version numbers that your issue occurs with:
- CLI: [PR 3356 Kddimitrov/nsconfig app folder](https://github.com/NativeScript/nativescript-cli/pull/3356)
- Cross-platform modules: `3.4.1`
- Runtime(s): 
"tns-android" `3.4.1-2018-02-27-03`
"tns-ios" `3.4.1`

### Please tell us how to recreate the issue in as much detail as possible.
1. `tns create SomeApp`
2. Rename `App_Resources` folder
3. In project root folder create `nsconfig.json` with the new path to `App_Resources` folder like this:
4. tns run android

```
{
    "appResourcesPath": "app/My_App_Resources"
}
```

### Is there code involved? If so, please share the minimal amount of code needed to recreate the problem.
Here is the log
```
Searching for devices...
Preparing project...
Successfully prepared plugin nativescript-theme-core for android.
Successfully prepared plugin tns-core-modules for android.
Successfully prepared plugin tns-core-modules-widgets for android.
Project successfully prepared (Android)
Building project...
Gradle build...
         + applying user-defined configuration from .../AppNSConfig3/app/My_App_Resources/Android/app.gradle
         + adding nativescript runtime package dependency: nativescript-optimized
         + adding aar plugin dependency: .../AppNSConfig3/node_modules/tns-core-modules-widgets/platforms/android/widgets-release.aar
false
.../AppNSConfig3/platforms/android/app/build/intermediates/manifests/full/debug/AndroidManifest.xml:28:23-37: AAPT: No resource found that matches the given name (at 'icon' with value '@drawable/icon').
    
.../AppNSConfig3/platforms/android/app/build/intermediates/manifests/full/debug/AndroidManifest.xml:35:28-52: AAPT: No resource found that matches the given name (at 'theme' with value '@style/LaunchScreenTheme').
    
.../AppNSConfig3/platforms/android/app/build/intermediates/manifests/full/debug/AndroidManifest.xml:25: error: Error: No resource found that matches the given name (at 'icon' with value '@drawable/icon').

[...]

PR #3403, opened 2018-02-28
fix(prepare): disregard app_resources's dir name when preparing in platforms/

Pip3r4o:
patches https://github.com/NativeScript/nativescript-cli/pull/3356

addresses #3400 and #3401

### DIFF
--- a/test/app-files-updates.ts
+++ b/test/app-files-updates.ts
@@ -7,7 +7,7 @@ require("should");
 function createTestInjector(): IInjector {
 	const testInjector = new yok.Yok();
 
-	testInjector.register("projectData", { appResourcesDirectoryPath: "App_Resources"});
+	testInjector.register("projectData", { appResourcesDirectoryPath: "App_Resources" });
 
 	return testInjector;
 }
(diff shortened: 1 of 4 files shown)


Pip3r4o:
@KristianDD please review

Pip3r4o:
run ci

test/app-files-updates.ts
```diff
@@ -71,7 +71,7 @@ describe("App files copy", () => {
 
 	it("copies all app files but app_resources when not bundling", () => {
 		const updater = new CopyAppFilesUpdater([
-			"file1", "dir1/file2", "App_Resources/Android/blah.png"
```
rosen-vladimirov: why is this changed? If the behavior is changed, the testname and expected results should be changed

Pip3r4o: @rosen-vladimirov that's an oversight on my part. I changed the only failing test, because the app_resources were no longer filtered as part of the `copy()` method but in [public updateApp()](https://github.com/NativeScript/nativescript-cli/pull/3403/files#diff-d7315453ac31262b45d0c2530d69cd94R14) which returns void and I found no straightforward way of mocking it in order to update the test properly.

PR #3410, opened 2018-03-01
fix(livesync): respect rapid file changes

Mitko-Kerezov:
Fix LiveSync not respecting rapid file changes (like transpilation, or copying directories for example).

Fix this by not passing `filesToSync` to `updateApp` method of `AppFilesUpdater`. This way each file will be copied upon every LiveSync action and we won't miss out any files.

Ping @rosen-vladimirov


dtopuzov:
run ci

Issue #3413, opened 2018-03-02
Doctor do not show correct message if everything is OK on Windows and Linux


dtopuzov:
`tns doctor` show this message on Linux/Windows machines when everything is ok:

```
NOTE: You can develop for iOS only on Mac OS X systems.
To be able to work with iOS devices and projects, you need Mac OS X Mavericks or later.

There seem to be issues with your configuration.
Your components are up-to-date: 
nativescript,tns-core-modules,tns-android,tns-ios

When you file an issue, these warnings will help the NativeScript team to investigate, identify, and resolve the report.
Please, ignore them if you are not experiencing any issues with NativeScript.
````

I don't think we should say:
```
There seem to be issues with your configuration.
```

CLI: @next

PR #3417, opened 2018-03-03
fix(doctor): Show correct message when everything is setup correctly on WIndows and Linux

Fatme:
In case when doctor command is executed on Windows or Linux and everything is correctly installed, {N} prints the following warning:
"NOTE: You can develop for iOS only on Mac OS X systems.
To be able to work with iOS devices and projects, you need Mac OS X Mavericks or later."

Fixes https://github.com/NativeScript/nativescript-cli/issues/3413


Fatme:
I'm closing this because the fix is merged with this PR https://github.com/NativeScript/nativescript-cli/pull/3418

PR #3425, opened 2018-03-06
feat: fail if bundling both platforms

Mitko-Kerezov:
This is currently not possible so throw an error if this is the case

Ping @sis0k0 @rosen-vladimirov


dtopuzov:
run ci

dtopuzov:
run ci

Issue #3428, opened 2018-03-07
on macOS: App_Resources/ folder is not deleted in platforms/ when rename App_Resources folder or change it's location


Natalia-Hristova:
### Tell us about the problem
After build of the app  `App_Resources/` exists in `src/main/assets/app` but it shouldn't. Have in mind that this happens **only on Mac**. It's works as expected on Windows.

### Which platform(s) does your issue occur on?
Android and iOS

### Please provide the following version numbers that your issue occurs with:
- CLI: `4.0.0-2018-03-07-10905`
- Runtime(s): `tns-android` `3.4.2` or current `next`
`tns-ios` `3.4.1` or current `next`

### Please tell us how to recreate the issue in as much detail as possible.
1. Rename `App_Resources/` and set up the path in `nsconfig.json` OR put `App_Respurces/` in different location like `app/res/App_Respurces/`
2. Run `tns build android` or `tns build ios`
3. Check for the existence of the renamed `App_Respurces/` in
**for Android:** `../platforms/android/app/src/main/assets/app`
**for iOS:** `../platforms/ios/MyApp/app`

PR #3429, opened 2018-03-08
fix(nsconfig):app resources not filtered on prepare

KristianDD:
Fixes https://github.com/NativeScript/nativescript-cli/issues/3428

### DIFF
--- a/lib/services/app-files-updater.ts
+++ b/lib/services/app-files-updater.ts
@@ -17,7 +17,7 @@ export class AppFilesUpdater {
 
 		// exclude the app_resources directory from being enumerated
 		// for copying if it is present in the application sources dir
-		const appResourcesPathNormalized = path.normalize(projectData.appResourcesDirectoryPath + "\\");
+		const appResourcesPathNormalized = path.normalize(projectData.appResourcesDirectoryPath + path.sep);
 		sourceFiles = sourceFiles.filter(dirName => !path.normalize(dirName).startsWith(appResourcesPathNormalized));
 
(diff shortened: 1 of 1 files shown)


Pip3r4o:
Fixes https://github.com/NativeScript/nativescript-cli/issues/3428

lib/services/app-files-updater.ts
```diff
@@ -17,7 +17,7 @@ export class AppFilesUpdater {
 
 		// exclude the app_resources directory from being enumerated
 		// for copying if it is present in the application sources dir
-		const appResourcesPathNormalized = path.normalize(projectData.appResourcesDirectoryPath + "\\");
+		const appResourcesPathNormalized = path.normalize(projectData.appResourcesDirectoryPath + path.sep);
```
rosen-vladimirov: why do you need `/` at the end ?

Pip3r4o: So that there are no false-positives when there is a directory which contains similar name.

For example, if your resources remain as `App_Resources`, but you've got another `App_Resources123` directory, you don't want that filtered out. Similarly if you rename the resources directory to something more generic like `resources`, and if you have another `resources_home_page`, the latter should not be filtered out of the application files.

KristianDD: @Pip3r4o added it so we can filter more accurately. For example if you have Resources folder named "native" and folder named "native_classes" we want to filter only files inside "native".

Pip3r4o:
run ci

KristianDD:
run ci

PR #3433, opened 2018-03-08
fix: stop printing double spaces

Mitko-Kerezov:
Whenever installing the platform `tns-android` or `tns-ios` do not print `Installing  <platform>` (with two spaces) but instead print only a single space between the words

Ping @rosen-vladimirov @Fatme


Mitko-Kerezov:
run ci

PR #3437, opened 2018-03-09
fix(userSettings): Don't fail when user settings file is not a valid json

Fatme:
In case when user executes some command and by some reasons user settings file is not a valid json, {N} CLI throws "Unexpected token in JSON at position 0" error. After that user is not able to execute any cli command until the file exists.


dtopuzov:
run ci

rosen-vladimirov:
run ci

rosen-vladimirov:
run ci

PR #3448, opened 2018-03-12
fix(plugin-build): strip npm scope names from android plugin names

Pip3r4o:
Don't merge before: https://github.com/telerik/mobile-cli-lib/pull/1063

Problem:
Plugins contained within an npm scope, like `@nota/nativescript-accessibility-ext` fail to build because their android plugin name becomes `@nota/nativescript-accessibility-ext`, both `@` and `/` are illegal characters in an Android package name.

Proposed solution:
Split the plugin name string by a forward slash (`/`) and take the last element, effectively removing all forward slashes, and omitting the npm scope name, if any.


Plamen5kov:
run CI

PR #3452, opened 2018-03-13
chore(nsconfig): export constants and getDefautNConfig to public API

KristianDD:


KristianDD:
Ping @rosen-vladimirov @Mitko-Kerezov

lib/services/project-data-service.ts
```diff
@@ -124,5 +124,13 @@ export class ProjectDataService implements IProjectDataService {
 			projectFilePath
 		};
 	}
+
+	@exported("projectDataService")
+	public getNsConfigDefaultContent(data: Object): string {
```
rosen-vladimirov: can you please add the method to the lib tests: https://github.com/NativeScript/nativescript-cli/blob/master/test/nativescript-cli-lib.ts#L19

lib/project-data.ts
```diff
@@ -169,6 +169,10 @@ export class ProjectData implements IProjectData {
 		return this.$fs.readText(configNSFilePath);
 	}
 
+	public getNsConfigRelativePath() {
```
Fatme: Missing return type
```
public getNsConfigRelativePath(): string { ... }
```

lib/services/project-data-service.ts
```diff
@@ -124,5 +124,13 @@ export class ProjectDataService implements IProjectDataService {
...
+		const config: INsConfig = {};
+		Object.assign(config, data);
+
+		return JSON.stringify(config);
```
Fatme: Can we use `return JSON.stringify(data);` If it is possible `config` will not be needed.

PublicAPI.md
```diff
@@ -187,6 +188,17 @@ Returns an IProjectData object that is initialized with the provided package.jso
 getProjectDataFromContent(packageJsonContent: string, nsconfigContent: string, projectDir?: string): IProjectData
 ```
 
+### getNsConfigDefaultContent
+Returns the default content of "nsconfig.json" merged with the properties provided by the the `data` argument.
+* Definition:
+```TypeScript
+/**
+ * Returns the default content of "nsconfig.json" merged with the properties provided by the the "data" argument.
```
sis0k0: typo: double `the`

PublicAPI.md
```diff
@@ -187,6 +188,17 @@ Returns an IProjectData object that is initialized with the provided package.jso
...
+ * @param {string} data Properties that should not be defaulted.
```
sis0k0: ` * @param {string} `, isn't the type `Object`?

PublicAPI.md
sis0k0: double `the`
`provided by the the "data"` -> `provided by the "data"`

lib/project-data.ts
```diff
@@ -169,6 +169,10 @@ export class ProjectData implements IProjectData {
...
+		return  constants.CONFIG_NS_FILE_NAME;
```
sis0k0: There are two white spaces between `return` and `constants.`

lib/services/project-data-service.ts
KristianDD: The main idea of the method is to merge default values with some overrides. For now nsconfig's default state is an empty object but writing the method this way makes it a bit more readable.

KristianDD:
run ci

KristianDD:
run ci

PR #3455, opened 2018-03-13
Get sysInfos based on specified platform

Fatme:
Related to this PR https://github.com/NativeScript/nativescript-doctor/pull/30
Should be merged after this PR https://github.com/telerik/mobile-cli-lib/pull/1064

Fixes cloud builds regression problem.

### DIFF
--- a/lib/services/doctor-service.ts
+++ b/lib/services/doctor-service.ts
@@ -20,7 +20,7 @@ class DoctorService implements IDoctorService {
 		private $terminalSpinnerService: ITerminalSpinnerService,
 		private $versionsService: IVersionsService) { }
 
-	public async printWarnings(configOptions?: { trackResult: boolean }): Promise<boolean> {
+	public async printWarnings(configOptions?: { trackResult: boolean }): Promise<void> {
 		const infos = await this.$terminalSpinnerService.execute<NativeScriptDoctor.IInfo[]>({
 			text: `Getting environment information ${EOL}`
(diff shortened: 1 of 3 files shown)


lib/services/doctor-service.ts
```diff
@@ -70,10 +70,8 @@ class DoctorService implements IDoctorService {
 	}
 
 	public async canExecuteLocalBuild(platform?: string): Promise<boolean> {
-		let infos = await doctor.getInfos();
-		if (platform) {
-			infos = this.filterInfosByPlatform(infos, platform);
-		}
+		let infos = await doctor.getInfos({ platform: platform });
```
rosen-vladimirov: You can use short syntax here:
```TypeScript
const infos = await doctor.getInfos({ platform });
```

lib/services/doctor-service.ts
```diff
@@ -42,15 +42,15 @@ class DoctorService implements IDoctorService {
 		if (hasWarnings) {
 			this.$logger.info("There seem to be issues with your configuration.");
 			await this.promptForHelp();
+		} else {
+			this.$logger.out("No issues were detected.".bold);
```
rosen-vladimirov: will we print this on each `tns build ...` when there are no issues?

Fatme: No. This will be printed only on doctor command

rosen-vladimirov:
run ci

PR #3457, opened 2018-03-14
Add spinner when checking for NativeScript components versions

Fatme:
Currently when doctor command is executed, NativeScript components versions information is displayed as table. With this PR NativeScript components versions will be displayed with terminal spinners.

In case when component is up to date with latest available version, success spinner will be displayed.
In case when for component update is available, warn spinner will be displayed.
In case when component is not installed, fail spinner will be displayed.

This PR changes also the behaviour of `tns info` command. When `tns info` command is executed,  NativeScript components versions information also will be displayed with terminal spinners.

Should be merged with this PR https://github.com/telerik/mobile-cli-lib/pull/1066

### DIFF
--- a/lib/services/versions-service.ts
+++ b/lib/services/versions-service.ts
@@ -1,22 +1,28 @@
-import { EOL } from "os";
 import * as constants from "../constants";
+import * as helpers from "../common/helpers";
 import * as semver from "semver";
 import * as path from "path";
-import { createTable } from "../common/helpers";
+
+export enum VersionInformationType {
+	UpToDate = "UpToDate",
+	UpdateAvailable = "UpdateAvailable",
+	NotInstalled = "NotInstalled"
+}
 
 class VersionsService implements IVersionsService {
-	private static UP_TO_DATE_MESSAGE = "Up to date".green.toString();
-	private static UPDATE_AVAILABLE_MESSAGE = "Update available".yellow.toString();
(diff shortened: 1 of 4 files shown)


lib/services/versions-service.ts
```diff
@@ -1,22 +1,28 @@
-import { EOL } from "os";
 import * as constants from "../constants";
+import * as helpers from "../common/helpers";
 import * as semver from "semver";
 import * as path from "path";
-import { createTable } from "../common/helpers";
+
+export enum VersionInformationType {
+	UpToDate,
```
Mitko-Kerezov: You could assign values here, should you decide to follow my suggestion in `mobile-cli-lib`'s PR

lib/services/versions-service.ts
```diff
@@ -127,30 +109,50 @@ class VersionsService implements IVersionsService {
...
+
+		if (!helpers.isInteractive()) {
+			versionsInformation.map(componentInformation => this.$logger.out(componentInformation.message));
+		}
+
+		_.forEach(versionsInformation, componentInformation => {
+			const spinner = this.$terminalSpinnerService.createSpinner();
+			spinner.text = componentInformation.message;
+
+			if (componentInformation.type === VersionInformationType.UpToDate) {
```
Mitko-Kerezov: We could use a `switch` here instead of multiple `if` statements

Fatme:
run ci

Issue #3451, opened 2018-03-13
tns plugin remove fails with : Cannot convert undefined or null to object


zbranzov:
### Tell us about the problem
tns plugin remove fails with: Cannot convert undefined or null to object when trying to remove "nativescript-socket.io" plugin.

### Which platform(s) does your issue occur on?
Both

### Please provide the following version numbers that your issue occurs with:
- CLI: 4.0.0-2018-03-09-11006
- Cross-platform modules: 3.4.1
- Runtime(s): "tns-android": {
      "version": "4.0.0-2018.3.12.1"
    }
- Plugin(s): "nativescript-socket.io": "^0.9.0"

### Please tell us how to recreate the issue in as much detail as possible.
1. tns plugin add nativescript-socket.io
2. tns plugin remove nativescript-socket.io      => Cannot convert undefined or null to object
3.  The error below is shown when using --log trace

mcsofbranzov:demo branzov$ ../node_modules/.bin/tns plugin remove nativescript-socket.io --log trace
Loading extensions.
System information:
{ platform: 'darwin',
  shell: '/bin/bash',
  os: 'Darwin mcsofbranzov 17.4.0 Darwin Kernel Version 17.4.0: Sun Dec 17 09:19:54 PST 2017; root:xnu-4570.41.2~1/RELEASE_X86_64 x86_64\n',
  procArch: 'x64',
  nodeVer: '6.10.0',
  npmVer: '3.10.10',
  nodeGypVer: null,
  dotNetVer: null,
  javacVersion: '1.8.0_121',
  xcodeVer: '9.2.0',
  xcodeprojLocation: '/usr/local/bin/xcodeproj',
  itunesInstalled: true,
  cocoaPodsVer: '1.4.0',
  adbVer: '1.0.39',
  androidInstalled: true,
  monoVer: null,
  gitVer: '2.14.3 (Apple Git-98)',
  gradleVer: null,
  isCocoaPodsWorkingCorrectly: true,
  nativeScriptCliVersion: '3.4.3',
  isCocoaPodsUpdateRequired: false,
  isAndroidSdkConfiguredCorrectly: true,
  pythonInfo: { isInstalled: true, isSixPackageInstalled: true } }
Looking for project in '/Users/branzov/Git/testRC/demo'
Project directory is '/Users/branzov/Git/testRC/demo'.
Initializing analytics statuses.
[...]

PR #3459, opened 2018-03-14
fix: Plugin remove command fails in case plugin name has dot

rosen-vladimirov:
In case plugin name has dot (`.`) in its name, the `plugin remove` command fails.
The reason is the way we are trying to remove plugin related data from the `nativescript` key of application's package.json.
We are using `.` to concatenate the keys inside nativescript key and later we are trying to split by `.` in order to find which key should be deleted.
This works in most of the cases, but when the plugin name has `.` in it, our logic fails as the split operation finds the dot.
Fix this by using some symbols that are not allowed in npm for plugins names. Use symbols that will rarely be used in the same sequence.

Fixes https://github.com/NativeScript/nativescript-cli/issues/3451

### DIFF
--- a/lib/constants.ts
+++ b/lib/constants.ts
@@ -139,3 +139,8 @@ export const enum BuildStates {
 }
 
 export const NATIVESCRIPT_CLOUD_EXTENSION_NAME = "nativescript-cloud";
+
+/**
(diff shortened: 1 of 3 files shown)


rosen-vladimirov:
run ci

PR #3464, opened 2018-03-15
fix-next(prepare): clean platforms/.../app/ on bundle

sis0k0:
If prepare of the project is required and the bundle option is passed,
the platforms/.../app/ dir should be deleted as there may be old assets
left.

Ex.
When the app bundled with snapshot enabled:
```
tns build android --bundle --env.snapshot
```
this produces some assets:
```
platforms/android/.../app/vendor.js
platforms/android/.../app/_embedded_script.js
// ...
```

Then, if the project is bundled without snapshot:
```
tns build android --bundle
```
the produced assets will override the ones that are already in
`platforms/android/.../app`. However, since the build is without
snapshot, an `_embedded_script.js` won't be generated to override the one
that's left in `platforms/android/.../app` from the previous build.

fixes https://github.com/NativeScript/nativescript-dev-webpack/issues/463

### DIFF
--- a/lib/services/platform-service.ts
+++ b/lib/services/platform-service.ts
@@ -209,7 +209,7 @@ export class PlatformService extends EventEmitter implements IPlatformService {
 			// Always clear up the app directory in platforms if `--bundle` value has changed in between builds or is passed in general
 			// this is done as user has full control over what goes in platforms when `--bundle` is passed
 			// and we may end up with duplicate symbols which would fail the build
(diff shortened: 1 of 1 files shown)


dtopuzov:
run ci

sis0k0:
This method is ran whenever a change is detected and that's why the dist dir shouldn't be cleaned here. CLosing in favor of https://github.com/NativeScript/nativescript-dev-webpack/pull/465

PR #3476, opened 2018-03-20
Fix issue with missing projectName.

ginev:


ginev:
Addressing: https://github.com/NativeScript/ios-runtime/issues/893

PR #3485, opened 2018-03-21
fix(assets): Fix asset genration 

rosen-vladimirov:
Fix the generation of asset - the filter is incorrect and fails with `Cannot read property 'splashImages' of null, TypeError: Cannot read property 'splashImages' of null`
Also fix the scale - it has been incorrectly declared as number, when it is string (1x, 2x, etc.). Use the scale correctly by parsing it and use it when generate the image.
Also fix the using of resizeOperation property - it was not passed correctly for iOS assets.

chore: Update to latest common lib

### DIFF
--- a/lib/services/assets-generation/assets-generation-service.ts
+++ b/lib/services/assets-generation/assets-generation-service.ts
@@ -1,6 +1,7 @@
 import * as Jimp from "jimp";
 import * as Color from "color";
 import { exported } from "../../common/decorators";
+import { AssetConstants } from "../../constants";
 
 export const enum Operations {
 	OverlayWith = "overlayWith",
@@ -39,12 +40,12 @@ export class AssetsGenerationService implements IAssetsGenerationService {
 		const assetsStructure = await this.$projectDataService.getAssetsStructure(generationData);
 
 		const assetItems = _(assetsStructure)
-			.filter((assetGroup: IAssetGroup, platform: string) => {
-				return !generationData.platform || platform.toLowerCase() === generationData.platform.toLowerCase();
-			})
+			.filter((assetGroup: IAssetGroup, platform: string) =>
+				!generationData.platform || platform.toLowerCase() === generationData.platform.toLowerCase()
+			)
 			.map((assetGroup: IAssetGroup) =>
 				_.filter(assetGroup, (assetSubGroup: IAssetSubGroup, imageTypeKey: string) =>
-					propertiesToEnumerate.indexOf(imageTypeKey) !== -1 && !assetSubGroup[imageTypeKey]
+					assetSubGroup && propertiesToEnumerate.indexOf(imageTypeKey) !== -1 && assetSubGroup[imageTypeKey]
 				)
 			)
 			.flatten<IAssetSubGroup>()
@@ -55,7 +56,14 @@ export class AssetsGenerationService implements IAssetsGenerationService {
 
 		for (const assetItem of assetItems) {
 			const operation = assetItem.resizeOperation || Operations.Resize;
-			const scale = assetItem.scale || 0.8;
+			let tempScale: number = null;
+			if (assetItem.scale && !_.isNumber(assetItem.scale)) {
+				const splittedElements = `${assetItem.scale}`.split(AssetConstants.sizeDelimiter);
+				tempScale = splittedElements && splittedElements.length && splittedElements[0] && +splittedElements[0];
+			}
+
+			const scale = tempScale || 0.8;
+
 			const outputPath = assetItem.path;
 
 			switch (operation) {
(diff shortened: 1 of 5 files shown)


lib/services/assets-generation/assets-generation-service.ts
```diff
@@ -55,7 +56,14 @@ export class AssetsGenerationService implements IAssetsGenerationService {
 
 		for (const assetItem of assetItems) {
 			const operation = assetItem.resizeOperation || Operations.Resize;
-			const scale = assetItem.scale || 0.8;
+			let tempScale: number = null;
```
Fatme: In case when `assetItem.scale` is a number, `scale` will always be `0.8` and `assetItem.scale` will be disregarded. Is it the desired behaviour?

If not it should be `let tempScale = assetItem.scale`

lib/services/project-data-service.ts
```diff
@@ -137,25 +137,32 @@ export class ProjectDataService implements IProjectDataService {
...
-			} else {
-				// Find the image size based on the hardcoded values in the image-definitions.json
-				_.each(imageDefinitions, (assetSubGroup: IAssetItem[]) => {
-					_.each(assetSubGroup, assetItem => {
-						if (assetItem.filename === image.filename && path.basename(assetItem.directory) === path.basename(dirPath)) {
-							image.width = assetItem.width;
-							image.height = assetItem.height;
-							image.size = `${assetItem.width}${AssetConstants.sizeDelimiter}${assetItem.height}`;
+			// Find the image size based on the hardcoded values in the image-definitions.json
+			_.each(imageDefinitions, (assetSubGroup: IAssetItem[]) => {
```
Fatme: Just a suggestion (not a merge stopper), not sure if it is working correctly.
```
_(imageDefinitions)
				.filter((assetSubGroup: IAssetItem[]) => _.find(assetSubGroup, assetElement => assetElement.filename === image.filename && path.basename(assetElement.directory) === path.basename(dirPath)))
				.map((assetItem: IAssetItem) => {
					if (image.size) {
						// size is basically <width>x<height>
						const [width, height] = image.size.toString().split(AssetConstants.sizeDelimiter);
						if (width && height) {
							image.width = +width;
							image.height = +height;
						}
					}

					if (!image.width || !image.height) {
						image.width = assetItem.width;
						image.height = assetItem.height;
						image.size = image.size || `${assetItem.width}${AssetConstants.sizeDelimiter}${assetItem.height}`;
					}
				});
```

lib/services/assets-generation/assets-generation-service.ts
```diff
@@ -55,7 +56,14 @@ export class AssetsGenerationService implements IAssetsGenerationService {
...
+			if (assetItem.scale && !_.isNumber(assetItem.scale)) {
+				const splittedElements = `${assetItem.scale}`.split(AssetConstants.sizeDelimiter);
+				tempScale = splittedElements && splittedElements.length && splittedElements[0] && +splittedElements[0];
+			}
+
+			const scale = tempScale || 0.8;
```
ivanovit: The scale must be applied to all resize operations.

lib/services/assets-generation/assets-generation-service.ts
rosen-vladimirov: Thanks , I've fixed it in the code below

lib/services/assets-generation/assets-generation-service.ts
```diff
@@ -55,20 +56,34 @@ export class AssetsGenerationService implements IAssetsGenerationService {
...
+			if (assetItem.scale) {
+				if (_.isNumber(assetItem.scale)) {
+					tempScale = assetItem.scale;
+				} else {
+					const splittedElements = `${assetItem.scale}`.split(AssetConstants.sizeDelimiter);
+					tempScale = splittedElements && splittedElements.length && splittedElements[0] && +splittedElements[0];
+				}
+			}
+
+			const scale = tempScale || 1;
```
Fatme: 1 can be extracted as a constant.

lib/services/assets-generation/assets-generation-service.ts
```diff
@@ -55,20 +56,34 @@ export class AssetsGenerationService implements IAssetsGenerationService {
...
+
 			const outputPath = assetItem.path;
+			const width = assetItem.width * scale;
+			const height = assetItem.height * scale;
 
 			switch (operation) {
 				case Operations.OverlayWith:
-					const imageResize = Math.round(Math.min(assetItem.width, assetItem.height) * scale);
+					const overlayImageScale = assetItem.overlayImageScale || 0.8;
```
Fatme: Also 0.8 can be extracted as a constant

PR #3492, opened 2018-03-23
fix(AppResources): try recover if resources update fails

KristianDD:
Fixes https://github.com/NativeScript/nativescript-cli/issues/3510

### DIFF
--- a/lib/services/android-resources-migration-service.ts
+++ b/lib/services/android-resources-migration-service.ts
@@ -1,6 +1,6 @@
 import * as path from "path";
-import * as shell from "shelljs";
 import * as constants from "../constants";
+import { EOL } from "os";
 
 export class AndroidResourcesMigrationService implements IAndroidResourcesMigrationService {
 	private static ANDROID_DIR = "Android";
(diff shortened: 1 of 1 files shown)


KristianDD:
@rosen-vladimirov @Pip3r4o

lib/services/android-resources-migration-service.ts
```diff
@@ -22,6 +22,22 @@ export class AndroidResourcesMigrationService implements IAndroidResourcesMigrat
...
+
+		try {
+			await this.tryMigrate(originalAppResources, appResourcesDestination, appResourcesBackup);
+			this.$logger.out(`Successfully updated your project's application resources '/Android' directory structure.${EOL}The previous version of your Android application resources has been renamed to '/${AndroidResourcesMigrationService.ANDROID_DIR_OLD}'`);
+		} catch (error) {
+			try {
+				this.recover(originalAppResources, appResourcesDestination, appResourcesBackup);
+				this.$logger.out("Failed to update resources. They should be in their initial state.");
+			} catch (error) {
+				this.$logger.out(`Failed to update resources.${EOL} Backup of original content is inside "${appResourcesBackup}".${EOL}If "${originalAppResources} if missing copy from backup folder."`);
```
rosen-vladimirov: `If "${originalAppResources} iS missing...`

rosen-vladimirov: Why don't we throw the original error here? Currently the command exits with code 0, as we've caught all errors, which is incorrect.

rosen-vladimirov: Can you also add `this.$logger.trace` with the currently caught error?

PR #3500, opened 2018-03-26
fix: Move @types/ora as devDependency

Fatme:
Ora's definition files are not needed in production package, so move it as devDependency.


Fatme:
run ci

Fatme:
run ci

Fatme:
run ci

PR #3501, opened 2018-03-26
feat(commands): Introduce postCommandAction

rosen-vladimirov:
Introduce postCommandAction that is executed when the command succeeds.
Use the action to print additional information to user how to continue after command is finished.
Show some information on successful postinstall - instruct the users how to use local builds, cloud builds and playground.

### DIFF
--- a/lib/commands/post-install.ts
+++ b/lib/commands/post-install.ts
@@ -18,6 +18,27 @@ export class PostInstallCliCommand extends PostInstallCommand {
 
 		await this.$subscriptionService.subscribeForNewsletter();
 	}
+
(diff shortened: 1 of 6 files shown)


lib/commands/post-install.ts
```diff
@@ -9,7 +9,7 @@ export class PostInstallCliCommand extends PostInstallCommand {
 		$settingsService: ISettingsService,
 		$doctorService: IDoctorService,
 		$analyticsService: IAnalyticsService,
-		$logger: ILogger) {
+		protected $logger: ILogger) {
```
Fatme: `$protected` is not needed

PR #3542, opened 2018-04-22
feat: Allow templates to be full applications

rosen-vladimirov:
Currently when a project is created the content of the template is placed inside the `app` directory of the newly created project.
This leads to some issues when you want to support more complex scenarios, for example it is difficult to add configuration file (like nsconfig.json or webpack.config.js) in the root of the project.
The suggested solution to allow templates to be the full application is to check the template from which the application is created. In case the template contains a nativescript key and templateVersion property in it, check its value.
In case it is v1, use the old way, i.e. place the content of the template in the app directory of the created project.
In case it is v2 place the content of the template at the root of the application.
In case it is anything else - throw an error.
In case it is missing, use v1 as default.

The solution ensures backwards compatiblity with existing templates and allows creating new types of templates.

## PR Checklist

### DIFF
--- a/lib/services/project-service.ts
+++ b/lib/services/project-service.ts
@@ -2,14 +2,15 @@ import * as constants from "../constants";
 import * as path from "path";
 import * as shelljs from "shelljs";
 import { exported } from "../common/decorators";
+import { Hooks } from "../constants";
 
 export class ProjectService implements IProjectService {
 
-	constructor(private $npm: INodePackageManager,
+	constructor(private $hooksService: IHooksService,
+		private $npm: INodePackageManager,
 		private $errors: IErrors,
 		private $fs: IFileSystem,
 		private $logger: ILogger,
-		private $projectData: IProjectData,
 		private $projectDataService: IProjectDataService,
 		private $projectHelper: IProjectHelper,
 		private $projectNameService: IProjectNameService,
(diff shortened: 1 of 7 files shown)


Fatme:
run ci

rosen-vladimirov:
run ci

dtopuzov:
run ci

dtopuzov:
run ci

lib/services/project-service.ts
```diff
@@ -99,21 +106,32 @@ export class ProjectService implements IProjectService {
...
-	private async extractTemplate(projectDir: string, realTemplatePath: string): Promise<void> {
+	private async extractTemplate(projectDir: string, realTemplatePath: string, templateVersion: string): Promise<void> {
 		this.$fs.ensureDirectoryExists(projectDir);
 
-		const appDestinationPath = this.$projectData.getAppDirectoryPath(projectDir);
-		this.$fs.createDirectory(appDestinationPath);
+		this.$logger.trace(`Template version is ${templateVersion}`);
+		let destinationDir = "";
+		switch (templateVersion) {
+			case constants.TemplateVersions.v2:
```
KristianDD: Because we always defatult to v1 [here](https://github.com/NativeScript/nativescript-cli/pull/3542/files#diff-123e2a8a4e8daddf9ac8a60b8c79546dR65), maybe will be better that the `default` case is v2 and above.
```Typescript
case constants.TemplateVersions.v1:
    const appDestinationPath = this.$projectData.getAppDirectoryPath(projectDir);
    this.$fs.createDirectory(appDestinationPath);
    destinationDir = appDestinationPath;
    break;
default:
   destinationDir = projectDir;
```

lib/services/project-service.ts
```diff
@@ -57,17 +57,24 @@ export class ProjectService implements IProjectService {
...
+			// Install devDependencies and execute all scripts:
 			await this.$npm.install(projectDir, projectDir, {
 				disableNpmInstall: false,
 				frameworkPath: null,
 				ignoreScripts: projectOptions.ignoreScripts
 			});
 
-			const templatePackageJson = this.$fs.readJson(path.join(templatePath, "package.json"));
 			await this.$npm.uninstall(templatePackageJson.name, { save: true }, projectDir);
+			if (templateVersion === constants.TemplateVersions.v2) {
```
KristianDD: `if (templateVersion !== constants.TemplateVersions.v2)` will allow bumping versions without having to alter code.

KristianDD:
Maybe we can use semver for templateVersion. Just curious why we chose custom versioning.

lib/services/project-service.ts
rosen-vladimirov: We cannot be sure what will be required in the next versions, so I prefer keeping the strictness here - for v2 do this. In case we decide to use v3 and to make it work in a similar manner , we'll alter the code.

KristianDD: I find it more likely that v3 will be derived from v2 structure rather than from the one of v1.

rosen-vladimirov:
@KristianDD  I've used REST API versioning as it seems more suitable for the current case where we do no have a specific logic for patches and minor versions. We have absolutely different logic for the two types of templates.
Using semver means we have different logic for minor and patch versions, which is not the case at the moment. Also I do not see why we would have a different logic for template's minor version for example.

PR #3557, opened 2018-05-02
fix(assets-generation): Do not fail in case some App_Resources are missing

rosen-vladimirov:
In case you delete `App_Resources/iOS` directory, the `getAssetsStructure` method fails as it tries to read `Content.json` files from the deleted directory. Instead of failing, just return empty arrays for the images. This is already working this way for Android.
Add test for the case

## PR Checklist

Trying to generate/get assets structure fails when App_Resources/iOS is deleted

## What is the new behavior?
The generation of assets will work for Android in this case.

Fixes https://github.com/NativeScript/sidekick-feedback/issues/168 .

### DIFF
--- a/lib/services/project-data-service.ts
+++ b/lib/services/project-data-service.ts
@@ -125,7 +125,7 @@ export class ProjectDataService implements IProjectDataService {
 
 	private async getIOSAssetSubGroup(dirPath: string): Promise<IAssetSubGroup> {
 		const pathToContentJson = path.join(dirPath, AssetConstants.iOSResourcesFileName);
-		const content = <IAssetSubGroup>this.$fs.readJson(pathToContentJson);
+		const content = this.$fs.exists(pathToContentJson) && <IAssetSubGroup>this.$fs.readJson(pathToContentJson) || { images: [] };
 
 		const imageDefinitions = this.getImageDefinitions().ios;
 
--- a/test/services/project-data-service.ts
+++ b/test/services/project-data-service.ts
@@ -2,8 +2,9 @@ import { Yok } from "../../lib/common/yok";
 import { assert } from "chai";
 import { ProjectDataService } from "../../lib/services/project-data-service";
 import { LoggerStub } from "../stubs";
-import { NATIVESCRIPT_PROPS_INTERNAL_DELIMITER } from '../../lib/constants';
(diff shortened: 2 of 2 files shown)


lib/services/project-data-service.ts
```diff
@@ -125,7 +125,7 @@ export class ProjectDataService implements IProjectDataService {
 
 	private async getIOSAssetSubGroup(dirPath: string): Promise<IAssetSubGroup> {
 		const pathToContentJson = path.join(dirPath, AssetConstants.iOSResourcesFileName);
-		const content = <IAssetSubGroup>this.$fs.readJson(pathToContentJson);
+		const content = this.$fs.exists(pathToContentJson) && <IAssetSubGroup>this.$fs.readJson(pathToContentJson) || { images: [] };
```
Fatme: What will happen in case when `pathToContentJson` exists but the file is not a valid json? I'm expecting to throw an error in this case.

test/services/project-data-service.ts
```diff
@@ -244,4 +264,39 @@ describe("projectDataService", () => {
...
+
+	describe("getAssetsStructure", () => {
+		it("does not fail when App_Resources/Android and App_Resources/iOS do not exist", async () => {
+			const defaultEmptyData: any = {};
+			defaultEmptyData[CLIENT_NAME_KEY_IN_PROJECT_FILE] = {};
+			const testInjector = createTestInjector(JSON.stringify(defaultEmptyData));
+			const fs = testInjector.resolve<IFileSystem>("fs");
+			fs.readJson = (filePath: string): any => {
+				if (basename(filePath) === AssetConstants.imageDefinitionsFileName) {
+					return { android: {}, ios: {} };
```
Fatme: In case when `imageDefinitionsFileName` file does not exist, an error will be thrown. I'm wondering if we should check this case.

lib/services/project-data-service.ts
rosen-vladimirov: Yeah, I prefer to throw the error in this case as it is not expected to have a non-json file here.

test/services/project-data-service.ts
rosen-vladimirov: The file is part of CLI resources, so in case it does not exist, something terrible happened :smile_cat:

Issue #3570, opened 2018-05-09
Add deprecation message for macOS Sierra


Natalia-Hristova:
### Tell us about the problem
Message have to be added to notify users about future deprecation of macOS Sierra.

PR #3588, opened 2018-05-15
feat: Deprecate support for all macOS versions below High Sierra

rosen-vladimirov:
Add deprecation message in case the current macOS version is below High Sierra. As the latest Xcode versions can be installed only on this OS and we state that we support only latest Xcode, we have to deprecate the support for other OS versions.
The deprecation message will be shown on each CLI command.
As the deprecation messages from CLI should be shown in Sidekick as well, expose new methods in CLI's `sysInfo` moodule, so Sidekick can get the warnings and show them.
In order to get the warnings for Node.js version, change the way `verifyNodeVersion` is called.
Expose method for getting the currently supported Node.js version range.
Add tests for the new methods.

## PR Checklist

All macOS versions are supported

## What is the new behavior?
When macOS is not HighSierra, CLI prints a warning that this version will not be supported in the future.

Implements https://github.com/NativeScript/nativescript-cli/issues/3570


PublicAPI.md
```diff
@@ -1151,6 +1156,47 @@ tns.androidProcessService.getAppProcessId("4df18f307d8a8f1b", "org.nativescript.
...
+console.log(nodeJsRange);
+```
+
+### getSystemWarnings
+The `getSystemWarnings` methods returns all deprecation warnings for current environment. For example, in case the support for the current OS is deprecated by CLI, the method will return array with one message describing the deprecation.
+
+* Definition
+```TypeScript
+/**
+ * Gets all globall warnings for the current environment, for example Node.js version compatibility, OS compatibility, etc.
```
Fatme: `globall` -> `global`

PR #3592, opened 2018-05-16
fix(GDPR): Do not track local project paths

rosen-vladimirov:
Do not track paths to projects when local template is used. Instead track `localTemplate_<name from template's package.json or dirname>`.
Also fix the prompter for emails as we need to show information how to unsubscribe from the newsletter.
Add tests for the behavior.

## PR Checklist

Fixes https://github.com/NativeScript/nativescript-cli/issues/3595

### DIFF
--- a/lib/services/project-templates-service.ts
+++ b/lib/services/project-templates-service.ts
@@ -22,11 +22,14 @@ export class ProjectTemplatesService implements IProjectTemplatesService {
 
 		await this.$analyticsService.track("Template used for project creation", templateName);
 
-		await this.$analyticsService.trackEventActionInGoogleAnalytics({
-			action: constants.TrackActionNames.CreateProject,
-			isForDevice: null,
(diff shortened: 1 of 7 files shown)


lib/services/project-templates-service.ts
```diff
@@ -46,5 +46,21 @@ export class ProjectTemplatesService implements IProjectTemplatesService {
 		this.$logger.trace(`Using NativeScript verified template: ${templateName} with version ${version}.`);
 		return this.$npmInstallationManager.install(templateName, projectDir, { version: version, dependencyType: "save" });
 	}
+
+	private getTemplateNameToBeTracked(templateName: string, realTemplatePath: string): string {
```
Fatme: Not sure, but maybe this should be in `try { } catch { }`. In case something fails here we don't want to stop the command execution.

rosen-vladimirov: Nice idea, I've added the try-catch and test for this case.

PR #3624, opened 2018-05-27
fix: platform remove command fails with old Android runtime

rosen-vladimirov:
After upgrading `nativescript-doctor`, the version of the Java is
validated against the current Android runtime version. In case the
runtime version cannot work with Java 10, `nativescript-doctor` does not
allow the operation to continue. This breaks the `platform remove`
command, as it checks the system requirements first, before executing
the actual removal. However, this is no longer required as the code for
removing the platform is `try-catched` and in case the current
environment is not setup correctly, it will try to remove the
platforms/<platform> dir and the respective key in package.json.
So remove the environment validation from this command. Also remove the
`validatePlatformInstalled` call from the `canExecute` method of the
command, as it is actually called in the `removePlatforms` method.

## PR Checklist

- [] There is an issue for the bug/feature this PR is for. To avoid wasting your time, it's best to open a suggestion issue first and wait for approval before working on it.

Trying to execute `tns platform remove android` on an older project, when Java version on local machine is 10, fails with error that you need to update your Android runtime version.
## What is the new behavior?
You can successfully remove older platforms with `tns platform remove android` even when Java version is 10.

### DIFF
--- a/lib/commands/remove-platform.ts
+++ b/lib/commands/remove-platform.ts
@@ -3,8 +3,7 @@ export class RemovePlatformCommand implements ICommand {
 
 	constructor(private $platformService: IPlatformService,
 		private $projectData: IProjectData,
-		private $errors: IErrors,
-		private $platformsData: IPlatformsData) {
+		private $errors: IErrors) {
 		this.$projectData.initializeProjectData();
 	}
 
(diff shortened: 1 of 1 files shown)


Natalia-Hristova:
issue: https://github.com/NativeScript/nativescript-cli/issues/3632

PR #3626, opened 2018-05-28
fix: platform update/clean commands may fail with Java 10

rosen-vladimirov:
Add logic to validate environment based on the specified runtime version. This is required for the cases where the runtime version is not specified in the project's package.json but we need to verify if a runtime version can be used
For `tns platform update <platform>` command validate the runtime version that will be installed based on the passed argument. This has two cases:
1. In case version is specified: `tns platform update android@4.0.1` - we'll get the 4.0.1 version and validate if it can be used with currently installed Java compiler version.
2. In case version is not specified: `tns platform update android` - we'll not validate the java version here as we know we'll install at least 4.1.0 runtime, which is compatible with all Java versions

For `tns platform clean <platform>` command we need to validate the current platform immediately as the command removes it and adds the same version after that.

Depends on: https://github.com/NativeScript/nativescript-doctor/pull/37 and https://github.com/telerik/mobile-cli-lib/pull/1096

### DIFF
--- a/lib/commands/platform-clean.ts
+++ b/lib/commands/platform-clean.ts
@@ -5,7 +5,7 @@ export class CleanCommand implements ICommand {
 		private $projectData: IProjectData,
 		private $platformService: IPlatformService,
 		private $errors: IErrors,
-		private $platformsData: IPlatformsData) {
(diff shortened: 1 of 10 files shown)


rosen-vladimirov:
run ci

Natalia-Hristova:
run ci

PR #3711, opened 2018-06-27
fix(livesync): check node_modules for changes only with `--sync-all-files`

### LINKED ISSUE #3710: CLI regenerates ALL aar files on changes in `node_modules`
### Did you verify this is a real problem by searching the [NativeScript Forum](http://forum.nativescript.org) and the [other open issues in this repo](https://github.com/NativeScript/nativescript/issues)?
:+1: 

#### Setup:
Application with at least one plugin that requires the CLI to generate aar file. For example - the [sdk samples](https://github.com/nativescript/nativescript-sdk-examples-ng) that use the `nativescript-camera` plugin.

##### First case (webpack):
```
tns run android --bundle
```
##### Second case (no webpack):
```
tns run android
```

#### First case (webpack)
When you change a file in the `node_modules/` folder, for example `node_modules/tns-core-modules/bundle-entry-points.js`:
1. The `mtime` (modified time) of the file is updated.
1. Webpack detects the change and recompiles the JS bundles (for example, `bundle.js` or `vendor.js`). Then it starts the livesync process. 
1. The CLI [detects](https://github.com/NativeScript/nativescript-cli/blob/master/lib/services/project-changes-service.ts#L73) that there was a change in node_modules.
1. Because of the change in `node_modules`, the `modulesChanged` flag is raised and the CLI regenerates all aar files. [see here](https://github.com/NativeScript/nativescript-cli/blob/master/lib/services/prepare-platform-native-service.ts#L34)
1. The changed JS file is deployed on the device and the app is restarted/refreshed.

##### Second case (no webpack)
You have to make two file changes here, because the CLI keeps track of the `mtimes` of the files in `node_modules`, but the watcher detects changes only in the `app/` folder.

1. [User action] Modify a file in `node_modules`, for example - `node_modules/tns-core-modules/bundle-entry-points.js`.
1. The `mtime` of the file is updated.
1. [User action] Modify a file in the `app/` folder, for example - `app/app.css`.
1. The CLI watcher detects that there is a change in the `app/` folder and starts the livesync process, which includes preparing the project.
1. The CLI [detects](https://github.com/NativeScript/nativescript-cli/blob/master/lib/services/project-changes-service.ts#L73) that there was a change in node_modules.
1. Because of the change in `node_modules`, the `modulesChanged` flag is raised and the CLI regenerates all aar files. [see here](https://github.com/NativeScript/nativescript-cli/blob/master/lib/services/prepare-platform-native-service.ts#L34)
1. The changed JS bundles are deployed on the device and the app is restarted.

### Expected behavior
The CLI should keep track of `node_modules` only when the `sync-all-files` flag is used. [Checking for modified files](https://github.com/NativeScript/nativescript-cli/blob/master/lib/services/project-changes-service.ts#L229) in `node_modules` on every change can be quite slow.
> Note: Rebuilding the aar files for every change in `node_modules` (even if that's done only when using the `sync-all-files` flag) is also slow and not necessary. This probably should be logged as separate issue. @rosen-vladimirov, what do you think?

### Why does this happen?
Currently, the `node_modules` are traversed and checked for modified files explicitly. That's because of this change: https://github.com/NativeScript/nativescript-cli/commit/04f9737a13b43e5cd5bd08505dcfbfe18d9243f2#diff-919912d52b0e0303e76febe7c84cacfdR67.
If we get the old code back, the CLI will only check node_modules when the `--sync-all-files` flag is provided.

#### Cons of the proposed change
If we stop keeping track of `node_modules` when using normal livesync or livesync with bundle, we will lose the following functionality:
- Adding a new NativeScript plugin that doesn't have a prebuilt aar file will not work until you stop the CLI process and run a clean build.
- Modifying the native files (`platform/android`) of a plugin in `node_modules` won't rebuild its aar file until you stop the CLI process and run a clean build.

### Which platform(s) does your issue occur on?
Android

### Please provide the following version numbers that your issue occurs with:
- CLI: 4.2.0-2018-06-26-11829 (the same thing happens with the official version - 4.1.2)

sis0k0:
The reasons for this change are described in the issue below.

fixes #3710

### DIFF
--- a/lib/services/project-changes-service.ts
+++ b/lib/services/project-changes-service.ts
@@ -70,13 +70,11 @@ export class ProjectChangesService implements IProjectChangesService {
 			const platformResourcesDir = path.join(projectData.appResourcesDirectoryPath, platformData.normalizedPlatformName);
 			this._changesInfo.appResourcesChanged = this.containsNewerFiles(platformResourcesDir, null, projectData);
 			/*done because currently all node_modules are traversed, a possible improvement could be traversing only the production dependencies*/
-			this._changesInfo.nativeChanged = projectChangesOptions.skipModulesNativeCheck ?
-        false :
-        this.containsNewerFiles(
-          path.join(projectData.projectDir, NODE_MODULES_FOLDER_NAME),
-          path.join(projectData.projectDir, NODE_MODULES_FOLDER_NAME, "tns-ios-inspector"),
-          projectData,
-          this.fileChangeRequiresBuild);
+			this._changesInfo.nativeChanged = projectChangesOptions.skipModulesNativeCheck ? false : this.containsNewerFiles(
+				path.join(projectData.projectDir, NODE_MODULES_FOLDER_NAME),
+				path.join(projectData.projectDir, NODE_MODULES_FOLDER_NAME, "tns-ios-inspector"),
+				projectData,
+				this.fileChangeRequiresBuild);
 
 			if (this._newFiles > 0 || this._changesInfo.nativeChanged) {
 				this._changesInfo.modulesChanged = true;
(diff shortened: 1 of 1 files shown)


dtopuzov:
run ci

PR #3735, opened 2018-07-08
feat: Speed up adding native platform

rosen-vladimirov:
The current logic for adding native platform is:
1. Install the runtime package (tns-ios or tns-android) as a dependency of the project
2. Copy all files from the `<project dir>/node_modules/<runtime package>/framework` to `<project dir>/plaforms/<platform>/...` (specific places there).
3. Uninstall the runtime package, so it is no longer dependency of the project.

This solution becomes really slow when the project has a lot of dependencies. For example, for `tns-template-master-detail-ng`, all these steps take around 2 mins and 5 seconds on Windows (without SSD) and around 40 seconds on macOS (with SSD).

In order to speed up this process, remove the installing and uninstalling of the runtime package. Instead of this, use the `pacote` package to extract the runtime to a directory in the `temp` and copy
files from there to `platforms` directory. As we do not execute `npm install` and `npm uninstall` with this solution, the operation is not dependent on the number of dependencies used in the project. It takes
around 5-9 second on both Windows and macOS.

Also remove the usage of `$options` from `npmInstallationManager` - its purpose was to work with `--frameworkPath` on `tns platform add <platform>` command, but we no longer call `npmInstallationManager` for this case.

Fix the tests for `platorm-service`'s addPlatform method as they were incorrectly passing.

## PR Checklist

`tns platform add <platform>` takes around 40 seconds

## What is the new behavior?
`tns platform add <platform>` takes around 6 seconds.

### DIFF
--- a/lib/npm-installation-manager.ts
+++ b/lib/npm-installation-manager.ts
@@ -6,7 +6,6 @@ export class NpmInstallationManager implements INpmInstallationManager {
 	constructor(private $npm: INodePackageManager,
 		private $childProcess: IChildProcess,
 		private $logger: ILogger,
-		private $options: IOptions,
 		private $settingsService: ISettingsService,
 		private $fs: IFileSystem,
 		private $staticConfig: IStaticConfig,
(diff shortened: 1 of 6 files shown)


Natalia-Hristova:
run ci

Issue #3549, opened 2018-04-26
Podfile generation broken on livesync


vcooley:
### Which platform(s) does your issue occur on?
iOS

### Please provide the following version numbers that your issue occurs with:
- CLI: 4.0.0 
- Cross-platform modules: 4.0.0
- Runtime(s): 4.0.1
- Plugin(s): 
```
    "nativescript-advanced-webview": "^1.1.3",
    "nativescript-angular": "~5.3.0",
    "nativescript-async": "^1.0.2",
    "nativescript-bottombar": "^3.0.8",
    "nativescript-directions": "^1.2.0",
    "nativescript-drop-down": "^3.2.4",
    "nativescript-floatingactionbutton": "^3.0.0",
    "nativescript-geolocation": "^3.0.0",
    "nativescript-google-maps-sdk": "^2.6.0",
    "nativescript-gradient": "^2.0.1",
    "nativescript-input-mask": "^1.0.0",
    "nativescript-iqkeyboardmanager": "~1.1.0",
    "nativescript-ng-shadow": "^2.0.0",
    "nativescript-plugin-firebase": "^5.3.0",
    "nativescript-pro-ui": "^3.3.0",
    "nativescript-snackbar": "^2.0.0",
    "nativescript-theme-core": "^1.0.4",
    "nativescript-unit-test-runner": "^0.3.4",
```

### Please tell us how to recreate the issue in as much detail as possible.
Running `tns run ios --clean` with this configuration produces a working build, however any changes made to the code cause the Podfile to be rebuilt in a way that causes a recursion error.

Here is my Podfile (from `platforms/ios`) after the initial build:
```
use_frameworks!

target "myproject" do
# Begin Podfile - /Users/vincecooley/Projects/myproject/node_modules/nativescript-bottombar/platforms/ios/Podfile 
 use_frameworks!
pod 'MiniTabBar', :git => 'https://github.com/rhanb/MiniTabBar.git'
def post_install1 (installer)
    installer.pods_project.targets.each do |target|
        target.build_configurations.each do |config|
            config.build_settings['SWIFT_VERSION'] = '3.0'
        end
    end
end 
 # End Podfile 
# Begin Podfile - /Users/vincecooley/Projects/myproject/node_modules/nativescript-floatingactionbutton/platforms/ios/Podfile 
 pod 'MNFloatingActionButton', '~> 0.1.5' 
 # End Podfile 
[...]

PR #3768, opened 2018-07-23
fix: Skip preparation of plugins native files in case they are not changed

rosen-vladimirov:
Skip the preparation of plugins' native files in case there's no change of the files since last operation when the files have been prepared.
Do this by using hash sums of files in each plugin's `platforms/<platform>` dir in case such exists. This way, in case you apply a change outside of the `<plugin dir>/platforms/<platform>` dir, the plugin's native files will not be prepared again. This way application will not be rebuilt as there's no need to do this.

### chore: Add logging in projectChangesService
Add logging in projectChangesService's checkForChanges method in order to know which files are changed. This will help us to understand why prepare is triggered.

## PR Checklist

Whenever a change in `node_modules` is detected, CLI prepares the native resources. For example this leads to regeneration of Podfile whenever a change in `.js` or even `.css` inside `node_modules` is detected.

## What is the new behavior?
CLI persists shasums for each plugin's native part. In case the plugin's native part is not changed, CLI will not prepare the native resources of the plugin itself.

Partial fix for #3549

### DIFF
--- a/lib/services/project-changes-service.ts
+++ b/lib/services/project-changes-service.ts
@@ -54,6 +54,7 @@ export class ProjectChangesService implements IProjectChangesService {
 		private $devicePlatformsConstants: Mobile.IDevicePlatformsConstants,
 		private $fs: IFileSystem,
 		private $filesHashService: IFilesHashService,
+		private $logger: ILogger,
 		private $injector: IInjector) {
 	}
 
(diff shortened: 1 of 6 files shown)


lib/services/project-changes-service.ts
```diff
@@ -112,9 +120,11 @@ export class ProjectChangesService implements IProjectChangesService {
 			this._prepareInfo.bundle = projectChangesOptions.bundle;
 		}
 		if (this._changesInfo.packageChanged) {
+			this.$logger.trace("Set modulesChanged to true as packageChanged is true");
 			this._changesInfo.modulesChanged = true;
 		}
 		if (this._changesInfo.modulesChanged || this._changesInfo.appResourcesChanged) {
+			this.$logger.trace(`Set configChanged to true, current value of moduleChanged is: ${this._changesInfo.modulesChanged}, appResourcesChanged is: ${this._changesInfo.appResourcesChanged}`);
```
Fatme: Maybe it is better to print all changesInfo.

test/plugins-service.ts
```diff
@@ -541,4 +546,90 @@ describe("Plugins service", () => {
...
+
+		it("prepares the files when plugin has platforms dir and has not been built before", async () => {
+			const newPluginHashes = { "file": "hash" };
+			const testData = setupTest({ newPluginHashes, hasPluginPlatformsDir: true });
+			await testData.pluginsService.preparePluginNativeCode(testData.pluginData, platform, projectData);
+			assert.isTrue(testData.isPreparePluginNativeCodeCalled);
+			assert.deepEqual(testData.dataPassedToWriteJson, { [testData.pluginData.name]: newPluginHashes });
+		});
+
+		it("does not prepare the files when plugin has platforms dir and has files has not changed since then", async () => {
```
Fatme: Maybe
has files that are not changed since then?

lib/services/project-changes-service.ts
rosen-vladimirov: It is printed at the end of the method. Or you want to know the value of the properties before setting all of them to true?

Fatme: Actually I just want to see all changesInfo and didn't notice it is printed at the end of the method.

PR #3783, opened 2018-07-31
fix: socket not cleared on time for second sync

KristianDD:
## PR Checklist

## What is the new behavior?


lib/services/livesync/android-livesync-tool.ts
```diff
@@ -381,12 +389,21 @@ export class AndroidLivesyncTool implements IAndroidLivesyncTool {
 	private handleSocketError(socketId: string, errorMessage: string) {
 		const error = this.getErrorWithMessage(errorMessage);
 		if (this.socketConnection && this.socketConnection.uid === socketId) {
-			this.end();
+			this.socketError = error;
+			this.end(error);
```
Fatme: Why do we need to pass `error` here as parameter. We can use it directly in `end()` function
```
public end() {
const error = this.socketError || this.getErrorWithMessage()
}
```

KristianDD:
run ci

lib/services/livesync/android-livesync-tool.ts
KristianDD: It is a personal preference. If someone moves `this.socketError = error` one line down or the error is not cleared at the right moment, the method might be broken without us noticing.

PR #3786, opened 2018-08-02
fix: Test command runs tests twice and often fails on Android

rosen-vladimirov:
With the new LiveSync with sockets functionality, the `tns test android` command executes the tests twice and often exits with a non-zero exit code. The problem is that the command uses the LiveSync logic to upload the files first, restart the application after that and runs the tests. However, the new LiveSync requires the application to be running on device in order to have a socket for file transfer. When we start the application in order to execute the LiveSync, the tests are run. During the run the LiveSync may end and CLI will restart the app. If this happens during running the tests, Karma will mark them as failure and the next execution may not be triggered.
Fix this by creating a temp file on device before starting the LiveSync operation. Runtime will check if the file exists and if it does, it will sleep the application for 30 seconds, i.e. it will not load the JS implementation of the app. During this time CLI will transfer all files, remove the temp file from device and restart the application. Runtime will not find the file and it will load the application.

###  fix: LiveSync to Android hangs for 30 seconds
When LiveSync operation is executed, CLI tries to connect to device socket for 30 seconds. In case it succeeds, the timer is not stopped and this keeps the process alive for 30 seconds.

## PR Checklist

`tns test android --justlaunch` executes tests twice, often fails and in many cases does not release the terminal.

## What is the new behavior?
`tns test android --justlaunch` executes tests only once and exits the process immediately after finishing the execution.


dtopuzov:
run ci

dtopuzov:
run ci

PR #3800, opened 2018-08-07
feat: add create plugin command

lini:
## PR Checklist

- [x] The PR title follows our guidelines: https://github.com/NativeScript/NativeScript/blob/master/CONTRIBUTING.md#commit-messages.
- [x] There is an issue for the bug/feature this PR is for. To avoid wasting your time, it's best to open a suggestion issue first and wait for approval before working on it.
- [ ] All existing tests are passing: https://github.com/NativeScript/nativescript-cli/blob/master/CONTRIBUTING.md#contribute-to-the-code-base
- [x] Tests for the changes are included.

## What is the new behavior?

`tns plugin create` command added to CLI. The command will clone the plugin seed and execute the initial seed configuration script to prepare a new project for developing a NativeScript plugin.

PR #3836, opened 2018-08-23
feat: add template option to plugin create command

lini:
Extension of https://github.com/NativeScript/nativescript-cli/pull/3800
Accepting .tar.gz path/URL for custom plugin seed.
+ test
+ help

### DIFF
--- a/docs/man_pages/lib-management/plugin-create.md
+++ b/docs/man_pages/lib-management/plugin-create.md
@@ -6,7 +6,8 @@ position: 1
 
 Usage | Synopsis
 ---|---
-Create a new plugin | `$ tns plugin create <Plugin Repository Name> [--path <Directory>]`
+Create from the default plugin seed | `$ tns plugin create <Plugin Repository Name> [--path <Directory>]`
+Create from a custom plugin seed | `$ tns plugin create <Plugin Repository Name> [--path <Directory>] --template <Template>`
 
 Creates a new project for NativeScript plugin development. The project uses the [NativeScript Plugin Seed](https://github.com/NativeScript/nativescript-plugin-seed) as a base and contains the following directories:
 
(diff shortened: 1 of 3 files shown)


test/plugin-create.ts
```diff
@@ -34,7 +42,11 @@ function createTestInjector() {
 
 	testInjector.register("pacoteService", {
 		manifest: () => Promise.resolve(),
-		extractPackage: () => Promise.resolve()
+		extractPackage: (packageName: string, destinationDirectory: string) => {
+			dummyPacote.destinationDirectory = destinationDirectory;
+			dummyPacote.packageName = packageName;
+			Promise.resolve();
```
rosen-vladimirov: you need to return this Promise

docs/man_pages/lib-management/plugin-create.md
```diff
@@ -20,9 +21,19 @@ The project is setup for easy commit in Github, which is why the command will as
...
\ No newline at end of file
+* `<Plugin Repository Name>` is the name of repository where your plugin will reside. A directory with the same name will be created. For example: `nativescript-awesome-list`. If a directory with the name already exists and is not empty, the plugin create command will fail.<% if(isHtml) { %>
+* `<Template>` can be a URL or a local path to a `.tar.gz` file with the contents of a seed repository. This must be a clone of the [NativeScript Plugin Seed](https://github.com/NativeScript/nativescript-plugin-seed) and must contain a `src` directory with a package.json file and a script at `src/scripts/postclone.js`. After the archive is extracted, the postclone script will be executed with the username (`gitHubUsername`) and plugin name (`pluginName`) parameters given to the `tns plugin create` command prompts. For more information, visit the default plugin seed repository and [examine the source script](https://github.com/NativeScript/nativescript-plugin-seed/blob/master/src/scripts/postclone.js) there. Examples:
```
rosen-vladimirov: `can be a URL or a local path to a .tar.gz` -> in fact it can be anything that can be npm installed, you can pass a local dir path with correct structure (i.e. local copy of the plugin seed for example) and it should work,

lini: I tried it and it only worked with .tar.gz archive. I suppose this is because the plugin seed does not have a package.json file in the root of the repository and it cannot be packed directly as an npm package.

miroslavaivanova:
run ci

PR #3872, opened 2018-09-06
Replace forum references in the CLI with Stack Overflow

rdlauer:
As we are looking to deprecate usage of the community forum by the end of September, this PR replaces most references of the forum with appropriate Stack Overflow links.


rosen-vladimirov:
Hey @rdlauer ,
Can you please retarget this PR to release branch, so we can include it in the incoming 4.2.4 version of CLI?

rdlauer:
@rosen-vladimirov my git skills are lacking...but I think I just did what you asked...?

rosen-vladimirov:
Hey @rdlauer ,
I think you have changed the base branch of the PR, which is almost what I've requested. The problem with this approach is that your branch already contains a lot of commits from master branch. So when you've retargeted it to release branch, you branch introduces a lot of new commits instead of only one (the one we would like to have in the release branch).
This is normal and in such cases you have to apply some changes manually. For example, my own workflow in such cases (I wouldn't say it is perfect, but I'm used with it) is the following:
```
$ git checkout <my branch>
$ git log -1 # this gives me the sha of my last commit, the one I want to move in the release branch
$ git fetch origin # update the local copies of the remote branches
$ git reset --hard origin/release # resets your local branch to the state of the branch you want to use as a base
$ git cherry-pick <sha taken from git log> # gets the commit you want and places it at the top of the current tree
$ git push -f origin <my branch>
```

I've already applied a similar approach for your changes and I've created another PR: https://github.com/NativeScript/nativescript-cli/pull/3881, so I'm closing this one. Thanks for your help!

rdlauer:
Thanks @rosen-vladimirov!

PR #3881, opened 2018-09-11
replacing forum references with stack overflow

rosen-vladimirov:
As we are looking to deprecate usage of the community forum by the end of September, this PR replaces most references of the forum with appropriate Stack Overflow links.

Replaces https://github.com/NativeScript/nativescript-cli/pull/3872

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -8,7 +8,7 @@ questions regarding the behavior/reproduction for more than 20 days "dead". All
 
 ## Please, provide the details below:
 
-### Did you verify this is a real problem by searching the [NativeScript Forum](http://forum.nativescript.org) and the [other open issues in this repo](https://github.com/NativeScript/nativescript/issues)?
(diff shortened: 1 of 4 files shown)


.github/ISSUE_TEMPLATE.md
```diff
@@ -8,7 +8,7 @@ questions regarding the behavior/reproduction for more than 20 days "dead". All
 
 ## Please, provide the details below:
 
-### Did you verify this is a real problem by searching the [NativeScript Forum](http://forum.nativescript.org) and the [other open issues in this repo](https://github.com/NativeScript/nativescript/issues)?
+### Did you verify this is a real problem by searching [Stack Overflow](https://stackoverflow.com/questions/tagged/nativescript), the [NativeScript Forum](https://discourse.nativescript.org/), and the [other open issues in this repo](https://github.com/NativeScript/nativescript/issues)?
```
rosen-vladimirov: @rdlauer why do we want to keep the discourse link here? Should I change it to something else ?

rdlauer: Only because it's another source of resolution for folks...even if it'll be read-only. I'd be ok removing it though.

PR #3882, opened 2018-09-11
fix: read npm registry from npm config

### LINKED ISSUE #3866: Read registry from npm config instead of hard-wiring to npmjs.org in node-package-manager
We use artifactory as a our internal npm repository. It mirrors npmjs.org and also allow us to push internal npm packages to it.

However, when we do a clean install/build using `tns build android` the nativescript-cli itself starts a http client to pull the tns-android package from npmjs.org:

https://github.com/NativeScript/nativescript-cli/blob/f0e150ca8b2988bb23caafa2e239ee91cc8cde14/lib/node-package-manager.ts#L123

this does not work for us. 

Ideally, inside the `getRegistryPackageData` the registry to use should be read from the npm config (.npmrc)

lambourn:
## PR Checklist

nativescript-cli assumes that npmjs.org is used. If another (local) repo like artifactory is used, `tns add platform` (and probably other steps) fails as it hard-wired tries to pull packages from npmjs.org

## What is the new behavior?

read npm registry from npm config instead of hard-wiring it to npmjs.org, fixes #3866

### DIFF
--- a/lib/node-package-manager.ts
+++ b/lib/node-package-manager.ts
@@ -120,7 +120,8 @@ export class NodePackageManager implements INodePackageManager {
 	}
 
 	public async getRegistryPackageData(packageName: string): Promise<any> {
-		const url = `https://registry.npmjs.org/${packageName}`;
+		const registry = await this.$childProcess.exec(`npm config get registry`);
+		const url = registry.trim() + packageName;
 		this.$logger.trace(`Trying to get data from npm registry for package ${packageName}, url is: ${url}`);
 		const responseData = (await this.$httpClient.httpRequest(url)).body;
 		this.$logger.trace(`Successfully received data from npm registry for package ${packageName}. Response data is: ${responseData}`);


rosen-vladimirov:
run ci

lib/node-package-manager.ts
```diff
@@ -120,7 +120,8 @@ export class NodePackageManager implements INodePackageManager {
 	}
 
 	public async getRegistryPackageData(packageName: string): Promise<any> {
-		const url = `https://registry.npmjs.org/${packageName}`;
+		const registry = await this.$childProcess.exec(`npm config get registry`);
```
rosen-vladimirov: What do you think about falling back to registry.npmjs.org in case `npm config get registry` fails for some reason?
For example:
```TypeScript
let registry = "https://registry.npmjs.org/";
try {
    registry = await this.$childProcess.exec(`npm config get registry`);
} catch (err) {
    this.$logger.trace("Unable to get information for current npm registry", err);
}

const url = registry.trim() + packageName;
```

lambourn: I thought about this as well - defensive programming FTW ;-) But even if I wipe my `.npmrc` file, the value npm defaults to is always `https://registry.npmjs.org/`

rosen-vladimirov: Ok, thanks for the explanation. We'll run our tests now and if the build is green, I'll merge this PR.

rosen-vladimirov:
@lambourn thanks a lot for your Contribution! We've merged the suggested fix and it will be included in our 5.0.0 release. The fix will be available soon in CLI's next version - you can install it by executing `npm i -g nativescript@next` (it may take 15-20 minutes to publish the next version with the fix).

PR #3900, opened 2018-09-17
fix: Project's Podfile is regenerated incorrectly

rosen-vladimirov:
In case multiple plugins have Podfile with post_install hook, CLI modifies the hook of each of them by renaming the hook's function and calling it in the single post_install that is allowed for Podfile (placed at the bottom of the project's Podfile). However, when some change in node_modules is detected, CLI processes all Podfiles again and tries to apply them to project's Podfile. The "apply" is check if the plugin's Podfile is already added to project's one and in case not - adding it again. Whenever all Podfiles are processed, the "afterPrepareAllPlugins" method checks the final project's Podfile and replaces the post_install hooks from all plugins with a new functions (by adding index to the name) and calls them inside a newly added post_install function.
This is not working as when CLI checks if a plugin's Podfile is already added to the project's one, it uses the original content of the plugin's Podfile, where the post_install hook is called `post_install`, while in the project's one it is renamed. So CLI always thinks that plugin's Podfile, which has post_install hook is not added to the project. CLI tries to add it again and messes the whole structure.

In order to fix the issue, apply the following logic:
1. When a plugin has a Podfile, during prepare of the plugin read the content of this file. After that apply modifications over the file by replacing the post_install with a well known name - `post_install<plugin_name>_<index>`.
2. Check if the replaced content is part of the project's Podfile. In case yes - do nothing.
3. In case the replaced content is not part of the project's Podfile, it means that either we have never added this Podfile or its content has been changed. In order to handle both cases first remove the plugin's Podfile from the project's one - this includes removing the whole content between `# Begin Podfile <podfile>` and `# End Podfile` for this Podfile to be removed from the project's one and all post_install hooks with name that would have been used for this plugin to be removed. After that add the replaced content in the project's Podfile and call all post_install replaced functions in the project's Podfile's post_install hook.
4. On `afterPrepareAllPlugins` do nothing with the Podfile, it is already processed.

This solution ensures the Podfile is correct after processing each plugin. The previous one relyed on the `afterPrepareAllPlugins` to fix the project's Podfile. Also, by using well known pattern for generating the post_install hooks, we ensure we can remove them from the final Podfile whenever is required.

## PR Checklist

Applying a change in a plugin's Podfile, when it has postinstall hook, causes the iOS build to fail.

## What is the new behavior?
Applying a change in a plugin's Podfile, when it has postinstall hook, removes the old plugin's Podfile from the project and adds the new content.

Fixes https://github.com/NativeScript/nativescript-cli/issues/3549

### DIFF
--- a/lib/services/cocoapods-service.ts
+++ b/lib/services/cocoapods-service.ts
@@ -1,11 +1,11 @@
 import { EOL } from "os";
-
-interface IRubyFunction {
-	functionName: string;
-	functionParameters?: string;
-}
+import * as path from "path";
+import { PluginNativeDirNames, PODFILE_NAME } from "../constants";
 
 export class CocoaPodsService implements ICocoaPodsService {
+	private static PODFILE_POST_INSTALL_SECTION_NAME = "post_install";
+	private static INSTALLER_BLOCK_PARAMETER_NAME = "installer";
+
 	constructor(private $fs: IFileSystem) { }
 
 	public getPodfileHeader(targetName: string): string {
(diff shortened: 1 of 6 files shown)


lib/services/cocoapods-service.ts
```diff
@@ -16,20 +15,132 @@ export class CocoaPodsService implements ICocoaPodsService {
...
+	}
+
+	private removePostInstallHook(pluginData: IPluginData, projectPodFileContent: string): string {
+		const regExp = new RegExp(`^.*?${this.getHookBasicFuncNameForPlugin(CocoaPodsService.PODFILE_POST_INSTALL_SECTION_NAME, pluginData.name)}.*?$\\r?\\n`, "gm");
+		projectPodFileContent = projectPodFileContent.replace(regExp, "");
+		return projectPodFileContent;
+	}
+
+	private getHookBasicFuncNameForPlugin(hookName: string, pluginName: string): string {
+		return `${hookName}${pluginName.replace(/[^A-Za-z0-9]/g, "")}`;
```
Fatme: As we discussed personally we need to change this to:
```
`${hookName}${pluginName.replace(/[^A-Za-z0-9]/g, "_")}`;
```

rosen-vladimirov:
run ci

PR #3910, opened 2018-09-19
fix: `tns platform add ios` should not be executed on non-macOS

rosen-vladimirov:
`tns platform add ios` should not be executed on non-macOS system. Currently in the master branch it exits silently, add correct error, so the user will know that it is not okay.


rosen-vladimirov:
run ci

Issue #3912, opened 2018-09-19
Build fails with Xcode 10 with error could not find included file '../plugins-debug.xcconfig' in search paths


rosen-vladimirov:
### Did you verify this is a real problem by searching the [NativeScript Forum](http://forum.nativescript.org) and the [other open issues in this repo](https://github.com/NativeScript/nativescript/issues)?
Yes

### Tell us about the problem
Build fails for iOS in case the following circumstances are fulfilled:
* Xcode 10 is used
* application does not have build.xcconfig file in `App_Resources/iOS/build.xcconfig`
* none of the plugins have build.xcconfig file in `<plugin dir>/platforms/ios/build.xcconfig`

The error is:
```
error: <path to project>/platforms/ios/<app name>/build-debug.xcconfig:3: could not find included file '../plugins-debug.xcconfig' in search paths (in target 'tests')
```

### Which platform(s) does your issue occur on?
iOS

### Please provide the following version numbers that your issue occurs with:
- CLI: 4.2.4
- Cross-platform modules: Not applicable
- Runtime(s): Not applicable
- Plugin(s): Ensure all plugins do not have build.xcconfig in `<plugin dir>/platforms/ios`

### Please tell us how to recreate the issue in as much detail as possible.
```
$ tns create app1
$ cd app1
$ rm app/App_Resources/iOS/build.xcconfig
$ tns build ios
```

### Is there code involved? If so, please share the minimal amount of code needed to recreate the problem.
No

### Workaround
Create empty `build.xcconfig` in `<path to project's App_Resources>/iOS/build.xcconfig`

anaszgh:
Is there a way to fix this issue on version 4.2.4?

anaszgh:
@rosen-vladimirov build.xcconfig already is there in the `<App_Resources>/IOS` but still the issue occur

MarkPieszak:
After doing workaround above and deleting build-debug file, it builds a little bit further but throws a new error

```
The sandbox is not in sync with the Podfile.lock. Run 'pod install' or update your CocoaPods installation.
```

Fatme:
@MarkPieszak,

You need to update your CocoaPods. Just execute `pod install`.

FranciZ:
Any suggestions on how to solve this for 4.2.4? Can't build with Xcode10 at this point.

anaszgh:
You can downgrade your XCode Version 9, thats what I've done and its working fine.

[...]

PR #3913, opened 2018-09-19
fix: build with Xcode 10 may fail for missing plugins...xcconfig file

rosen-vladimirov:
In the iOS Project template we have include for plugins-debug.xcconfig (or plugins-release.xcconfig when building in release mode). CLI should produce such file, but it does not do it when the project does not have build.xcconfig in its App_Resources and neither of the plugins have such file. With previous Xcode versions, this case was leading to warnings in the output, however they are errors now and the build fails.
Fix this by adding an empty file in such case.

## PR Checklist

Build fails for iOS in case the following circumstances are fulfilled:
* Xcode 10 is used
* application does not have build.xcconfig file in `App_Resources/iOS/build.xcconfig`
* none of the plugins have build.xcconfig file in `<plugin dir>/platforms/ios/build.xcconfig`

## What is the new behavior?
The application can be build in case the mentioned circumstances are fulfilled.

Fixes https://github.com/NativeScript/nativescript-cli/issues/3912

### DIFF
--- a/lib/services/ios-project-service.ts
+++ b/lib/services/ios-project-service.ts
@@ -1301,6 +1301,13 @@ We will now place an empty obsolete compatability white screen LauncScreen.xib f
 			await this.mergeXcconfigFiles(appResourcesXcconfigPath, pluginsXcconfigFilePath);
 		}
 
+		if (!this.$fs.exists(pluginsXcconfigFilePath)) {
+			// We need the pluginsXcconfig file to exist in platforms dir as it is required in the native template:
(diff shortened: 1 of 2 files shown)


lib/services/ios-project-service.ts
```diff
@@ -778,7 +778,7 @@ We will now place an empty obsolete compatability white screen LauncScreen.xib f
 		this.$fs.deleteDirectory(this.getAppResourcesDestinationDirectoryPath(projectData));
 	}
 
-	public async processConfigurationFilesFromAppResources(release: boolean, projectData: IProjectData): Promise<void> {
+public async processConfigurationFilesFromAppResources(release: boolean, projectData: IProjectData): Promise<void> {
```
Fatme: missing space?

SvetoslavTsenov:
test

dtopuzov:
test

Issue #3773, opened 2018-07-25
bug(android) plugin built with wrong sdk/tools versions


farfromrefug:
I have a [plugin](https://github.com/Akylas/nativescript-material-components) which needs the very last support library.
Consequently I set up ```include.gradle``` with the correct version. You can see it [here](https://github.com/Akylas/nativescript-material-components/blob/master/plugin/platforms/android/include.gradle)

However when I build through my app it fails because the plugin is built with the wrong versions.
If I print the corresponding gradle call:
```
/gradlew -p /myappfolder/platforms/tempPlugin/nativescript_material_components assembleRelease -PcompileSdk=android-27 -PbuildToolsVersion=27.0.3 -PsupportVersion=26.0.0-alpha1
```

The build.gradle within the temp folder does not take into account the targetSDk or compileSDKVersion that I have set-up in ```include.gradle```. Only the dependencies are correct.

```

buildscript {
    repositories {
        google()
        jcenter()
    }
    dependencies {
        classpath 'com.android.tools.build:gradle:3.1.2'

        // NOTE: Do not place your application dependencies here; they belong
        // in the individual module build.gradle files
    }
}

allprojects {
    repositories {
        jcenter()
        mavenLocal()
        mavenCentral()
        maven {
            url 'https://maven.google.com/'
            name 'Google'
        }
    }
}

apply plugin: 'com.android.library'

def computeCompileSdkVersion = { -> project.hasProperty("compileSdk") ? compileSdk : 24 }
def computeBuildToolsVersion = { ->
    project.hasProperty("buildToolsVersion") ? buildToolsVersion : "27.0.3"
}

android {
    compileSdkVersion computeCompileSdkVersion()
    buildToolsVersion computeBuildToolsVersion()

    defaultConfig {
        targetSdkVersion 26
        versionCode 1
        versionName "1.0"
    }
}

dependencies {
    def supportVer = "27.0.1"
    if (project.hasProperty("supportVersion")) {
        supportVer = supportVersion
    }
    compileOnly "com.android.support:support-v4:$supportVer"
    compileOnly "com.android.support:appcompat-v7:$supportVer"
}

dependencies {
    compile "com.android.support:appcompat-v7:28+"
    compile 'com.android.support:design:28+'
}
```

That ```build.gradle``` should have targetSDKVersion, compileSDKVersion, buildToolsVersion setup in that order:
* if the app app.gradle has a version use this one
* if not and if the plugin has a version, use the plugin version
[...]

PR #3923, opened 2018-09-21
feat: remove requirement for Android Support Repository local installation

rosen-vladimirov:
In the past there was a requirement to have a local installation of the Android Support Repository in order to use libraries from there. However, since some time, the new versions are not downloaded locally and they must be used from the Google repository.
Currently this is not possible in NativeScript as {N} CLI always passes the Support Library version to gradle based on the local installation.
Remove the requirement for local installation and never pass the version to gradle. This way the apps will use the value specified in the `build.gradle` and they can overwrite it in `app.gradle`.
Update nativescript-doctor to latest version, which allows building project without local Androud Support Repository

## PR Checklist

You cannot use new versions of Android Support Library. Defining it in `app.gradle` still does not allow you to use newer version, as CLI overwrites the value. Also you cannot build NativeScript projects in case you do not have local version of the Android Support Repository.

## What is the new behavior?
1. You can specify your own version of the Android Support Library in your `app.gradle`. In case you do not specify it, a default value will be used (set in `build.gradle`).
2. You can build projects without installing local version of Android Support Repository.

Fixes issue https://github.com/NativeScript/nativescript-cli/issues/3773 

BREAKING CHANGES:
Existing applications that are using older runtime (not 5.0.0 one), but are built with CLI 5.0.0, may experience some changes - until now CLI was always passing parameter to gradle `-PsupportVersion=26.0.0-alpha1`. As CLI no longer passes this version, the default one from `build.gradle` will be used (for example 27.0.1). In case you want to use the old version in your application, add the following in your `app.gradle`:
```Groovy
project.ext.supportVersion = "26.0.0-alpha1"
```

### DIFF
--- a/lib/android-tools-info.ts
+++ b/lib/android-tools-info.ts
@@ -31,7 +31,6 @@ export class AndroidToolsInfo implements IAndroidToolsInfo {
 			infoData.compileSdkVersion = this.getCompileSdkVersion();
 			infoData.buildToolsVersion = this.getBuildToolsVersion();
 			infoData.targetSdkVersion = this.getTargetSdk();
-			infoData.supportRepositoryVersion = this.getAndroidSupportRepositoryVersion();
 			infoData.generateTypings = this.shouldGenerateTypings();
 
 			this.toolsInfo = infoData;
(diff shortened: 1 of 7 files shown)


manoldonev:
We would need to set 26.0.0 as a threshold for allowed older support versions though. In order to clean up https://github.com/NativeScript/NativeScript/pull/6129#issuecomment-409852341 (restore animator API usage) and limit the breaking changes in core modules we need android.support.v4.app.Fragment that supports both animations and animators and it seems this was introduced with 26.0.0: https://developer.android.com/sdk/support_api_diff/26.0.0/changes (older support library Fragments worked only with animations).

rosen-vladimirov:
run ci

rosen-vladimirov:
run ci

PR #3929, opened 2018-09-25
fix: skip platform add on every change when `tns preview --bundle` command is executed

Fatme:
## PR Checklist

"Platform android successfully added. V4.2.0" message is shown on every change when `tns preview --bundle` command is executed.

## What is the new behavior?
"Platform android successfully added. V4.2.0" message is not shown on every change.

### DIFF
--- a/lib/services/platform-service.ts
+++ b/lib/services/platform-service.ts
@@ -41,7 +41,8 @@ export class PlatformService extends EventEmitter implements IPlatformService {
 		private $projectChangesService: IProjectChangesService,
 		private $analyticsService: IAnalyticsService,
 		private $terminalSpinnerService: ITerminalSpinnerService,
-		private $pacoteService: IPacoteService
+		private $pacoteService: IPacoteService,
+		private $usbLiveSyncService: any
 	) {
 		super();
 	}
@@ -726,7 +727,7 @@ export class PlatformService extends EventEmitter implements IPlatformService {
 				platform = data[0],
 				version = data[1];
 
-			if (this.isPlatformInstalled(platform, projectData)) {
+			if (this.hasPlatformDirectory(platform, projectData)) {
 				await this.updatePlatform(platform, version, platformTemplate, projectData, config);
 			} else {
 				await this.addPlatform(platformParam, platformTemplate, projectData, config);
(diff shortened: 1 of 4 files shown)


lib/services/platform-service.ts
```diff
@@ -780,7 +780,9 @@ export class PlatformService extends EventEmitter implements IPlatformService {
 		const platformData = this.$platformsData.getPlatformData(platform, projectData);
 		const prepareInfo = this.$projectChangesService.getPrepareInfo(platform, projectData);
 		// In case when no platform is added and webpack plugin is started it produces files in platforms folder. In this case {N} CLI needs to add platform and keeps the already produced files from webpack
-		if (appFilesUpdaterOptions.bundle && this.isPlatformInstalled(platform, projectData) && !this.$fs.exists(platformData.configurationFilePath) && (!prepareInfo || !prepareInfo.nativePlatformStatus || prepareInfo.nativePlatformStatus !== constants.NativePlatformStatus.alreadyPrepared)) {
+		if (appFilesUpdaterOptions.bundle && this.isPlatformInstalled(platform, projectData)
+			&& !this.$fs.exists(platformData.configurationFilePath)
+			&& (!prepareInfo || !prepareInfo.nativePlatformStatus || prepareInfo.nativePlatformStatus !== constants.NativePlatformStatus.requiresPlatformAdd)) {
```
rosen-vladimirov: shouldn't this be `prepareInfo.nativePlatformStatus === constants.NativePlatformStatus.requiresPlatformAdd`

Fatme: No. In case when no platform is added and the user executes cloud build, {N} CLI will produce `.nsprepareinfo` file with the following content:
```
{
    nativePlatformStatus: "1"
}
```

If the cloud build is ran for second time, {N} CLI should'n add the platform again so the check should be
```
prepareInfo.nativePlatformStatus !== constants.NativePlatformStatus.requiresPlatformAdd
```

rosen-vladimirov: Okay, I agree with the described behavior, but what will happen in case the `nativePlatformStatus` has value "2" - does it mean we have to call `addPlatform` again?

PR #3936, opened 2018-09-25
fix: passing --hmr must always set bundle option to "webpack" 

rosen-vladimirov:
In case `--hmr` is passed, CLI should set the `bundle` option to `webpack`. This works for normal command, but in case it is command that has its own specific `--options` (commands from extensions), CLI calls `validateOptions` again and resets the value set to `--bundle`.
In order to fix the behavior, set the `bundle` option to "webpack" directly in the setArgv method which is called every time.

### refactor: remove OptionsBase and move logic to Options
As we do not have submodule anymore, remove the `OptionsBase` class and move the implementation to `Options` class.
Remove the `lib/common/options.ts` file and the `ICommonOptions` interface.
Replace the usage of `ICommonOptions` with `IOptions`.
Fix the unit tests of `lib/common/test/unit-tests/common-options` to use `Options` instead of `OptionsBase` and apply some changes in the passed options in the tests, so they can be valid now. Move the tests to `test/options.ts`

## PR Checklist

`tns cloud run android --hmr --accountId 1` does not use webpack build.

## What is the new behavior?
`tns cloud run android --hmr --accountId 1` uses webpack build and HMR is applied.


lib/options.ts
```diff
@@ -1,11 +1,46 @@
...
+		_: { type: OptionType.String }
+	};
+
+	public options: IDictionary<IDashedOption>;
+
+	constructor(private $errors: IErrors,
+		private $staticConfig: Config.IStaticConfig,
+		private $settingsService: ISettingsService) {
+
+		this.options = _.extend({}, this.commonOptions, this.options, this.globalOptions);
```
Fatme:
```
this.options = _.extend({}, this.commonOptions, this.globalOptions);
```
`this.options` is always an empty dictionary, so no need to use it.

PR #3941, opened 2018-09-26
feat: make log parser logic to work with multiple rules

KristianDD:
Until now the log parsing was used only for getting debug port for iOS. Now the service is generic and can execute multiple rules on each log. This will help us in other scenarios - detecting hot module reload fail for example.


lib/common/mobile/ios/simulator/ios-simulator-application-manager.ts
```diff
@@ -104,6 +104,6 @@ export class IOSSimulatorApplicationManager extends ApplicationManagerBase {
 
 	@cache()
 	private startDeviceLog(): Promise<void> {
-		return this.device.openDeviceLogStream({predicate: 'senderImagePath contains "NativeScript"'});
+		return this.device.openDeviceLogStream({predicate: 'senderImagePath contains "NativeScript" || eventMessage contains[c] "NativeScript"'});
```
Fatme: Maybe it is better to extract the predicate as a constant.

lib/definitions/log-parser-service.d.ts
```diff
@@ -1,7 +1,14 @@
...
+interface ILogParserService extends NodeJS.EventEmitter {
 	/**
 	 * Starts looking for debugger port. Attaches on device logs and processes them. In case when port is found, DEBUGGER_PORT_FOUND event is emitted.
 	 * @param {Mobile.IDevice} device - Describes the device which logs will be processed.
 	 */
-	startParsingLog(device: Mobile.IDevice, data: IProjectName): Promise<void>;
+	addParseRule(rule: ILogParseRules): void;
 }
+
+interface ILogParseRules {
```
Fatme: `ILogParseRule` ?

lib/services/ios-debugger-port-service.ts
```diff
@@ -58,13 +58,26 @@ export class IOSDebuggerPortService implements IIOSDebuggerPortService {
...
+	private handlePortFound(matches: RegExpMatchArray, deviceId: string): void {
+		const data = {
+			port: parseInt(matches[1]),
+			appId: matches[2],
+			deviceId,
+			projectDir: ""
+		};
+
+		this.$logger.trace(DEBUGGER_PORT_FOUND_EVENT_NAME, data);
+		this.setData(data, { port: data.port });
```
Fatme: missing `this.clearTimeout(data);` ?

lib/services/ios-debugger-port-service.ts
```diff
@@ -58,13 +58,26 @@ export class IOSDebuggerPortService implements IIOSDebuggerPortService {
...
+			platform: this.$devicePlatformsConstants.iOS.toLowerCase()
 		});
 	}
 
+	private handlePortFound(matches: RegExpMatchArray, deviceId: string): void {
+		const data = {
+			port: parseInt(matches[1]),
+			appId: matches[2],
+			deviceId,
+			projectDir: ""
```
Fatme: Why this is empty? Not sure if this will be ok for sidekick. Maybe we need to use
```
projectDir: this.$projectDataService.getProjectData().projectDir
```

or
```
projectDir: this.$projectData.projectDir
```

KristianDD: This project data might not be valid in Sidekick. Moreover until now this property was totally missing. Maybe it is better to set it to null/undefined.

KristianDD:
run ci

PR #3955, opened 2018-09-30
chore: enable debugging of unit test runner process

rosen-vladimirov:
When CLI executes unit tests (for the NativeScript application), it forks a new Node.js process and communicates with it via IPC.
However, the Node.js fork method has an issue when you want to debug your main process - it passes the `--inspect[-brk]` option to the forked process as well. The main process had already used the port, so debugging fails.
In order to fix this, replace the `fork` with `spawn`.


rosen-vladimirov:
run ci

Issue #4007, opened 2018-10-10
Application built in release has TypeScript files


rosen-vladimirov:
**Environment**
Provide version numbers for the following components (information can be retrieved by running `tns info` in your project folder or by inspecting the `package.json` of the project):
 - CLI: 4.2.4
 - Cross-platform modules: not applicable
 - Android Runtime: not applicable
 - iOS Runtime: not applicable
 - Plugin(s): nativescript-facebook

**Describe the bug**
When building in release, application should not contain any TypeScript files, as they cannot be used (in debug you can use them for debugging). However, `.d.ts` files from plugins are included in the built app.

**To Reproduce**
```
tns create myApp
cd myApp
npm i --save nativescript-facebook
tns build android --release --keyStorePath <path> --keyStorePassword <pass> --keyStoreAlias <alias> --keyStoreAliasPassword <alias pass>
tns build ios --release --for-device
```
Now check the produced `.apk` and `.ipa` - they both contain `.d.ts` files inside `<path to app>/tns_modules/nativescript-facebook`.

**Expected behavior**
TypeScript files should not be included in the application built in release mode.

PR #4008, opened 2018-10-10
fix: remove all .ts files when building in release

rosen-vladimirov:
When building application in release mode, the `.ts` files of all plugins should be removed from the final application. Currently this is happening only for `tns-core-modules`. Do this for all the plugins.
Remove logic that deletes the `node_modules` directory of the `tns-core-modules` when it is copied to `platforms`. This is not required anymore and it breaks use-case when linking tns-core-modules.

## PR Checklist

Applications built in release have `.ts` files inside them.
Also you are unable to use a linked copy of `tns-core-modules`.

## What is the new behavior?
Applications built in release does NOT have `.ts` files inside them.
You can use symlink for tns-core-modules.

Fixes issue https://github.com/NativeScript/nativescript-cli/issues/4007

### DIFF
--- a/lib/tools/node-modules/node-modules-dest-copy.ts
+++ b/lib/tools/node-modules/node-modules-dest-copy.ts
@@ -10,33 +10,21 @@ export interface ILocalDependencyData extends IDependencyData {
 export class TnsModulesCopy {
 	constructor(
 		private outputRoot: string,
-		private $options: IOptions,
 		private $fs: IFileSystem,
 		private $pluginsService: IPluginsService
 	) {
 	}
 
-	public copyModules(dependencies: IDependencyData[], platform: string): void {
-		for (const entry in dependencies) {
-			const dependency = dependencies[entry];
+	public copyModules(opts: { dependencies: IDependencyData[], release: boolean }): void {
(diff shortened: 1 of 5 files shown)


lib/tools/node-modules/node-modules-dest-copy.ts
```diff
@@ -10,33 +10,21 @@ export interface ILocalDependencyData extends IDependencyData {
...
 		private $fs: IFileSystem,
 		private $pluginsService: IPluginsService
 	) {
 	}
 
-	public copyModules(dependencies: IDependencyData[], platform: string): void {
-		for (const entry in dependencies) {
-			const dependency = dependencies[entry];
+	public copyModules(opts: { dependencies: IDependencyData[], release: boolean }): void {
+		const filePatternsToDelete = opts.release ? "**/*.ts" : "**/*.d.ts";
```
Fatme: Should we exclude `.scss` files here as well?

rosen-vladimirov: I was thinking the same, but it is not a common practice to publish packages with `.scss` ,while all plugins have `.d.ts`. We need a more generic way to handle this and let the user decides what to exclude.

PR #4044, opened 2018-10-19
fix(preview): skip check if local plugin version is invalid

### LINKED ISSUE #4043: tns preview fails when local plugin is referenced with tag in package.json
**Environment**
 - CLI: 5.0.0-2018-10-11-12463
 - Cross-platform modules: rc

**Describe the bug**
If the version of some plugin in the package.json is specified with tag (f.e. `rc`) instead of version (f.e. `5.0.0`), the `tns preview` command fails with:
```
TypeError: Cannot read property 'major' of null
    at PreviewAppPluginsService.getWarningForPluginCore (/home/sis0k0/nativescript-cli/lib/services/livesync/playground/preview-app-plugins-service.ts:82:30)
```

The `PreviewAppPluginsService` uses `semver` to check some stuff with the versions of the plugins in the package.json. However, `semver.coerce('rc')` returns `null` and breaks the following logic.
https://github.com/NativeScript/nativescript-cli/blob/b9bfc685c8ab3b1f03d78e820661aa0f472075b3/lib/services/livesync/playground/preview-app-plugins-service.ts#L74-L91

**To Reproduce**
1. Create a hello-world app.
2. Add this dependency to your package.json:
````
...
"dependencies": {
	"tns-core-modules": "rc",
	...
}
````
3. Run `tns preview` and scan the barcode.

**Expected behavior**
To ignore the above check if the tag is `beta`, `rc` or `next`.
I can submit a PR if you agree that's the correct behaviour.

sis0k0:
If the local version of the plugin cannot be parsed by `semver`, the
check should be skipped and no warning shown.

fixes #4043

### DIFF
--- a/lib/services/livesync/playground/preview-app-plugins-service.ts
+++ b/lib/services/livesync/playground/preview-app-plugins-service.ts
@@ -74,20 +74,25 @@ export class PreviewAppPluginsService implements IPreviewAppPluginsService {
 	private getWarningForPluginCore(localPlugin: string, localPluginVersion: string, devicePluginVersion: string, deviceId: string): string {
 		this.$logger.trace(`Comparing plugin ${localPlugin} with localPluginVersion ${localPluginVersion} and devicePluginVersion ${devicePluginVersion}`);
 
-		if (devicePluginVersion) {
-			const localPluginVersionData = semver.coerce(localPluginVersion);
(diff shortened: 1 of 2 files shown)


endarova:
Can we still find a way to give the user a message that something is wrong with the plugin.

Fatme:
We decided to merge this PR with current fixes and to handle such plugins in another PR.

PR #4056, opened 2018-10-24
fix: don't throw `TypeError: Invalid Version: null` in sidekick when emulator is stopped immediately after start

Fatme:
`TypeError: Invalid Version: null` error is shown is Sidekick when emulator is stopped immediately after start. This happens because emulator is stopped too early and it is not fully initialized. In this case reported emulator is with `version: null` and the semver comparison fails. So we need to ensure the device's version is not `null` before comparison.

## PR Checklist

`TypeError: Invalid Version: null` error is thrown in Sidekick when emulator is stopped immediately after start.

## What is the new behavior?
No error is thrown in Sidekick when emulator is stopped immediately after start.

### DIFF
--- a/lib/common/mobile/android/logcat-helper.ts
+++ b/lib/common/mobile/android/logcat-helper.ts
@@ -58,20 +58,6 @@ export class LogcatHelper implements Mobile.ILogcatHelper {
 		}
 	}
 
-	private async getLogcatStream(deviceIdentifier: string, pid?: string) {
-		const device = await this.$devicesService.getDevice(deviceIdentifier);
-		const minAndroidWithLogcatPidSupport = "7.0.0";
-		const isLogcatPidSupported = !!device.deviceInfo.version && semver.gte(semver.coerce(device.deviceInfo.version), minAndroidWithLogcatPidSupport);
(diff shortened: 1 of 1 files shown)


lib/common/mobile/android/logcat-helper.ts
```diff
@@ -101,6 +87,20 @@ export class LogcatHelper implements Mobile.ILogcatHelper {
 			delete this.mapDevicesLoggingData[deviceIdentifier];
 		}
 	}
+
+	private async getLogcatStream(deviceIdentifier: string, pid?: string) {
+		const device = await this.$devicesService.getDevice(deviceIdentifier);
+		const minAndroidWithLogcatPidSupport = "7.0.0";
+		const isLogcatPidSupported = !!device.deviceInfo.version && semver.gte(semver.coerce(device.deviceInfo.version), minAndroidWithLogcatPidSupport);
```
rosen-vladimirov: what if we do not have device? It will still fail maybe? Or if the deviceInfo is null ?

Fatme: I was wondering about the same problem.

Actually when `adb devices` returns some data - for example `emulator-5554 device`, we know for sure in this moment the device's identifier and device's status. In other words we should have at least 
```
deviceInfo: {
    identifier: "emulator-5554",
    status: "connected"
}
``` 
So it is not possible to be here with `deviceInfo` or `device` with value `null`.

PR #4099, opened 2018-11-05
Apply before-plugins.gradle file in the plugin build.gradle

vtrifonov:
Related to #4100
If there's a **before-plugins.gradle** file in the app it should be called before the plugin build.gradle


DimitarTachev:
run ci

PR #4150, opened 2018-11-20
fix(preview-api): raise deviceLost event after timeout of 5 seconds

Fatme:
In case when some `.js` file is changed, preview app is restarted on device e.g preview app is stopped and started again. When the preview app is stopped, Pubnub reports the device as lost and when preview app is started, Pubnub reports the device as found. With this fix, we want to delay the emitting of deviceLost event. This way we'll give a chance to find the device before reporting it as lost.

## PR Checklist

### DIFF
--- a/lib/services/livesync/playground/devices/preview-devices-service.ts
+++ b/lib/services/livesync/playground/devices/preview-devices-service.ts
@@ -4,6 +4,7 @@ import { DeviceDiscoveryEventNames, DEVICE_LOG_EVENT_NAME } from "../../../../co
 
 export class PreviewDevicesService extends EventEmitter implements IPreviewDevicesService {
 	private connectedDevices: Device[] = [];
+	private deviceTimers: IDictionary<NodeJS.Timer> = {};
 
 	constructor(private $previewAppLogProvider: IPreviewAppLogProvider,
 		private $previewAppPluginsService: IPreviewAppPluginsService) {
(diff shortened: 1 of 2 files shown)


lib/services/livesync/playground/devices/preview-devices-service.ts
```diff
@@ -4,6 +4,7 @@ import { DeviceDiscoveryEventNames, DEVICE_LOG_EVENT_NAME } from "../../../../co
 
 export class PreviewDevicesService extends EventEmitter implements IPreviewDevicesService {
 	private connectedDevices: Device[] = [];
+	private deviceTimers: IDictionary<NodeJS.Timer> = {};
```
DimitarTachev: deviceTimers -> deviceLostTimers

lib/services/livesync/playground/devices/preview-devices-service.ts
```diff
@@ -45,6 +46,10 @@ export class PreviewDevicesService extends EventEmitter implements IPreviewDevic
 	}
 
 	private raiseDeviceFound(device: Device) {
+		if (this.deviceTimers[device.id]) {
+			clearTimeout(this.deviceTimers[device.id]);
+		}
+
 		this.emit(DeviceDiscoveryEventNames.DEVICE_FOUND, device);
```
DimitarTachev: Shouldn't this be in `else` (we could also add a test for such case)?

PR #4207, opened 2018-12-06
feat: automate publish of patch tag

rosen-vladimirov:
Automate publish of patch tag when PR is merged in `release-patch` tag.


Fatme:
run ci

PR #4216, opened 2018-12-11
fix: debugging in VSCode is not working on iOS Simulator

rosen-vladimirov:
When you've stopped on breakpoint in VSCode and apply a change in the code, CLI tries to restart application. However, at this point the command for killing the application may not kill it immediately and it needs 30 seconds. The problem is that the runtime loops indefinitely while on breakpoint. After some seconds (looks like 30), the OS kills the application. In order to resolve the issue, disconnect all sockets before stopping the application, so the runtime can successfully break the indefinite loop and after that application can be safely killed.

## PR Checklist

## What is the new behavior?

### DIFF
--- a/lib/common/mobile/ios/device/ios-application-manager.ts
+++ b/lib/common/mobile/ios/device/ios-application-manager.ts
@@ -8,7 +8,7 @@ export class IOSApplicationManager extends ApplicationManagerBase {
 
 	constructor(protected $logger: ILogger,
 		protected $hooksService: IHooksService,
-		private device: Mobile.IDevice,
+		private device: Mobile.IiOSDevice,
 		private $errors: IErrors,
 		private $iOSNotificationService: IiOSNotificationService,
 		private $iosDeviceOperations: IIOSDeviceOperations,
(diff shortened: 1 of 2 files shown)


lib/common/mobile/ios/device/ios-application-manager.ts
```diff
@@ -86,6 +86,8 @@ export class IOSApplicationManager extends ApplicationManagerBase {
 	}
 
 	public async stopApplication(appData: Mobile.IApplicationData): Promise<void> {
+		this.device.destroyAllSockets();
```
DimitarTachev: this.device.destroyAllSockets(); -> this.device.destroySocket(appId);

lib/common/mobile/ios/simulator/ios-simulator-application-manager.ts
```diff
@@ -47,7 +47,9 @@ export class IOSSimulatorApplicationManager extends ApplicationManagerBase {
 	}
 
 	public async stopApplication(appData: Mobile.IApplicationData): Promise<void> {
-		return this.iosSim.stopApplication(this.device.deviceInfo.identifier, appData.appId, appData.projectName);
+		this.device.destroyAllSockets();
```
DimitarTachev: this.device.destroyAllSockets(); -> this.device.destroySocket(appId);

PR #4242, opened 2018-12-20
Kddimitrov/track perf decorator

KristianDD:
## PR Checklist

We have no means of consistent profiling and tracking performance of our methods, while executing commands.

## What is the new behavior?

We will send performance analytics if usage-reporting is enabled. If `--performance` flag is added to the command all profiling information will be included in the command output. If `--performance` options is set to a file path, the profiling information will be saved inside the file.

Fixes/Implements/Closes https://github.com/NativeScript/nativescript-cli/issues/4243

### DIFF
--- a/lib/common/decorators.ts
+++ b/lib/common/decorators.ts
@@ -1,3 +1,5 @@
+const { performance } = require('perf_hooks');
+
 /**
  * Caches the result of the first execution of the method and returns it whenever it is called instead of executing it again.
  * Works with methods and getters.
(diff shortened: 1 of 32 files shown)


lib/common/decorators.ts
```diff
@@ -83,3 +85,48 @@ export function exported(moduleName: string): any {
...
+					.catch((err) => {
+						end = performanceService.now();
+						performanceService.processExecutionData(trackName, start, end, args);
+					});
+				}
+
+				return result;
+			}
+		};
+		descriptor.value = tempObject[originalMethod.name];
```
DimitarTachev: This looks a little bit risky, we need to cover it with unit tests for not hiding exceptions, results, not affecting the call stack and so on.

lib/services/performance-service.ts
```diff
@@ -0,0 +1,96 @@
...
+			methodInfo,
+			executionTime,
+			timestamp: getFormattedDate(),
+			methodArgs: JSON.parse(methodArgs)
+		};
+
+		try {
+			this.$fs.appendFile(filePath, `${JSON.stringify(info)}${EOL}`);
+		} catch (e) {
+			this.$logger.trace(PerformanceService.FAIL_LOG_MESSAGE_TEMPLATE, methodInfo);
```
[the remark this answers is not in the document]
KristianDD: Will change naming and add comment.

lib/options.ts
```diff
@@ -119,7 +119,8 @@ export class Options {
 			analyticsLogFile: { type: OptionType.String, hasSensitiveValue: true },
 			hooks: { type: OptionType.Boolean, default: true, hasSensitiveValue: false },
 			link: { type: OptionType.Boolean, default: false, hasSensitiveValue: false },
-			aab: { type: OptionType.Boolean, hasSensitiveValue: false }
+			aab: { type: OptionType.Boolean, hasSensitiveValue: false },
+			performance: { type: OptionType.Object, hasSensitiveValue: true }
```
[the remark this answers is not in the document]
KristianDD: Not sure if we should. It is more for an internal use and troubleshooting. Users shouldn't need to use it unless we specifically ask for additional information.

lib/common/decorators.ts
KristianDD: Will add tests for the decorator once the implementation is final.

lib/common/helpers.ts
```diff
@@ -567,6 +567,33 @@ export function stringify(value: any, replacer?: (key: string, value: any) => an
 	return JSON.stringify(value, replacer, space || 2);
 }
 
+export function getFormattedDate(): string {
```
DimitarTachev: `getFormattedDate` in which format? As far as I see it makes the date strings orderable and it could be something like getOrderableDateString. Also, we could make it accept the Date as an argument in order to make it unit testable.

lib/common/decorators.ts
```diff
@@ -83,3 +85,48 @@ export function exported(moduleName: string): any {
 		return descriptor;
 	};
 }
+
+export function performanceLog(injector?: IInjector): any {
+	injector = injector || $injector;
+	return function (target: any, propertyKey: string, descriptor: PropertyDescriptor): any {
```
DimitarTachev: shouldn't we skip this logic based on `$options.performance`?

PR #4299, opened 2019-01-18
chore: remove installPods property from platform's project interface

Fatme:
## PR Checklist

## What is the new behavior?

### DIFF
--- a/lib/services/ios-project-service.ts
+++ b/lib/services/ios-project-service.ts
@@ -986,6 +986,10 @@ We will now place an empty obsolete compatability white screen LauncScreen.xib f
 		}
 	}
 
+	public beforePrepareAllPlugins(): Promise<void> {
+		return Promise.resolve();
+	}
+
 	public async checkForChanges(changesInfo: IProjectChangesInfo, { provision, teamId }: IProjectChangesOptions, projectData: IProjectData): Promise<void> {
 		const hasProvision = provision !== undefined;
 		const hasTeamId = teamId !== undefined;
--- a/lib/services/ios-project-service.ts
+++ b/lib/services/ios-project-service.ts
@@ -800,6 +800,7 @@ We will now place an empty obsolete compatability white screen LauncScreen.xib f
 		}
 
 		this.$pluginVariablesService.interpolateAppIdentifier(this.getPlatformData(projectData).configurationFilePath, projectData.projectIdentifiers.ios);
+		await this.mergeProjectPodFile(projectData);
 	}
 
 	private getInfoPlistPath(projectData: IProjectData): string {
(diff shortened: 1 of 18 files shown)


lib/services/ios-project-service.ts
```diff
@@ -980,34 +977,10 @@ We will now place an empty obsolete compatability white screen LauncScreen.xib f
...
-			await this.$cocoapodsService.applyPodfileToProject(constants.NS_BASE_PODFILE, mainPodfilePath, projectData, this.getPlatformData(projectData).projectRoot);
-
-			await this.$cocoapodsService.executePodInstall(projectRoot, xcodeProjPath);
-		}
-	}
-
-	public beforePrepareAllPlugins(): Promise<void> {
-		return Promise.resolve();
+		const xcodeProjPath = this.getXcodeprojPath(projectData);
+		await this.$cocoapodsService.executePodInstall(projectRoot, xcodeProjPath);
```
rosen-vladimirov: this command should be executed only when we actually have cocoapods dependencies, i.e. when we have Podfile in platforms dir.

Fatme:
run ci

lib/services/ios-project-service.ts
```diff
@@ -980,28 +977,11 @@ We will now place an empty obsolete compatability white screen LauncScreen.xib f
...
-
-				const createSchemeRubyScript = `ruby -e "require 'xcodeproj'; xcproj = Xcodeproj::Project.open('${projectData.projectName}.xcodeproj'); xcproj.recreate_user_schemes; xcproj.save"`;
-				await this.$childProcess.exec(createSchemeRubyScript, { cwd: this.getPlatformData(projectData).projectRoot });
-			}
-
-			await this.$cocoapodsService.applyPodfileToProject(constants.NS_BASE_PODFILE, mainPodfilePath, projectData, this.getPlatformData(projectData).projectRoot);
-
+		const xcodeProjPath = this.getXcodeprojPath(projectData);
+		const projectPodfilePath = this.$cocoapodsService.getProjectPodfilePath(projectRoot);
+		if (this.$fs.exists(projectPodfilePath)) {
```
KristianDD: This maybe should be executed also if `this.$fs.exists(mainPodfilePath)` is true.
```typescript
if (this.$fs.exists(projectPodfilePath) || this.$fs.exists(mainPodfilePath)) {
```

Fatme: Actually we don't have `mainPodfilePath` variable in this method and we don't need it.
The idea here is `pod install` command to be executed on pod file in platforms folder. This pod file is the output from merge of all pod files from `node_modules` and the pod file from `App_Resources` if such. So if we have pod file in `App_Resources` (`mainPodfilePath`) it will be merged in pod file in platforms folder. If we don't have pod file in `App_Resources`, `pod install` will be executed only when pod file in platforms dir exists e.g the app contains some plugins with pod files.

PR #4317, opened 2019-01-28
feat: skip postinstall steps in case CLI is not installed globally

rosen-vladimirov:
In case the CLI package is not installed globally, probably it is installed as a dependency of a project or as a dependency of other dependency.
In this case the postinstall actions have no meaning for the user, as probably CLI will be used as a library.
Skip all of the postinstall actions in such case. Remove the execution code of the old `post-install` command and move it to `post-install-cli` command.
Add unit tests. Remove a property from staticConfig - it has no meaning to be there, just place it in the postinstall command directly.
Handle case when CLI is installed globally with npm or yarn - in both cases the `npm_config_argv` will be populated with the original command. For example:
* `npm i -g nativescript`  - the `npm_config_argv` will be: `["i", "-g", "nativescript"]`
* `yarn global add nativescript`  - the `npm_config_argv` will be: `["global", "add", "nativescript"]`

## PR Checklist

When executing `npm i nativescript`, you are prompted to enable autocompletion and all other post install tasks.

## What is the new behavior?
CLI will ask you to enable autocompletion (and other questions asked on post-install) only when `npm i -g nativescript` or `yarn global add nativescript` is used.

Implements issue https://github.com/NativeScript/nativescript-cli/issues/4323

### DIFF
--- a/lib/common/test/unit-tests/helpers.ts
+++ b/lib/common/test/unit-tests/helpers.ts
@@ -9,6 +9,14 @@ interface ITestData {
 }
 
 describe("helpers", () => {
+	let originalProcessEnvNpmConfig: any = null;
+	beforeEach(() => {
+		originalProcessEnvNpmConfig = process.env.npm_config_argv;
+	});
+
+	afterEach(() => {
+		process.env.npm_config_argv = originalProcessEnvNpmConfig;
+	});
 
 	const assertTestData = (testData: ITestData, method: Function) => {
 		const actualResult = method(testData.input);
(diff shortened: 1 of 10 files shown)


lib/common/test/unit-tests/helpers.ts
```diff
@@ -699,4 +707,120 @@ describe("helpers", () => {
...
+
+	describe("doesCurrentNpmCommandMatch", () => {
+		[
+			{
+				name: "returns true when searching for --global flag and it is passed on terminal",
+				input: ["install", "--global", "nativescript"],
+				expectedOutput: true
+			},
+			{
+				name: "returns true when searching for -g flag and --global is passed on terminal",
```
DimitarTachev: this case is passing `-g`, not `--global`, also, as far as I see we are always searching for both of the global flags in each of these cases.

rosen-vladimirov: done

Issue #4383, opened 2019-02-22
`tns doctor` command does not report any information on CircleCI MacOS virtual machines


cowboyd:
**Environment**
Provide version numbers for the following components (information can be retrieved by running `tns info` in your project folder or by inspecting the `package.json` of the project):

The `tns info` command does not actually report either on the CircleCI vm:

```
bash-3.2$ yarn tns info
yarn run v1.12.1
$ /Users/distiller/project/node_modules/.bin/tns info
✨  Done in 6.89s.
bash-3.2$
```

But this is what I get when running it on my local Mac:

```
❯❯❯❯ yarn tns info                                                 ~/C/F/n/vacation-tracker jc/circle-workflow
yarn run v1.13.0
$ /Users/cowboyd/Code/Frontside/nativescript-playground/vacation-tracker/node_modules/.bin/tns info
✔ Getting NativeScript components versions information...
✔ Component nativescript has 5.2.0 version and is up to date.
✔ Component tns-core-modules has 5.2.1 version and is up to date.
✔ Component tns-android has 5.2.0 version and is up to date.
✔ Component tns-ios has 5.2.0 version and is up to date.
✨  Done in 4.55s.
```

**Describe the bug**

`tns doctor` and other commands do not actually render critical output. For example, this is output of the `tns doctor` command on a Circle MacOS VM:

```
bash-3.2$ yarn tns doctor
yarn run v1.12.1
TIP: To avoid setting up the necessary environment variables, you can use the Homebrew package manager to install the Android SDK and its dependencies.

There seem to be issues with your configuration.
You are missing the nativescript-cloud extension and you will not be able to execute cloud builds. Your environment is not configured properly and you will not be able to execute local builds. To continue, choose one of the following options:
Select "Sync to Playground" to enjoy NativeScript without any local setup. All you need is a couple of companion apps installed on your devices.
Select "Configure for Cloud Builds" to install the nativescript-cloud extension and automatically configure your environment for cloud builds.
Select "Configure for Local Builds" to run the setup script and automatically configure your environment for local builds.
Select "Configure for Both Local and Cloud Builds" to automatically configure your environment for both options.
Select "Configure for Both Local and Cloud Builds" to automatically configure your environment for both options.
? To continue, choose one of the following options:  Skip Step and Configure Manually
[...]

PR #4404, opened 2019-02-27
fix: errors from `tns doctor` are not visible in CI environment

rosen-vladimirov:
When `tns doctor` detects errors in the configuration, it should print them to stdout and prompt the user to select how to resolve them.
In non-interactive terminal, it should just print the errors and fails as the prompter are not meaningful in non-interactive terminal - there's noone to answer them.

The logic works fine when you pipe the output of `tns doctor` to file for example (`tns doctor > out.txt`). CLI correctly detects the terminal as non-interactive, prints the errors and exits.
However, most of the CI environments are determined as interactive terminals by CLI. That's because CLI checks if the stdout and stdin of the current process are text terminal (TTY). CI environments set required flags, so the process seems like running in such text terminal.
When CLI thinks the process is running in text terminal, it uses some external package (ora) to print pretty lines. However, `ora` package also checks if the process is running in text terminal (which both CLI and `ora` think is true), but it also checks if the environment variable `CI` is set.
When it is set, `ora` package decides that it cannot print colored messages and just doesn't print anything.

To resolve this, improve the check if interactive terminal in CLI to respect the known environment variables that define the process as running in CI:
- for Travis the environment varible `CI` is set.
- for CircleCI the environment variable `CI` is set
- for Jenkins the environment variable `JENKINS_HOME` is set

Whenever one of those environment variables is set, CLI will decide it is running in non-interactive terminal and will not use `ora` (for `tns doctor`). It will also not show any prompters in this case.

> NOTE: Deleted duplicate test from other service

## PR Checklist

Errors from `tns doctor` are not shown in Travis/CircleCI. Instead the CI build just hangs as CLI had shown prompter.

## What is the new behavior?
Errors from `tns doctor` are shown in Travis/CircleCI and the CI Build fails directly instead of hanging.

Fixes issue https://github.com/NativeScript/nativescript-cli/issues/4383

### DIFF
--- a/lib/common/helpers.ts
+++ b/lib/common/helpers.ts
@@ -309,7 +309,27 @@ export function versionCompare(version1: string | IVersionData, version2: string
 }
 
 export function isInteractive(): boolean {
-	return process.stdout.isTTY && process.stdin.isTTY;
+	const isInteractive = isRunningInTTY() && !isCIEnvironment();
+	return isInteractive;
+}
+
+/**
+ * Checks if current process is running in Text Terminal (TTY)
+ */
+function isRunningInTTY(): boolean {
+	return process.stdout &&
+		process.stdout.isTTY &&
+		process.stdin &&
+		process.stdin.isTTY;
+}
+
+function isCIEnvironment(): boolean {
+	// The following CI environments set their own environment variables that we respect:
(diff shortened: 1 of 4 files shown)


yagey:
thank you!

PR #4412, opened 2019-03-01
fix: avoid leaked lock files for more than 10 seconds

DimitarTachev:
Replace the `lockfile` package in order to depend on `modified`, instead of `created` date when checking for `stale` locks.

Replacing the package, we are able to keep smaller `stale` values (e.g. 10 seconds) and avoid losing the lock during long-running operations. In this way, when the CLI process crashes (or is killed with SIGKILL), we won't have leaked lock files for more than 10 seconds.

## PR Checklist

If the process is killed with SIGKILL (kill -9), the lock files will leak for up to 90 seconds.

## What is the new behavior?
If the process is killed with SIGKILL (kill -9), the lock files will leak for 5-10 seconds.


lib/common/services/lock-service.ts
```diff
@@ -31,41 +31,49 @@ export class LockService implements ILockService {
 		private $processService: IProcessService) {
 		this.$processService.attachToProcessExitSignals(this, () => {
 			const locksToRemove = _.clone(this.currentlyLockedFiles);
-			_.each(locksToRemove, lock => {
-				this.unlock(lock);
-			});
+			for (let i = 0; i < locksToRemove.length; i++) {
```
rosen-vladimirov:
```TypeScript
for (const lockToRemove of locksToRemove) {
    lockfile.unlockSync(lockToRemove);
    this.cleanLock(lockToRemove);
}
```

Issue #4440, opened 2019-03-18
`NativeScript can only run in Xcode version 6.0 or greater` error is shown on every iOS command


rosen-vladimirov:
**Environment**
Provide version numbers for the following components (information can be retrieved by running `tns info` in your project folder or by inspecting the `package.json` of the project):
 - CLI: 5.2.3
 - Cross-platform modules: Not applicable
 - Android Runtime: Not applicable
 - iOS Runtime: Not applicable
 - Plugin(s): Not applicable

**Describe the bug**
Whenever I try to execute `tns run ios`, `tns prepare ios`, `tns build ios`, `tns debug ios`, etc. (i.e. every iOS related command), I receive an error: `NativeScript can only run in Xcode version 6.0 or greater`. However, my Xcode version is 10.1 (output of `xcodebuild -version` is: 
```
Xcode 10.1
Build version 10B61
```

**To Reproduce**
The issue happens when the `head` command is overwritten by some other executable with `head` name. This happens in some cases related to Python, Go or XAMPP installations. The easiest way to reproduce the same behavior is to follow the steps below:
1. `tns create myApp`
2. `cd myApp` 
3. `echo "" > head` (This will create file called `head` in your project directory.
4.  `chmod +x head` (This will make the currently created `head` file executable.
5. Execute any iOS related command in the following manner:
```
PATH=`pwd`:$PATH tns prepare ios
```
This will set the PATH variable to contain the current working directory in the first place of your PATH, this way the head file we've just created will be used instead of the original head. NOTE: This command will set the PATH only for current command, so no need to worry about messing your machine.

**Expected behavior**
CLI should allow me to execute iOS related commands.

Related to: https://github.com/NativeScript/nativescript-cli/issues/3189#issuecomment-471077037

PR #4442, opened 2019-03-18
fix: get Xcode version correctly

rosen-vladimirov:
In some cases NativeScript CLI fails on all iOS related commands (build, run, debug, etc.) with error: `NativeScript can only run in Xcode version 6 or greater`.
This happens when user has correct Xcode version, i.e. 9 or greater, but the `head` command on their machine does not support `-n` flag. This happens when the `head` command is overwritten by some tool, so the `head` executable is not the original one used by macOS.
CLI has some obsolete code to check the Xcode version AFTER it has verified it in the doctorService. Delete this code and use sysInfo's getXcodeVersion, which does not rely on `head` command.

## PR Checklist

CLI fails to execute iOS related commands when the `head` command is not the original one from macOS.

## What is the new behavior?
CLI is able to execute iOS related commands when the `head` command is not the original one.

Fixes issue https://github.com/NativeScript/nativescript-cli/issues/4440


rosen-vladimirov:
test cli-smoke

PR #4483, opened 2019-03-28
Device/emulator detection fixes

rosen-vladimirov:
### fix: do not start emulator detection interval
The emulator detection interval is not required for CLI and in fact it just uses CPU and memory. Also, it causes some issues when you do not hava JAVA installed.
So remove start of emulator detection interval and allow it to be started from the API, when CLI is used as a library.

### fix: stop executing avdmanager in case java is not available
In case JAVA_HOME is not set and there's no java executable in PATH, calls to `avdmanager` fail with error. However, on macOS this also leads to system prompt to install JAVA.
To handle this, do not execute `avdmanager` in case there's no JAVA_HOME or java in PATH.

## PR Checklist

## What is the new behavior?

### DIFF
--- a/lib/common/mobile/mobile-core/devices-service.ts
+++ b/lib/common/mobile/mobile-core/devices-service.ts
@@ -10,7 +10,6 @@ import { CONNECTED_STATUS } from "../../constants";
 import { isInteractive } from "../../helpers";
 import { DebugCommandErrors } from "../../../constants";
 import { performanceLog } from "../../decorators";
-import { Timers } from "../../timers";
 
 export class DevicesService extends EventEmitter implements Mobile.IDevicesService {
 	private static DEVICE_LOOKING_INTERVAL = 200;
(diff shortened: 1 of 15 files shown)


rosen-vladimirov:
test cli-smoke

lib/common/mobile/mobile-core/devices-service.ts
```diff
@@ -287,51 +289,63 @@ export class DevicesService extends EventEmitter implements Mobile.IDevicesServi
...
 
-		emulatorDetectionInterval.unref();
-		this.deviceDetectionIntervals.push(emulatorDetectionInterval);
+		this.emulatorDetectionInterval.unref();
+	}
+
+	@exported("devicesService")
+	public stopDeviceDetectionInterval(): void {
+		if (this.deviceDetectionInterval) {
+			clearInterval(this.deviceDetectionInterval);
```
Fatme: Maybe it'll be better to expose a public `clearInterval` method from `timers` service

rosen-vladimirov: Done in latest commit.

rosen-vladimirov:
test cli-smoke

PR #4490, opened 2019-04-01
chore update the package locks

DimitarTachev:


DimitarTachev:
test cli-smoke package_version#latest

PR #4510, opened 2019-04-08
fix: handle HTTP 304 response status code

rosen-vladimirov:
In case 304 is returned, the cached version of the result should be used, so consider this response as success. NOTE: It can be returned only when specific header is sent in the request, so the caller actually expects to receive either 304 or 200 with new content.

## PR Checklist

Http requests returning 304 are considered as failed.

## What is the new behavior?
Http requests returning 304 are considered as successful.

### DIFF
--- a/lib/common/http-client.ts
+++ b/lib/common/http-client.ts
@@ -168,7 +168,7 @@ export class HttpClient implements Server.IHttpClient {
 							this.setResponseResult(promiseActions, cleanupRequestData, { err: new Error(HttpClient.STUCK_RESPONSE_ERROR_MESSAGE) });
 						}
 					}, HttpClient.STUCK_RESPONSE_CHECK_INTERVAL);
-					const successful = helpers.isRequestSuccessful(responseData);
+					const successful = helpers.isRequestSuccessful(responseData) || responseData.statusCode === HttpStatusCodes.NOT_MODIFIED;
 					if (!successful) {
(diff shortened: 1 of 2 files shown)


lib/common/http-client.ts
```diff
@@ -168,7 +168,7 @@ export class HttpClient implements Server.IHttpClient {
 							this.setResponseResult(promiseActions, cleanupRequestData, { err: new Error(HttpClient.STUCK_RESPONSE_ERROR_MESSAGE) });
 						}
 					}, HttpClient.STUCK_RESPONSE_CHECK_INTERVAL);
-					const successful = helpers.isRequestSuccessful(responseData);
+					const successful = helpers.isRequestSuccessful(responseData) || responseData.statusCode === HttpStatusCodes.NOT_MODIFIED;
```
Fatme: Maybe the check `responseData.statusCode === HttpStatusCodes.NOT_MODIFIED` should be inside `isRequestSuccessful` function

rosen-vladimirov: I've considered this option, but I'm not sure if it is applicable for everyone who would use `isRequestSuccessful` helper method. By definition [only requests between 200 and 300 are success](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#2xx_Success), so it will seem strange to handle 304 inside that method

Issue #3993, opened 2018-10-08
Improve `ctrl + c` handling


DimitarTachev:
**Is your feature request related to a problem? Please describe.**
Currently, the CLI cannot be easily killed with `ctrl + c`. 

**Describe the solution you'd like**
Stop handling the SIGINT signals and let the nodejs do the job.

**Describe alternatives you've considered**
Extract a common helpers for managing the timeouts, intervals and child processes and keep the exit logic in a single place.

rosen-vladimirov:
The issue should be fixed in latest `next` version of nativescript CLI. To get the full benefits of it, you can install CLI's next version by executing `npm i -g nativescript@next`.
The new approach of handling Ctrl + C is described in the related PR. The main concept is that now CLI does not handle SIGINT signal. Instead, it starts a new child process that will cleanup the resources once CLI's process dies. As the child process is detached and executes most of its work after CLI's dead, you can pass `--cleanupLogFile <path to file>` where the logs of the execution will be stored. The child process will execute the cleanup actions, write information to the specified file (in case the `cleanupLogFile` option is not passed, it will not create log) and after that it will exit.

PR #4533, opened 2019-04-12
feat: introduce better way for handling Ctrl+C

rosen-vladimirov:
During its work, CLI uses different resources that should be cleaned once CLI process finishes its work. There are several ways for CLI to exit:
 - short living command, like `tns version`, `tns run <platform> --justlaunch` are executed and at the end of the execution, `$injector.dispose` is called. At this point the `dispose` methods of each of the injected dependencies is called. That's where you can clean the used resources. The main purpose of the `dispose` event is to cleanup some resources like timers, which will prevent CLI process from exit gracefully.
 - long living commands, like `tns run`, `tns debug <platform>` are usually killed with `Ctrl + C` (SIGINT signal is sent to CLI). To handle this we've added handling for SIGINT signal in CLI.

However, handling SIGINT in CLI led to a problem. As per [Node.js documentation](https://nodejs.org/dist/latest-v11.x/docs/api/process.html):
```
'SIGTERM' and 'SIGINT' have default handlers on non-Windows platforms that reset the 
terminal mode before exiting with code 128 + signal number. 
If one of these signals has a listener installed, its default behavior will 
be removed (Node.js will no longer exit).
```

So after we've added the handling, we had to manually cleanup all resources that prevent Node.js from exiting gracefully, i.e. pending Promises, pending timers, child processes, etc. These operations take time and users started complaining that pressing Ctrl + C does not exit the process and they have to press it several times. Also many users started pressing Ctrl + Z, which just sends CLI process in background, but they thought this kills the process as the terminal is not blocked by CLI anymore.

So we needed a solution how to cleanup some resources that Node.js does not clean when exiting on its own, but also we did not want to take care of killing all child_processes and timers.

By default, when Node.js receives SIGINT it kills all child processes started from the current process, clears the pending timers and exits. Detached processes are not killed as by default they are not related to current process.

So this PR adds a new detached child process that executes the cleanup procedures once CLI process exits. The child process is detached, so Node.js will not kill it when Ctrl + C is used to stop CLI's process. The child process communicates with CLI via IPC communication. Once CLI process dies, the child process will receive `disconnect` event and will start execution of cleanup tasks.

You can add/remove cleanup actions by using the newly introduced `$cleanupService` in the CLI. With it you can add cleanup actions which are actually processes that should be spawned from the cleanup service, for example `adb` command with some args. For each action you can specify command, args and timeout. In case the cleanup process is unable to execute the cleanup action for specified time (default is 3 seconds), the current cleanup action is stopped and next one starts.
As for our `lockfile` we need to ensure some locks are deleted once CLI dies, the `$cleanupService` allows you to specify file/directory that should be deleted from the cleanup process.

As the cleanup process is detached, we do not have any logs from it. This makes it hard to understand if it actually does its job, so add a way to have logs from this file. In CLI this happens by passing `--cleanupLogFile <path to file>`. As CLI can be used as a library, introduce public method that allows setting the logs file location.

The whole effort made current implementation of handling exit signals (`$processService`) unusable, so it is deleted and where it is appropriate it is replaced by `$cleanupService`. 

More information for the changes is available in each commit.

## PR Checklist

Running some long living CLI commands, like `tns run`, `tns debug <platform>`, etc. and trying to stop them with `Ctrl + C` does not stop the process immediately or causes different unexpected logs. In some cases the process does not die at all.

## What is the new behavior?
Pressing `Ctrl + C` kills CLI process immediately.

Fixes issue: https://github.com/NativeScript/nativescript-cli/issues/3993


rosen-vladimirov:
test cli-smoke cli-preview

lib/detached-processes/cleanup-process-definitions.d.ts
```diff
@@ -0,0 +1,36 @@
...
+	 * Type of the action
+	 */
+	actionType: CleanupProcessMessageType;
+}
+
+interface ICleanupActionMessage extends ICleanupProcessMessage {
+	/**
+	 * Describes the action that must be executed
+	 */
+	action: ICleanupAction;
```
DimitarTachev: action -> command

lib/detached-processes/cleanup-process.ts
```diff
@@ -0,0 +1,116 @@
...
+const actionsToExecute: ICleanupAction[] = [];
+const filesToDelete: string[] = [];
+
+const executeCleanup = async () => {
+	const $childProcess = $injector.resolve<IChildProcess>("childProcess");
+	for (const action of actionsToExecute) {
+		try {
+			fileLogService.logData({ message: `Start executing action: ${JSON.stringify(action)}` });
+
+			// TODO: Add timeout for each action here
```
DimitarTachev: forgotten TODO?

lib/detached-processes/detached-process-enums.d.ts
```diff
@@ -0,0 +1,47 @@
...
+	 */
+	Info = "Info",
+
+	/**
+	 * Error message - used to indicate that some action did not succeed.
+	 */
+	Error = "Error"
+}
+
+declare const enum CleanupProcessMessageType {
```
DimitarTachev: Respect the interface naming changes here

rosen-vladimirov:
test cli-smoke cli-preview

rosen-vladimirov:
test cli-smoke cli-preview cli-misc

rosen-vladimirov:
Merge after https://github.com/NativeScript/nativescript-cloud/pull/167

rosen-vladimirov:
test cli-preview

rosen-vladimirov:
test cli-templates

Issue #4553, opened 2019-04-21
Unable to determine Swift version for the following pods


sandoche:
- CLI: 5.3.2

When building with Sidekick for iOS in the cloud for the release I get the following error:
```
[19-04-21 17:33:16.698] (CLI)                                                  App_Resources/iOS/baseline_share_black_24@2x.png   14.7 KiB          [emitted]  
[19-04-21 17:33:16.698] (CLI)                                                  App_Resources/iOS/baseline_share_black_24@3x.png   15.2 KiB          [emitted]  
[19-04-21 17:33:16.698] (CLI)                                                                  App_Resources/iOS/build.xcconfig  504 bytes          [emitted]  
[19-04-21 17:33:16.698] (CLI)                                                        App_Resources/iOS/GoogleService-Info.plist   1.38 KiB          [emitted]  
[19-04-21 17:33:16.698] (CLI)                                                                      App_Resources/iOS/Info.plist   1.47 KiB          [emitted]  
[19-04-21 17:33:16.698] (CLI)                                                         App_Resources/iOS/LaunchScreen.storyboard   3.84 KiB          [emitted]  
[19-04-21 17:33:16.698] (CLI)                                                                         App_Resources/iOS/PodFile  207 bytes          [emitted]  
[19-04-21 17:33:16.698] (CLI)                                                                            assets/images/fork.png   18.8 KiB          [emitted]  
[19-04-21 17:33:16.698] (CLI)                                                                            assets/images/hero.png    113 KiB          [emitted]  
[19-04-21 17:33:16.698] (CLI)                                                                            assets/images/star.png   18.5 KiB          [emitted]  
[19-04-21 17:33:16.698] (CLI)                                                                                      package.json  129 bytes          [emitted]  
[19-04-21 17:33:16.698] (CLI)                                                                                        starter.js   39 bytes          [emitted]  
[19-04-21 17:33:16.698] (CLI) Entrypoint bundle = vendor.js bundle.js
[19-04-21 17:33:16.698] (CLI) [../node_modules/babel-loader/lib/index.js!../node_modules/vue-loader/lib/index.js?!./views/News.vue?vue&type=script&lang=js&] ../node_modules/babel-loader/lib!../node_modules/vue-loader/lib??vue-loader-options!./views/News.vue?vue&type=script&lang=js& 8.53 KiB {bundle} [built]
[...]

PR #4559, opened 2019-04-22
fix: prepare xcconfig files for all conigurations

rosen-vladimirov:
Currently CLI merges and prepares xcconfig files only for the current build configuration, i.e. debug or release. However, the `pod install` command works with both configurations. This leads to problem when some property is set in the project's (plugin's) build.xcconfig file as CLI will prepare the native iOS project only for one configuration. When `pod install` is executed, if the Pods project uses the property set in the build.xcconfig file, it will fail with error, as the value for debug and release will be different (i.e. it will be set only for one of them). Pods project does not support this behavior, so the build operation fails.

So, ensure CLI merges all xcconfig files for both debug and release configurations. This way, the Pods project will be in sync, so it should continue its execution. This is important when using Cocoapods 1.6.0+ where some changes are applied for Swift support, which can be workarounded by setting SWIFT_VERSION in project's build.xcconfig file.

## PR Checklist

## What is the new behavior?

Related to: https://github.com/NativeScript/nativescript-cli/issues/4553 
Fixes issue: https://github.com/NativeScript/nativescript-cli/issues/4561

### DIFF
--- a/lib/declarations.d.ts
+++ b/lib/declarations.d.ts
@@ -890,12 +890,11 @@ interface IXcprojInfo {
 
 interface IXcconfigService {
 	/**
-	 * Returns the path to the xcconfig file
+	 * Returns the paths to the xcconfig files for build configuration (debug/release)
 	 * @param projectRoot The path to root folder of native project (platforms/ios)
-	 * @param opts
-	 * @returns {string}
+	 * @returns {IStringDictionary}
 	 */
-	getPluginsXcconfigFilePath(projectRoot: string, opts: IRelease): string;
+	getPluginsXcconfigFilePaths(projectRoot: string): IStringDictionary;
 
 	/**
 	 * Returns the value of a property from a xcconfig file.
--- a/lib/services/cocoapods-service.ts
+++ b/lib/services/cocoapods-service.ts
@@ -55,14 +55,16 @@ export class CocoaPodsService implements ICocoaPodsService {
 		return podInstallResult;
 	}
 
-	public async mergePodXcconfigFile(projectData: IProjectData, platformData: IPlatformData, opts: IRelease) {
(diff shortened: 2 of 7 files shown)


rosen-vladimirov:
test cli-smoke cli-templates cli-plugin

rosen-vladimirov:
test cli-smoke cli-templates cli-plugin

.vscode/launch.json
```diff
@@ -7,15 +7,15 @@
 		{
 			"type": "node",
 			"request": "launch",
-			"cwd": "${workspaceRoot}",
+			"cwd": "${workspaceRoot}/scratch/app1",
```
rosen-vladimirov: revert this file

lib/declarations.d.ts
```diff
@@ -895,7 +895,7 @@ interface IXcconfigService {
 	 * @param opts
```
rosen-vladimirov: fix JSDocs based on changes in the method

lib/services/cocoapods-service.ts
```diff
@@ -55,14 +55,16 @@ export class CocoaPodsService implements ICocoaPodsService {
 		return podInstallResult;
 	}
 
-	public async mergePodXcconfigFile(projectData: IProjectData, platformData: IPlatformData, opts: IRelease) {
+	public async mergePodXcconfigFile(projectData: IProjectData, platformData: IPlatformData) {
```
rosen-vladimirov: add return type

.vscode/launch.json
DimitarTachev: ℹ️ `git update-index --assume-unchanged .vscode/launch.json`

rosen-vladimirov:
test cli-smoke cli-plugin

PR #4573, opened 2019-04-30
fix: logcat process is not restarted in some cases

rosen-vladimirov:
In case you have `keepSingleProcess` passed to logcat Stream (this option is passed when using CLI as a library - in this case we keep a long living logcat process alive and change only the filter for it by specifying different PIDs), in case the logcat process dies (for example device is disconnected), the process is not cleaned from CLI's internal cache.
This way, in case you try to start the logs again (in the example above - just attach the device again), they are not started as CLI thinks it has a process for this case. So, in case the logcat stream process dies, ensure we clean all the resources related to it.
Add unit tests for this case and remove calls to `type`/`cat` from the tests - there's no need to spawn actual child process in the tests.

## PR Checklist

When using CLI as a library attaching Android device, disconnecting it after that and reconnecting it does not show any logs for this device.

## What is the new behavior?
Logs from Android devices are properly shown.

Fixes issue in Sidekick (logged in JIRA)

### DIFF
--- a/lib/common/mobile/android/logcat-helper.ts
+++ b/lib/common/mobile/android/logcat-helper.ts
@@ -39,7 +39,8 @@ export class LogcatHelper implements Mobile.ILogcatHelper {
 
 			logcatStream.on("close", (code: number) => {
 				try {
-					this.stop(deviceIdentifier);
+					this.forceStop(deviceIdentifier);
+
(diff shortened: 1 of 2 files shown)


rosen-vladimirov:
test cli-smoke

PR #4592, opened 2019-05-08
feat: improve logger

rosen-vladimirov:
Currently CLI's logger uses different ways to print messages on stdout and stderr:
* using `log4js` in some cases
* using `console.log`
* using `process.stdout.write`
* using `process.stderr.write`

Also, whenever we want to colorize the message, we include the colorization of the message in the message itself, before passing it to one of the methods above. This is a problem when CLI is used as a library, when it might not be used in a terminal environment - the messages contain terminal colorization. Also, in order to get the messages, you need to parse stdout and stderr of the process in which you've required CLI as a library.

To resolve the above, streamline the logger ways for handling messages - use `log4js` everywhere. This allows using different logging mechanism when CLI is used as a command line and when it is used as a library. 
`log4js` has the concept of using `appenders` and `layout` when printing messages.
* `appenders` are responsible for output of log events. They may write events to files, send emails, store them in a database, or anything. Most appenders use layouts to serialise the events to strings for output.
* `layout` is a function for converting a LogEvent into a string representation.

Currently we have the several use cases for logging messages:
* print on `stdout` or `stderr` - for some of the methods we need to print errors and we want them to be shown on `stderr`. This can be handled in the appender, as it is the one which actually prints the message. To handle the case, pass `useStderr` property in the context of the LogEvent.
* add new line at the end of the message or skip it - in some cases we do not want to append new line at the end of the message. This logic is for the presentation of the message, so it can be handled in the layout. Check the value of `skipNewLine` property of the LogEvent context.
* wrap message with borders - in some cases, especially for command line usage, we want to wrap the message in some borders, to make it more visible to the user. This is a layout specific logic that can be handled based on the `wrapMessageWithBorders` property of the LogEvent context.

Each of the properties can be passed directly to each of the methods exposed from `logger`, i.e.:
```TypeScript
const a = 1;
this.$logger.info("This is info message");
this.$logger.info("This is info message with placeholder %s", a);
this.$logger.info("This is info message without new line at the end", { skipNewLine: true });
this.$logger.info("This is info message with placeholder %s and without new line at the end,", a, { skipNewLine: true });
this.$logger.info("This is info message without new line at the end and printed on stderr", { skipNewLine: true, useStderr: true });
this.$logger.info("This is another info message without new line at the end and printed on stderr", { skipNewLine: true}, { useStderr: true });
this.$logger.info("This is info message with placeholder %s and without new line at the end printed on stderr", a, { skipNewLine: true, useStderr: true });
this.$logger.info("This is another info message with placeholder %s and without new line at the end printed on stderr", a, { skipNewLine: true }, { useStderr: true });
```

Full list of the properties that can be set is available in `constants.LoggerConfigData`.

When using CLI as a library, it makes sense to skip the colorization and instead of printing to `stdout`/`stderr`, to allow the consumer to take care of the messages on its own. In order to handle this, introduce an easy way for initializing the logger with different settings, i.e. `appender`, `layout`, `level`. This can be achieved through the `logger.initialize` method or `initializeService.initialize` method.
To make it easy when CLI is used as a library, introduce a custom `emit-appender` that emits the logged data instead of printing it. Its usage is fairly simple:
```JavaScript
const tns = require("nativescript");
const { EventEmitter } = require("events");
const { EMIT_APPENDER_EVENT_NAME, LoggerAppenders } = tns.constants;
const emitter = new EventEmitter();
// IMPORTANT: Add the event handler before calling logger's initialize method.
// This is required as log4js makes a copy of the appenderOptions, where the emitter is passed
// NOTE: In case you want to debug the event handler, place `debugger` in it.
emitter.on(EMIT_APPENDER_EVENT_NAME, (logData) => {
	if (logData.loggingEvent.level.levelStr === LoggerLevel.WARN) {
		console.log(`WARNING: ${logData.formattedMessage}`);
	}
});

const logger = tns.logger;
logger.initialize({
	appenderOptions: {
		type: LoggerAppenders.emitAppender,
		emitter
	}
});
```

### Additional code changes
During the implementation of the logger, several changes in other services were required:
* rename `initService` to `projectInitService` - it was required to make it easier to determine if the service is for initializing CLI or the project.
* remove `progress-indicator` - it was not used anywhere
* remove `printMsgWithTimeout` method from logger - it was used only in `progress-indicator`
* introduce `initializeService` - it is used to initialize CLI specific modules (like logger) and show all system warnings (for example for not supported macOS version). This way we no longer need to expose different methods when we want to show some warnings in the command line and when CLI is used as a library. We can just place them in the initialize method.
* remove `warnWithLabel` method from logger - it has been used on only two places and did not make sense to keep it.
* remove `out` method from logger - its usage is replaced with `logger.info`
* remove `write` method from logger - its intention was to print message on stdout, without modifying it (i.e. without adding new line, etc.), so it is replaced with `info` method and passing `skipNewLine` to true:
```TypeScript
// OLD:
this.logger.write(line, platform, deviceIdentifier);
// NEW:
this.logger.info(line, platform, deviceIdentifier, { [LoggerConfigData.skipNewLine]: true });
```
* remove `printInfoMessageOnSameLine` method from logger - it was used to print messages on stdout without adding new line and only when log level is info. However, the implementation was incorrect (i.e. it was not checking the log level correctly). Replace the method with `logger.info` calls:
```TypeScript
// OLD:
this.$logger.printInfoMessageOnSameLine("Waiting for emulator device initialization...");
// NEW:
this.$logger.info("Waiting for emulator device initialization...", { [LoggerConfigData.skipNewLine]: true });
```
* change `logger.error` to print to `stderr` instead of `stdout`.
* remove `printOnStderr` method from logger. This can be achieve with any log level by passing `useStderr: true`. Usage is replaced with `logger.error`.
* introduce `initializeCliLogger` method in logger, so it will be easy to setup CLI's log4js options
* introduce special layout function for CLI - based on the LoggingEvent context it changes the way message is printed. This layout should not be used when CLI is as a library.
* move all `logger` related files in a `logger` directory.

## PR Checklist

## What is the new behavior?


rosen-vladimirov:
test cli-smoke cli-create cli-plugin cli-preview cli-misc

rosen-vladimirov:
test cli-smoke cli-misc

rosen-vladimirov:
test cli-smoke cli-misc

PublicAPI.md
```diff
@@ -1514,6 +1524,155 @@ const tns = require("nativescript");
 tns.cleanupService.setCleanupLogFile("/Users/username/cleanup-logs.txt");
 ```
 
+## initializeService
+The `initializeService` is used to initialize CLI's configuration and the beginning and print all warnings related to current environment.
```
KristianDD: configuration **at** the beginning

rosen-vladimirov:
test cli-smoke cli-create cli-plugin

rosen-vladimirov:
test cli-smoke cli-misc

PR #4679, opened 2019-06-05
fix: logger fails to print null objects

rosen-vladimirov:
In case you try printing the following:
```
const obj1 = null;
this.$logger.info("value %s", obj1);
```
Our implementation fails with error: `Cannot read property hasOwnProperty of null`.
The problem is in the way we check for passed logger options in the args passed to `logger` methods. The `typeof null` returns object, so next parts of the code fail.

Also passed logger options are handled by priority, i.e. the first passed option should be used. This allows callers of methods that specify some logger options internaly, to overwrite their values. Add unit tests for this behavior.
Fix the case when same logger option is passed multiple times in different objects - currently only one of the instances is removed and the other one is included in the message visible to the user. Now all such objects are removed.
Add unit tests for all of the described cases.

## PR Checklist

## What is the new behavior?

### DIFF
--- a/lib/common/logger/logger.ts
+++ b/lib/common/logger/logger.ts
@@ -154,31 +154,40 @@ export class Logger implements ILogger {
 	}
 
 	private getLogOptionsForMessage(data: any[]): { data: any[], [key: string]: any } {
-		const opts = _.keys(LoggerConfigData);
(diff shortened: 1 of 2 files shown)


rosen-vladimirov:
test cli-smoke

PR #4711, opened 2019-06-11
fix: fix plugin's development workflow

Fatme:
Currently CLI provides the following watch patterns `node_modules/**/platforms/<platform>` to `chokidar`. The plugin developers have `symlinked` plugin's source code inside the plugin's demo projects. When the plugin's demo project is ran and a `.js` or `.ts` file is changed in `src` folder of a plugin, `chokidar` reports it as a changed file. It seems as an issue with the way how `symlinked` files are processed from `chokidar` itself. This led to a problem that when a file from plugin's src folder is changed, CLI receives changed event for this file twice - ones from `chokidar` and ones from `webpack's watcher`. After that CLI restarts the application as a change from "native watcher" is received. In order to fix this issue, we changed the patterns provided to choki as we removed the glob pattern and provides only full paths to all nativescript's plugins of the application.

Also this PR removes the check if a native build should be executed and the project is built every time when chokidar reports changed files. This check was based on the result of `prepareNativePlatform` method. This method checks if there are newer files in nativescript's plugins of the application - it traverses all nativescript's plugins file by file and compares the modify time of each file with the modify time of `.nsprepareinfo` file from `platforms` folder. If a newer files from nativescript's plugins of the application is found, CLI returns that there is a native change. This actually doesn't work in case when a file from plugin's src folder is changed, as CLI checks only files from node_modules/<plugin-name>. In this case CLI reports that there is no native change.

## PR Checklist

## What is the new behavior?

### DIFF
--- a/lib/controllers/prepare-controller.ts
+++ b/lib/controllers/prepare-controller.ts
@@ -4,7 +4,7 @@ import { hook } from "../common/helpers";
 import { performanceLog } from "../common/decorators";
 import { EventEmitter } from "events";
 import * as path from "path";
-import { PREPARE_READY_EVENT_NAME, WEBPACK_COMPILATION_COMPLETE, PACKAGE_JSON_FILE_NAME } from "../constants";
+import { PREPARE_READY_EVENT_NAME, WEBPACK_COMPILATION_COMPLETE, PACKAGE_JSON_FILE_NAME, PLATFORMS_DIR_NAME } from "../constants";
 
 interface IPlatformWatcherData {
 	webpackCompilerProcess: child_process.ChildProcess;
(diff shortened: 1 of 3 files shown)


lib/controllers/prepare-controller.ts
```diff
@@ -114,12 +115,16 @@ export class PrepareController extends EventEmitter {
 			return false;
 		}
 
+		const pluginsNativeDirectories = this.$nodeModulesDependenciesBuilder.getProductionDependencies(projectData.projectDir)
+			.filter(dep => dep.nativescript)
+			.map(dep => path.join(dep.directory, "platforms", "ios"));
```
rosen-vladimirov: platforms and ios are available in constants (I think), so you can use them from there. Ensure you've used `toLowerCase()` of the iOS constant.

PR #4722, opened 2019-06-14
Add time(hours, minutes and seconds) to cli version.

vhristov5555:
All other packages in Nativescript(e.g. Runtime, Angular...) have time in their version. We are using the time in the version to determine if we need to run tests. We run tests if the package has been released  within the 24 hours period since running the triggering job.


vhristov5555:
All other packages format is  `6.0.0-2019-06-13-112614-01` so this should be right.

PR #4724, opened 2019-06-15
fix: fix  the way android appResources are processed

Fatme:
Currently we copy all App_Resources from `./app/App_Resources/Android/*` to `./platforms/../app/src`. As there is src folder in `./app/App_Resources/Android/*`, we receive as a result inside platforms `./platforms/app/../src/src`.

## PR Checklist

## What is the new behavior?

### DIFF
--- a/lib/services/android-project-service.ts
+++ b/lib/services/android-project-service.ts
@@ -289,8 +289,8 @@ export class AndroidProjectService extends projectServiceBaseLib.PlatformProject
 
 		this.$fs.ensureDirectoryExists(platformsAppResourcesPath);
 
-		this.$fs.copyFile(path.join(projectAppResourcesPath, platformData.normalizedPlatformName, "*"), platformsAppResourcesPath);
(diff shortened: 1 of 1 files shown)


lib/services/android-project-service.ts
```diff
@@ -289,8 +289,8 @@ export class AndroidProjectService extends projectServiceBaseLib.PlatformProject
 
 		this.$fs.ensureDirectoryExists(platformsAppResourcesPath);
 
-		this.$fs.copyFile(path.join(projectAppResourcesPath, platformData.normalizedPlatformName, "*"), platformsAppResourcesPath);
-			const appResourcesDirStructureHasMigrated = this.$androidResourcesMigrationService.hasMigrated(projectAppResourcesPath);
+		this.$fs.copyFile(path.join(projectAppResourcesPath, platformData.normalizedPlatformName, "*"), path.dirname(platformsAppResourcesPath));
```
rosen-vladimirov: maybe I'm mistaken, but isn't the whole row on a wrong place? Currently, in case the Android App_Resources are migrated, we will copy them here and on line 295.
I think the correct fix is to move the current line in the else statement below.

PR #4761, opened 2019-06-24
fix: device logs are parsed slowly

rosen-vladimirov:
Currently it takes around half a second to parse each console message from the application. For callstacks, which contain a lot of lines, which we try to map to original file, it takes several seconds.
Also, at the moment, when the application reports that some log is from bundle.js/vendor.js, we parse the local files, which may not be the same files uploaded on the device. This may happen in cases where we have applied a hot module, the local bundle.js is changed and its source map is changed, but we have not uploaded it on the device yet.

The reason for the slowness is that we read the whole original file every time when we see lines from the logs that points to it. We also create a new SourceMapConsumer for it. This means, that if the stacktrace contains 10 lines pointing to vendor.js, we'll read it 10 times.

To handle both issues, set the sourceMapConsumer (and read the original file location) at the point when we upload files on the device. This way we'll always have in memory the same file as the one currently running on device and the source map will be correct. Also, this way the files will be parsed only once.

## PR Checklist

## What is the new behavior?

### DIFF
--- a/lib/common/definitions/mobile.d.ts
+++ b/lib/common/definitions/mobile.d.ts
@@ -165,6 +165,14 @@ declare module Mobile {
 	 */
 	interface IDeviceLogProvider extends NodeJS.EventEmitter {
 		/**
+		 * Sets the path to source file from which the logs are produced,
+		 * i.e. the original file location of the file running on device.
+		 * @param {string} pathToSourceFile Path to the source file.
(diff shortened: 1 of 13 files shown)


rosen-vladimirov:
test cli-smoke

DimitarTachev:
test

DimitarTachev:
test cli-smoke

Issue #4770, opened 2019-06-26
incorrectly execution of hooks where there is unresolved injected dependency


Fatme:
**Environment**
Provide version numbers for the following components (information can be retrieved by running `tns info` in your project folder or by inspecting the `package.json` of the project):
 - CLI: next
 - Cross-platform modules:
 - Android Runtime:
 - iOS Runtime:
 - Plugin(s):

**Describe the bug**

**To Reproduce**
1. Create new application
2. Create `hooks` directory if no such
3. Create `after-prepare` directory under hooks directory if no such
4. Create `hook1.js`, `hook2.js` and `hook3.js` inside `after-prepare` directory
#### hook1.js
```
module.exports = function() {
    console.log("=================== HOOK1 ================ ");
}
```

#### hook2.js
```
module.exports = function ($platformsData) {
    console.log("=================== HOOK2 =================== ");
}
```

#### hook3.js
```
module.exports = function() {
    console.log("=================== HOOK3 ================ ");
}
```
5. Execute `tns prepare ios`

**Actual behavior** 
You'll see the following output:
```
=================== HOOK1 ================ 
after-prepare will NOT be executed because it has invalid arguments - $platformsData.
```

**Expected behavior**
You should see the following output:
```
=================== HOOK1 ================ 
after-prepare will NOT be executed because it has invalid arguments - $platformsData.
=================== HOOK3 ================ 
```

PR #4771, opened 2019-06-26
fix: fix the execution of hooks when there is unresolved injected dependency

Fatme:
Rel to: https://github.com/NativeScript/nativescript-cli/issues/4770

## PR Checklist

## What is the new behavior?

### DIFF
--- a/lib/common/services/hooks-service.ts
+++ b/lib/common/services/hooks-service.ts
@@ -121,7 +121,7 @@ export class HooksService implements IHooksService {
 
 				if (invalidArguments.length) {
 					this.$logger.warn(`${hookName} will NOT be executed because it has invalid arguments - ${invalidArguments.join(", ").grey}.`);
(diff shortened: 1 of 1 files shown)


lib/common/services/hooks-service.ts
```diff
@@ -121,7 +121,7 @@ export class HooksService implements IHooksService {
 
 				if (invalidArguments.length) {
 					this.$logger.warn(`${hookName} will NOT be executed because it has invalid arguments - ${invalidArguments.join(", ").grey}.`);
```
vakrilov: Suggestion: Log the actual file name of the hook with invalid arguments.
Currently the log says: `after-prepare will NOT be executed because it has invalid arguments - platformsData, $usbLiveSyncService.` , but if you have multiple after-prepare hook it would be useful to know which exactly is causing the problem.

PR #4802, opened 2019-07-03
fix: preview and run do not work correctly 

rosen-vladimirov:
…
In case there are multiple devices with the same platform, `run command` fails with error `Cannot read property hasNativeChanges of undefined`. The problem is that we return incorrect prepare data.
Also we have a problem in `tns preview` when multiple platforms are used - all files are send multiple times to each device as we have too many started webpack processes (in case you scan fast with devices) and we are also adding handler of the prepare ready event per each device instead per platform. Fix this by handling the event only once per platform

## PR Checklist

## What is the new behavior?

### DIFF
--- a/lib/controllers/prepare-controller.ts
+++ b/lib/controllers/prepare-controller.ts
@@ -1,4 +1,3 @@
-import * as child_process from "child_process";
 import * as choki from "chokidar";
 import { hook } from "../common/helpers";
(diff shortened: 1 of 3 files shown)


rosen-vladimirov:
test cli-smoke cli-preview cli-device package_version#rc

PR #4827, opened 2019-07-10
fix: `tns debug ios` should work on device

rosen-vladimirov:
Currently `tns debug ios` hangs when executed for device. The problem is that we pass `"waitForDebugger": false` to ios-device-lib binary, which can parse only strings. The value `false` is boolean, not string, so ios-device-lib binary crashes, but CLI does not understand this and hangs.
Fix this by ensuring we pass waitForDebugger as string.

## PR Checklist

## What is the new behavior?

Fixes issue https://github.com/NativeScript/nativescript-cli/issues/4828


rosen-vladimirov:
test cli-device package_version#rc

PR #4836, opened 2019-07-11
fix: don't build every time the application from sidekick when using cloud build

Fatme:
Currently the application is built every time from sidekick when opening the app and starting cloud build. This happens as we decided that there are changes in app_resources as `prepareData.release` is undefined but `prepareInfo.release` is false. This way the check `prepareData.release !== this._prepareInfo.release` passes and we set `this._changesInfo.appResourcesChanged` to `true`.

## PR Checklist

- [ ] The PR title follows our guidelines: https://github.com/NativeScript/NativeScript/blob/master/CONTRIBUTING.md#commit-messages.
- [ ] There is an issue for the bug/feature this PR is for. To avoid wasting your time, it's best to open a suggestion issue first and wait for approval before working on it.
- [ ] You have signed the [CLA](http://www.nativescript.org/cla).
- [ ] All existing tests are passing: https://github.com/NativeScript/nativescript-cli/blob/master/CONTRIBUTING.md#contribute-to-the-code-base
- [ ] Tests for the changes are included.

## What is the new behavior?

Fixes/Implements/Closes #[Issue Number].

PR #4843, opened 2019-07-12
fix: fix the missing icons in newly created apps on iOS devices

Fatme:
Currently the icons of newly created apps are not displayed when running the application on iOS device. This happens as the icons are not included into `.pbxproject` file. This happens as the `prepareProject` method is not called due to this check https://github.com/NativeScript/nativescript-cli/blob/release/lib/services/platform/prepare-native-platform-service.ts#L35. This check is based on hasChangesRequirePrepare from `projectChangesService` which value is based on `appResourcesChanged` and `signingChanged` - https://github.com/NativeScript/nativescript-cli/blob/release/lib/services/project-changes-service.ts#L27. The values of those properties are not populated correctly initially when the application is without `platforms` directory.

This functionality was broken with this PR https://github.com/NativeScript/nativescript-cli/pull/4836/files. As this PR changes the behavior of this if, now it never passes and respectively appResourcesChanged and signingChanged are with false values. In other words, we've broken the functionality that set initial values of those properties.

In order to set correctly the initial values of those properties, we decided deleting the setNativePlatformStatus call after adding the platform. This functionally was added in this PR https://github.com/NativeScript/nativescript-cli/pull/4139. So here come the question how calling setNativePlatformStatus after adding the platform reflects the current issue with missing icons on device.

`setNativePlatformStatus` method sets the nativePlatformStatus property in _prepareInfo and saves it to `.nsprepareinfo` file which happens when adding the platform. After that, CLI calls `ensurePrepareInfo` method https://github.com/NativeScript/nativescript-cli/blob/release/lib/services/project-changes-service.ts#L160. As we've already saved prepareInfo to `.nsprepareinfo` file, this if always passes https://github.com/NativeScript/nativescript-cli/blob/release/lib/services/project-changes-service.ts#L162 and we've never reached the code that sets the default values here https://github.com/NativeScript/nativescript-cli/blob/release/lib/services/project-changes-service.ts#L183. This happens as we've already saved `.nsprepareinfo` from `setNativePlatformStatus` method. As calling `setNativePlatformStatus` after adding the platform was introduced as a bug fix when supporting legacy and bundle workflow, we decided to delete this call. This way, we'll ensure that initially when the project is without platforms directory, this code we'll be called https://github.com/NativeScript/nativescript-cli/blob/release/lib/services/project-changes-service.ts#L183 and the initial values of properties will be set correctly.

## PR Checklist

## What is the new behavior?

### DIFF
--- a/lib/services/platform/add-platform-service.ts
+++ b/lib/services/platform/add-platform-service.ts
@@ -1,13 +1,12 @@
 import * as path from "path";
 import * as temp from "temp";
-import { PROJECT_FRAMEWORK_FOLDER_NAME, NativePlatformStatus } from "../../constants";
+import { PROJECT_FRAMEWORK_FOLDER_NAME } from "../../constants";
 import { performanceLog } from "../../common/decorators";
 
 export class AddPlatformService implements IAddPlatformService {
 	constructor(
 		private $fs: IFileSystem,
 		private $pacoteService: IPacoteService,
-		private $projectChangesService: IProjectChangesService,
 		private $projectDataService: IProjectDataService,
 		private $terminalSpinnerService: ITerminalSpinnerService
 	) { }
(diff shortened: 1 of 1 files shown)


rosen-vladimirov:
test cli-smoke cli-device package_version#rc

PR #4905, opened 2019-07-26
fix(npm/pacote): allow tns to be able to use npm configuration properly

NathanaelA:
## PR Checklist

If you have a npm registry that is not hosted by npm, the TNS cli will fail to install npm modules.
Doing a `npm i --save @proplugins/blah` will work 100% properly.
Doing a `tns plugins add @proplugins/blah` will partially work.   It actually downloads the package (because it apparently uses npm under the hood and pulls the file and unpacks it, yay!).  
So `node_modules/@proplugins/blah` is unpacked and exists on the drive.   

However, apparently the tns cli uses pacote to verify the existence of the registry entry, BEFORE it adds it to the `package.json` file.  Which of course because it is a older verison of pacote and does not read the npm configuration -- it has NO idea how to resolve `@scoped` to the private server.  So tns then fails and so then tns throws
image

As you can see from the image; npm actually found the package and installed it; but TNS failed and either rolls back the package.json change, or doesn't commit the package.json change.

## What is the new behavior?
Upgraded to latest pacote library; and added ability to read npm configuration files (using the official libnpmconfig, which is then passed into pacote; so of course it now works properly.  :-)

BREAKING CHANGES:
None


NathanaelA:
@rosen-vladimirov - Any chance this can be merged into 6.03 (or whatever the very next release is)?

test/services/pacote-service.ts
```diff
@@ -36,6 +38,20 @@ interface ITestCase extends ITestSetup {
 	expectedArgs: any[];
 }
 
+function createPacoteOptions(source: Object): Object {
+		const options: { [index: string]: any } = {};
+		npmconfig.read().forEach((value: any, key: string) => {
```
DimitarTachev: Use a hardcoded config obj here instead of reading the real config from the file system and applying the same business logic in the unit tests.

DimitarTachev:
Hi @NathanaelA,

Thanks for reporting and handling this!

It looks OK for a patch release as it's really a bug fix, you just have to rebase the PR on our `release` branch as the `master` branch will be released in 6.1.0.

We will run some tests after you rebase the changes on `release` and if everything is OK, we will include it in `nativescript@6.0.3`.

test/services/pacote-service.ts
NathanaelA: You can't because the actual Pacote driver will be loading the npm config; which it passes back and won't match your hard coded one.  So we have to pass in anything we expect it to get by itself.  :-)

NathanaelA: Maybe a better way of explaining is it takes the npm configuration, then adds your passed in object to override them.   So if you pass in `{"hi": "blah"}`.  The Pacote driver because it reads the actual npm config will have the other 100 or so settings in addition to the `{"hi": "blah"}` as it loaded it itself.

NathanaelA:
@DimitarTachev - Are you just wanting me to just move this over to release, or do you still want the test issue changed up a bit?    Trying to understand exactly what my next steps on this needs to be.  :grinning:

DimitarTachev:
@NathanaelA,

Just rebase it on `release` and we will run the required tests on the updated PR.

test/services/pacote-service.ts
DimitarTachev: Thanks for the details. As far as I understand, the problem is caused by `libnpmconfig` which is directly used in the Pacote service. In this way, we cannot inject a Stub returning a hardcoded value in the unit tests. If you refactor this part and extract the `npmconfig` related logic into a service, you will be able to mock it in the unit tests and avoid slow disk operations there. Also, have you checked this feature with `yarn`? Are they using the same config file?

NathanaelA: Yarn will pull the same .npmrc file for server information.  So we should be safe here --  the main values for where servers live and the authentication to talk with them will be pulled from the ~.npmrc file.   

As for moving things into a service and mocking it -- you totally lost me.   Pretend I'm 2yo; and let me know exactly want you need and I'll get it done.   :grinning:    I'll be rebasing this onto Release, and we can move the discussing over their.  :grinning:

NathanaelA: #4919 is the PR against Release

rosen-vladimirov:
Closing this PR in favor of https://github.com/NativeScript/nativescript-cli/pull/4992

PR #4984, opened 2019-08-28
feat: add tracking in analytics for company using NativeScript CLI

rosen-vladimirov:
#### feat: add tracking for companies 
Add tracking with information about the company from which CLI is being used. Use the insights endpoint to get information about the company and track it as custom dimensions on each hit.
>NOTE: Information will be tracked only when CLI's usage-reporting is enabled. In case you do not want to have this information tracked, execute `tns usage-reporting disable`.

#### feat: add ipService to get current ip address
Add a new ipService to get the current ip address. As this information is critical for us, use three different ways to determine the ip address, so even if one of them fails, use the result from the next one.
Use the `play.nativescript.org/api/whoami` endpoint to get data for the public ip address.

#### fix: handle correctly non-shared instances in injector
Currently `$injector` can have non-shared instances by setting the third argument of the `register` method to false, i.e.:
```
$injector.register("myClass", MyClass, false);
```

However, in this case, the injector caches only the last created instance and this way it will not call the dispose method of all created instances. Fix this by caching all instances in an array and iterrate over them when disposing them.
This way we can have correct resolving of instances by name when constructor has non-injectable dependencies:
```
class A {
    constructor(private stringArg: string) { }
}
$injector.register("a", A, false);

// in another file:
$injector.resolve("a", { stringArg: "stringValue" }); //returns new instance;
$injector.resolve("a", { stringArg: "different stringValue" }); //returns new instance;
```

#### feat: introduce jsonFileSettingsService to work with JSON files 
Intorduce a new service - `jsonFileSettingsService` to work with JSON files in which you need to persist specific settings. The read/write operation of the file are locked, which makes the service usable in multiple cases.
Having a separate service allows us to use multiple files to store data instead of writing everything in the user-settings.json file. To create the service reuse the logic from the old UserSettingsServiceBase and make it available through `$injector`.
Add logic in the service to allow writing/reading data within specific cache timeout. In case you want to save data with cache information, just pass this setting to saveSetting method and the service will automatically add required information in the file.
When you read information from the file, the service will check if you have passed cacheTimeout and will respect it when checking the value. In case you've passed cacheTimeout, but the value in the file does not have cache information, the service will return null as it is unable to determine if the cache is valid.
In case there's cache information, the service will return the value in case the cache had not expired and null otherwise. In case the value in the file has cache information, but the readSetting is called without cacheTimeout, the value is returned immediately without checking the cache.

The service is marked as non-shared, so for each `$injector.resolve("jsonFileSettingsService", { jsonFileSettingsPath: "myPath.json" })`, CLI will create a new instance of the service. As all actions are locked via lockfile, this shouldn't be a problem.

#### feat: add caching for getting company information
Add local caching in a file when getting the company information. This allows us to minimize the number of calls to external services. As the companyInsightsService now uses multiple other services and is also exposed publically, rename it to companyInsightsController.
Set the expiration of the cache to 2 days
## PR Checklist

CLI does not track company related information in Google Analytics.

## What is the new behavior?
In case analytics are enabled, CLI tracks information for the company from which it is being executed. This can be checked by passing `--analyticsLogFile <path to some file>` to any CLI command.


lib/common/test/unit-tests/services/json-file-settings-service.ts
```diff
@@ -0,0 +1,372 @@
...
+describe("jsonFileSettingsService", () => {
+	const jsonFileSettingsPath = "abc.json";
+	let dataPassedToWriteFile: { filename: string, data: string }[] = [];
+	let dataPassedToWriteJson: { filename: string, data: any }[] = [];
+	let dataInFile: IDictionary<any> = {};
+	let deletedFiles: string[] = [];
+
+	const createTestInjector = (): IInjector => {
+		const testInjector = new Yok();
+		/**
```
DimitarTachev:
```suggestion
```

lib/common/test/unit-tests/services/json-file-settings-service.ts
```diff
@@ -0,0 +1,372 @@
...
+		 * constructor(jsonFileSettingsPath: string,	}
```
DimitarTachev:
```suggestion
```

lib/common/test/unit-tests/services/json-file-settings-service.ts
```diff
@@ -0,0 +1,372 @@
...
+		 */
```
DimitarTachev:
```suggestion
```

rosen-vladimirov:
test cli-device cli-smoke cli-run cli-debug

rosen-vladimirov:
test cli-device cli-run cli-debug

rosen-vladimirov:
test cli-device cli-debug

rosen-vladimirov:
test cli-run cli-device

rosen-vladimirov:
test cli-device cli-debug

PR #5101, opened 2019-10-29
fix: handle correctly the situation when the compilation hash is the same as the previous one

Fatme:
Currently CLI considers the hash that is the same as the previous as invalid and this way the application is restarted on device. As that hash is already handled, we should consider it as a valid hash.
Steps to reproduce :
```
`tns create ts_proj`  - choose ts tabs project

npm i tns-core-modules@rc --save --save-exact

npm i nativescript-dev-webpack@rc --save-dev --save-exact

node_modules/.bin/update-ns-webpack --deps –configs

tns run android

Change something in app.android.scss - the change is applied, no restart

Change something in app.ios.scss - application restarts
```

Rel to: https://github.com/NativeScript/nativescript-cli/issues/5105

## PR Checklist

## What is the new behavior?

### DIFF
--- a/lib/services/webpack/webpack-compiler-service.ts
+++ b/lib/services/webpack/webpack-compiler-service.ts
@@ -46,6 +46,11 @@ export class WebpackCompilerService extends EventEmitter implements IWebpackComp
 							return;
 						}
 
+						// the hash of the compilation is the same as the previous one
+						if (this.expectedHashes[platformData.platformNameLowerCase] === message.hash) {
+							return;
+						}
+
 						let result;
 
 						if (prepareData.hmr) {


lib/services/webpack/webpack-compiler-service.ts
```diff
@@ -260,7 +260,10 @@ export class WebpackCompilerService extends EventEmitter implements IWebpackComp
 		// if files will be emitted or not. This way, the first successful compilation after fixing the compilation error generates
 		// a hash that is not the same as the one expected in the latest emitted hot-update.json file.
 		// As a result, the hmr chain is broken and the changes are not applied.
-		const isHashValid = nextHash ? this.expectedHashes[platform] === currentHash : true;
+		let isHashValid = nextHash ? this.expectedHashes[platform] === currentHash : true;
+		if (this.expectedHashes[platform] === nextHash) {
+			isHashValid = true;
```
DimitarTachev: IMO, this is not a valid hash, we've already handled it. I belive this logic should be directly in the watch callback and if `currentHash` and `nextHash` are equal, we should not call `getUpdatedEmittedFiles` and `emmit` anything at all.

Fatme:
test cli-smoke cli-run

Issue #5132, opened 2019-11-14
livescync stops working with --no-hmr if changes are reverted


endarova:
**Environment**
 - CLI: 6.2.1
 - Cross-platform modules:
 - Android Runtime: 
 - iOS Runtime: 
 - Plugin(s):

**Describe the bug**
When run with --no-hmr flag after same changes which are synced if you revert the app to initial state this change is not synced.

**To Reproduce**
1. `tns create TestApp --ng` 
2. `tns run android --path TestApp  --no-hmr`
3. make some set of changes - they are synced
4. Revert the changes - the last change which applies the app's initial state wont be synced 
Log: 
```File change detected. Starting incremental webpack compilation...
Hash: 569f7f6a75c9f70ac865
Version: webpack 4.27.1
Time: 123ms
Built at: 11/14/2019 3:06:48 PM
    Asset      Size  Chunks             Chunk Names
bundle.js  40.9 KiB  bundle  [emitted]  bundle
 + 2 hidden assets
Entrypoint bundle = runtime.js vendor.js bundle.js
[../$$_lazy_route_resource lazy recursive] ../$$_lazy_route_resource lazy namespace object 160 bytes {bundle} [built]
[./app.css] 2.01 KiB {bundle} [built]
    + 541 hidden modules
Webpack compilation complete. Watching for file changes.
Webpack build done! ```
 

And nothing is synced

**Expected behavior**
All change should be synced

PR #5137, opened 2019-11-15
fix: fix livesync when `--no-hmr` option is provided and changes are reverted to the initial state of application

Fatme:
Livesync doesn't work when `tns run ios --no-hmr` command is executed, same changes are synced and the application is reverted to its initial state. On first webpack compilation CLI stores the initial hash of the compilation but the stored hash is updated only in `hmr` mode. On the other side, CLI has a logic to return when the stored hash is the same as the current hash. So, when the changes are reverted to the initial state of the application, the current hash is the same as the first reported hash. This way, CLI doesn't execute any livesync logic after the reported changes from webpack process.

Rel to: https://github.com/NativeScript/nativescript-cli/issues/5132

## PR Checklist

## What is the new behavior?


Fatme:
test cli-device

PR #5140, opened 2019-11-16
[issue] Marking mode message is probably wrong

NathanaelA:
Marking mode upgrade message is probably wrong.  :grinning:   I would guess by NativeScript 7.0; the iOS v8 engine would ship -- I would also **assume** that the iOS v8 engine would also require marking mode none.      So the **Android** part of the message should probably be removed.    If my assumptions are incorrect please feel free to close the PR.  :grinning:

## PR Checklist

## What is the new behavior?

### DIFF
--- a/lib/services/marking-mode-service.ts
+++ b/lib/services/marking-mode-service.ts
@@ -9,7 +9,7 @@ const enum MarkingMode {
 }
 
 const MARKING_MODE_PROP = "markingMode";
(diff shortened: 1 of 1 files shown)


DimitarTachev:
@NathanaelA,

Your assumption is correct, the iOS Runtime with V8 also requires marking mode `none`. The difference is that in the iOS Runtime, the marking mode is already hardcoded as `none` and it's not configurable even before 7.0.

The Android part of the message is there as the message is related to the `markingMode` configuration supported only by the Android runtime (the one in the `android` section in `app/package.json`).

Both of the messages are correct, we just want to be more clear that the Android Runtime won't support such configuration anymore.

Issue #5214, opened 2020-01-14
[IOS]Unexpected duplicate tasks


farfromrefug:
When i am in a working env dev with npm links i might end up with an XCode build error `Unexpected duplicate tasks`
It is due to native deps being added twice.
For example my plugin ```nativescript-gesturehandler``` comes with a framework.
In my current case i have it in my app package.json but also as a dependency of another plugin.
Consequently i have ```nativescript-gesturehandler``` in:
* node_modules/nativescript-gesturehandler
* node_modules/nativescript-chart/node_modules/nativescript-gesturehandler

Could you filter duplications when creating the XCode project and adding modules targets?

rosen-vladimirov:
Hey @farfromrefug ,
Thanks for the suggestions. I've created a PR that should resolve this issue.

rosen-vladimirov:
After latest fixes, CLI will detect if NativeScript plugin exists multiple times in node_modules. In case that's the case, CLI will check the versions of the plugin in each occurrence and will apply the following rules:
 - in case all versions are the same, CLI will use the native code from the version which is topmost in node_modules. You can reproduce this by executing:
```
tns create myApp --js
cd myApp
npm i --save-exact nativescript-ui-listview@8.0.1 nativescript-ui-core@4.0.0
cd node_modules/nativescript-ui-listview
npm i --production
cd ../../
tns prepare ios/android
```
This must produce warning that nativescript-ui-core exists in multiple locations.
 - in case there's difference in the versions, CLI will throw an error:
You can reproduce this by executing:
```
tns create myApp --js
cd myApp
npm i --save-exact nativescript-ui-listview@8.0.1 nativescript-ui-core@4.0.0 nativescript-ui-autocomplete@5.0.0
tns prepare ios/android
```
NOTE: the error message will guide you to delete package-lock.json (or other configuration file if it exists in your project). CLI checks if there's npm-shrinkrwrap.json, package-lock.json, yarn.lock and pnpm-lock.yaml.

farfromrefug:
@rosen-vladimirov awesome!  will test it and report any error.

rosen-vladimirov:
After some additional investigation, we've decided to implement the following:
1. For Android - analyze the dependencies tree and show debug messages (visible when you pass `--log debug` in case there are same versions of plugin in node_modules.
2. For iOS search for `.framework` in the NativeScript plugins.
  2.1 In case the same framework comes from different plugins - show error message
[...]

PR #5237, opened 2020-02-03
fix: improve check for duplicated NativeScript plugins

rosen-vladimirov:
Currently CLI fails when different versions of the same plugin are used. However, this introduces a lot of issues for Android, where the `nativescript-permissions` plugin is used by most of the other plugins. When they hardcode different version, CLI stops the build. However, this plugin is pure JS and does not have native part.
To resolve this, separate the checks for iOS and Android. For Android, the SBG may fail in case different versions of the plugin are detected. However, there's no easy way to determine this, so just make the checks and show debug messages.
For iOS the build will fail in case framework is linked multiple times from different locations, so group the plugins by frameworks and fail in case the same framework is listed multiple times by different plugins or by different versions of the same plugin

## PR Checklist

## What is the new behavior?

Related to issue https://github.com/NativeScript/nativescript-cli/issues/5214


rosen-vladimirov:
test cli-run cli-plugin

Issue #5105, opened 2019-10-30
App restarts when changing platform specific scss file from other platform


Fatme:
**Environment**
Provide version numbers for the following components (information can be retrieved by running `tns info` in your project folder or by inspecting the `package.json` of the project):
 - CLI: 6.2.0-rc
 - Cross-platform modules: 6.2.0-rc
 - Android Runtime:
 - iOS Runtime:
 - Plugin(s):

**Describe the bug**

The application is restarted when changing platform specific scss file from other platform. It shouldn't transfer any files in such a situation.

**To Reproduce**
1. `tns create ts_proj`  - choose ts tabs project 
2. `npm i tns-core-modules@rc --save --save-exact`
3. `npm i nativescript-dev-webpack@rc --save-dev --save-exact`
4. `./node_modules/.bin/update-ns-webpack --deps –configs` 
5.  `tns run android`

Change something in app.android.scss - the change is applied, no restart 

Change something in app.ios.scss - application restarts 

**Expected behavior**
The application shouldn't be restarted.

DimitarTachev:
The issue is also reproducible by saving any file without changes (e.g. just save the `main-view-model.js` in a Hello World js app and the CLI will send `bundle.js` and restart the application).

PR #5250, opened 2020-02-17
fix: external files are not livesynced

rosen-vladimirov:
In case you have external files set in the webpack.config.js (or resources copied with CopyWebpackPlugin), the `tns run ...` command does not LiveSync them to device. Webpack compilation is triggered, but as the emitted files are not hot updates, the hot update hash of the compilation is the same as the previous one. In this case CLI decides there's nothing to sync.
To fix this extend the check in webpack-compiler-service to be based not only on the comparison between previous and current hash, but on the emitted files as well - if there are files which are not hot updates, continue the execution even if the previous and current hash are the same. That's exactly the case with externals.

## PR Checklist

## What is the new behavior?

Fixes issue https://github.com/NativeScript/nativescript-cli/issues/5251
> NOTE: Current changes may cause regression for issues described in https://github.com/NativeScript/nativescript-cli/issues/5105 They should be checked as well.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,14 @@
 NativeScript CLI Changelog
 ================
 
+6.4.1 (2020, February 19)
+===
+
+### Fixed
+
+* [Fixed #5251](https://github.com/NativeScript/nativescript-cli/issues/5251): External files are not livesynced
+
+
 6.4.0 (2020, February 11)
 ===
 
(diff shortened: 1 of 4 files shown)


rosen-vladimirov:
test cli-smoke cli-run cli-templates cli-preview

rosen-vladimirov:
test cli-smoke cli-run cli-templates cli-preview

Issue #5252, opened 2020-02-19
Logs from platform specific files point to incorrect file


rosen-vladimirov:
**Environment**
Provide version numbers for the following components (information can be retrieved by running `tns info` in your project folder or by inspecting the `package.json` of the project):
 - CLI: 6.4.0
 - Cross-platform modules: Not applicable
 - Android Runtime: Not applicable
 - iOS Runtime: Not applicable
 - Plugin(s): Not applicable

**Describe the bug**
CLI parses device logs and replaces the file path locations in them with local paths to the project. However, in case the filepath is from platform specific file, the path in the logs points to a non-existent file instead of the platform specific one.

**To Reproduce**
```
tns create myApp --js
cd myApp
mv app/main-page.js app/main-page.ios.js
# edit app/main-page.ios.js and add `console.trace("MAIN PAGE"); in the `onNavigatingTo` method.
tns run ios
```
Check the logs:
```
CONSOLE TRACE file:///app/main-page.js:21:0: MAIN PAGE
onNavigatingTo(file:///app/main-page.js:21:0)
```
However, we do not have such file, it should point to main-page.ios.js file.

### Another way to reproduce the issue
```
tns create myApp --js
cd myApp
# edit main-view-model.js and in the onTap handler add throw new Error("some error");
tns run ios
```
Check the logs and the callstack - the callstack contains some lines about button:
```
file:///node_modules/@nativescript/core/ui/button/button.js:216:0
```
However, there's no button.js at this location, it should be button.ios.js

**Expected behavior**
File paths should be correct.

**Sample project**

**Additional context**
The issue is not reproducible with Angular project.

Issue #5236, opened 2020-02-03
custom file prefix in error stack


farfromrefug:
When you "translate" an error stack line with source maps [here](https://github.com/NativeScript/nativescript-cli/blob/4fa7e65071bb527a3b3784b1be79dd3d019667e4/lib/services/log-source-map-service.ts#L86) you add a `file:///` prefix to the file path.
This makes it impossible to use "CTRL+Click" to open the file at line/column which would be incredibly useful.
I dont know why you do this but i suppose it is to get this to work on another IDE.

Can you make this customizable ?

rosen-vladimirov:
Hey @farfromrefug ,
Thanks for pointing this UX problem, I've created a PR to replace the `file:///` with `file: ` when CLI prints the logs. 
We've tried to persist the same output coming from the device itself, that's why we've kept the `file:///` until now. However, we've discussed your suggestion and we've decided to implement it as it will be really useful for everyone. Thanks for the idea.

farfromrefug:
@rosen-vladimirov thanks! do you know if `file:` works with vscode? if not would be good for it to be "customizable"

rosen-vladimirov:
Hey @farfromrefug ,
I've used `file: ` (note the space at the end), just to keep the output as similar as possible to the current one.
I've also tested it in VSCode's terminal and I was able to click on the link and it navigated to the correct file and even the exact line.

dtopuzov:
Available in `nativescript@rc`, will be shipped with `nativescript@64.1`.

farfromrefug:
@rosen-vladimirov Seems to be good but not completely:
<img width="971" alt="Screen Shot 2020-02-24 at 14 32 54" src="https://user-images.githubusercontent.com/655344/75156574-b0797880-5712-11ea-9fe6-e0c69178e02e.png">
I don't know why the `Object.p ...` is still using `file:///`

farfromrefug:
@rosen-vladimirov found another issue
<img width="1243" alt="Screen Shot 2020-02-25 at 14 16 17" src="https://user-images.githubusercontent.com/655344/75250846-98b8f780-57d9-11ea-96d4-a0c9076fbb5b.png">
the file in webpack is `../../src/opencv.android.ts`. Though the cli still adds `app/` in front so vscode cant pick up the link

PR #5253, opened 2020-02-19
fix: device logs parsing

rosen-vladimirov:
### fix: file paths shown for device logs should be clickable
Currently CLI tries to persist the formating of logs as close as possible to the real logs coming from device.
The logs from the applications contain `file:///<filePath>`. CLI replaces the `<filePath>` with the real local file path. However, due to the `file:///` prefix, the link is not clickable. For example, in case you use VSCode's terminal, you should be able to use Ctrl + Click on the link and this will lead you to the exact line where the stack trace points to.
To resolve this issue, replace the `file:///` with just `file: `.

### fix: paths to platform specific files are not correct in logs
When CLI parses the device logs, it tries to replace the file location with the real file path location. However, this does not work for platform specific files (in non-Angular projects) as we are using a virtual file system in webpack, which "lies" the file `button.js` exists, while in fact the file is `button.<platform>.js`. So in the source maps, the location of the file is `button.js`.
To resolve this, check if there's a platform specific file and use it instead. Cache the original file location, so we'll not execute multiple file operations for the same file.

## PR Checklist

## What is the new behavior?

Fixes issues https://github.com/NativeScript/nativescript-cli/issues/5252 and https://github.com/NativeScript/nativescript-cli/issues/5236

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,9 @@ NativeScript CLI Changelog
 
 ### Fixed
 
+* [Fixed #5236](https://github.com/NativeScript/nativescript-cli/issues/5236): File paths from device logs are not clickable
 * [Fixed #5251](https://github.com/NativeScript/nativescript-cli/issues/5251): External files are not livesynced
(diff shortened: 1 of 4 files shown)


rosen-vladimirov:
test cli-smoke cli-run

Issue #5264, opened 2020-02-20
tns package-manager --help fails with error


rosen-vladimirov:
**Environment**
Provide version numbers for the following components (information can be retrieved by running `tns info` in your project folder or by inspecting the `package.json` of the project):
 - CLI: 
 - Cross-platform modules:
 - Android Runtime:
 - iOS Runtime:
 - Plugin(s):

**Describe the bug**
When you try `tns package-manager --help` you receive an error:
```
/Users/vladimirov/.nvm/versions/node/v10.18.1/bin/tns package-manager --help
Unknown command 'package-manager'. Try '$ tns help' for a full list of supported commands.
```
It is not a big deal, except for the case when CLI itself tells you to run this command. In case you try `tns package-manager` you receive:
```
The input is not valid sub-command for 'package-manager' command.
Run 'tns package-manager --help' for more information.
```
**To Reproduce**
`tns package-manager --help`

**Expected behavior**
Print the help of tns package-manager get command or some other useful information.

Issue #5263, opened 2020-02-20
`tns package-manager` docs does not list pnpm as supported value


rosen-vladimirov:
**Environment**
Provide version numbers for the following components (information can be retrieved by running `tns info` in your project folder or by inspecting the `package.json` of the project):
 - CLI: 6.4.0
 - Cross-platform modules: Not applicable
 - Android Runtime: Not applicable
 - iOS Runtime: Not applicable
 - Plugin(s): Not applicable

**Describe the bug**
Help of `tns package-manager get` and `tns package-manager set` commands does not list the pnpm as supported value.

**To Reproduce**
`tns package-manger set --help` and `tns package-manager get --help` - no `pnpm` is described there.

**Expected behavior**
`pnpm` should be listed as supported value.

Issue #5262, opened 2020-02-20
`tns package-manager` fails with error


rosen-vladimirov:
**Environment**
Provide version numbers for the following components (information can be retrieved by running `tns info` in your project folder or by inspecting the `package.json` of the project):
 - CLI: 6.4.0
 - Cross-platform modules: Not applicable
 - Android Runtime: Not applicable
 - iOS Runtime: Not applicable
 - Plugin(s): Not applicable

**Describe the bug**
`tns package-manager` fails with error:
The input is not valid sub-command for 'package-manager' command.
Run 'tns package-manager --help' for more information.

**To Reproduce**
`tns package-manager` 

**Expected behavior**
Print the currently used package-manager.

Issue #5260, opened 2020-02-20
`tns package-manager set invalid_value` does not say pnpm is supported


rosen-vladimirov:
**Environment**
Provide version numbers for the following components (information can be retrieved by running `tns info` in your project folder or by inspecting the `package.json` of the project):
 - CLI: 6.4.0
 - Cross-platform modules: Not applicable
 - Android Runtime: Not applicable
 - iOS Runtime: Not applicable
 - Plugin(s): Not applicable

**Describe the bug**
When calling `tns package-manager set ` with invalid value, the command says only npm and yarn are supported, while in fact `pnpm` is also supported.

**To Reproduce**
```
tns package-manager set invalid_value
invalid_value is not a valid package manager. Only yarn or npm are supported.
```
**Expected behavior**
The error message should list all supported options.

**Sample project**

**Additional context**

PR #5265, opened 2020-02-20
fix: package-manager and pnpm support fixes

rosen-vladimirov:
This PR contains multiple fixes for package-manager issues and pnpm particularly:

### fix: pnpm support should work on macOS and Linux 
Currently we are unable to build applications on macOS and Linux when pnpm is set as package manager. The problem seems to be caused by the way pnpm produces files in node_modules and the way webpack expects them. However, the issue is resolved in case we do not pass `--preserve-symlinks` option to Node.js when starting the webpack process. According to docs (https://pnpm.js.org/en/faq) the preserve-symlinks does not work very well and pnpm does not require it.
So, to resolve the issue, skip the `--preserve-symlinks` option when pnpm is used. To achieve this, introduce a new method in the packageManager that returns the package manager used for the current process.
Fixes issue https://github.com/NativeScript/nativescript-cli/issues/5259

### fix: CLI should print help of commands with dash in the name
When trying to print the help of a default hierarchical command, CLI expects the format of the command name to be `<letters>|*<something else>`. However, some commands have dash in the name, like `tns package-manager get`. Fix the expectation of the help service.
Fixes issue https://github.com/NativeScript/nativescript-cli/issues/5264 

### fix: help of package-manager commands does not list pnpm
The markdown files of package-manager commands does not list pnpm as supported value. Add it where is required and add help file for package-manager command (without passig get to it).
Fixes issue https://github.com/NativeScript/nativescript-cli/issues/5263

### fix: register package-manager get as default hierarchical command
Register the package-manager get as default hierarchical command, so when you write `tns package-manager` we'll print the currently selected package-manager.
Fixes issue https://github.com/NativeScript/nativescript-cli/issues/5262

### fix: print information message when package-manager is set
Currently when calling `tns package-manager set <valid value>` you do not see any output or information if the operation is successful. Add such message and clean the code from hardcoded strings. 
Fixes issues https://github.com/NativeScript/nativescript-cli/issues/5260  https://github.com/NativeScript/nativescript-cli/issues/5261

## PR Checklist

## What is the new behavior?


rosen-vladimirov:
test cli-smoke cli-misc

lib/services/webpack/webpack-compiler-service.ts
```diff
@@ -155,6 +156,15 @@ export class WebpackCompilerService extends EventEmitter implements IWebpackComp
 		}
 	}
 
+	private async shouldUsePreserveSymlinksOption(projectData: IProjectData): Promise<boolean> {
```
Fatme: Do you need to pass `projectData` here? 😸

rosen-vladimirov: Nice catch.

PR #5296, opened 2020-04-14
feat(options): use the 'nsconfig.json' file to set tns options

bgrand-ch:
## PR Checklist

**Tns options** are to be written each time to build, run, deploy, debug and test an Android and iOS application.

## What is the new behavior?

In addition to the current behavior, all tns options can be defined in the **nsconfig.json** file.

In the **nsconfig.json** file, you need to add the `tnsOptions` key with `android` and `ios` subkeys.

Option names are unchanged, they are those of the documentation of the CLI.

#### nsconfig.json

```json
{
    "tnsOptions": {
        "android": {
            "copyTo": "./_build/android",
            "keyStorePath": "./keys/android.keystore",
            "keyStoreAlias": "my_ns_project",
            "release": true,
            "aab": true,
            "env": {
                "uglify": true
            }
        }
    }
}
```

#### CLI

```shell
tns build android --nsconfig --key-store-password "my_pwd" --key-store-alias-password "my_pwd"
```

([see the original idea](https://github.com/elvticc/nativescript-build-release))

### Fixes

* [./lib/services/build-artefacts-service.ts](https://github.com/NativeScript/nativescript-cli/pull/5296/commits/c582bbe157824fed2a8dddce1c92c30cbcfded4d)
Create up to the last folder in the path with the **copyTo** option.

### DIFF
--- a/docs/man_pages/project/testing/build-android.md
+++ b/docs/man_pages/project/testing/build-android.md
@@ -17,6 +17,7 @@ General | `$ tns build android [--compileSdk <API Level>] [--key-store-path <Fil
 
 ### Options
 
+* `--nsconfig` - If set, all the options below can be defined in the [nsconfig.json file](https://docs.nativescript.org/core-concepts/project-structure-app#the-nsconfigjson-file) at the root of your project. In the `nsconfig.json` file, you need to add the `tnsOptions` key with the `android` subkey.
 * `--compileSdk` - Sets the Android SDK that will be used to build the project. `<API Level>` is a valid Android API level. For example: 28, 29. The minimum supported SDK is 28. <% if(isHtml) { %> For a complete list of the Android API levels and their corresponding Android versions, click [here](http://developer.android.com/guide/topics/manifest/uses-sdk-element.html#platform).<% } %>
 * `--clean` - If set, forces the complete rebuild of the native application.
 * `--release` - If set, produces a release build. Otherwise, produces a debug build. When set, you must also specify the `--key-store-*` options.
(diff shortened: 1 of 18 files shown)


elvticc:
@DimitarTachev @rosen-vladimirov Hello, I am at your disposal for any changes. Have a nice day 😊

rigor789:
Hi @elvticc - I have wanted something like this for a long time! With the recent introduction of `nativescript.config.ts` I think we can go a step further, and allow setting many of CLI configuration there, including the options.

```ts
import { NativeScriptConfig } from '@nativescript/core'

export default {
  id: 'org.nativescript.app',
  // ...
  cli: {
    packageManager: 'yarn',
    android: {
      copyTo: './_build/android'
    }
  },
} as NativeScriptConfig
```
The syntax/api is subject to change, merely just a quick example of what I imagined.

elvticc:
Hi @rigor789, I will update this to include the new `nativescript.config.ts` file and preserve backward compatibility with the `nsconfig.json` file.  I will try with all the options 👍 I'll be back soon.

rigor789:
@elvticc the CLI should automatically fall back to `nsconfig` when there is no new config found. The logic has been extracted to `project-config-service` which you can inject wherever and read values, or the whole config. 

I think in `options.ts` we can inject/resolve the config service:

```ts
const projectConfig = injector.resolve('projectConfigService')
// add guards if it's not set/found etc.
const cliOptions = projectConfig.getValue('cli') // I think this should work, and return the object

// do any processing of the options (can probably be enclosed in a function within `options.ts`)
// just an example, doesn't have to be called this - or even called this way
this.argv = this.processConfigOptions(cliOptions)

// probably better to just have a single line that calls a function that 
// does all the config fetching, manipulation and setting argv
```

Other than that, I think we want these to be used by default - and provide a flag to opt-out? `--skip-config` (don't like that name, but can't think of anything at the moment)
And

PR #5319, opened 2020-05-30
fix(chrome): new debug url with latest chrome 83+ versions

NathanWalker:
closes https://github.com/NativeScript/NativeScript/issues/8605

### DIFF
--- a/lib/controllers/debug-controller.ts
+++ b/lib/controllers/debug-controller.ts
@@ -157,7 +157,9 @@ export class DebugController extends EventEmitter implements IDebugController {
 				this.emit(DEBUGGER_ATTACHED_EVENT_NAME, debugInformation);
 			}
 
-			this.$logger.info(`To start debugging, open the following URL in Chrome:${EOL}${debugInformation.url}${EOL}`.cyan);
+			this.$logger.info(`To start debugging, open the following URL in Chrome:${EOL}${debugInformation.url}${EOL}`.green);
+
+			this.$logger.info(`If you're using an older Chrome version 82 and below, use this instead:${EOL}${debugInformation.legacyUrl}${EOL}`.cyan);
 		}
 
 		return debugInformation;
--- a/lib/controllers/debug-controller.ts
+++ b/lib/controllers/debug-controller.ts
@@ -158,8 +158,6 @@ export class DebugController extends EventEmitter implements IDebugController {
 			}
 
 			this.$logger.info(`To start debugging, open the following URL in Chrome:${EOL}${debugInformation.url}${EOL}`.green);
-
-			this.$logger.info(`If you're using an older Chrome version 82 and below, use this instead:${EOL}${debugInformation.legacyUrl}${EOL}`.cyan);
 		}
 
 		return debugInformation;
--- a/lib/declarations.d.ts
+++ b/lib/declarations.d.ts
@@ -494,7 +494,6 @@ interface IGenerateOptions {
 
 interface IDebugInformation extends IPort, Mobile.IDeviceIdentifier {
 	url: string;
-	legacyUrl?: string;
 }
 
 interface IPort {
--- a/lib/declarations.d.ts
+++ b/lib/declarations.d.ts
@@ -494,6 +494,7 @@ interface IGenerateOptions {
 
 interface IDebugInformation extends IPort, Mobile.IDeviceIdentifier {
 	url: string;
+	legacyUrl?: string;
 }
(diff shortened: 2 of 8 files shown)


lib/services/debug-service-base.ts
```diff
@@ -28,9 +28,9 @@ export abstract class DebugServiceBase extends EventEmitter implements IDeviceDe
 				return true;
 			}
 		};
-	}
+  }
 
-	protected getChromeDebugUrl(debugOptions: IDebugOptions, port: number): string {
+	protected getChromeDebugUrl(debugOptions: IDebugOptions, port: number, legacy?: boolean): string {
```
NathanaelA, line 33: I think adding `Legacy` here and passing it through the engine adds a lot more code which isn't needed as this url should be removed in the future version.   I would just fix it here to return the proper url with `devtools://`

lib/controllers/debug-controller.ts
```diff
@@ -157,7 +157,9 @@ export class DebugController extends EventEmitter implements IDebugController {
 				this.emit(DEBUGGER_ATTACHED_EVENT_NAME, debugInformation);
 			}
 
-			this.$logger.info(`To start debugging, open the following URL in Chrome:${EOL}${debugInformation.url}${EOL}`.cyan);
+			this.$logger.info(`To start debugging, open the following URL in Chrome:${EOL}${debugInformation.url}${EOL}`.green);
+
+			this.$logger.info(`If you're using an older Chrome version 82 and below, use this instead:${EOL}${debugInformation.legacyUrl}${EOL}`.cyan);
```
NathanaelA, line 162: I love using separate colors!   Awesome,  I think here I would just do :    `use this instead of: chrome-{$EOL}`....   And just hard code the prefix `chrome-` here....   As in a few versions from now we will just eliminate this message once chrome 83 is the standard...

lib/declarations.d.ts
```diff
@@ -494,6 +494,7 @@ interface IGenerateOptions {
 
 interface IDebugInformation extends IPort, Mobile.IDeviceIdentifier {
 	url: string;
+	legacyUrl?: string;
```
NathanaelA, line 497: Not needed.

lib/definitions/debug.d.ts
```diff
@@ -116,6 +116,7 @@ interface IDeviceDebugService extends IPlatform, NodeJS.EventEmitter {
 
 interface IDebugResultInfo {
 	debugUrl: string;
+	legacyDebugUrl?: string;
```
NathanaelA, line 119: Not needed.

lib/services/android-device-debug-service.ts
```diff
@@ -99,6 +99,7 @@ export class AndroidDeviceDebugService extends DebugServiceBase implements IDevi
 		await this.printDebugPort(this.deviceIdentifier, debugPort);
 
 		result.debugUrl = this.getChromeDebugUrl(debugOptions, debugPort);
+		result.legacyDebugUrl = this.getChromeDebugUrl(debugOptions, debugPort, true);
```
NathanaelA, line 102: Not Needed.

lib/services/ios-device-debug-service.ts
```diff
@@ -32,12 +32,9 @@ export class IOSDeviceDebugService extends DebugServiceBase implements IDeviceDe
 
 	@performanceLog()
 	public async debug(debugData: IDebugData, debugOptions: IDebugOptions): Promise<IDebugResultInfo> {
-		const result: IDebugResultInfo = { debugUrl: null };
 		await this.validateOptions(debugOptions);
 
-		result.debugUrl = await this.wireDebuggerClient(debugData, debugOptions);
-
-		return result;
+		return await this.wireDebuggerClient(debugData, debugOptions);
```
NathanaelA, line 37: I love the simplification here.

NathanaelA:
Based on some additional research we discovered that the reasons for the url change from "chrome-devtools" to "devtools" occurred on the Chrome path: 

https://github.com/chromium/chromium/commit/28b21a67f09a13c465b73776aecd9995d8a07737

Which occurred mid-2019 in Chrome 76.    So all modern chrome based engines should understand the replacement "devtools://" url.   

Chrome 83, in this PR, they dropped the old obsolete "chrome-devtools://" url. 
https://chromium.googlesource.com/chromium/src/+/6700d12448f76712c62a6d2372a95b97a26d4779  as part of their cleanup.

PR #5347, opened 2020-08-02
feat(cli): {N} 7 package scoping, nativescript.config and nsc

NathanWalker:


NathanaelA:
@NathanWalker - Can we do `ns` rather than `nsc`?

NathanWalker:
@NathanaelA we could go with `ns` - wasn't sure if collided with existing OS level binary but we'll add alias to that as well and can start using it.

NathanWalker:
closing in favor of https://github.com/NativeScript/nativescript-cli/pull/5350

PR #5376, opened 2020-09-06
use androidx libs to build plugins

farfromrefug:
## PR Checklist

## What is the new behavior?

### DIFF
--- a/vendor/gradle-plugin/build.gradle
+++ b/vendor/gradle-plugin/build.gradle
@@ -46,20 +46,8 @@ android {
 }
 
 dependencies {
-    def supportVer = "28.0.0"
-    if (project.hasProperty("supportVersion")) {
-        supportVer = supportVersion
-    }
-    compileOnly "com.android.support:support-v4:$supportVer"
-    compileOnly "com.android.support:appcompat-v7:$supportVer"
-
-    configurations.all {
-       resolutionStrategy.eachDependency { DependencyResolveDetails details ->
-           if (details.requested.group == "com.android.support" && !details.requested.name.startsWith("multidex")) {
-               details.useVersion supportVer
-           }
-       }
-    }
+    def androidXAppCompatVersion =  project.hasProperty("androidXAppCompatVersion") ? project.androidXAppCompatVersion : "1.2.0"
+	compile "androidx.appcompat:appcompat:$androidXAppCompatVersion"
 }
 
 def getAppResourcesPath() {


vendor/gradle-plugin/build.gradle
```diff
@@ -46,20 +46,8 @@ android {
 }
 
 dependencies {
-    def supportVer = "28.0.0"
-    if (project.hasProperty("supportVersion")) {
-        supportVer = supportVersion
-    }
-    compileOnly "com.android.support:support-v4:$supportVer"
-    compileOnly "com.android.support:appcompat-v7:$supportVer"
```
rigor789, line 54: Does 
```gradle
compile "androidx.appcompat:appcompat:$androidXAppCompatVersion"
```
replace both of these?

vendor/gradle-plugin/build.gradle
```diff
@@ -46,20 +46,8 @@ android {
...
-
-    configurations.all {
-       resolutionStrategy.eachDependency { DependencyResolveDetails details ->
-           if (details.requested.group == "com.android.support" && !details.requested.name.startsWith("multidex")) {
-               details.useVersion supportVer
-           }
-       }
-    }
```
rigor789, line 62: Should we keep this for  `androidX` too? In case a dep specifies a different androidX version, we likely want to use the same version across the board?

farfromrefug: The thing is all androidx packages have different versioning rules. That s why i removed it

rigor789: Ok, if that has no side-effects then that should be fine. I'm a bit worried it may lead to larger apk/aab's if we include multiple versions of the same lib. Would it make more sense to use the most recent of them all? (the one coming from the main gradle file)

farfromrefug: I dont think so. You cant use multiple versions in the apk (i dont think you can). Gradle will do the job

vendor/gradle-plugin/build.gradle
farfromrefug: I dont think the first one is needed anymore

farfromrefug:
@rigor789 any news on this ?

rigor789:
@farfromrefug I don't know the effect of the changes, perhaps if @NathanaelA or @triniwiz could take a look we could get this merged.

vendor/gradle-plugin/build.gradle
triniwiz: most of support-v4 was replaced by `androidx.core:core`

janoshrubos:
@triniwiz do you think this change would cause any side effects?

triniwiz:
We can remove it now

rigor789:
@triniwiz so good to merge as-is, or should we use `androidx.core:core`?

triniwiz:
The previous used appcompat + support-v4 which now maps to following , but I feel we can get away with using appcompat.

```gradle
androidx.appcompat:appcompat
androidx.legacy:legacy-support-v4
```

PR #5377, opened 2020-09-06
allow plugins to use widgets lib

farfromrefug:
I have plugins that use classes from the nativescript widget libs.
Now i want to publish the plugin with the java sources directly.
If i do that i need to add `widget-release` to the dependencies.

Now i could do that in the plugin include.gradle, it will make plugin build to work however then app building wont work because the widgets lib will be included twice

So the best solution is automatically add it here.
I am not happy about the path thing, however it will work in all cases.

## PR Checklist

## What is the new behavior?

### DIFF
--- a/vendor/gradle-plugin/build.gradle
+++ b/vendor/gradle-plugin/build.gradle
@@ -60,6 +60,7 @@ dependencies {
            }
        }
     }
+    implementation files('../../../node_modules/@nativescript/core/platforms/android/widgets-release.aar')
 }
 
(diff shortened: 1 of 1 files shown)


farfromrefug:
Ok so actually find a way to declare the dependency without needing this.
Anyway that might get useful.
Let me know if you still want that PR

janoshrubos:
@triniwiz do you see anything wrong with this change? (i.e. increased apk size, etc..)

PR #5378, opened 2020-09-07
feat: migrate tsconfig.json for ns7 projects

### LINKED ISSUE #5364: ns migrate does not update tsconfig file
**Environment**
 - CLI: 7.0.4

**Describe the bug**
I just run the migrate over a demo angular project of a plugin and no changes have been made in the tsconfig file of the app. 

**To Reproduce**
Run `ns migrate` for a pre 7.0 project

**Expected behavior**
The migration should update the tsconfig to use the new required settings (like module: ESNext, target: es2015, etc.)

**Sample project**

**Additional context**

rigor789:
fixes #5364
Makes `ns migrate` update the `tsconfig.json` (if it exists).


lib/controllers/migrate-controller.ts
```diff
@@ -1182,6 +1192,41 @@ export class MigrateController
...
+
+			configContents.compilerOptions.lib = [
+				...new Set([...(configContents.compilerOptions.lib || []), "es2017"]),
+			];
+			configContents.compilerOptions.plugins = [
+				...new Set([
+					...(configContents.compilerOptions.plugins || []),
+					{
+						transform:
+							"@nativescript/webpack/transformers/ns-transform-native-classes",
```
NathanWalker, line 1215: This is only needed for plugins since projects get the transformer automatically through webpack.config.

NathanWalker (changes requested):

NathanWalker (approved):

PR #5382, opened 2020-09-08
fix: chmod gradle executable before building android plugins

### LINKED ISSUE #1836: Error in tns build android: 'spawn gradle.bat ENOENT'
I'm able to run NativeScript simple example. However, I am getting 'spawn gradle.bat ENOENT' when I'm trying to build a project where multidex is enabled. Here is my build log:
http://pastebin.com/mf6ZHVA0

Thanks,

### LINKED ISSUE #5368: Since NS7 update, some plugins fail building with "Error: spawn ./gradlew EACCES"
I personally haven't tested it on Windows, so there might not be any problems on that platform because the build scripts are handled separately.

Since the update to NS7 a few days ago, there have been multiple people on the NSCommunity Slack reporting this. I personally have encountered the problem in relation to `nativescript-socket.io`, but it's not specific to it, as it actually boils down to an issue in the CLI.

The problem is two-fold: first, the EACCES error is caused by the `gradlew` file not being being marked as executable. Once this is fixed, errors relating to the carriage return character start popping up. The issue now is that the `gradlew` file uses windows style newline (`\r\n`) instead of *nix style newline (`\n`). I've implemented a quick and dirty fix, but this only solves the symptoms, it doesn't solve the problem at the root cause as I'm not familiar enough with NS to know which part of the tooling generates the `gradlew` file when building plugins.

The temporary fix involves editing a file in the NS CLI: 
1. in the global node_modules folder (unless NS is locally installed under the project's devDependencies), find `nativescript/lib/services/android-plugin-build-service.js`
2. anywhere at the top of the file (maybe between lines 19-24 for consistency's sake) add `const fs = require("fs");`
3. at line 413 (formerly 412 before the above addition) add:
```javascript
if (!this.$hostInfo.isWindows) {
    const gradleFile = path.join(pluginBuildSettings.pluginDir, gradlew);
    let content = fs.readFileSync(gradleFile, {encoding:'utf8', flag:'r'});
    content = content.replace(/\r/g, '');
    fs.writeFileSync(gradleFile, content);
    fs.chmodSync(gradleFile, 0o755, (err) => {
        if (err) {
            console.log("Error changing setting gradlew as executable");
            throw err;
        }
    });
}
```

rigor789:
## PR Checklist

The CLI relies on the files being ececutable

## What is the new behavior?

The CLI makes the file executable before execution.

Fixes #5368

### DIFF
--- a/lib/services/android-plugin-build-service.ts
+++ b/lib/services/android-plugin-build-service.ts
@@ -710,6 +710,19 @@ export class AndroidPluginBuildService implements IAndroidPluginBuildService {
 		}
 
 		try {
+			if (!this.$hostInfo.isWindows) {
+				// make sure gradlew is executable
+				this.$fs.chmod(
+					path.join(pluginBuildSettings.pluginDir, gradlew),
+					"0755"
+				);
+			}
+		} catch (error) {
+			this.$logger.trace("Failed to chmod gradlew. Error is: ", error);
+			// ignore - worst case it fails further down
+		}
+
+		try {
 			await this.$childProcess.spawnFromEvent(gradlew, localArgs, "close", {
 				cwd: pluginBuildSettings.pluginDir,
 				stdio: "inherit",


lib/services/android-plugin-build-service.ts
```diff
@@ -709,6 +709,19 @@ export class AndroidPluginBuildService implements IAndroidPluginBuildService {
...
+		try {
+			if (!this.$hostInfo.isWindows) {
+				// make sure gradlew is executable
+				this.$fs.chmod(
+					path.join(pluginBuildSettings.pluginDir, gradlew),
+					"0755"
+				);
+			}
+		} catch (error) {
+			this.$logger.trace("Failed to chmod gradlew. Error is: ", error);
```
Codex-, line 721: I assume trace logs are not piped to stdout by default, having this as a trace log would be problematic in the case where `gradlew` is not executable, and the call to `chmod` fails.

imo at least make this a `warn` log if not an `error` log

lib/services/android-plugin-build-service.ts
```diff
@@ -709,6 +709,19 @@ export class AndroidPluginBuildService implements IAndroidPluginBuildService {
 			localArgs.push("--quiet");
 		}
 
+		try {
+			if (!this.$hostInfo.isWindows) {
+				// make sure gradlew is executable
+				this.$fs.chmod(
```
Codex-, line 715: imo we shouldn't just impose a specific permission on here without first at least checking whether or not the file in question is executable. I'd propose that we politely check if the file has an executable permission already before performing a chmod call:
```ts
if (!this.$hostInfo.isWindows) {
  const gradlewPath = path.join(pluginBuildSettings.pluginDir, gradlew);
  let gradlewCanExecute = false;
  try {
    fs.accessSync(gradlewPath, fs.constants.X_OK);
    this.$logger.trace('gradlew: Has executable permission');
    gradlewCanExecute = true;
  } catch (err) {
    this.$logger.trace('gradlew: Does not have executable permission');
    gradlewCanExecute = false;
  }

  if (!gradlewCanExecute) {
    try {
      // Attempt to make gradlew executable.
      this.$fs.chmodSync(gradlewPath, '0755');
    } catch (error) {
      this.$logger.warn('gradlew: Failed to chmod');
      this.$logger.trace(
        `gradlew: chmod error ${
          error && error.message ? error.message : error
        }`,
      );
    }
  }
}
```

FestplattenSchnitzel:
Hi, are there any plans to merge this in the (near) future?

rigor789:
> 
> 
> Hi, are there any plans to merge this in the (near) future?
> Kind Regards

Should no longer be required.

PR #5478, opened 2021-01-19
perf: initialize project data if needed

### LINKED ISSUE #5398: Optimize ProjectDataService.getProjectData()
**Is your feature request related to a problem? Please describe.**
 
While developing a before-prepare hook that should adjust some CI aspects (use different bundle IDs, setup dynamic code signing. etc.) I wanted to modify some content on the injected `$projectData` object from inside the hook. 

However, my changes applied in the hook were all gone after the hook was done. In the end, the CLI worked on the same `ProjectData` object instance but all modified data was wiped.

I debugged the CLI and found that `ProjectDataService.getProjectData()` is called ridicously often by the CLI. While that still might ok, the internal caching mechanism itself is just not used at all. This leads to unnecessary re-initialization of the projectData over and over again and the cache has no effect at all. 

Relevant code snippet https://github.com/NativeScript/nativescript-cli/blob/686135ca8a11ad224873864a2c0c254f79c4b8ce/lib/services/project-data-service.ts#L117-L125

**Describe the solution you'd like**

The `getProjectData` is missing a simple check if the project data has been initialized already. If not, initialize it, otherwise return the cached value. 

I monkey-patched this locally and it works for me. Not sure of any side-effects, though. It goes like this:

```
public getProjectData(projectDir: string): IProjectData {
    projectDir = projectDir || this.defaultProjectDir;
    this.projectDataCache[projectDir] =
        this.projectDataCache[projectDir] ||
        this.$injector.resolve<IProjectData>(ProjectData);
    if (!this.projectDataCache[projectDir].initialized) {
        this.projectDataCache[projectDir].initializeProjectData(projectDir);
        // either have initializeProjectData taking care of setting the flag or set it here
        // (or use any of the existing properties that get initialized in initializeProjectData...)
        this.projectDataCache[projectDir].initialized = true;
    }
    return this.projectDataCache[projectDir];
}
```

janoshrubos:
## PR Checklist

`projectData` gets re-initialized over and over again and the cache has no effect.

## What is the new behavior?
Initialize `projectData` only if it hasn't been initialized yet.

Fixes #5398

### DIFF
--- a/lib/services/project-data-service.ts
+++ b/lib/services/project-data-service.ts
@@ -119,8 +119,10 @@ export class ProjectDataService implements IProjectDataService {
 		this.projectDataCache[projectDir] =
 			this.projectDataCache[projectDir] ||
 			this.$injector.resolve<IProjectData>(ProjectData);
-		this.projectDataCache[projectDir].initializeProjectData(projectDir);
-
+		if (!this.projectDataCache[projectDir].initialized) {
+			this.projectDataCache[projectDir].initializeProjectData(projectDir);
+			this.projectDataCache[projectDir].initialized = true;
+		}
 		return this.projectDataCache[projectDir];
 	}
 
--- a/lib/services/project-data-service.ts
+++ b/lib/services/project-data-service.ts
@@ -119,10 +119,7 @@ export class ProjectDataService implements IProjectDataService {
 		this.projectDataCache[projectDir] =
 			this.projectDataCache[projectDir] ||
 			this.$injector.resolve<IProjectData>(ProjectData);
-		if (!this.projectDataCache[projectDir].initialized) {
-			this.projectDataCache[projectDir].initializeProjectData(projectDir);
-			this.projectDataCache[projectDir].initialized = true;
-		}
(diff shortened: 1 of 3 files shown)


rigor789 (approved):

lib/services/project-data-service.ts
```diff
@@ -119,8 +119,10 @@ export class ProjectDataService implements IProjectDataService {
 		this.projectDataCache[projectDir] =
 			this.projectDataCache[projectDir] ||
 			this.$injector.resolve<IProjectData>(ProjectData);
-		this.projectDataCache[projectDir].initializeProjectData(projectDir);
-
+		if (!this.projectDataCache[projectDir].initialized) {
+			this.projectDataCache[projectDir].initializeProjectData(projectDir);
+			this.projectDataCache[projectDir].initialized = true;
+		}
```
rigor789, line 125: I think we should move this check into `initializeProjectData()` in `lib/project-data.ts` - that way the callers don't need to do any checking - just call `initializeProjectData()`.

rigor789 (changes requested):

rigor789 (approved):

PR #5524, opened 2021-05-07
feat: allow to ignore native dependencies

farfromrefug:
Along with that we simply need to have a cli parameter and an option in nativescript.config
 @rigor789 could you help with the cli params? not sure where to do that.
 
 Also i think we could even add ignored dependencies through the webpack config right? I dont remember how but we can modify the project data from there?
 
 With this people can simply make it so that some deps are only used in dev (like vue-dev-tools and socketio) while not being present in production.
 
 You can also make different releases for Google Play and Fdroid (what i am going for) where one would use non open source libs and the other would not (flavors on android)

### DIFF
--- a/lib/controllers/prepare-controller.ts
+++ b/lib/controllers/prepare-controller.ts
@@ -375,7 +375,7 @@ export class PrepareController extends EventEmitter {
 		projectData: IProjectData
 	): Promise<string[]> {
 		const dependencies = this.$nodeModulesDependenciesBuilder
-			.getProductionDependencies(projectData.projectDir)
+			.getProductionDependencies(projectData.projectDir, projectData.dependenciesToIgnore)
 			.filter((dep) => dep.nativescript);
 		const pluginsNativeDirectories = dependencies.map((dep) =>
 			path.join(
--- a/lib/controllers/prepare-controller.ts
+++ b/lib/controllers/prepare-controller.ts
@@ -375,7 +375,7 @@ export class PrepareController extends EventEmitter {
 		projectData: IProjectData
 	): Promise<string[]> {
 		const dependencies = this.$nodeModulesDependenciesBuilder
-			.getProductionDependencies(projectData.projectDir, projectData.dependenciesToIgnore)
+			.getProductionDependencies(projectData.projectDir, projectData.ignoredDependencies)
 			.filter((dep) => dep.nativescript);
 		const pluginsNativeDirectories = dependencies.map((dep) =>
 			path.join(
--- a/lib/definitions/project.d.ts
+++ b/lib/definitions/project.d.ts
@@ -153,7 +153,7 @@ interface IProjectData extends ICreateProjectData {
 	projectId: string;
 	projectIdentifiers?: Mobile.IProjectIdentifier;
 	dependencies: any;
-	dependenciesToIgnore?: string[];
+	ignoredDependencies?: string[];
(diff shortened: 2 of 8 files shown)


lib/controllers/prepare-controller.ts
```diff
@@ -375,7 +375,7 @@ export class PrepareController extends EventEmitter {
 		projectData: IProjectData
 	): Promise<string[]> {
 		const dependencies = this.$nodeModulesDependenciesBuilder
-			.getProductionDependencies(projectData.projectDir)
+			.getProductionDependencies(projectData.projectDir, projectData.dependenciesToIgnore)
```
rigor789, line 378:
```suggestion
			.getProductionDependencies(projectData.projectDir, projectData.ignoredDependencies)
```

lib/definitions/project.d.ts
```diff
@@ -153,6 +153,7 @@ interface IProjectData extends ICreateProjectData {
 	projectId: string;
 	projectIdentifiers?: Mobile.IProjectIdentifier;
 	dependencies: any;
+	dependenciesToIgnore?: string[];
```
rigor789, line 156:
```suggestion
	ignoredDependencies?: string[];
```

lib/services/plugins-service.ts
```diff
@@ -352,7 +352,7 @@ export class PluginsService implements IPluginsService {
 		dependencies =
 			dependencies ||
 			this.$nodeModulesDependenciesBuilder.getProductionDependencies(
-				projectData.projectDir
+				projectData.projectDir, projectData.dependenciesToIgnore
```
rigor789, line 355:
```suggestion
				projectData.projectDir, projectData.ignoredDependencies
```

lib/services/project-changes-service.ts
```diff
@@ -93,7 +93,7 @@ export class ProjectChangesService implements IProjectChangesService {
 			);
 
 			this.$nodeModulesDependenciesBuilder
-				.getProductionDependencies(projectData.projectDir)
+				.getProductionDependencies(projectData.projectDir, projectData.dependenciesToIgnore)
```
rigor789, line 96:
```suggestion
				.getProductionDependencies(projectData.projectDir, projectData.ignoredDependencies)
```

lib/tools/node-modules/node-modules-builder.ts
```diff
@@ -19,7 +19,7 @@ export class NodeModulesBuilder implements INodeModulesBuilder {
 		projectData,
 	}: IPrepareNodeModulesData): Promise<void> {
 		const dependencies = this.$nodeModulesDependenciesBuilder.getProductionDependencies(
-			projectData.projectDir
+			projectData.projectDir, projectData.dependenciesToIgnore
```
rigor789, line 22:
```suggestion
			projectData.projectDir, projectData.ignoredDependencies
```

rigor789 (approved):
LGTM, but to stay consistent with the other "types of dependencies", I'd rename the property to `ignoredDependencies`

farfromrefug:
@cla-bot check

farfromrefug:
@rigor789 I am ok with ignoring them only in production if used through `nativescript.config.ts`. But what i would really want is to have total control when defined through webpack config file. For example when using `sentry` i actually want to "ignore" it in dev too (and in f-droid production builds).
So the question is could i override it through webpack config?

PR #5544, opened 2021-06-25
fix: race condition when getting iOS inspector port

### LINKED ISSUE #3701: "NativeScript debugger has opened inspector socket on port.." is not shown for every simulator when run multiple simulators and livesyncing
### Tell us about the problem
`NativeScript debugger has opened inspector socket on port..` is not shown for every simulator when run multiple simulators and livesyncing

### Which platform(s) does your issue occur on?
iOS

### Please provide the following version numbers that your issue occurs with:
- CLI: next

### Please tell us how to recreate the issue in as much detail as possible.
1. `tns create SomeApp --ng`
2. Run few simulators
3. `tns run ios`
4. Make change in some file.

### Is there code involved? If so, please share the minimal amount of code needed to recreate the problem.
Consider using the [playground](https://play.nativescript.org) to provide the code you have issues with.

BenSjoberg:
## PR Checklist

- [x] There is an issue for the bug/feature this PR is for. To avoid wasting your time, it's best to open a suggestion issue first and wait for approval before working on it. #3701

When running an iOS simulator app, the NativeScript CLI sends an `AttachRequest` notification to the application right after launch. If the runtime hasn't had time to set up the event handler for this notification yet, it's silently lost, and the CLI never gets the port number for debugging. This causes HMR to not work, and worse, the application auto-restarts after a minute. (The console shows the error "NativeScript debugger was not able to get inspector socket port")

This seems to be a race condition based on how quickly the iOS runtime sets up its event handler vs. how quickly the CLI sends the `AttachRequest` notification after launching the app. For some reason, I went from not experiencing this issue at all to experiencing it nearly 100% of the time when using the simulator on my Mac. (I haven't tested with an actual device. Also, I'm using the V8 iOS runtime.)

## What is the new behavior?

I set up a timer to re-send the `AttachRequest` notification once every 5 seconds until `getDebuggerPort()` either returns or throws an error. In my testing, I was able to resolve it by adding a simple one-second delay before sending `AttachRequest`, but this seems more robust.

Fixes/Implements/Closes #3701. (Seems to be the same issue.)

### DIFF
--- a/lib/common/mobile/ios/simulator/ios-simulator-device.ts
+++ b/lib/common/mobile/ios/simulator/ios-simulator-device.ts
@@ -99,6 +99,8 @@ export class IOSSimulator extends IOSDeviceBase implements Mobile.IiOSDevice {
 				.catch((e) => this.$logger.error(e));
 		}, 5e3);
 
+		// the internal retry-mechanism of getDebuggerPort will ensure the above 
+		// interval has a chance to execute multiple times
 		const port = await super.getDebuggerPort(appId).finally(() => {
 			clearInterval(postNotificationRetryInterval);
 		});
--- a/lib/common/mobile/ios/simulator/ios-simulator-device.ts
+++ b/lib/common/mobile/ios/simulator/ios-simulator-device.ts
@@ -87,7 +87,21 @@ export class IOSSimulator extends IOSDeviceBase implements Mobile.IiOSDevice {
 			attachRequestMessage,
 			this.deviceInfo.identifier
 		);
-		const port = await super.getDebuggerPort(appId);
+
+		// Retry posting the notification every five seconds, in case the AttachRequest
+		// event handler wasn't registered when the first one was sent
+		const postNotificationRetryInterval = setInterval(() => {
+			this.$iOSEmulatorServices
+				.postDarwinNotification(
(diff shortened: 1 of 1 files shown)


NathanWalker:
Thank you @BenSjoberg ! We'll include this with the upcoming 8.1 release.

lib/common/mobile/ios/simulator/ios-simulator-device.ts
```diff
@@ -87,7 +87,21 @@ export class IOSSimulator extends IOSDeviceBase implements Mobile.IiOSDevice {
...
+		const postNotificationRetryInterval = setInterval(() => {
+			this.$iOSEmulatorServices
+				.postDarwinNotification(
+					attachRequestMessage,
+					this.deviceInfo.identifier
+				)
+				.catch((e) => this.$logger.error(e));
+		}, 5e3);
+
+		const port = await super.getDebuggerPort(appId).finally(() => {
```
rigor789, line 102:
```suggestion
		// the internal retry-mechanism of getDebuggerPort will ensure the above 
		// interval has a chance to execute multiple times
		const port = await super.getDebuggerPort(appId).finally(() => {
```

rigor789 (approved):

PR #5554, opened 2021-07-27
feat: `config` option

farfromrefug:
to set custom `nativescript.config` file name (relative, absolute, with or without .ts or .js)

read `ignoredNativeDependencies` from `nativescript.config``nativescript.config`

I added a test. What is weird to me is that some tests are failing.But i cant see how they could pass, like this one https://github.com/farfromrefug/nativescript-cli/blob/cdf912ecdbd01c34ab405df2e9e39268f982d71f/test/tools/node-modules/node-modules-dependencies-builder.ts#L536

### DIFF
--- a/docs/man_pages/start.md
+++ b/docs/man_pages/start.md
@@ -75,5 +75,6 @@ Option | Description
 -------|---------
 --help, -h, /? | Prints help about the selected command in the console.
 --path `<Directory>` | Specifies the directory that contains the project. If not set, the project is searched for in the current directory and all directories above it.
+--config | Specifies the name of the Nativescript configuration file to load (relative to the project directory). The default is.
 --version | Prints the client version.
 --log trace | Prints a detailed diagnostic log for the execution of the current command.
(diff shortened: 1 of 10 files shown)


rigor789:
Interesting side-effect of passing `--config` is that when the CLI spawns the webpack process for building it passes `--config=path/to/webpack.config.js` (or the one set in the config).

With webpack5, we require the CLI as a library and expose it through the Utils - but since the spawned webpack process now has `--config=...` in `process.argv` the CLI resolves the project config as `webpack.config.js`.

Workaround is to splice/delete the `--config` option from `process.argv` inside the `@nativescript/webpack` bin - which then makes everything work - except the case where you want to use a custom config file and read it through `Utils.config.getValue()`...

farfromrefug:
@rigor789 would nt it be simpler to rename this option ? Like nconfig maybe .

rigor789:
Did some tweaking with both the CLI and the webpack bin to handle this case correctly.

A small issue I need to fix is the config lookup - right now it will set all 3 paths to the resolved config (`TSConfigPath`, `JSConfigPath`, `NSConfigPath`). WIP

rigor789:
TODO: Add `ignoredNativeDependencies: string[]` to the NativeScript Config interface: https://github.com/NativeScript/NativeScript/blob/master/packages/core/config/config.interface.ts

PR #5555, opened 2021-07-29
fix allow plugins to use kt files

farfromrefug:
I am writing a plugin which comes with kotlin files directly in its platforms folder (android/java/...)
Right now it wont build while building an app (building the plugin in tempPlugin). This enables kotlin for all!


triniwiz (approved):
LGTM

PR #5578, opened 2021-09-26
fix: update plugin gradle script

farfromrefug:
This updates the plugin gradle script:
* no more android support, use androidx
* use java 1.8 (removes warning in the console
* koltin, androidx versions through project properties


rigor789:
> no more android support, use androidx

Would that be considered a BREAKING CHANGE?

farfromrefug:
Do we still support old versions without androidx?

rigor789:
I'm not sure - some plugins may still use it?

farfromrefug:
Seems unlikely. And anyway they can deps on their include.gradle. 
Honestly we should remove them, plus they end up in the .dat metadata...

rigor789:
👍 - just don't want to break things by accident in a patch release

farfromrefug:
@rigor789 i think it should got with android runtime 8.2.0 as we move to gradle 7. But i understand your point

farfromrefug:
@rigor789 @triniwiz that PR now goes further and makes generated plugins aar like 10 times smaller. The consequence is apps dex files getting also much smaller (mine dont even need multidex anymore)

triniwiz:
All support packages would be jetpacked (if you wanna call it that) .. this looks good for 8.2

vendor/gradle-plugin/build.gradle
```diff
@@ -6,6 +6,12 @@ apply plugin: 'com.android.library'
 apply plugin: 'kotlin-android'
 apply plugin: 'kotlin-android-extensions'
 
+def computeKotlinVersion = { -> project.hasProperty("kotlinVersion") ? kotlinVersion : "1.4.21" }
```
rigor789, line 9: Can we bump kotlinVersion without too much testing, or is 1.4.21 safer for now?

vendor/gradle-plugin/build.gradle
```diff
@@ -6,6 +6,12 @@ apply plugin: 'com.android.library'
...
+def computeCompileSdkVersion = { -> project.hasProperty("compileSdk") ? compileSdk : 28 as int }
```
rigor789, line 10: Should we change the default `compileSdk` from 28 to 30 (or perhaps newer)?

vendor/gradle-plugin/build.gradle
```diff
@@ -6,6 +6,12 @@ apply plugin: 'com.android.library'
...
+def computeTargetSdkVersion = { -> project.hasProperty("targetSdk") ? targetSdk : 28 as int }
```
rigor789, line 11: Should we change `targetSdk` to 30 or newer? Believe 30 is minimum from google right now.

vendor/gradle-plugin/build.gradle
```diff
@@ -6,6 +6,12 @@ apply plugin: 'com.android.library'
...
+def computeBuildToolsVersion = { ->
+    project.hasProperty("buildToolsVersion") ? buildToolsVersion : "28.0.3" as String
```
rigor789, line 13: With 8.2 - should we default to build tools 32?

vendor/gradle-plugin/build.gradle
```diff
@@ -88,39 +94,36 @@ allprojects {
...
 
 
-def computeCompileSdkVersion = { -> project.hasProperty("compileSdk") ? compileSdk : 28 }
-def computeBuildToolsVersion = { ->
-    project.hasProperty("buildToolsVersion") ? buildToolsVersion : "28.0.3"
-}
-
 android {
+    kotlinOptions {
+        jvmTarget = '1.8'
```
rigor789, line 99: Do we need to set the target to 1.8 with gradle7+? Officially recommended jdk is 11, so this might need to match that?

vendor/gradle-plugin/build.gradle
```diff
@@ -88,39 +94,36 @@ allprojects {
...
+    }
     applyBeforePluginGradleConfiguration()
 
     compileSdkVersion computeCompileSdkVersion()
     buildToolsVersion computeBuildToolsVersion()
 
+    compileOptions {
+        sourceCompatibility JavaVersion.VERSION_1_8
+        targetCompatibility JavaVersion.VERSION_1_8
```
rigor789, line 108: Same question as `jvmTarget`

vendor/gradle-plugin/build.gradle
```diff
@@ -88,39 +94,36 @@ allprojects {
...
-
-    configurations.all {
-       resolutionStrategy.eachDependency { DependencyResolveDetails details ->
-           if (details.requested.group == "com.android.support" && !details.requested.name.startsWith("multidex")) {
-               details.useVersion supportVer
-           }
-       }
-    }
+    def androidXMultidexVersion =  project.hasProperty("androidXMultidexVersion") ? project.androidXMultidexVersion : "2.0.1"
+    def androidxVersion =  project.hasProperty("androidxVersion") ? project.androidxVersion : "1.2.0"
```
rigor789, line 120: Can we use 1.7.0?

vendor/gradle-plugin/build.gradle
```diff
@@ -88,39 +94,36 @@ allprojects {
...
+    def androidXAppCompatVersion =  project.hasProperty("androidXAppCompat") ? project.androidXAppCompat : "1.3.1"
```
rigor789, line 121: Can we use 1.4.1?

rigor789 (commented):
@farfromrefug @triniwiz if you could help clear up questions I posted in the review, that would help a ton with landing this.

vendor/gradle-plugin/build.gradle
farfromrefug: not sure about 1.7.0. Maybe 1.6.0

vendor/gradle-plugin/build.gradle
farfromrefug: we should be able to use more but was not tested https://kotlinlang.org/docs/faq.html#which-versions-of-jvm-does-kotlin-target
let s make it an ext property so we can test easily?

vendor/gradle-plugin/build.gradle
farfromrefug: yes i think so

farfromrefug:
superseded by #5774

PR #5601, opened 2021-10-27
fix(clean) report failures to remove item

### LINKED ISSUE #5588: ns clean windows shows that everything has gone well but does not delete the platforms folder
When you run ns clean and the platform folder is in use "for whatever reason" the cli message is that everything has gone well (next image). However, it has not been able to clean the platform folder and then in the next ns run android it always returns an error type:
```

* What went wrong:
Task 'assembleDebug' not found in root project 'pricelight'.

* Try:
Run gradlew tasks to get a list of available tasks. Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.
```

This is very confusing, since you think that the directory has been cleaned and when you start you have an error. For experienced people you may realize it but for new people you can go crazy.

I think that ns clean should check if it has really deleted the platform folder and if it has not been able to put a very large error indicating that it must be eliminated manually or the project will not work

jcassidyav:
## PR Checklist

- [X] The PR title follows our guidelines: https://github.com/NativeScript/NativeScript/blob/master/CONTRIBUTING.md#commit-messages.
- [X] There is an issue for the bug/feature this PR is for. To avoid wasting your time, it's best to open a suggestion issue first and wait for approval before working on it.
- [X] You have signed the [CLA](http://www.nativescript.org/cla).
- [X] All existing tests are passing: https://github.com/NativeScript/nativescript-cli/blob/master/CONTRIBUTING.md#contribute-to-the-code-base

If ```ns clean``` failed to remove an item ( which is common on windows where the platforms directory is not removed properly), there was no indication to the user.

## What is the new behavior?
Now the output of  ```ns clean``` will indicate which item was not removed.
e.g.

_ Cleaning project...√ Cleaned directory hooks
× Failed to Clean directory platforms
√ Cleaned directory node_modules
√ Cleaned file package-lock.json
× Project unsuccessfully cleaned.

Fixes/Implements/Closes #5588.

### DIFF
--- a/lib/services/project-cleanup-service.ts
+++ b/lib/services/project-cleanup-service.ts
@@ -32,11 +32,11 @@ export class ProjectCleanupService implements IProjectCleanupService {
 
 	public async cleanPath(pathToClean: string): Promise<boolean> {
 		this.spinner.clear();
-		let result = true;
+		let success = true;
 		let fileType: string;
 		if (!pathToClean || pathToClean.trim().length === 0) {
 			this.$logger.trace("cleanPath called with no pathToClean.");
-			return result;
+			return success;
 		}
 
 		const filePath = path.resolve(this.$projectHelper.projectDir, pathToClean);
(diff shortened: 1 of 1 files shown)


vallemar:
I applaud this PR, I would like to force the deletion of the platform folder. Some time ago someone commented in discord this script to eliminate the process that has it busy. Couldn't this be added before deleting this folder to ensure its deletion?
```cmd
$process=(Get-WmiObject Win32_Process -Filter "name = 'java.exe' AND CommandLine like '%gradle%'")
if(-NOT ( $process  -EQ $Null)){
    echo Terminating $process.CommandLine
    $process.Terminate() | Out-Null
}
```

jcassidyav:
@vallemar I posted that script to discord, but I think that would be a step too far to add to the cli, as who knows what people have running on their machines either now or in the future which would match that query.

A little later in discord I posted that from what I can see, this particular case is caused by The Gradle Daemon, the details on this process are documented here https://docs.gradle.org/current/userguide/gradle_daemon.html.

There are multiple options to turn it off, e.g.  setting then environment variable documented here https://docs.gradle.org/current/userguide/gradle_daemon.html#daemon_faq 
which prevents this happening for me.

vallemar:
@jcassidyav amazing.  So if this can be disabled, do you think it would be okay to put something in the ns clean documentation?  I'm also wondering if it would be okay to put a --force flag or something like that so that it launches the above query and closes the gradle process.  What do you think?  I do not know if they work with Windows in NS but it is a real nuisance

Issue #5550, opened 2021-07-13
Provide `--gradle` option for using a custom Gradle executable


FestplattenSchnitzel:
**Is your feature request related to a problem? Please describe.**
 

Apps can't be build for [F-Droid](https://f-droid.org/) directly with the CLI (`ns build android`) at the moment. This is due to the fact, that F-Droid 1) uses a custom Gradle executable and 2) deletes all additional Gradle wrappers before starting the build process.

**Describe the solution you'd like**

An `--gradle` option (or similarly named) for the CLI would be great. It should provide the ability to use a custom Gradle executable instead of the one from the {N} Gradle wrapper.

E.g. `ns build android --gradle /usr/bin/gradle`

**Describe alternatives you've considered**

The current workaround is:
- `ns prepare android`
- `cd platforms/android/app`
- `gradle "assembleRelease" "-PappPath=app" "-PappResourcesPath=app/resources"` (or similar)

However some [Webpack|JS] features like `AOT` or `uglify` might not work.

**Additional context**

Feel free to ask for more information at https://gitlab.com/fdroid/fdroiddata/-/issues/2309.

CC @farfromrefug

rigor789:
Would these apps need building for both regular android and F-Droid, or should we move this to the `nativescript.config.(ts|js)` where it's persistent (ie. doesn't require passing an extra flag every time)?

_PS. The proposed feature of passing CLI args from the config ( from #5296 ) is something we'll be looking at very soon, just slightly a different direction to utilize the config instead of a json file..._

FestplattenSchnitzel:
I think you got me wrong. F-Droid is not a custom Android variant, it's "just" an app store that does only contain free software apps for "regular" android (and therefore builds everything from source).

So there is no persistence needed, since the app developer (or someone who clones & builds from the repository) might build with the Gradle wrapper shipped with the {N} CLI but F-Droid or another user might want to use a Gradle executable installed in a custom location.

Nevertheless: thank you for your fast reply! :rocket:

farfromrefug:
@FestplattenSchnitzel i created and tested the PR https://github.com/NativeScript/nativescript-cli/pull/5628

FestplattenSchnitzel:
Thank you very much @farfromrefug!

PR #5628, opened 2022-02-15
feat: --gradlePath argument support

farfromrefug:
This simple PR add `--gradle` option to specify a custom gradle binary path
Solves https://github.com/NativeScript/nativescript-cli/issues/5550 

EDIT: renamed the option to `gradlePath`. makes more sense as i am about to add gradle option for args


rigor789 (approved):
LGTM

PR #5630, opened 2022-02-15
feat: `--gradleArgs` option to pass gradle parameters

farfromrefug:
In the same vibe as `gradlePath` this allows to pass gradle command parameters.
One note is that they have to be double quoted like this `--gradle '"-PsplitEnabled"'` or using `=` like this `--gradle="-PsplitEnabled"`
This is due to how `yargs` parses arguments


rigor789 (approved):
LGTM

PR #5631, opened 2022-02-15
fix: correctly allow plugins build to detect before-plugins.gradle

farfromrefug:
and other gradle scripts
This is a copy paste of the methods from the android runtime


rigor789 (approved):
LGTM

PR #5635, opened 2022-02-20
feat: only install corresponding abi package when possible

farfromrefug:
This PR changes the way the `installOnDevice` works.
The idea is to be compatible with apk split on android.
The reason for this is multiple:
* much smaller apps on device/emulator. It gets really boring to get warning about emulator not having enough space.
* it is the first step to other features like only build device ABI

The way it works is this:
* track device abis (might be multiple as a 64 device almost always accept 32bit apks). 
* When deploying see how many `.apk` files were built:
      - 1: always install this one
      - >1: if 64bit abi or 32bit abi package found use this one
      - >1: if not look for universal one found use this one
      - otherwise => error
 
I also applied the same rule for `copy copyLatestAppPackage To` (thus renamed to `copyAppPackages `)
* if `copyTo` is an existing directory or a path which does not end with `aab|ipa|apk` copy all packages to `copyTo` as a directory
* if `copyTo` is a file name
    - if only one package in output dir then copy it
    - copy only `universal` package if existing

There is one thing bugging be for which i have not found a solution. Because of [this](https://github.com/NativeScript/nativescript-cli/blob/master/lib/controllers/build-controller.ts#L113) line if there are multiple packages there will be warning about it which is wrong in its meaning when using `run`. Might also be wrong in other cases and we can simply remove the warning ...

### DIFF
--- a/lib/common/definitions/mobile.d.ts
+++ b/lib/common/definitions/mobile.d.ts
@@ -99,6 +99,12 @@ declare global {
 			 * For iOS simulators - same as the identifier.
 			 */
 			imageIdentifier?: string;
+
+			/**
+			 *  Optional property describing the architecture of the device
+			 *  Available for Android only
+			 */
+			 abis?: string[];
 		}
 
 		interface IDeviceError extends Error, IDeviceIdentifier {}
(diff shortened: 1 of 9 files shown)


farfromrefug:
@rigor789 i updated that PR with my latest changes. I now check for changes on next builds to see if the arch of the device is different from the arch of what we built previously. It now works in all cases

PR #5636, opened 2022-02-22
fix: handle `gradleArgs` as an array option

farfromrefug:
This changes the way `gradleArgs` works. 
Now you can use either `--gradleArgs=-PsplitEnabled  --gradleArgs=-PabiFilters=arm64-v8a,armeabi-v7a` or `--gradleArgs=\"-PsplitEnabled -PabiFilters=arm64-v8a,armeabi-v7a\"` 

there is just duplicate code in the parsing handle multiple args in the same `gradleArgs` (necessary because otherwise gradle wont read them)


farfromrefug:
superseded by #5774

PR #5644, opened 2022-03-07
Smaller plugin aar

farfromrefug:
There were a ton of unecessary data added to each aar. Those data would make apps bigger and slower because they also add sbg metadata


farfromrefug:
@rigor789 indeed might not be needed with gradle 7.0

PR #5667, opened 2022-05-03
feat: Added support for generating android adaptive icons

### LINKED ISSUE #5571: `ns resources generate icons` - cli script update needed to support latest Android vector graphic launchers.
**Environment**
zsh on macOS Catalina 10.15.7

✔ Component nativescript has 8.0.1 version and is up to date.
✔ Component @nativescript/core has 8.0.3 version and is up to date.
✖ Component @nativescript/ios is not installed.
✔ Component @nativescript/android has 8.0.0 version and is up to date.

**Describe the bug**
Created icon.png with 1024x1024 size.
```
tns resources generate icons ~/icon.png
Generating icons ...
Icons generation completed.
```
.. however, Android icons did not get created. Even when creating the mipmap icons, they don't get applied and the nativescript icon is associated with the app. No error messages.
Splashes are generated correctly.

**Expected behavior**
The launcher icon should be generated from the image.

CatchABus:
## PR Checklist

Command `ns resources generate icons <source image path>` cannot generate app launcher icons ever since the adaptive icons addition (ic_launcher). Right now, developers have no ways to generate android launcher icons apart from using Android Studio.

## What is the new behavior?
Command `ns resources generate icons <source image path>` will generate all needed icons.

According to google, adaptive icons have few requirements:

- Both layers must be sized at 108 x 108 dp.
- The inner 72 x 72 dp of the icon appears within the masked viewport.
- The system reserves the outer 18 dp on each of the 4 sides to create interesting visual effects, such as parallax or pulsing.

In order to comply with these requirements, I updated `image-definitions.json` using the following concept:

1. Source image will also be used to create an `ic_launcher_foreground` resized to 72 dp converted to pixels.
2. A new resize operation called `outerScale` was added to CLI. Using this operation, we can scale the outer part of the image (it's a bottom layer actually), which will become an opaque layer using hex `#00000000`.
3. Scale is set to 1.5x  to resize image (but not the inner content) to pixels that equal to 108dp according to device dpi. (72dp x 1.5 = 108dp)

Fixes/Implements/Closes #5571.

### DIFF
--- a/docs/man_pages/project/configuration/resources/resources-generate-icons.md
+++ b/docs/man_pages/project/configuration/resources/resources-generate-icons.md
@@ -15,6 +15,10 @@ Usage | Synopsis
 ------|-------
 `$ tns resources generate icons <Path to image>` | Generate all icons for Android and iOS based on the specified image.
 
+### Options
+
+* `--background` Sets the background color of the icon. When no color is specified, a default value of `transparent` is used. `<Color>` is a valid color and can be represented with string, like `white`, `black`, `blue`, or with HEX representation, for example `#FFFFFF`, `#000000`, `#0000FF`. NOTE: As the `#` is special symbol in some terminals, make sure to place the value in quotes, for example `$ tns resources generate icons ../myImage.png --background "#FF00FF"`.
+
 ### Arguments
 
 * `<Path to image>` is a valid path to an image that will be used to generate all icons.
--- a/lib/commands/generate-assets.ts
+++ b/lib/commands/generate-assets.ts
@@ -61,6 +61,7 @@ export class GenerateIconsCommand
 	): Promise<void> {
 		await this.$assetsGenerationService.generateIcons({
 			imagePath,
+			background,
 			projectDir: this.$projectData.projectDir,
 		});
 	}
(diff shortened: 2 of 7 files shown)


vallemar:
this PR is quite useful, right now the generation of icons is broken and it is quite annoying to execute the command and not see the change in the application

CatchABus:
@NathanWalker I know there's been a discussion about a different approach but perhaps we could give this one a try until then.

This PR is trying to follow the principles described in this guide: https://developer.android.com/guide/practices/ui_guidelines/icon_design_adaptive

These new dimensions equal to size of inner content of foreground layer which has to be 72dp.
New values inside image-definitions file make sense according to "dp to pixels" conversion board:
image

Multiplying those by 1.5 during asset generation gives us the right pixels for both layers that equal to 108dp.

CatchABus:
@NathanWalker I have applied my final changes on this one.

Few details regarding android icons.

CLI did not generate icons unless it found old icons with identical names in file system.
This was really confusing and frustrating, and was the reason for all open tickets to begin with.
I took the liberty and removed that check so that asset definitions are always generated but I can re-add it if needed.

Regarding the output content, it generates simple icons whose rendering is similar to the one of the 'legacy icons' used in old app templates (icon.png).
If a developer wants to create more complex adaptive icons, Android Studio is still the preferred choice.

CatchABus:
> this PR is quite useful, now the icon generation is broken and it is quite annoying to execute the command and not see the change in the application

You can test and post feedback if you like using my branch: https://github.com/CatchABus/nativescript-cli/tree/android-generate-icons-fix
It will need building TS and using command from repo `bin` directory.

bretep:
Hi! I'm also running into this issue. Is there a plan to release these fixes?

PR #5755, opened 2023-08-17
feat(logs): improve android logging

ammarahm-ed:
## PR Checklist

Logging on android stops working if the app is restarted. This happens mainly due to logging directly linked with app PID. When app restarts, the pid changes and logs stop.

## What is the new behavior?
I wanted to completely ditch PID for something else on android but I realized there are good reasons to keep using PID as it helps us filter logs precisely for the current app. In the new approach we restart adb logcat with the new PID if the app is restarts automatically. This we we always keep the CLI in sync with the current PID of the running NativeScript app.

This PR also makes logging much more performant on android by using native android filters of TAGs of concern instead of filtering on the host device.

- Removed `JS:` from logs
- Logging works on app restart
- Improved logging performance on android and reduced the filtering needed on host device

### DIFF
--- a/lib/common/mobile/android/android-log-filter.ts
+++ b/lib/common/mobile/android/android-log-filter.ts
@@ -7,7 +7,8 @@ export class AndroidLogFilter implements Mobile.IPlatformLogFilter {
 
 	// sample line is "11-23 12:39:07.310  1584  1597 I art     : Background sticky concurrent mark sweep GC freed 21966(1780KB) AllocSpace objects, 4(80KB) LOS objects, 77% free, 840KB/3MB, paused 4.018ms total 158.629ms"
 	// or '12-28 10:45:08.020  3329  3329 W chromium: [WARNING:data_reduction_proxy_settings.cc(328)] SPDY proxy OFF at startup'
-	private static API_LEVEL_23_LINE_REGEX = /.+?\s+?(?:[A-Z]\s+?)([A-Za-z \.]+?)\s*?\: (.*)/;
+	private static API_LEVEL_23_LINE_REGEX =
+		/.+?\s+?(?:[A-Z]\s+?)([A-Za-z \.]+?)\s*?\: (.*)/;
 
 	constructor(private $loggingLevels: Mobile.ILoggingLevels) {}
 
@@ -23,7 +24,9 @@ export class AndroidLogFilter implements Mobile.IPlatformLogFilter {
 			);
 			if (log) {
 				if (log.tag) {
-					return `${log.tag}: ${log.message}` + EOL;
+					return (
+						`${log.tag === "JS" ? "" : `${log.tag}: `}${log.message}` + EOL
+					);
 				} else {
 					return log.message + EOL;
 				}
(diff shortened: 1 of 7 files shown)


mukaschultze:
https://github.com/NativeScript/nativescript-cli/blob/ba3293c174155b1143e87662ed49c978e9cac32d/lib/common/mobile/device-log-provider.ts#L88-L90

I guess this can be removed now that the `JS: ` tag is not forwarded to the provider?

ammarahm-ed:
> https://github.com/NativeScript/nativescript-cli/blob/ba3293c174155b1143e87662ed49c978e9cac32d/lib/common/mobile/device-log-provider.ts#L88-L90
> 
> I guess this can be removed now that the `JS: ` tag is not forwarded to the provider?

Probably yes, will have to see which one is better to keep.

lib/common/mobile/android/android-log-filter.ts
```diff
@@ -35,33 +38,19 @@ export class AndroidLogFilter implements Mobile.IPlatformLogFilter {
...
-		];
-
-		let consoleLogMessage: { tag?: string; message: string };
-
+		let consoleLogMessage;
 		const match =
-			lineText.match(AndroidLogFilter.LINE_REGEX) ||
-			lineText.match(AndroidLogFilter.API_LEVEL_23_LINE_REGEX);
+			lineText.match(AndroidLogFilter.API_LEVEL_23_LINE_REGEX) ||
+			lineText.match(AndroidLogFilter.LINE_REGEX);
```
rigor789, line 49: Note: I flipped this condition to speed up matching by optimizing for API23 first and then falling back to legacy regex.

lib/common/mobile/android/logcat-helper.ts
```diff
@@ -71,6 +75,41 @@ export class LogcatHelper implements Mobile.ILogcatHelper {
...
+					if (
+						// action
+						line.includes("START") &&
+						// appId
+						line.includes(options.appId) &&
+						// pid - only if it's not the current pid...
+						!line.includes(options.pid)
+					) {
+						this.forceStop(deviceIdentifier);
+						options.onAppRestarted?.();
```
rigor789, line 109: Note: not `break;`-ing here, since it's possible (though unlikely) that two subsequent START lines come in the same lineBuffer. Though this doesn't really matter - this stream is not noisy at all because we're pre-filtering with the adb logcat regex.

lib/common/mobile/android/logcat-helper.ts
```diff
@@ -108,39 +147,92 @@ export class LogcatHelper implements Mobile.ILogcatHelper {
...
+			"TNS.Java",
+		];
 
 		if (pid && isLogcatPidSupported) {
 			logcatCommand.push(`--pid=${pid}`);
+
+			acceptedTags.forEach((tag) => {
+				// -s <tag> - shows only logs with the specified tag
+				logcatCommand.push("-s", tag);
+			});
```
rigor789, line 203: @ammarahm-ed this is a great idea to move here - technically if `isLogcatPidSupported` is false - we'd log out all tags even if not accepted - since the filtering has been removed from android-log-filter, however I don't think this would ever be a problem.

I've extracted the `isLogcatPidSupported` to a function since it's used by both methods here - and I marked is `@deprecated` because I don't think we need to worry about that soon anymore (if I understood correctly, that's for API24 or below - android 7.0)...

lib/common/mobile/android/logcat-helper.ts
```diff
@@ -108,39 +147,92 @@ export class LogcatHelper implements Mobile.ILogcatHelper {
...
+		appId?: string
+	) {
+		const adb: Mobile.IDeviceAndroidDebugBridge = this.$injector.resolve(
+			DeviceAndroidDebugBridge,
+			{ identifier: deviceIdentifier }
+		);
+
+		// -b system  - shows the system buffer/logs only
+		// -T 1 			- shows only new logs after starting adb logcat
+		const logcatCommand = [`logcat`, `-b`, `system`, `-T`, `1`];
```
rigor789, line 224: Note: Added the `-b system` flag. Probably not a huge difference, but only showing the `system` buffer should theoretically be faster than showing the default (system, crash, main).

rigor789 (approved):
Great additions! Added some notes about some cleanup changes I've done for future reference.

vallemar:
@rigor789 Is there a version to test these changes?

rigor789:
The next tag should have it.

PR #5757, opened 2023-08-27
Feat: Quality of life improvements

ammarahm-ed:
This PR adds some quality of life improvements to NativeScript CLI.

 - A new command, `ns start` that can run and manage Android & iOS apps side by side. The goal is to avoid restarting the CLI for most basic things like cleaning project/restarting the app.
 - The CLI is interactive when you run `ns start` or `ns run *` commands. Initially I have enabled the following key commands:
  
  ```
   a — Run android app
   i — Run iOS app
   A — Open android project in Android Studio
   I — Open iOS project in Xcode
   r — Rebuild native app if needed and restart
   R — Force rebuild native app and restart
   w — Toggle file watcher (Disables/Enables webpack/native changes watch etc)
   c — Clean project
   n — Install dependencies
  ```
Do let me know if there are any other key commands you would like to see.

One thing still missing is ability to connect to an exisiting app on a device without rebuilding, just connecting HMR and reloading the app that I really want to add.
  
  Also for the ns start command, there could be a way to toggle logging and show minimum logs other than those from the app with the option to enable complete logs.
  
- The CLI also has a new header similar to preview CLI that prints whenever you run any command.

Here's the new look:

Screenshot 2023-08-27 at 5 07 10 PM

## PR Checklist


NathanWalker:
Couple initial comments:

- `ns start` as interactive menu is excellent - and consistent with other workflows that have been enabled throughout {N} open source in general - super helpful
- The prompt order and grouping I may suggest the following (attempts to better group things to clarify):

```
Debug with Chrome DevTools Enabled:
a — Debug Android app 
i — Debug iOS app

Dev Workflow:
 r — Rebuild native app if needed and restart
 R — Force rebuild native app and restart
 w — Toggle file watcher (Disables/Enables webpack/native changes watch etc)

Open Platform Projects:
 A — Open Android project in Android Studio
 I — Open iOS project in Xcode

Project Management:
 c — Clean project
 n — Install dependencies
```

- Changing 'run' to 'debug' since run/debug have different meanings and uses in CLI. This start menu can use `debug` as default since most useful. One could still `ns run` on their own - it's more advanced and would argue less used command but allows multiples connected devices to be run at same time. Whereas `debug` always enabled chrome debugger and uses a targeted device.
- The debug usage I think might be useful to auto open chrome devtools rather than simply printing the url. This is debateable but given most are developing with the need to have devtools active/enabled from start, opening url on debug runs may be useful.

lib/common/header.ts
```diff
@@ -0,0 +1,28 @@
...
+	const middle = [
+		color.dim("│  "),
+		color.cyanBright.bold(" {N} NativeScript "),
+		color.whiteBright.bold("CLI"),
+		color.dim(` [v${version}] `),
+		color.dim("  │"),
+	].join("");
+	const middle2 = [
+		color.dim("│ "),
+		color.whiteBright.bold(" Empowering JS with Native APIs "),
```
NathanWalker, line 15:
```suggestion
		color.whiteBright.bold(" Empower JavaScript with native APIs "),
```

This will match heading on homepage 1-1 which helps clarify in full.

ammarahm-ed:
> The prompt order and grouping 

We can improve the order but I am not sure if it's a good idea to group them. I understand that we want to make things more clear but it will make the initial header + key command help quick long and take more space. Maybe just keeping the relevant commands together is good enough.

> Changing 'run' to 'debug' since run/debug have different meanings and uses in CLI. This start menu can use debug as default since most useful. One could still ns run on their own - it's more advanced and would argue less used command but allows multiples connected devices to be run at same time. Whereas debug always enabled chrome debugger and uses a targeted device.

I personally have been using `run` command most of the time. I think you aren't always in the `debug` mode and needing a chrome tab opened to debug the app. Most of the time `console.log` does it's job until you hit something that needs debugging in chrome. A key command like press "d" to start debugging can work too for that case.

lib/key-commands/index.ts
```diff
@@ -0,0 +1,278 @@
...
+				"Android Studio",
+				"bin",
+				"studio64.exe"
+			);
+			if (!fs.existsSync(studioPath)) {
+				this.$logger.error("Android Studio is not installed");
+				return;
+			}
+
+			this.$childProcess.exec(`${studioPath} ${androidDir}`);
```
jcassidyav, line 84: It needs to escape any spaces in the path by adding quotes, 
```suggestion
			this.$childProcess.exec(`'${studioPath}" "${androidDir}"`);
```

lib/services/start-service.ts
```diff
@@ -0,0 +1,115 @@
...
+	}
+
+	format(data: Buffer, platform: string) {
+		return data;
+	}
+
+	async runForPlatform(platform: string) {
+		const platformLowerCase = platform.toLowerCase();
+		(this as any)[platformLowerCase] = this.$childProcess.spawn(
+			"../nativescript-cli/bin/ns",
```
jcassidyav, line 38: This seems to always resolve relative to the project folder.  should be just `ns`?

farfromrefug:
@ammarahm-ed i never had the time to actually test that, but now i would like to use it for a secific case.
As you say when i `ns run ...` it should be interactive and i should be able to type `R` to restart in the console?
I tried but it does not seem interactive

ammarahm-ed:
@farfromrefug I haven't tried, you can check if it works like that. It should.

farfromrefug:
@ammarahm-ed it does not seem to work. Will to see if i can add an argument to run to make it interactive

ammarahm-ed:
> @ammarahm-ed it does not seem to work. Will to see if i can add an argument to run to make it interactive

It probably needs to pass the arguments to the `runForPlatform` function so it can forward those env vars. It does forward `process.env` to the commands that run.

farfromrefug:
@ammarahm-ed was actually easy:
```
NS_IS_INTERACTIVE=1 ns run ios --no-hmr --env.watchNodeModules --env.devlog --env.buildstyle --env.buildpeakfinder --env.build3dmap
```
It works! thanks a lot

ammarahm-ed:
@farfromrefug Awesome, I guess it would be nice if we can forward the arguments to internal `runForPlatform` function so env vars can work when you use `ns start`.

PR #5762, opened 2023-10-10
fix: setup docs links

vallemar:
when it has an error due to the environment, we have the error
```
Your environment is not configured properly and you will not be able to execute local builds.
Verify that your environment is configured according to the system requirements described at
https://docs.nativescript.org/environment-setup.html#macos-ios.
```

This does not correspond to the new documentation, since before we had everything on the same page and now on separate pages

Also this PR needs more changes, since we need
```
/setup/macos#setting-up-macos-for-ios

rather

/environment-setup.html#macos-ios
```
Here we can add the change to put
```
const anchor = platform ? `/${os}#setting-up-${os}-for-${platform.toLowerCase()}` : "";
```

Or we can change in the documentation

```
/setup/macos#setting-up-macos-for-ios

To

/setup/macos#ios
```

### DIFF
--- a/lib/services/platform-environment-requirements.ts
+++ b/lib/services/platform-environment-requirements.ts
@@ -92,7 +92,7 @@ export class PlatformEnvironmentRequirements
 
 		return (
 			`Verify that your environment is configured according to the system requirements described at\n` +
-			`https://docs.nativescript.org/setup${anchor}.`
+			`https://docs.nativescript.org/setup/${os}${anchor}.`
 		);
 	}
 }
--- a/lib/services/platform-environment-requirements.ts
+++ b/lib/services/platform-environment-requirements.ts
@@ -88,11 +88,11 @@ export class PlatformEnvironmentRequirements
 			darwin: "macos",
 		} as any)[process.platform];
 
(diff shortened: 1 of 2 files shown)


lib/services/platform-environment-requirements.ts
```diff
@@ -88,11 +88,11 @@ export class PlatformEnvironmentRequirements
 			darwin: "macos",
 		} as any)[process.platform];
 
-		const anchor = platform ? `#${os}-${platform.toLowerCase()}` : "";
+		const anchor = platform ? `/${os}#${platform.toLowerCase()}` : "";
 
 		return (
 			`Verify that your environment is configured according to the system requirements described at\n` +
-			`https://docs.nativescript.org/environment-setup.html${anchor}.`
+			`https://docs.nativescript.org/setup${anchor}.`
```
rigor789, line 95:
```suggestion
			`https://docs.nativescript.org/setup/${os}${anchor}.`
```

rigor789 (commented):
I think that should do it

vallemar:
@rigor789 apply changes!

PR #5764, opened 2023-10-17
feat(android): allow to force gradle version throught nativescript-co…

farfromrefug:
…nfig (`android.gradleVersion`).

We also now allows `buildToolsVersion` to be correctly overriden Also we now replace `{{pluginNamespace}}` for plugins as requirement for gradle 8 Same `__PACKAGE__` is replaced in app build.gradle (for namespace or app id)


farfromrefug:
superseded by #5774

PR #5772, opened 2023-12-01
feat: new `syncChangedDataOnDevices`

farfromrefug:
also send the changed files in `IFilesChangeEventData` so that hooks can know what files have changed

I needed those changes to handle all cases in localization plugin. I need to detect json file change to update native string files. Though i dont want to update the files on let s say js,jpg ...
With those changes i can now do whatever i want


rigor789:
replaced by #5774

PR #5783, opened 2024-01-28
feat: visionOS support

NathanWalker:
## PR Checklist

## What is the new behavior?


lib/services/ios/spm-service.ts
```diff
@@ -44,11 +50,14 @@ export class SPMService implements ISPMService {
 				ios: {
 					path: ".",
 				},
+				visionos: {
+					path: ".",
+				},
```
rigor789, line 55: Don't think this is needed or does anything - from Trapeze's perspective we're dealing with an ios project's pbxproj files - there is no `visionos` in trapeze.

NathanWalker: Build fails without this.

rigor789: That's strange - I believe that would be related to something else then - because there's no `visionos` in `MobileProjectConfig` - it doesn't do anything in trapeze. (I mean specifically these 3 lines, not the rest of the changes.)
image

NathanWalker: Hmm that is odd, I'll remove it as I don't see the failure any longer (may have been something from branch changes not sure?). I wonder if we do want to add that in future eventually with visionOS support (we can consider later though).

NathanWalker: Cleaned up here: https://github.com/NativeScript/nativescript-cli/commit/7c8a46a1ced5f87770c9ed6b259c412e1bda9246

rigor789 (approved):

PR #5845, opened 2025-07-16
fix: include process environment variables in webpack options

edusperoni:
We were overriding the environment on iOS so node options and others were not passed along to the webpack command


edusperoni:
Apparently I was too slow haha

PR #5937, opened 2025-12-16
feat: generate plugin outputs into .plugins folder inside platforms/android

edusperoni:
instead of writing the generated .aar into the plugin's node_modules folder, write it to `platforms/android/.plugins`. This prevents us from modifying node_modules

### DIFF
--- a/lib/services/android-project-service.ts
+++ b/lib/services/android-project-service.ts
@@ -89,7 +89,7 @@ function topologicalSortNativeDependencies(
 	dependencies: NativeDependency[],
 	start: NativeDependency[] = [],
 	depth = 0,
-	total = 0 // do not pass in, we calculate it in the initial run!
+	total = 0, // do not pass in, we calculate it in the initial run!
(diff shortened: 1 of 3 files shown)


edusperoni:
@farfromrefug

farfromrefug:
@edusperoni I like the idea! it will break plugins though. 
right now plugins having dependency on other plugins (or widget-release) had to had direct dependency on the jar. and we had to use it with path to the node modules folder.

actually what you are doing could make things simpler if the plugins folder is added to dependencies repositories. 
then i could simply add the jar name as a dép and it should be found

farfromrefug:
@edusperoni if i understand the output aar is now in platforms/android/plugins instead of in each node_modules/{plugin}/platforms/android ?
If so my plugins need an update. Here is an example https://github.com/nativescript-community/gesturehandler/blob/master/packages/gesturehandler/platforms/android/include.gradle#L21

PR #6012, opened 2026-04-15
fix: devtools csp to use 127.0.0.1

NathanWalker:
- Resolves the latest CSP 
Screenshot 2026-04-15 at 9 04 31 AM


jcassidyav (approved):
Strickly speaking only the changes to the server side are uneeded, it's going to listed on the loopback, it is just the client address that needs to change.

NathanWalker:
@jcassidyav pushed up with that isolated to clientside only for good measure.

jcassidyav (approved):

PR #6056, opened 2026-06-04
fix(bundler): copy the vite bundle to native in non-watch builds

JumpLink:
### Problem

With `bundler: 'vite'`, a **non-watch** build leaves the app's JS out of the APK/IPA, so the build fails.

The Vite output (`.ns-vite-build/`) is copied into the native project's `assets/app` only from the **watch-mode** IPC handler in `compileWithWatch` — the `childProcess.on("message", …)` branch that reads `message.emittedFiles` and calls `copyViteBundleToNative`. A non-watch build (`ns build`, or `ns run --justlaunch`) runs `compileWithoutWatch`, which:

- has **no IPC channel** — `startBundleProcess` sets `stdio: "inherit"` when `prepareData.watch` is `false`, so `process.send(...)` in the Vite child is a no-op and the `"message"` handler never fires; and
- never calls `copyViteBundleToNative` itself.

So the Vite output stays in `.ns-vite-build/`, `<platform>/.../assets/app` stays empty, and the Static Binding Generator fails:

```
Error executing Static Binding Generator: Couldn't find '…/platforms/android/build-tools/sbg-bindings.txt' bindings input file.
…
Error processing 'sbg-js-parsed-files.txt': no file was found in …/platforms/android/app/
```

Webpack/rspack are unaffected — they write directly into the platform folder rather than into `.ns-vite-build/`.

### Repro

Any `bundler: 'vite'` app:

```sh
ns build android        # or: ns run android --justlaunch
```

→ gradle fails at the SBG step because `assets/app` has only the runtime `package.json` and no bundle. `ns run android` (watch mode) works, which is what masks the bug.

### Fix

Copy the full Vite bundle in `compileWithoutWatch`'s `close` handler on a successful (exit code 0) build, mirroring the full-build branch of the watch-mode copy (`copyViteBundleToNative(distOutput, destDir)`), guarded on `this.getBundler() === "vite"`.

### Notes

- Minimal change; mirrors the existing watch-mode copy. `tsc` is clean.
- The `compileWithoutWatch`/`compileWithWatch` unit tests currently cover only the config-path error path (not the build/copy flow), so this matches the existing coverage. Happy to add a test that stubs the child-process `close` + asserts `copyViteBundleToNative` is called if you'd like.
- Found while building NativeScript apps with a Vite-8 config via [`@gjsify/nativescript-vite`](https://github.com/gjsify/gjsify); reproduced with stock `@nativescript/vite` + `ns build android`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved asset handling for Vite-based builds: non-watch builds now reliably copy Vite output into platform assets so native apps get the bundled files.
  * Watch-mode behavior improved: incremental and full updates from Vite are now propagated consistently to native assets during development.

### DIFF
--- a/lib/services/bundler/bundler-compiler-service.ts
+++ b/lib/services/bundler/bundler-compiler-service.ts
@@ -126,13 +126,9 @@ export class BundlerCompilerService
 						}
 
 						// Copy Vite output files directly to platform destination
-						const distOutput = path.join(
-							projectData.projectDir,
-							".ns-vite-build",
-						);
-						const destDir = path.join(
-							platformData.appDestinationDirectoryPath,
-							this.$options.hostProjectModuleName,
+						const { distOutput, destDir } = this.getVitePaths(
+							platformData,
+							projectData,
 						);
 
 						if (debugLog) {
@@ -389,13 +385,9 @@ export class BundlerCompilerService
 						// the Static Binding Generator with no input. Copy the full
 						// Vite bundle here instead.
 						if (this.getBundler() === "vite") {
-							const distOutput = path.join(
-								projectData.projectDir,
-								".ns-vite-build",
-							);
-							const destDir = path.join(
-								platformData.appDestinationDirectoryPath,
-								this.$options.hostProjectModuleName,
+							const { distOutput, destDir } = this.getVitePaths(
+								platformData,
+								projectData,
 							);
 							this.copyViteBundleToNative(distOutput, destDir);
 						}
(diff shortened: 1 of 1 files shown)


lib/services/bundler/bundler-compiler-service.ts
```diff
@@ -380,6 +376,21 @@ export class BundlerCompilerService
...
+						// runs and `<platform>/.../assets/app` stays empty, leaving
+						// the Static Binding Generator with no input. Copy the full
+						// Vite bundle here instead.
+						if (this.getBundler() === "vite") {
+							const { distOutput, destDir } = this.getVitePaths(
+								platformData,
+								projectData,
+							);
+							this.copyViteBundleToNative(distOutput, destDir);
+						}
```
NathanWalker, line 393: Thank you for these vite focused pr's @JumpLink - this one is actually already handled in the pending pr here: https://github.com/NativeScript/nativescript-cli/pull/6010/changes/8aef823df39df4816bb4d6ec6eb42db3cf207645

NathanWalker:
You can use `npm i -g nativescript@9.1.0-alpha.5` to give latest a try - this along with other niceties will be shipping with 9.1 this summer.

JumpLink:
Thanks @NathanWalker! I see `feat/vite-bundler-improvements` already implements this exact fix — the `exitCode === 0` → `if (isVite)` → `copyViteBundleToNative(...)` block in `bundler-compiler-service.ts` covers the non-watch case (same root cause: `stdio: "inherit"` in non-watch builds means the `emittedFiles` IPC copy never fires), and your version adds the try/catch plus the stale-artifact note, which is nicer than mine.

Since the branch supersedes this, I'll close this PR. Looking forward to 9.1 — I'll give `nativescript@9.1.0-alpha.5` a spin with our Vite setup and report back if anything comes up. Thanks for folding it in! 🙏

PR #6083, opened 2026-07-21
fix: copy Vite output for production builds

tannerlinsley:
## Problem

Vite writes non-watch builds to `.ns-vite-build`, but the CLI only copied that output into the native app directory from the watch-mode IPC handler. `ns build ios` and `ns build android` could therefore package stale or missing JavaScript.

## Fix

- share Vite source/destination path resolution between watch and non-watch builds
- copy a successful non-watch Vite build before resolving preparation
- fail the production build when the Vite output is missing or cannot be copied
- verify the source exists before replacing the destination

## Validation

- `npm test` (1509 passing, 9 pending)
- release iOS simulator build; verified a unique production endpoint in the final `.app` bundle
- signed Android release build; verified a unique production endpoint in the final `.apk` bundle

## Summary by CodeRabbit

* **Bug Fixes**
  * Vite builds now copy generated files to the native app destination immediately after successful non-watch builds.
  * Improved handling of missing or failed Vite output copies, with errors surfaced when required.
  * Destination directories are fully refreshed during Vite bundle copies to prevent stale files.

* **Tests**
  * Added coverage for successful Vite output copying and copy failure handling.

### DIFF
--- a/lib/services/bundler/bundler-compiler-service.ts
+++ b/lib/services/bundler/bundler-compiler-service.ts
@@ -126,13 +126,9 @@ export class BundlerCompilerService
 						}
 
 						// Copy Vite output files directly to platform destination
-						const distOutput = path.join(
-							projectData.projectDir,
-							".ns-vite-build",
-						);
-						const destDir = path.join(
-							platformData.appDestinationDirectoryPath,
-							this.$options.hostProjectModuleName,
+						const { distOutput, destDir } = this.getViteBuildPaths(
+							platformData,
+							projectData,
 						);
 
 						if (debugLog) {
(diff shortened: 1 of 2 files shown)


NathanWalker:
Thanks @tannerlinsley we have a lot of Vite improvements targeting 9.1 - this case is handled here:
https://github.com/NativeScript/nativescript-cli/pull/6010/changes#diff-8521a4dbe0da9ef2aed7d98bcabac41a7465f50c76019e2b539c9d214c4bf310
However you address a failure mode that's better so I'll bring both together soon.

You can use this (which includes 6010) now, with `npm i -g nativescript@alpha` and if using npm scripts can ensure it picks up a local scoped cli if needed `npm i --save-dev nativescript@alpha`

NathanWalker:
Ignore the merge conflicts, I'll handle those later today or tomorrow 👍

PR #6101, opened 2026-07-30
feat(commands): defineCommand — typed declarative commands via an ICommand adapter

edusperoni:
> Based on `main` — #6099 (the DI foundation) is merged.

## PR Checklist

- [x] The PR title follows our guidelines: https://github.com/NativeScript/NativeScript/blob/master/tools/notes/CONTRIBUTING.md#commit-messages.
- [ ] There is an issue for the bug/feature this PR is for. To avoid wasting your time, it's best to open a suggestion issue first and wait for approval before working on it.
- [x] You have signed the [CLA](http://www.nativescript.org/cla).
- [x] All existing tests are passing: https://github.com/NativeScript/nativescript-cli/blob/master/CONTRIBUTING.md#contribute-to-the-code-base
- [x] Tests for the changes are included.

## What is the current behavior?

A command is a class implementing `ICommand`, registered under a stringly-typed key, reading flags off the untyped global `$options` object. The validation semantics carry a trap: declaring `canExecute` silently disables `allowedParameters` validation, and an empty `allowedParameters` means "reject all positional arguments" — none of which the types express.

## What is the new behavior?

`defineCommand` — a declarative, typed command definition that plugs into the existing registry through an adapter (`createCommandFromDefinition` / `registerCommandDefinition`, routed through the `CommandRegistry` facet with a real `useFactory` provider). Fully additive: routing, help, hooks, and analytics behavior are untouched, and legacy `ICommand` classes remain fully supported.

```ts
export default defineCommand({
	name: "widget|add",
	options: { overwrite: booleanOption({ default: false }), output: stringOption({ alias: "o" }) },
	arguments: "none",
	async run(ctx) {
		// ctx.options.overwrite is boolean (has a default); ctx.options.output is string | undefined
		if (!ctx.options.output) ctx.fail("An output path is required.");
	},
});
```

- **Honest option types**: an option with a `default` is `T`; without one it is `T | undefined` — pinned by a strict-mode compile fixture (`test/type-fixtures/`), since the repo's own build has `strictNullChecks` off and any in-suite assertion would be vacuous. A no-options command's `ctx.options` rejects typos.
- **The `canExecute` trap is gone**: the adapter always enforces the declared `arguments` policy first, then calls a user `canExecute` as pure refinement — the fields compose instead of interacting.
[...]

PR #6102, opened 2026-07-30
feat(extensions): nativescript.commands map — per-command lazy loading for extensions

edusperoni:
> Based on `main` — #6099 (DI foundation), #6100 (defineHook) and #6101 (defineCommand) are all merged.

## PR Checklist

Every installed extension is eagerly `require()`d on every CLI invocation, before the command is even known — the extension's whole module tree loads so its top-level side effects can register commands against `global.$injector`. `nativescript.commands` in an extension's package.json is a `string[]` used only to suggest installs for unknown commands. Two extensions claiming the same command name crash at startup.

## What is the new behavior?

`nativescript.commands` also accepts a map of command name → module path, which becomes authoritative:

```jsonc
"nativescript": {
	"commands": {
		"widget|add": "./dist/commands/widget-add.js",
		"widget|new": { "path": "./dist/commands/widget-add.js" }  // alias: same module, second entry
	}
}
```

- **Per-command lazy loading**: nothing from the extension loads until one of its commands actually executes; the extension main is never required, and map-manifest extensions are not flagged by the deprecation tracer (the legacy array/eager path keeps working verbatim, tracer included). Values accept `string | { path }` so the envelope can grow additively.
- **The manifest key is authoritative**: routing works before any module loads. If a loaded `defineCommand` definition's `name` disagrees with its manifest key, the CLI warns naming both and runs under the key. Aliases are duplicate manifest entries pointing at the same module.
- **`registerDeferredCommand` on the `CommandRegistry` facet**: claiming a name and loading its implementation are now separate registry operations. The registry builds the command record, the parent's subcommand list, and the parent dispatcher *from the name alone* — a sibling's dispatch never drags in the first claimant's module — and returns a structured `DeferredCommandResult` (`claimed` / `built-in` / `subcommand-parent` / `invalid-name`) instead of exception text callers must match on. This is what keeps the future registry extraction a provider swap.
- **Failure UX**: loader failures name the command, the owning extension, and the module path; a loader that runs but registers nothing fails the same way; non-lowercase manifest keys (permanently unreachable — dispatch lower-cases input) warn and are skipped without sinking the extension's other commands. Built-in conflicts say "already provided by the CLI", no internals.
- **Deterministic defaults**: `*default` entries sort first per parent in code — JSON key order carries no meaning. First-wins conflict resolution is defined in the docs (extension load order, alphabetical; the mid-load `ns extension install` exception documented). Re-declaring a command under the same owner is a no-op, so `ns extension install <already-installed>` no longer warns about conflicting with itself. `"commands": {}` opts out of loading entirely.
- A command module may **self-register on load** (legacy shape) or simply **export a `defineCommand` definition** — one registration code path (`registerDefinitionAs`) serves both the manifest loader and `registerCommandDefinition`.
- `ILazyRequireProvider` is no longer part of the exported `Provider` union (container-internal).
- New authoring guide: [`extensions.md`](https://github.com/NativeScript/nativescript-cli/blob/feat/extension-manifests/extensions.md) — leads with the `peerDependency` + `devDependency` on `nativescript` and `inject()` from `nativescript/contracts`.

Public type names follow the new-API convention (no `I` prefix): `DeferredCommandOptions`, `DeferredCommandResult`, `DeferredCommandRejection`.

25 tests in `test/extension-manifests.ts` (lazy registration, eager-path preservation, malformed/conflict/self-conflict handling, both suggestion shapes, pure-definition modules incl. resolving the parent dispatcher before any child module has loaded, key-mismatch warning, alias entries, `{}` opt-out). Full stacked suite: 116 files, 1784 passed / 9 skipped; yok oracle, public-API test, and compat fixtures untouched.

## Summary by CodeRabbit

- **New Features**
  - Extensions can declare CLI commands directly in their manifests.
  - Commands load on demand, with support for command definitions and aliases.
  - Extension metadata now reports declared commands.
  - Legacy command manifest formats remain supported.
  - Added validation and clear handling for naming conflicts, malformed entries, and loading failures.
  - Command registration reports clear outcomes when commands cannot be registered.

- **Documentation**
  - Added comprehensive guidance for creating CLI extensions and configuring commands.
  - Added related-guide links and updated command declaration guidance.

### DIFF
--- a/defining-commands.md
+++ b/defining-commands.md
@@ -322,9 +322,10 @@ side-effect-free contracts entry point deliberately does not pull it in.
 `defineCommand`, the option helpers and all the types are exported from both
 `nativescript/contracts` and `lib/common/define-command`.
 
-Declaring commands from an extension manifest, so that an extension does not
-have to call a registration function at load time, is being added separately.
-Until then, extensions register definitions the same way the CLI does.
+Extensions do not need `registerCommandDefinition` at all: a
+`nativescript.commands` manifest entry may point straight at a module that
+exports a definition, and the CLI adapts and registers it lazily under the
+manifest key (see [extensions.md](extensions.md)).
 
 Relationship to `ICommand`
 --------------------------
--- a/dependency-injection.md
+++ b/dependency-injection.md
@@ -265,6 +265,13 @@ The first tranche, growing as services migrate:
 | `DoctorService` | `doctorService` |
 | `ProjectNameService` | `projectNameService` |
 
+Related guides
+--------------
+
+- [defining-commands.md](defining-commands.md) — declarative, typed commands via `defineCommand`.
+- [extensions.md](extensions.md) — extension authoring, including the `nativescript.commands` manifest.
+- [extending-cli.md](extending-cli.md) — hooks, including the typed `defineHook` API.
+
 Legacy → new quick reference
 ----------------------------
 
--- a/extensions.md
+++ b/extensions.md
@@ -15,7 +15,8 @@ of them. That makes the manifest below the most important file in an extension:
 it is what the CLI reads on startup, and it decides whether your code is loaded
 eagerly or only when one of your commands is actually executed.
 
-## Depending on the CLI
+Depending on the CLI
+--------------------
 
 An extension that imports anything from the CLI — `defineCommand`, `inject`, the
 types — needs `nativescript` declared twice:
@@ -63,7 +64,8 @@ searches npm for it when it needs to suggest an extension for an unknown command
 > heading; declaring the peer dependency now is what makes an extension ready
 > for it.
 
-## Declaring commands
+Declaring commands
+------------------
 
 Commands are declared in the `commands` key of the `nativescript` key of the
 extension's `package.json`. Two shapes are accepted.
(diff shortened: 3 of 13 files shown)


NathanWalker:
@copilot resolve the merge conflicts in this pull request

edusperoni:
Thanks — went through the review. Three of the four are addressed in 82309268f and 945528ab7:

- **Null-prototype maps** — valid, and it applied to `hierarchicalCommands` too, which was the worse of the two: a manifest key like `constructor|foo` skipped array initialisation on the truthy inherited value and then threw `TypeError: ... push is not a function`. Both maps now start from `Object.create(null)`. Audited the rest of `yok.ts` and `extensibility-service.ts`: the remaining name-keyed structures are `Map`s, `Set`s, or read only through own-key access, so nothing else was reachable.
- **Child of an existing parent command** — valid, and it broke the point of the structured result: the child record was registered and `{ registered: true }` returned, while `createHierarchicalCommand` then declined to install a dispatcher, leaving the command permanently unreachable and its ownership recorded. It is now rejected up front, before any state is touched, with a new `parent-is-command` reason carrying the parent name, so the caller reports a conflict naming the extension, the command and the parent.
- **Heading style / fence language** — valid, `extensions.md` was the odd file out; converted to setext to match `defining-commands.md` and `dependency-injection.md`, and the install-suggestion block is now labelled.

Not taking the fourth (`test/extension-manifests.ts` resolving `lib/contracts` through the Vitest loader): the suite runs the compiled output — `vitest.config.ts` includes `dist/test/**/*.js` — so from `dist/test/` that `require.resolve` already lands on `dist/lib/contracts`, and the generated fixture's plain `require()` of the absolute path is what we want. Those tests pass as written.

Three tests were added for the fixed behaviours, each verified to fail against the unfixed code first. Suite: 1815 passed / 9 skipped.

NathanWalker (approved):

PR #6121, opened 2026-08-05
fix(di): share the injection context across duplicated CLI copies

edusperoni:
## PR Checklist

The injection context (`inject()`'s hidden `current` slot) is a module-local variable. Node caches modules per resolved path, so when a hook or extension module resolves a *different copy* of the CLI than the one running — a nested `nativescript` install in an extension's tree, or a project-local copy under a globally-run CLI — that copy's `inject()` reads its own dead slot and throws "inject() can only be called from an injection context", even though the author is synchronously inside `run()` in a perfectly valid context. Verified in review with a realistic duplicated-copy install; it breaks the documented flagship examples of the stacked API PRs.

## What is the new behavior?

The context lives on `globalThis` under a `Symbol.for` key, so every loaded copy shares one slot — in the single-copy world the behavior is byte-identical. When a copy serves `inject()` through a frame it did not set (i.e. a duplicated copy is in play), it warns **once**, naming its own path and pointing at the fix: duplicated copies work but load the CLI twice; `peerDependency` on `nativescript` lets the running copy be shared. This keeps the ecosystem nudged toward the single-copy/local-install direction while making the documented API true in every install layout that exists today.

Tests include a genuine second-copy scenario: `inject.js` has no runtime imports, so a copied file loaded from another path is a real second module instance — it resolves through the running copy's context and produces exactly one warning.

Full suite green; no behavior change when a single copy is loaded.

### DIFF
--- a/dependency-injection.md
+++ b/dependency-injection.md
@@ -110,6 +110,13 @@ throws. `self` and `skipSelf` cannot be combined. There is deliberately no
 `host` option: it is an Angular component-tree concept with no analog in the
 CLI's injector hierarchy.
 
+The injection context is shared process-wide. If a hook or extension module
+ends up resolving a *duplicated* copy of the CLI (a nested `nativescript`
+install, or a project-local copy under a globally-run CLI), its `inject()`
+still resolves against the running CLI's context — with a one-time warning,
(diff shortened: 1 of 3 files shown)


edusperoni:
Landed directly on main as b52b8e051.

PR #6125, opened 2026-08-11
feat(catalyst): run and build Mac Catalyst apps with `ns run catalyst`

farfromrefug:
## PR Checklist

There is no way to build or run a NativeScript app as a Mac Catalyst app.

## What is the new behavior?

`catalyst` becomes a supported platform: `ns build catalyst` and `ns run catalyst`, the latter with full LiveSync.

> **On the name.** This deliberately does *not* claim `macos`. A separate effort adds `ns run macos` for a native macOS app built against a macOS runtime — a different product. Mac Catalyst is the iOS app rebuilt against the macOS SDK, so it takes the name that says what it is. The two can land independently.

A Catalyst app keeps every iOS convention — `App_Resources/iOS`, the iOS runtime package, each plugin's `platforms/ios` folder, the iOS Podfile, the iOS bundle — and diverges only in the directory it prepares into (`platforms/catalyst`) and the SDK it builds against.

- **Platform resolution.** `iOSProjectService` reports iOS as the platform name so every iOS convention falls out for free, and `projectRoot` becomes the single place Catalyst differs. Two call sites rebuilt the platform directory from the platform *name* instead of reading `projectRoot`; both now use `projectRoot`, which already resolves `hostProjectPath` identically. Left as-is, a Catalyst build would have probed and deleted `platforms/ios`.
- **Catalyst build.** `xcodebuild` gets `-destination 'platform=macOS,variant=Mac Catalyst'` and `SUPPORTS_MACCATALYST=YES`, with the deployment target clamped to the 13.1 minimum. Products land in `<Configuration>-maccatalyst`.
- **The Mac as a device.** Build, deploy and LiveSync drive it through the existing pipeline. Everything is local: the `.app` is a directory on this machine, so the file system is a plain copy, install only records the built bundle, launch is `open -n`, and the log stream is `log stream` narrowed to the app and the runtime.
- **Watch-mode platform matching.** Prepare events are now stamped with the platform the caller asked for rather than the platform data's name. `run-controller` pairs an event with a device by comparing the two, so for Catalyst — iOS platform data, Catalyst device — a file change recompiled but reached no device. The non-watch path already reported the requested platform; watchers and bundler processes are keyed by it too, since `stopWatchers` and `stopBundlerCompiler` are called with it. Nothing changes for ios/android/visionos, where both strings are identical.

## Testing

`tsc` and `npm run build` are clean, and the new files pass `prettier --check`.

End-to-end verification used a real app (OSS Weather) on Apple Silicon: `ns build catalyst` runs pod install, generates metadata for `arm64-apple-ios<version>-macabi` and links and signs the `.app`; `ns run catalyst` launches it, streams its logs, and syncs a file change into the running app — an added `console.log` came back through the CLI after the automatic restart.

**A caveat on where that ran.** That app depends on a fork-only bundler feature (a configurable webpack package name), so on this branch alone `ns run catalyst` gets as far as adding the platform (`Platform catalyst successfully added`, `platforms/catalyst` created), discovering the Mac device and starting prepare, then fails inside webpack. Command registration, platform resolution and device discovery are therefore exercised on this branch; the build/launch/LiveSync steps were verified by running this exact code — including the rename — on a branch where the app's bundler resolves. Happy to redo the whole run against a vanilla template, though the runtime caveat below has to land first.

No unit tests are included: the change is platform plumbing plus a device implementation that shells out to `open`, `pkill` and `log`, none of which the existing suite has a harness for.

For reference, `vitest run` reports 189 failures out of 1648 here — the suite is already red on `main` with what appears to be the same set, so I have left that checklist box unchecked rather than claim otherwise.

## Note for reviewers

Two runtime-side fixes in `NativeScript/ios` are needed before a clean checkout can build a working Catalyst app:

- the metadata generator needs the `iOSSupport` framework search paths, or the app launches into `ReferenceError: UIDevice is not defined` (opened as NativeScript/ios#433);
- the packaged xcframeworks must keep their versioned-bundle symlinks, or codesign rejects the app with "code object is not signed at all".

iOS never hits either, because it uses shallow framework bundles and its own SDK paths.

## Summary by CodeRabbit

* **New Features**
  * Added Mac Catalyst platform support.
  * Added commands to build and run Catalyst apps on macOS.
  * Added Catalyst device discovery, app management, file transfers, logging, and debugging.
  * Added Catalyst project configuration and automatic app bundle handling.
* **Bug Fixes**
  * Improved Catalyst platform path resolution and build output handling.
  * Added deployment-target validation, defaulting unsupported targets to iOS 13.1.


farfromrefug:
@NathanWalker should be ready for review. It is a first pass. might improve it in the future

PR #6130, opened 2026-08-18
feat(android): build only the ABIs of the devices being deployed to

farfromrefug:
### What

`ns run android` with a single arm64 device plugged in still builds x86, x86_64 and armeabi-v7a. This narrows the native build down to the ABIs of the devices it is about to deploy to, and installs the package that matches each device.

### How

`GradleBuildService` passes `-PabiFilters=<abis>` built from the devices the build targets, honouring `--device` and `--emulator`. Only the first (most preferred) ABI of each device is used, deduplicated across devices.

The CLI does not decide what that means for the build — the app's gradle configuration does, typically an `ndk.abiFilters` or `splits { abi { ... } }` block reading the property. A build that ignores it keeps building exactly what it built before, so this PR alone is a no-op until something consumes the property. An explicit `-PabiFilters` in `--gradleArgs` always wins.

**This needs #6129 to be fully functional.** The `app/build.gradle` that ships with the runtime today ignores `-PabiFilters`, so on its own this PR only helps projects that read the property from their own `App_Resources/Android/app.gradle`. #6129 ships the app gradle files with the CLI and its `app/build.gradle` consumes the property: it narrows `ndk.abiFilters` down, and for an apk debug build it splits so each abi gets its own package — which is what the install and rebuild logic below expects. Merged the other way round, #6129 gives the gradle side with nothing passing the property.

### When it does not apply

- `--no-filter-devices-arch` turns it off.
- `ns build` never narrows — its artifact is meant to be shipped. The command forces the flag off rather than relying on the default.
- App bundle (`--aab`) builds never narrow; a bundle carries every ABI.

### Device ABIs

`Mobile.IDeviceInfo` gained an optional `abis: string[]`. On android it comes from `ro.product.cpu.abilist64` + `ro.product.cpu.abilist32`, most preferred first, falling back to the single `ro.product.cpu.abi` on old devices that report neither. Both are already part of the `getprop` output the device details parser reads, so there is no extra adb round trip.

### Consuming the split output

Two things follow from a build that can now produce several packages:

- `DeviceInstallAppService` picks the package whose name contains one of the device's ABIs, falling back to the universal package and then to the newest one — which is what an unsplit build produces anyway, so single-package projects take the same path as before.
- `AndroidProjectService.checkForChanges` marks the native project as changed when a device in the current run has no package of its own in the build output. Without it a device plugged in after the first build would never get one: the sources did not change, so nothing else would ask for a native rebuild.
- `copyLatestAppPackage` became `copyAppPackages`. A directory `--copy-to` target receives every package the build produced; a single file target receives the universal one (or the only one). This is the one signature change on `IBuildArtifactsService`.

### Tests

`npm test` — 1862 passing. Added `test/services/android/gradle-build-service.ts` covering the ABI selection: all devices, `--device`, `--emulator`, deduplication, filtering off, and devices that report no ABIs.

### Notes

From https://github.com/Akylas/nativescript-cli, in production use there. Mergeable independently of #6129, but only useful together with it — see above. #6134 builds on this one to pass the same ABIs to plugin builds. Expect a small textual conflict in `lib/options.ts` and `lib/definitions/build.d.ts` depending on merge order.

## Summary by CodeRabbit

* **New Features**
  * Added Android ABI filtering for source-built plugins through `--filter-plugins-devices-arch`.
  * Android builds and plugin builds now better match packages to connected devices and selected emulators.
  * Improved installation by selecting device-specific packages, with universal-package fallback.
  * Builds can copy multiple generated application packages to directory destinations.
* **Bug Fixes**
  * Native Android changes now trigger rebuilds when required device-specific packages are unavailable.
* **Documentation**
  * Documented Android architecture-filtering options and their behavior.

### DIFF
--- a/docs/man_pages/project/testing/debug-android.md
+++ b/docs/man_pages/project/testing/debug-android.md
@@ -38,6 +38,7 @@ Attach the debug tools to a running app in the native emulator | `$ ns debug and
     *   `--env.sourceMap` - creates inline source maps.
     *   `--env.hiddenSourceMap` - creates sources maps in the root folder (useful for Crashlytics usage with bundled app in release).
 * `--aab` - Specifies that the command will produce and deploy an Android App Bundle.
+* `--no-filter-devices-arch` - If set, builds every ABI instead of only the ones the connected devices report. The narrowing only applies when the app's gradle configuration acts on the `abiFilters` property, and `ns build` never narrows.
 * `--force` - If set, skips the application compatibility checks and forces `npm i` to ensure all dependencies are installed. Otherwise, the command will check the application compatibility with the current CLI version and could fail requiring `ns migrate`.
 
 <% if(isHtml) { %>
--- a/docs/man_pages/project/testing/run-android.md
+++ b/docs/man_pages/project/testing/run-android.md
@@ -43,6 +43,7 @@ Start a default emulator if none are running, or run application on all connecte
     *   `--env.sourceMap` - creates inline source maps.
     *   `--env.hiddenSourceMap` - creates sources maps in the root folder (useful for Crashlytics usage with bundled app in release).
 * `--aab` - Specifies that the command will produce and deploy an Android App Bundle.
+* `--no-filter-devices-arch` - If set, builds every ABI instead of only the ones the connected devices report. The narrowing only applies when the app's gradle configuration acts on the `abiFilters` property, and `ns build` never narrows.
 * `--force` - If set, skips the application compatibility checks and forces `npm i` to ensure all dependencies are installed. Otherwise, the command will check the application compatibility with the current CLI version and could fail requiring `ns migrate`.
 
 <% if(isHtml) { %>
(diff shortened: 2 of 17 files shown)


farfromrefug:
Pushed `--filter-plugins-devices-arch` (5ecb7c0), and updated the description with a "Plugins (opt-in)" section.

It passes the same `-PabiFilters` this PR computes to the gradle build of every plugin built from source. It belongs here because it is the same ABI set from the same device selection — now shared in `lib/services/android/devices-abis.ts` — rather than a second copy of that logic in another PR.

The gradle files the CLI generates for a plugin still ignore the property, on purpose: what a plugin needs per ABI is the plugin's call. The flag only matters to a plugin whose own `include.gradle` reads `abiFilters`, where a long NDK/CMake build can drop to one ABI instead of four.

Off by default, unlike the app-side narrowing: a narrowed aar is a partial artifact and the aar cache is keyed by the plugin sources, which do not change when a device with another ABI joins. The ABIs are part of the plugin build data now, so switching devices — or turning the flag back off — rebuilds the aar.

farfromrefug:
Moved the plugin-side flag out of this PR — it is #6134 now, based on this branch. This PR is back to app-side narrowing only, and the description above no longer mentions it. Sorry for the churn.

The device selection both need still moves into `lib/services/android/devices-abis.ts`, but that happens in #6134, so nothing here changed.

